PR #5, opened 2021-08-02
Modularize and add CITE-seq

jashapiro:
This PR splits up the scpca-nf main workflow into a couple of modules. One is mostly the previous workflow of salmon/alevin-fry for rnaseq data, but with the whitelist/collate/quant steps combined into a single process, as in https://github.com/AlexsLemonade/alsf-scpca/pull/120. The other module is the feature barcode workflow from https://github.com/AlexsLemonade/alsf-scpca/pull/120.

The modularization makes `main.nf` much more streamlined, so it is now hopefully a bit easier to parse for somebody new to the workflow. 
I also added a step that points to how one of the next steps of the workflow will procede, joining CITE-seq and RNA-seq data to merge those into singleCellExperiment objects. 

I expect that this will still need tweaking for handling cell hashing, as that will end up with splitting SCE data, but I think the first step should be similar; the cell hash data and RNA-seq will be merged first into a SingleCellExperiment, then split in a subsequent step.

I also made some tweaks to the configuration file, based on the AWS instances that we have been seeing. Specifically, I added a 12 processor job, with 30GB RAM, which is less than the 32GB previously used, but will hopefully be sufficient for most cases. The reasoning was that the c4 instances that Batch has been choosing have 36 cpus and 60GB RAM, which means that there can only be one job per machine with 32GB requests, but 2 fit with 30 GB. (Side note, I am not sure why Batch is choosing c4 machines,  as it seems like m4 machines with 16 cpus and 64GB would have been a better choice with the 8cpu/32GB requests, but I guess spot pricing must be making a different choice. Not sure the best optimization here, to be honest, but I thought I would try this for now.


modules/af-features.nf
```diff
@@ -0,0 +1,137 @@
...
+      -o ${run_dir} \
+      -p ${task.cpus} 
+    """
+}
+
+// quantify features from rad input
+process fry_quant_feature{
+  container ALEVINFRY_CONTAINER
+  label 'cpus_8'
+  publishDir "${params.outdir}/features"
```
allyhawkins, line 65:
```suggestion
  publishDir "${params.outdir}/${run_id}"
```
If we use this as is right now, then we are going to have folders within the `processed` folder with `rna` and `features` and then each of those will have individual sample folders within that. This means the rna and features for each sample are separated. I personally like the idea of having a folder for each sample directly within the processed folder and then subfolders within each sample folder for rna and features so that the files for each sample are together. This is just my personal preference but either works.

allyhawkins (approved):
I think this looks good to me! My only minor comment is about the file structure and whether or not we want to store the rna and features counts output in separate folders or if it would be more beneficial to have one folder for each sample and then have a subfolder for rna, cite, and then any other data (like bulk) that may be associated with that sample. Personally, I like the idea of having all the counts data for each sample together in one folder. 

Also, this sets us up nicely for combining data as well as grabbing file locations to create sce objects.

modules/af-features.nf
jashapiro: We could do something like this, but the problem is we won't want to use `run_id` as that is different for the CITE-seq and RNA-seq data, and I have a fear that `sample_id` is going to go south once we get to cell hashes, where the sample ID is a list.

So ultimately we will want to combine these, but for now I don't really see a good way to do it, and I was thinking this would be a later in the pipeline kind of problem. I actually expect we will _not_ publish this step, but we need it for now to evaluate the outputs and prepare next steps.

The cell hash sample id question is more general problem that I was kind of putting off for now, but thinking ahead:
What might make most sense is for us to create another ID column that is explicitly used for grouping related experiments (RNA + cell hashes or RNA + CITE-seq). This could then be used for the join later, rather than sample IDs.

modules/af-rna.nf
```diff
@@ -0,0 +1,106 @@
+
+
+// containers
+SALMON_CONTAINER = 'quay.io/biocontainers/salmon:1.5.2--h84f40af_0'
+ALEVINFRY_CONTAINER = 'quay.io/biocontainers/alevin-fry:0.4.1--h7d875b9_0'
```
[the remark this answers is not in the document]
jashapiro: This does sound like a good idea. I might turn the cell barcode map into a param too, just to save passing it to the modules explicitly as I do now.

modules/af-features.nf
allyhawkins: You are right about not using the `run_id`, I completely forgot that they would have different ones so it would be better to combine by sample and not run. I like the idea of adding in another column that would help with the join later on.

jashapiro: I found that this is going to be required, because there is at least one sample that appears in two different RNA/CITEseq pairs. So I am adding this as I complete https://github.com/AlexsLemonade/ScPCA-admin/issues/154. My current plan is to add a `scpca_libary_id` which will be the same numerically as the `scpca_run_id` in most cases, but for multimodal data, (i.e. CITE-seq or cell hashing) will be the same as that of the paired RNA run. So after that PR is in, I will modify this to output by library ID. This  probably still won't be the final output location though, as we will still need to handle the splitting of cell hashed libraries!

PR #9, opened 2021-08-16
Add support for cellhashing

### LINKED ISSUE #7: Add support for cellhashing
At the moment, we only support RNA-seq and CITE-seq data. Adding support for cellhashing should be as simple as adding the correct `technology` types to the nextflow script. In theory, the `feature` processes should work the same for CITE-seq and feature barcode libraries.

jashapiro:
This PR adds support for cellhash data, closes #7 

The change is pretty minimal: I added cellhash data to the cellbarcode list in `main.nf`, and made sure we included it asa type for "feature" data. 

I did simplify the UMI geometry map a bit as well, so that we can expect that to work in the future if there are other types that we want to support.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -21,13 +21,16 @@ cell_barcodes = [
   '10Xv3.1': '3M-february-2018.txt',
   'CITEseq_10Xv2': '737K-august-2016.txt',
   'CITEseq_10Xv3': '3M-february-2018.txt',
-  'CITEseq_10Xv3.1': '3M-february-2018.txt'
+  'CITEseq_10Xv3.1': '3M-february-2018.txt',
+  'cellhash_10Xv2': '737K-august-2016.txt',
+  'cellhash_10Xv3': '3M-february-2018.txt',
+  'cellhash_10Xv3.1': '3M-february-2018.txt'
   ]
 
 // supported single cell technologies
 tech_list = cell_barcodes.keySet()
 rna_techs = tech_list.findAll{it.startsWith('10Xv')}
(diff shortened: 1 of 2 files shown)


cgreene:
You may want to also add genetic demultiplexing as a complementary approach (or a ticket to have it as a possibility) if folks multiplexed samples. It may also reveal if there are free-floating nucleic acids around from one or more of the samples (genetic demultiplexing will fail on many cells with a mixture but cell hashing won't necessarily). I don't think this is a breaking issue (cell hashing _should_ be better unless the samples have some significant fraction of compromised tissue, in which case you'd just be filtering "cells" that ended up as a mixture with genetic demultiplexing.

jashapiro:
> You may want to also add genetic demultiplexing as a complementary approach (or a ticket to have it as a possibility) if folks multiplexed samples. It may also reveal if there are free-floating nucleic acids around from one or more of the samples (genetic demultiplexing will fail on many cells with a mixture but cell hashing won't necessarily). I don't think this is a breaking issue (cell hashing _should_ be better unless the samples have some significant fraction of compromised tissue, in which case you'd just be filtering "cells" that ended up as a mixture with genetic demultiplexing.

Thanks for the suggestion @cgreene. That is going to require a whole new set of processes (we would need to actually map the reads, if I understand things correctly, so our current alevin-fry workflow can't do this). Only one group is currently using multiplexing at all, so I don't think it is something we are going to support in the near future.

allyhawkins (approved):
Looks pretty straight forward to me!

PR #11, opened 2021-08-16
Add support for joining by library ID

### LINKED ISSUE #8: Add support for passing library IDs with workflows and joining on that id
Joining runs with multimodal data by sample IDs works for most projects, but does not cover the case where a sample might have multiple libraries with separately paired RNA and CITEseq or cellhash data.  To support this, in https://github.com/AlexsLemonade/ScPCA-admin/pull/202, I added a new column, `scpca_library_id` which can be used as an unambiguous id for joining such libraries.

We could simply swap where we pass `sample_id` in the workflows, but a more sustainable method might be to use the system that `nf-core` seems to use: Rather than passing individual sample values (along with things like data files) as we do currently (which means we pass a bunch of values to each process), we could take advantage of a groovy `map` (dictionary in some other languages) to pass metadata values like various ids in a single process value (following `nf-core`, we can call this `meta`). This would (partially) free us from the tyranny of argument order, and allow easier changes to individual processes that might require more or different metadata, without having to change unaffected processes.

So basically, we would us `val(meta)` in place of `val(sample_id), val(run_id)` etc. and `meta` then use values such as `meta.sample_id` within processes where we would previously have just used `sample_id`. 

In the simplest implementation, we might use all info from the input csv file, but it is probably worth doing a bit of transformation to simplify some column names andonly retain columns that we need.

jashapiro:
Closes #8

This PR add support for merging RNAseq and feature runs based on the new library_id. It also changes the output location (publishDir) to be nested by sample and library id, for an organization that will be closer to the final output we expect to want.

The other major change is how metadata is passed around within processes. Each process now expects a metadata map (dictionary) as the first element of the main channel input it receives. It then returns this as the first element of the output tuple in all cases. 

This should make it easier for us to add or remove metadata that might be used by processes and pass them along with less fuss. It did make joining by library a bit more involved, but I think the tradeoff is probably worth it. 

One nice extension (not currently implemented, but I'm pretty sure this is possible) is that a process could add to the metadata before returning. This would be done by adding a line like
```
meta.foo = "bar"
```
within any process `script:` block (before or after the shell script portion). Adding something like this would not affect any other script.

At the moment, I have not included in the metadata map the slide information for spatial data. Since we have not yet implemented processing those data, I thought that could wait until we needed it. Adding it should not affect any existing process!

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -72,14 +72,11 @@ workflow{
   map_quant_feature(feature_ch)
   
   // combine feature & RNA quants for feature reads
+  feature_rna_quant_ch = map_quant_feature.out
+    .map{[it[0]["library_id"]] + it } // add library_id from metadata as first element
+    .combine(map_quant_rna.out.map{[it[0]["library_id"]] + it }, by: 0) // combine by library_id 
+    .map{it.subList(1, it.size())} // remove library_id index
   // just print for now
-  // combine_meta = {ch1, ch2, meta_value -> 
-  //   ch1_indexed = ch1.map{[it[0][meta_value]] + it }
-  //   ch2_indexed = ch2.map{[it[0][meta_value]] + it }
-  //   ch1_indexed.combine(ch2_indexed, by: 0)
-  //     .map{it.subList(1, it.size())} // remove index
-  // }
-  // combine_meta(map_quant_feature.out, map_quant_rna.out, "library_id")
-  //   .view()
+  feature_rna_quant_ch.view()
 
 }
--- a/main.nf
+++ b/main.nf
@@ -47,10 +47,21 @@ workflow{
   run_all = run_ids[0] == "All"
   runs_ch = Channel.fromPath(params.run_metafile)
     .splitCsv(header: true, sep: '\t')
+    // convert row data to a metadata map, keeping only columns we will need (& some renaming)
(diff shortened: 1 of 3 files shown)


allyhawkins (approved):
This all seems reasonable to me and looks good. To make sure I have the directory structure right for a sample that has CITE-seq, there will be two output folders that are published: 
`out_directory/sample_id/library_id/run_id-rna`
`out_directory/sample_id/library_id/run_id-feature`

And then we will have the same overall structure to house the RDS files.

jashapiro:
> This all seems reasonable to me and looks good. To make sure I have the directory structure right for a sample that has CITE-seq, there will be two output folders that are published:
> `out_directory/sample_id/library_id/run_id-rna`
> `out_directory/sample_id/library_id/run_id-feature`
> 
> And then we will have the same overall structure to house the RDS files.

Yes, though I kind of assume that we will at that point stop publishing the alevin folders, so we will will have something like:
`out_directory/sample_id/library_id/library_id_qcreport.html`
`out_directory/sample_id/library_id/library_id_filtered.rds`
`out_directory/sample_id/library_id/library_id_unfiltered.rds`

as the final outputs. `run_id` will go away as features get merged.

PR #12, opened 2021-08-19
Add process to convert alevin output to rds files 

### LINKED ISSUE #10: Return alevin output as RDS files
Related to #6, after the alevin-fry process is complete, the output needs to be transformed into an RDS file, one with the unfiltered matrix and one with the filtered matrix.

allyhawkins:
Closes #10. This PR adds a process to convert the alevin output matrix to both the filtered and unfiltered rds files. Right now, I have only set this up for the RNA counts and we will later need to add in integration for if a sample has CITE-seq or not. 

It is currently set up as a separate module, `generate-rds.nf` that is called from the `main.nf` workflow and takes as input a tuple containing the meta map and the alevin-fry output directory. This is the same output from running the `map_quant_rna` workflow in `af-rna.nf`. Within the new module, I am then running an R script `generate_output_files.R` which takes as input the `seq_unit`, needed to determine which counts to include in the final count matrix, the `alevin_dir`, and then the paths for the unfiltered and filtered rds files. I then use the `import_quant_data()` function from `scpcaTools` to read in the data and either filter or not filter before writing the output. 

Currently, the workflow is able to start running, but then the actual script is failing at the first line and causing an error which is why I have this PR in draft form. I am not entirely sure what the issue is and feel like it probably is something to do with a package not being present or the script not running properly, but any ideas on how to solve this error or ways to edit the scripts would be very helpful! I included a screenshot of the error I am receiving when running currently. 
Screen Shot 2021-08-19 at 9 46 39 AM


modules/generate-rds.nf
```diff
@@ -0,0 +1,20 @@
...
+process generate_rds{
+    container 'ghcr.io/alexslemonade/scpca-r'
+    publishDir "${params.outdir}/${meta.sample_id}/${meta.library_id}"
+    input: 
+        tuple val(meta), path(alevin_dir)
+    output:
+        tuple val(meta), path(unfiltered_rds), path(filtered_rds)
+    script:
+        unfiltered_rds = "${params.outdir}/${meta.sample_id}/${meta.library_id}/unfiltered.rds"
+        filtered_rds = "${params.outdir}/${meta.sample_id}/${meta.library_id}/filtered.rds"
```
jashapiro, line 12: We don;t need to specify the folder here, as that will only come into play for the publishing. But I think we do want to have the `library_id` in the file name.
```suggestion
        unfiltered_rds = "${meta.library_id}_unfiltered.rds"
        filtered_rds = "${meta.library_id}_filtered.rds"
```

modules/generate-rds.nf
```diff
@@ -0,0 +1,20 @@
+
+// generate unfiltered and filtered RDS file using scpcaTools
+process generate_rds{
+    container 'ghcr.io/alexslemonade/scpca-r'
+    publishDir "${params.outdir}/${meta.sample_id}/${meta.library_id}"
```
jashapiro, line 5: If the output files have the library ids, then we shouldn't need it for the final output.
```suggestion
    publishDir "${params.outdir}/${meta.sample_id}"
```

modules/generate-rds.nf
```diff
@@ -0,0 +1,20 @@
...
+        tuple val(meta), path(unfiltered_rds), path(filtered_rds)
+    script:
+        unfiltered_rds = "${params.outdir}/${meta.sample_id}/${meta.library_id}/unfiltered.rds"
+        filtered_rds = "${params.outdir}/${meta.sample_id}/${meta.library_id}/filtered.rds"
+        """
+        generate_output_files.R \
+          -s ${meta.seq_unit} \
+          -a ${alevin_dir} \
+          -u ${unfiltered_rds} \
+          -f ${filtered_rds}
```
jashapiro, line 18: For a workflow, I like to use the full argument names for easier interpretation.
```suggestion
          --seq_unit ${meta.seq_unit} \
          --alevin_dir ${alevin_dir} \
          --unfiltered_file ${unfiltered_rds} \
          --filtered_file ${filtered_rds}
```

modules/generate-rds.nf
```diff
@@ -0,0 +1,20 @@
...
+    publishDir "${params.outdir}/${meta.sample_id}/${meta.library_id}"
+    input: 
+        tuple val(meta), path(alevin_dir)
+    output:
+        tuple val(meta), path(unfiltered_rds), path(filtered_rds)
+    script:
+        unfiltered_rds = "${params.outdir}/${meta.sample_id}/${meta.library_id}/unfiltered.rds"
+        filtered_rds = "${params.outdir}/${meta.sample_id}/${meta.library_id}/filtered.rds"
+        """
+        generate_output_files.R \
```
jashapiro, line 14: I think this may have been the main problem? I think it was trying to run the script as a shell script.
```suggestion
        Rscript --vanilla generate_output_files.R \
```

jashapiro (commented):
I think the main issue was that it was trying to run your R script as a bash script! Adding `Rscript` to the front should make it work.

bin/generate_output_files.R
```diff
@@ -0,0 +1,70 @@
+# This script takes the output folder from alevin-fry as input and
```
jashapiro, line 1: Since slack seems to be down... Using `/usr/bin/Rscript` is not guaranteed to work, but the following shebang should:
```suggestion
#!/usr/bin/env Rscript

# This script takes the output folder from alevin-fry as input and 
```

allyhawkins: yes it was the shebang that did it! So note to future selves, to run Rscripts from nextflow they need to be in a bash script with the `#!/usr/bin/env Rscript` shebang. Thank you!

allyhawkins:
Thanks @jashapiro for the helpful comments. All should be working now and is ready for a full review.

jashapiro:
I was just thinking about this, and I'm wondering if we might want to separate the filtering to a separate script. I would keep them as a single process, but call one R script to make the unfiltered, and then a second one to filter it.

The reason I am suggesting this is that we may well want to have a separate script (or separate options) for the initial conversion once we get to incorporating other data, but filtering should be largely the same. So having separate scripts will allow easier composition later. There is a bit of added overhead with separate scripts (R startup & reading in the unfiltered RDS), but I think it will likely be worth it.

bin/generate_output_files.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env Rscript
```
jashapiro, line 1: This is still an R script, so it should be named with a `.R` extension.

The shebang just tells the shell what to use to execute it, but doesn't actually make it into a shell script.

jashapiro (commented):
Looks good! I had a few minor comments, but the major one is still the splitting the R script in two, which will allow for more code reuse, as we are likely to need a separate script for reading in cite-seq data with features, but filtering it will follow the same process, so we should set things up to use the same code.

bin/generate_output_files.sh
```diff
@@ -0,0 +1,72 @@
...
+                                    tool = "alevin-fry", 
+                                    which_counts = which_counts, 
+                                    usa_mode = TRUE)
+
+
+filtered_sce <- import_quant_data(quant_dir = opt$alevin_dir, 
+                                    tool = "alevin-fry", 
+                                    which_counts = which_counts, 
+                                    usa_mode = TRUE, 
+                                    filter = TRUE)
```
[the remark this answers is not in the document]
allyhawkins: That makes sense. I went ahead and made those changes so the first script `generate_unfiltered_rds.R` outputs the unfiltered sce and then the second script, `generate_filtered_rds.R` takes as input the unfiltered.rds, filters it, and then outputs the filtered sce as an rds file.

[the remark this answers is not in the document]
allyhawkins: I will say, that I stuck with using the function `scpcaTools::filter_counts()` because I am anticipating that any decisions we make in changing filtering we may then go back and implement in that function. Unless you have strong feelings against using that.

modules/generate-rds.nf
```diff
@@ -0,0 +1,24 @@
...
+    script:
+        unfiltered_rds = "${meta.library_id}_unfiltered.rds"
+        filtered_rds = "${meta.library_id}_filtered.rds"
+        """
+        generate_unfiltered_rds.R \
+          --seq_unit ${meta.seq_unit} \
+          --alevin_dir ${alevin_dir} \
+          --unfiltered_file ${unfiltered_rds}
+
+        generate_filtered_rds.R \
```
jashapiro, line 20: Gonna suggest that we make this a bit more descriptive of the action here.
```suggestion
        filter_sce_rds.R \
```

bin/generate_unfiltered_rds.R
```diff
@@ -0,0 +1,54 @@
...
+                                 opt$seq_unit == "nucleus" ~ "unspliced")
+
+# get unfiltered sce
+unfiltered_sce <- import_quant_data(quant_dir = opt$alevin_dir,
+                                    tool = "alevin-fry",
+                                    which_counts = which_counts,
+                                    usa_mode = TRUE)
+
+# write to rds
+readr::write_rds(unfiltered_sce, opt$unfiltered_file)
```
jashapiro, line 53: A though here (and for filtered data) is whether we want to compress these files. They should be very compressible, and people will appreciate the files being compressed (might save us some storage costs too). Note: this is all internal to rds: we don't need to change any other code whatever we decide.
```suggestion
readr::write_rds(unfiltered_sce, opt$unfiltered_file, compress = "gz")
```

jashapiro (approved):
Looks good! Two very small comments. One about making things smaller.

modules/generate-rds.nf
```diff
@@ -0,0 +1,24 @@
+
+// generate unfiltered and filtered RDS file using scpcaTools
+process generate_rds{
+    container 'ghcr.io/alexslemonade/scpca-r'
```
jashapiro, line 4: Oh, one more thought was that we should make this one of the params in `main.nf` so we can put all of the docker images in one place in the original file. (We may well have other processes that use our R image: better to only have one place to change)

allyhawkins: Good point, will make that small change.

bin/generate_output_files.sh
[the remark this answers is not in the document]
jashapiro: I'm fine with this for now. It makes the change process a bit more involved/less transparent, because people will have to look to see what the filtering actually is. Before doing any real processing we will want to make sure we are tagging the docker image and scpcaTools for reproducibility.

PR #13, opened 2021-08-19
Fix feature mapping

jashapiro:
When converting to the standard `meta` value for processes, I missed a place where I needed to substitute that in, breaking the mapping of feature data. Now fixed, so feature mapping works again


allyhawkins (approved):
LGTM and yay for no more empty output!

PR #22, opened 2021-09-02
Generate merged SCE objects with feature (cite-seq) data 

### LINKED ISSUE #16: Move production of filtered.rds output to its own process within workflow
Currently, there is one process, `generate-rds.nf`, that produces both the unfiltered and filtered rds files. We would like to separate those into two separate processes, both of which take as input the alevin output directory and the metadata. 

This would also mean we would need to alter the script, `filter_sce_rds.R` to also read in the alevin output before filtering.

### LINKED ISSUE #21: Generate merged SCE objects with feature (cite-seq) data
With https://github.com/AlexsLemonade/scpca-nf/pull/5 and https://github.com/AlexsLemonade/scpcaTools/pull/31 complete, we now need to update the workflow to generate SCE objects (and rds files) with both the RNA-seq and feature data (usually cite-seq). This can be done with a modification to the current R script for DRY reasons, but will probably require a separate process within nextflow, as having two input directories will be required.

jashapiro:
This PR adds functionality for generating merged SCE objects (with alt experiments) for libraries with feature data. Along the way, it splits up the unfiltered and filtered rds file generation. 
As such, this
closes #21 
and closes #16 

There are a few particular changes worthy of note/comments:
- There is some potentially confusing logic in the main workflow that splits up by whether a library has alt features or not, and then does a little dance to create a channel with alt features and rna. Please let me know if it is clear enough as is, or needs additional comments. There is at least one place where there is a thing that I do because nextflow is odd and seems to have a bug which requires me to do a type conversion in a place I would not logically do it... (Type conversions, what nonsense is this? nextflow should handle that... but I digress)

- the final output channel includes both unfiltered and filtered rds files, but we could separate that out if we so desired. I figured we would probably want both when we went onward to generating QC metrics.

- When merging alt experiments, I decided to just throw the whole `meta` object from the features in, nested within the main `meta` (from the RNA run). I like this solution, so the data is all retained, but I am happy to hear disagreement and alternatives.

- I changed some file names... I can change them back if it causes trouble.


modules/generate-rds.nf
```diff
@@ -1,24 +1,85 @@
...
+    publishDir "${params.outdir}/${meta.sample_id}"
+    input: 
+        tuple val(feature_meta), path(feature_alevin_dir), val (meta), path(alevin_dir)
+    output:
+        tuple val(meta), path(unfiltered_rds)
+    script:
+        unfiltered_rds = "${meta.library_id}_unfiltered.rds"
+        // add feature metadata as an element of the main meta object
+        meta['feature_type'] = feature_meta.technology.split('_')[0]
+        meta['feature_meta'] = feature_meta
```
allyhawkins, line 34: Why do we need to do this? I'm not sure I'm following what adding this information is providing.

allyhawkins, line 34: So here is where you are adding in the entire metadata for the CITE library and appending it to the main metadata? I don't know if that's 100 percent necessary, but I think it might come in handy as we create the qc reports so I would agree that this could stay in.

main.nf
```diff
@@ -79,6 +86,10 @@ workflow{
     .map{[it[0]["library_id"]] + it } // add library_id from metadata as first element
     .combine(map_quant_rna.out.map{[it[0]["library_id"]] + it }, by: 0) // combine by library_id 
     .map{it.subList(1, it.size())} // remove library_id index
```
allyhawkins, line 88: Just to make sure I understand this correctly, because you are joining the rna out to the feature_ch which is already filtered by `feature_techs` the only rows that will be present in this channel are those with library id's that are in both rna and features, so the rna only will not be present? I just want to make sure that that is being removed.

allyhawkins (approved):
This looks good to me, I just had a question for clarification about making sure we are not running the `generate_merged_rds` only on the libraries with both CITE and RNA. 

In regards to your specific questions - I think the logic regards splitting up the libraries into RNA only channel and CITE seq channel made sense. I also like having the metadata, unfiltered, and filtered rds for the output since I think we will need all of them as input for the qc report. 

For the metadata, I don't know if we particularly need every field, but I think it makes sense to add it all in rather than picking and choosing at this point and as we get down the line if we find we only want certain parts and can find a better solution we can modify it.

modules/generate-rds.nf
jashapiro: Yep. This will be nested inside, so you can use something like `meta.feature_meta.run_id` to get the run id from the feature data. I had considered making an entry of `meta.feature_run_id`,  but that seemed like it would require too much forethought. Here I am basically saying, "let's keep everything because I don't know what we might need later", which is part of the whole philsophy of passing around the `meta` object. This data is comparatively tiny, so we might as well retain more info rather than throwing things away and then having to go back later when we realize there is more we need.

main.nf
jashapiro: Yeah, so `combine` is strange as it is basically a left join, but the name doesn't make sense and `join` actually does something a bit different. In any case, only the things that are in the `feature_rna_quant_ch` object will get paired up (which is why the features are first). Everything else gets dropped.

The hard part was actually the step before of figuring out how to get RNA-only samples...

allyhawkins: Ah yes this makes sense. What an uninformative name...

PR #24, opened 2021-09-03
Allow processing by project/submitter

### LINKED ISSUE #14: Allow processing by project/submitter
Right now we have flags to process individual enumerated `run_ids` or the whole table (for data types we can process). Since we are likely to want to fire off the workflow by project, we should probably add a flag to allow filtering by PI (`submitter`). that info is not currently captured in the metadata, but can easily be added before filtering.

jashapiro:
This PR adds a new `project` parameter to the workflow that allows running all of the runs from a submitter as a block. As implemented, this option supersedes the `run_id` option, which seems like what we will want most of the time, but if people think it should work differently, we could change that.

I also made the `run_id` option a bit more flexible, in that it can now be used to specify library ids or sample ids as well: if any match, it will process those runs. 

Closes #14

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -11,9 +11,12 @@ params.index_path = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103/
 params.t2g_3col_path = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103/annotation/Homo_sapiens.GRCh38.103.spliced_intron.tx2gene_3col.tsv'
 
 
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
I like this setup and it should easily be converted when we add in a unique project ID. The only other thought that I had was if we want to further segregate by disease (i.e. for projects like the Dyer group that may have a few hundred samples). But I think we could leave this as is and then when we create project ID's segregate at that point and we wouldn't have to alter this code to account for disease types.

Issue #19, opened 2021-09-01
Modify filepath for splici index


allyhawkins:
Based on changes that will be made in AlexsLemonade/alsf-scpca#132, we will need to update the file path that is used for the splici index in `main.nf`.

jaclyn-taroni:
This was closed by #25 (the language of that PR used `closes <issue> and #19` and I believe that's why this was not automatically closed)

PR #25, opened 2021-09-03
Add workflow for building index to scpca-nf

### LINKED ISSUE #23: Rename index files and regenerate indices using most recent Ensembl version
Currently the names of our indices are not informative, so we need to rename those files with the ensembl version so we know where the indices came from. In the process of doing this, we should regenerate the indices with the newest Ensembl version.

allyhawkins:
This PR closes #23 and #19. I moved the `build-index.nf` workflow over from `alsf-scpca/workflows` to the root directory of this repo and have streamlined the processes to only make the splici index using kmer=31. Here there are now only two processes, the first takes as input the ensembl gtf and primary assembly fasta file and uses the Rscript `make_splici_fasta.R` to grab only the spliced cDNA and intronic regions from the fasta. This script also outputs the needed tx2gene 3 column file and the corresponding mitochondrial gene list. 

The second process then takes as input the path to the splici fasta and creates the index. In this PR, I also moved the path to the splici index and the tx2gene file path as parameters in the config file and removed them from the `main.nf` file. 

In testing this I ran into an error part way through the R script, but after testing the Rscript locally with no issues and closer inspection it was exiting with command 137, which usually means a memory error. So I bumped up the memory and am re-running but in the meantime putting this up in draft form to elicit any feedback on the actual workflow. 

- Are there any parameters included here that we would want to move into the config file in addition to the splici index path? 
- Is the setup of the two processes logical, or are there any recommendations on a better way to do this? 

The input reference files used for this workflow are stored on s3 at `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104`.

### DIFF
--- a/bin/make_splici_fasta.R
+++ b/bin/make_splici_fasta.R
@@ -1,0 +1,154 @@
+#!/usr/bin/env Rscript
+
+# This script takes the input reference fasta of the genome and the 
+# corresponding gtf file, identifies the regions of interest corresponding
+# to spliced transcripts and introns, then subsets the genome and gtf for 
+# only those particular regions, and finally outputs a spliced_intron.txome.fa, 
+# spliced_intron.txome.gtf, and spliced_intron.txome.tx2gene.tsv all corresponding to the genomic 
+# regions with spliced transcripts and introns
+
+# This script is also used to generate the 3 column spliced_intron.tx2gene_3col.tsv needed for alevin-fry.
+
+# finally, a list of mitochondrial genes is output for QC
+
+# load needed packages
+library(Biostrings)
+library(BSgenome)
+library(eisaR)
+library(GenomicFeatures)
+library(magrittr)
+library(optparse)
+library(tidyverse)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-f", "--gtf"),
+    type = "character",
+    default = "annotation/Homo_sapiens.GRCh38.103.gtf.gz",
+    help = "File path for input gtf file",
+  ),
+  make_option(
+    opt_str = c("-g", "--genome"),
+    type = "character",
+    default = "fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz",
+    help = "File path for reference fasta file",
+  ), 
+  make_option(
+    opt_str = c("-a", "--annotation_output"),
+    type = "character",
+    default = "annotation",
+    help = "Directory to write output files",
+  ), 
+  make_option(
+    opt_str = c("-o", "--fasta_output"),
+    type = "character",
+    default = "fasta",
+    help = "Directory to write output files",
+  )
+)
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# Get base file name from input gtf file
+file_base <- stringr::str_replace(basename(opt$gtf), "\\.gtf(\\.gz)?$", "")
+
+## files with spliced + intron regions
+spliced_intron_fasta_file <- paste0(file_base, ".spliced_intron.txome.fa.gz")
+spliced_intron_gtf_file <- paste0(file_base, ".spliced_intron.txome.gtf")
+spliced_intron_tx2gene_file <- paste0(file_base, ".spliced_intron.tx2gene.tsv")
+spliced_intron_tx2gene_3col_file <- paste0(file_base, ".spliced_intron.tx2gene_3col.tsv")
+
+mito_file <- paste0(file_base, ".mitogenes.txt")
+
+# make final output file names needed
+
+spliced_intron_fasta <- file.path(opt$fasta_output, spliced_intron_fasta_file)
+spliced_intron_gtf <- file.path(opt$annotation_output, spliced_intron_gtf_file)
+spliced_intron_tx2gene <- file.path(opt$annotation_output, 
+                                    spliced_intron_tx2gene_file)
+
+spliced_intron_tx2gene_3col <- file.path(opt$annotation_output, 
+                                 spliced_intron_tx2gene_3col_file)
+
+mito_out <- file.path(opt$annotation_output, mito_file)
+
+# Check for output directory 
+if (!dir.exists(opt$fasta_output)) {
+  dir.create(opt$fasta_output)
+}
+
+if (!dir.exists(opt$annotation_output)) {
+  dir.create(opt$annotation_output)
+}
+
+
+# extract GRanges object containing genomic coordinates of each annotated transcript
+# both spliced transcripts and intronic regions
+grl <- eisaR::getFeatureRanges(
+  gtf = opt$gtf,
+  featureType = c("spliced", "intron"), 
+  flankLength = 86L,
+  joinOverlappingIntrons = FALSE,
+  verbose = TRUE
+)
+
+# load in primary genome
+genome <- Biostrings::readDNAStringSet(opt$genome)
+names(genome) <- stringr::word(names(genome), 1)
+
+
+# add the levels and lengths
+seqlevels(grl) <- seqlevels(genome)
+seqlengths(grl) <- seqlengths(genome)
+
+# trim grl used for splici
+splici_grl <- trim(grl)
+
+# extract seqs from trimmed grl for splici index
+splici_seqs <- GenomicFeatures::extractTranscriptSeqs(
+  x = genome, 
+  transcripts = splici_grl
+)
+
+# remove sequence duplicates
+splici_seqs <- unique(splici_seqs)
+splici_grl <- splici_grl[names(splici_seqs)]
+
+# write splici to fasta 
+Biostrings::writeXStringSet(
+  splici_seqs, filepath = spliced_intron_fasta, compress = TRUE
+)
+
+# write the associated annotations to gtf 
+eisaR::exportToGtf(
+  splici_grl, 
+  filepath = spliced_intron_gtf
+)
+
+# create text file mapping transcript and intron identifiers to corresponding gene identifiers
+# make 2 column Tx2Gene for all spliced and intron sequences
+splici_tx2gene_df <- eisaR::getTx2Gene(splici_grl)
+
+# write out 2 column Tx2 gene mapping
+readr::write_tsv(splici_tx2gene_df, spliced_intron_tx2gene, col_names = FALSE)
+
+# make 3 column Tx2 gene needed for alevin-fry USA mode
+splici_tx2gene_df_3col <- splici_tx2gene_df  %>%
+  # add status column
+  # remove extra -I* on gene ID
+  mutate(gene_id = stringr::word(gene_id,1,sep = '-'),
+         #if transcript_id contains an added -I*, then it is usnpliced
+         status = ifelse(stringr::str_detect(transcript_id, '-I'), 'U', 'S'))
+
+# write 3 column tx2gene
+readr::write_tsv(splici_tx2gene_df_3col, spliced_intron_tx2gene_3col, col_names = FALSE)
+
+
+# reimport gtf to get list of mito genes 
+gtf <- rtracklayer::import(spliced_intron_gtf)
+mitogenes <- gtf[seqnames(gtf) == 'MT']
+
+# write out mitochondrial gene list
+writeLines(mitogenes$gene_id, mito_out)
--- a/bin/make_splici_fasta.R
+++ b/bin/make_splici_fasta.R
@@ -25,7 +25,7 @@ option_list <- list(
   make_option(
     opt_str = c("-f", "--gtf"),
     type = "character",
-    default = "annotation/Homo_sapiens.GRCh38.103.gtf.gz",
+    default = "annotation/Homo_sapiens.GRCh38.104.gtf.gz",
     help = "File path for input gtf file",
   ),
   make_option(
--- a/bin/make_splici_fasta.R
+++ b/bin/make_splici_fasta.R
@@ -25,13 +25,11 @@ option_list <- list(
   make_option(
     opt_str = c("-f", "--gtf"),
     type = "character",
-    default = "annotation/Homo_sapiens.GRCh38.104.gtf.gz",
     help = "File path for input gtf file",
   ),
   make_option(
     opt_str = c("-g", "--genome"),
     type = "character",
-    default = "fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz",
     help = "File path for reference fasta file",
   ), 
   make_option(
@@ -49,7 +47,6 @@ option_list <- list(
   make_option(
     opt_str = c("-s", "--assembly"),
     type = "character",
-    default = "",
     help = "Prefix name containing organism and ensembl assembly version to be used for file naming"
   )
 )
@@ -92,7 +89,7 @@ if (!dir.exists(opt$annotation_output)) {
 grl <- eisaR::getFeatureRanges(
   gtf = opt$gtf,
   featureType = c("spliced", "intron"), 
-  flankLength = 86L,
+  flankLength = opts$flank_length,
   joinOverlappingIntrons = FALSE,
   verbose = TRUE
 )
@@ -154,4 +151,4 @@ gtf <- rtracklayer::import(spliced_intron_gtf)
 mitogenes <- gtf[seqnames(gtf) == 'MT']
 
 # write out mitochondrial gene list
-writeLines(mitogenes$gene_id, mito_out)
+writeLines(unique(mitogenes$gene_id), mito_out)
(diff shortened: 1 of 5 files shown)


build-index.nf
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+// basic parameters
+params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104'
+params.gtf = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
+params.fasta = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
+params.assembly = 'Homo_sapiens.GRCh38.104'
```
jashapiro, line 8: I would move all of these to the params file as well, since we will want to modify them all at once if we change assembly version.

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+}
+
+
+process salmon_index{
+  container 'quay.io/biocontainers/salmon:1.4.0--hf69c8f4_0'
+  publishDir "${params.ref_dir}/salmon_index", mode: 'copy'
+  memory { 28.GB * task.attempt}
+  cpus 8
+  errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+  maxRetries 1
```
jashapiro, line 41: You should be able to replace all of these with: 
```suggestion
  label 'cpus_8'
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+  input:
+    path(splici_fasta)
```
jashapiro, line 43: Keep this generic, for reasons I will suggest elsewhere.
```suggestion
    path(fasta)
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+  output:
+    path(params.index_path)
+  script:
+    """
+    salmon index \
+      -t ${splici_fasta} \
```
jashapiro, line 49:
```suggestion
      -t ${fasta} \
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+
+// basic parameters
+params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104'
+params.gtf = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
+params.fasta = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
+params.assembly = 'Homo_sapiens.GRCh38.104'
+
+// generate fastq files with spliced cDNA + intronic reads 
+process generate_fastq{
+  container params.SCPCATOOLS_CONTAINTER
```
jashapiro, line 12:
```suggestion
  container params.SCPCATOOLS_CONTAINER
```

nextflow.config
```diff
@@ -4,6 +4,8 @@ params{
   SALMON_CONTAINER = 'quay.io/biocontainers/salmon:1.5.2--h84f40af_0'
   ALEVINFRY_CONTAINER = 'quay.io/biocontainers/alevin-fry:0.4.1--h7d875b9_0'
   SCPCATOOLS_CONTAINTER = 'ghcr.io/alexslemonade/scpca-tools'
```
jashapiro, line 6: If I made this mistake elsewhere, we should correct it there as well!
```suggestion
  SCPCATOOLS_CONTAINER = 'ghcr.io/alexslemonade/scpca-tools'
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+nextflow.enable.dsl=2
+
+// basic parameters
+params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104'
+params.gtf = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
+params.fasta = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
+params.assembly = 'Homo_sapiens.GRCh38.104'
+
+// generate fastq files with spliced cDNA + intronic reads 
+process generate_fastq{
```
jashapiro, line 11:
```suggestion
// generate fasta and annotation files with spliced cDNA + intronic reads 
process generate_splici{
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+    path(gtf)
+    path(fasta)
+  output: 
+    path(splici_fasta)
+  script:
+    splici_fasta="${params.ref_dir}/fasta/${params.assembly}.spliced_intron.txome.fa.gz"
+    """
+    make_splici_fasta.R \
+      --gtf ${gtf} \
+      --genome ${fasta}
```
jashapiro, line 28: assumptions about paths are dangerous, so we should be explicit about where things come from and go to (even though these are the defaults)
```suggestion
    make_splici_fasta.R \
      --gtf ${gtf} \
      --genome ${fasta}
      --fasta_output fasta
      --annotation_output annotation
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+  memory { 28.GB * task.attempt}
+  errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+  maxRetries 1
+  input:
+    path(gtf)
+    path(fasta)
+  output: 
+    path(splici_fasta)
+  script:
+    splici_fasta="${params.ref_dir}/fasta/${params.assembly}.spliced_intron.txome.fa.gz"
```
jashapiro, line 24: Let nextflow handle the generic parts of the paths; the ref dir is not something we should need to specify except in the `publishDir`
```suggestion
    splici_fasta="fasta/${params.assembly}.spliced_intron.txome.fa.gz"
```

bin/make_splici_fasta.R
```diff
@@ -0,0 +1,154 @@
...
+    default = "fasta",
+    help = "Directory to write output files",
+  )
+)
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# Get base file name from input gtf file
+file_base <- stringr::str_replace(basename(opt$gtf), "\\.gtf(\\.gz)?$", "")
```
jashapiro, line 55: For clarity & flexibility, I think it makes sense to have this come in as a parameter. In the nextflow script, you are already using `params.assembly`, which should be the same as this, correct?

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+  memory { 28.GB * task.attempt}
+  cpus 8
+  errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+  maxRetries 1
+  input:
+    path(splici_fasta)
+  output:
+    path(params.index_path)
+  script:
+    """
```
jashapiro, line 47: Let's define this based on the input file name? I want to keep this step generic so we can use it for other salmon indices. The other way we could do this would be to have the index name as a `val` input and define from that (which might actually be better).
```suggestion
  script:
    index_dir = fasta.baseName.split("\\.(fasta|fa)")[0]
    """
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+    salmon index \
+      -t ${splici_fasta} \
+      -i ${params.assembly}.spliced_intron.txome \
```
jashapiro, line 50:
```suggestion
      -i ${index_dir} \
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+  container 'quay.io/biocontainers/salmon:1.4.0--hf69c8f4_0'
+  publishDir "${params.ref_dir}/salmon_index", mode: 'copy'
+  memory { 28.GB * task.attempt}
+  cpus 8
+  errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+  maxRetries 1
+  input:
+    path(splici_fasta)
+  output:
+    path(params.index_path)
```
jashapiro, line 45: Again, we don't want to refer to full paths for input and output, let nextflow define these.
```suggestion
    path(index_dir)
```

build-index.nf
```diff
@@ -0,0 +1,62 @@
...
+  // publish fasta and annotation files within reference directory 
+  publishDir params.ref_dir
+  memory { 28.GB * task.attempt}
+  errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+  maxRetries 1
+  input:
+    path(gtf)
+    path(fasta)
+  output: 
+    path(splici_fasta)
```
jashapiro, line 22: I think we may need to explicitly output the annotation directory here as well. Otherwise it won't save the output somewhere we can get it easily. That means we will need to output a tuple with both this file and the annotation directory.
```suggestion
    tuple path(splici_fasta), path("annotation")
```

jashapiro (commented):
So I think this review may have turned into a bit of a mess, as I had a lot of comments and there were a lot of disjoint places that needed to be changed to stay consistent. Let me know if we need to try a different approach (It might make sense for me to just file a PR to this branch with all my changes?)

There are some file path things that may have been the source of the errors you were seeing: the short version is that the processes should never explicitly use full S3 paths for files. Nextflow renames and moves things, so we can't really know where they are: use the variables from `input` for everything. (the one exception is `publishDir`)

The other thing is that I want to keep things that we might reuse a bit more generic. The main thing I am thinking of there is the salmon index generation process, which we will likely want to use again for the index we will use for bulk. The splici index can't be used for bulk, but we will need the closest thing for that. For that, we probably want to output the `spliced` index & annotation as well as the `spliced_intron`. We may be able to use the same script for that, but with an additional option? I don't think we need to do that in this PR, but we should be thinking about what might be reused, and the salmon index step certainly can be, if we have the right flexibility of inputs.

Anyway, let me know if you want to pair to go over any of the changes I am suggesting, or if you want me to take on some next steps, or however you want to proceed.

allyhawkins:
Thank you so much for the feedback on this, your comments were extremely helpful! I was able to work through them and implement your comments and the path issue definitely seemed to be the main problem. You have a good point about using the index generation for bulk, although it still may be different because we will probably want a decoy addition there where as with alevin-fry we cannot have that? Just some food for thought.

bin/make_splici_fasta.R
```diff
@@ -0,0 +1,157 @@
...
+library(magrittr)
+library(optparse)
+library(tidyverse)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-f", "--gtf"),
+    type = "character",
+    default = "annotation/Homo_sapiens.GRCh38.104.gtf.gz",
```
jashapiro, line 28: I think we probably don't want defaults here, as they are unlikely to be present in a fresh download situation, so better to leave them out.
```suggestion
```

bin/make_splici_fasta.R
```diff
@@ -0,0 +1,157 @@
...
+    help = "File path for input gtf file",
+  ),
+  make_option(
+    opt_str = c("-g", "--genome"),
+    type = "character",
+    default = "fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz",
```
jashapiro, line 34: Same comment as above.
```suggestion
```

bin/make_splici_fasta.R
```diff
@@ -0,0 +1,157 @@
...
+    help = "File path for reference fasta file",
+  ), 
+  make_option(
+    opt_str = c("-a", "--annotation_output"),
+    type = "character",
+    default = "annotation",
```
jashapiro, line 40: This is okay, as we will make these if they don't exist.

bin/make_splici_fasta.R
```diff
@@ -0,0 +1,157 @@
...
+  make_option(
+    opt_str = c("-o", "--fasta_output"),
+    type = "character",
+    default = "fasta",
+    help = "Directory to write output transcript fasta files",
+  ),
+  make_option(
+    opt_str = c("-s", "--assembly"),
+    type = "character",
+    default = "",
```
jashapiro, line 52: Here, we would want either no default or some text, so that the file names created don't start with `.`, which could result in hidden files. Since I can't come up with a good default, I'm going to suggest no default, but if you have an idea, feel free to use it!
```suggestion
```

bin/make_splici_fasta.R
```diff
@@ -0,0 +1,157 @@
...
+  dir.create(opt$annotation_output)
+}
+
+
+# extract GRanges object containing genomic coordinates of each annotated transcript
+# both spliced transcripts and intronic regions
+grl <- eisaR::getFeatureRanges(
+  gtf = opt$gtf,
+  featureType = c("spliced", "intron"), 
+  flankLength = 86L,
```
jashapiro, line 95: Do we want to make this an option? There was some discussion in the Alevin paper about choice of flank length, and it doesn't seem to matter much, but it seems reasonable to me to make it flexible. (Make sure to set `type= "integer"` when adding to options).
```suggestion
  flankLength = opts$flank_length,
```

bin/make_splici_fasta.R
```diff
@@ -0,0 +1,157 @@
...
+# write 3 column tx2gene
+readr::write_tsv(splici_tx2gene_df_3col, spliced_intron_tx2gene_3col, col_names = FALSE)
+
+
+# reimport gtf to get list of mito genes 
+gtf <- rtracklayer::import(spliced_intron_gtf)
+mitogenes <- gtf[seqnames(gtf) == 'MT']
+
+# write out mitochondrial gene list
+writeLines(mitogenes$gene_id, mito_out)
```
jashapiro, line 157: For some reason, we were getting triplicated outputs in the file (at least the one I looked at), so let's make sure that doesn't happen. 

Not sure if any of the other output tables have duplicates: we might want to add a `dplyr::distinct()` step make sure they are unique too. 

```suggestion
writeLines(unique(mitogenes$gene_id), mito_out)
```

nextflow.config
```diff
@@ -3,7 +3,13 @@ params{
...
   SALMON_CONTAINER = 'quay.io/biocontainers/salmon:1.5.2--h84f40af_0'
   ALEVINFRY_CONTAINER = 'quay.io/biocontainers/alevin-fry:0.4.1--h7d875b9_0'
-  SCPCATOOLS_CONTAINTER = 'ghcr.io/alexslemonade/scpca-tools'
+  SCPCATOOLS_CONTAINER = 'ghcr.io/alexslemonade/scpca-tools'
+  index_path = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/salmon_index/Homo_sapiens.GRCh38.104.spliced_intron.txome'
+  t2g_3col_path = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv'
+  ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104'
+  gtf = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
+  fasta = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
+  assembly = 'Homo_sapiens.GRCh38.104'
```
jashapiro, line 12: Reordering and aligning for easier editing/error catching later.
```suggestion
  assembly = 'Homo_sapiens.GRCh38.104'
  ref_dir       = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104'
  fasta         = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
  gtf           = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
  index_path    = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/salmon_index/Homo_sapiens.GRCh38.104.spliced_intron.txome'
  t2g_3col_path = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv'
```

jashapiro (approved):
This looks good. I had a few minor additional comments, but nothing that should require additional review.

bin/make_splici_fasta.R
allyhawkins: yes I was also reading that and thinking that having that flexibility could be useful. I'll go ahead and add to the options but set the default to 86.

PR #28, opened 2021-09-08
Change publishDir to split alevin output and rds files 

### LINKED ISSUE #17: Change publishDir for alevin output to be different from publishDir for rds files
In an effort to keep the file organization close to how the final product will look, we want to keep the alevin output directories separate from the directories that are holding the final filtered and unfiltered rds files. To do this we will want to alter the `publishDir` to be different between the `map_quant_rna` workflow and the `generate_rds` process. For now we still want to keep the alevin directories and store them (without the RAD files), so will just move them to a new place rather than removing the `publishDir` step completely.

allyhawkins:
Closes #17. Here I am splitting up the `publishDir` for the modules that run alevin-fry to put the alevin output in a separate folder than the rds files. This gets us closer to our desired output structure. I'm leaving this in draft form right now because we anticipate there to be more structure changes with the addition of the unique project ID discussed in https://github.com/AlexsLemonade/ScPCA-admin/issues/213. At that point, we would want to further change the published output files to be nested by project and sample. Now all files are nested only by sample.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -3,7 +3,8 @@ nextflow.enable.dsl=2
 
 // run parameters
 params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
-params.outdir = "s3://nextflow-ccdl-results/scpca/processed"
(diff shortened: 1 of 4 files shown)


jashapiro:
superseded by #44

PR #29, opened 2021-09-09
Add colData and rowData to filtered sce object

### LINKED ISSUE #18: Modify generation of filtered sce object to include basic cell statistics in colData
The filtered sce object should include `colData` in the output. This includes everything that is calculated using `scater::addPerCellQC()` with the subset of mito genes along with a column containing the posterior probability of a cell being compromised using miQC. 

We will also want to add in the `rowData` to the object using `scater::addPerFeatureQC()` and remove any genes that are not detected in that sample.

allyhawkins:
Closes #18. This PR adds in the `colData` and `rowData` to the filtered sce object. This includes adding any per cell and per feature data for alternative experiments, if present in the filtered data. To do this, I first removed the previous `rowData` that is attached to the unfiltered data since we will want to recalculate the per gene metrics using only the cells that are present in the filtered matrix. 

We don't need to recalculate the per cell metrics here, but we did want to add in a column containing the posterior probability of each cell being compromised as calculated by `miQC`. Because we want to do this after performing filtering with `emptyDrops`, I added in a step here to calculate the model then using `miQC::filterCells()` with a `posterior_cutoff=1` add in the `prob_compromised` column, just like we have in `scpcaTools::add_cell_mito_qc()`. After discussion with @jashapiro we chose to do it manually here rather than removing the old `colData` and recalculating using scpcaTools. We are anticipating adding additional steps here to add in the `ccdl_suggests` column in the future. 

Finally, to calculate the `rowData` for any alternative experiments if present, I'm grabbing the names of the altExp and then looping through those names to remove any previous `rowData` and then recalculate it using the filtered cells. After a code pairing session with @jashapiro, we came to the conclusion that use of the `for` loop was ultimately the best way to handle this for now.

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -39,26 +39,26 @@ unfiltered_sce <- readr::read_rds(opt$unfiltered_file)
 # filter sce
 filtered_sce <- scpcaTools::filter_counts(unfiltered_sce)
 
-# need to remove old rowData first 
+# need to remove old gene-level rowData first 
 rowData(filtered_sce) <- NULL
-
-# add prob_compromised to colData from miQC::mixtureModel 
-model <- miQC::mixtureModel(filtered_sce)
-filtered_sce <- miQC::filterCells(filtered_sce, model, posterior_cutoff = 1, verbose = FALSE)
 
 # recalculate rowData and add to filtered sce 
 filtered_sce <- filtered_sce %>%
   scater::addPerFeatureQC()
+  
+# add prob_compromised to colData from miQC::mixtureModel 
+model <- miQC::mixtureModel(filtered_sce)
+filtered_sce <- miQC::filterCells(filtered_sce, model, posterior_cutoff = 1, verbose = FALSE)
 
 # grab names of altExp, if any
-feature_names <- altExpNames(filtered_sce)
+alt_names <- altExpNames(filtered_sce)
 
-for (feature in feature_names) {
+for (alt in alt_names) {
   # remove old row data from unfiltered 
-  rowData(altExp(test, feature)) <- NULL
+  rowData(altExp(test, alt)) <- NULL
   
   # add alt experiment features stats for filtered data
-  altExp(test, feature) <- scater::addPerFeatureQC(altExp(test, feature))
+  altExp(test, alt) <- scater::addPerFeatureQC(altExp(test, alt))
 }
 
 # write filtered sce to output
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -18,6 +18,18 @@ option_list <- list(
     opt_str = c("-f", "--filtered_file"),
     type = "character",
     help = "path to output filtered rds file. Must end in .rds"
+  ),
+  make_option(
+    opt_str = c("-m", "--mito_file"),
+    type = "character",
+    default = "",
+    help = "path to list of mitochondrial genes"
+  ), 
+  make_option(
+    opt_str = c("-n", "--feature_name"),
+    type = "character",
+    default = "ALT",
+    help = "Feature type"
   )
 )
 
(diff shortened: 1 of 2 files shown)


bin/filter_sce_rds.R
```diff
@@ -39,6 +39,28 @@ unfiltered_sce <- readr::read_rds(opt$unfiltered_file)
 # filter sce
 filtered_sce <- scpcaTools::filter_counts(unfiltered_sce)
 
+# need to remove old rowData first
```
jashapiro, line 42: Just to be more clear about what data this is
```suggestion
# need to remove old gene-level rowData first 
```

bin/filter_sce_rds.R
```diff
@@ -39,6 +39,28 @@ unfiltered_sce <- readr::read_rds(opt$unfiltered_file)
...
+# need to remove old rowData first 
+rowData(filtered_sce) <- NULL
+
+# add prob_compromised to colData from miQC::mixtureModel 
+model <- miQC::mixtureModel(filtered_sce)
+filtered_sce <- miQC::filterCells(filtered_sce, model, posterior_cutoff = 1, verbose = FALSE)
+
+# recalculate rowData and add to filtered sce 
+filtered_sce <- filtered_sce %>%
+  scater::addPerFeatureQC()
```
jashapiro, line 51: Swapping order for organization
```suggestion
# recalculate rowData and add to filtered sce 
filtered_sce <- filtered_sce %>%
  scater::addPerFeatureQC()
  
# add prob_compromised to colData from miQC::mixtureModel 
model <- miQC::mixtureModel(filtered_sce)
filtered_sce <- miQC::filterCells(filtered_sce, model, posterior_cutoff = 1, verbose = FALSE)
```

bin/filter_sce_rds.R
```diff
@@ -39,6 +39,28 @@ unfiltered_sce <- readr::read_rds(opt$unfiltered_file)
...
+# grab names of altExp, if any
+feature_names <- altExpNames(filtered_sce)
+
+for (feature in feature_names) {
+  # remove old row data from unfiltered 
+  rowData(altExp(test, feature)) <- NULL
+  
+  # add alt experiment features stats for filtered data
+  altExp(test, feature) <- scater::addPerFeatureQC(altExp(test, feature))
+}
```
jashapiro, line 62: Two different uses of "feature" here seemed potentially confusing.
```suggestion
alt_names <- altExpNames(filtered_sce)

for (alt in alt_names) {
  # remove old row data from unfiltered 
  rowData(altExp(test, alt)) <- NULL
  
  # add alt experiment features stats for filtered data
  altExp(test, alt) <- scater::addPerFeatureQC(altExp(test, alt))
}
```

jashapiro (approved):
LGTM, just a couple little tweak suggestions.

bin/filter_sce_rds.R
```diff
@@ -39,6 +39,28 @@ unfiltered_sce <- readr::read_rds(opt$unfiltered_file)
...
+# need to remove old gene-level rowData first 
+rowData(filtered_sce) <- NULL
+
+# recalculate rowData and add to filtered sce 
+filtered_sce <- filtered_sce %>%
+  scater::addPerFeatureQC()
+  
+# add prob_compromised to colData from miQC::mixtureModel 
+model <- miQC::mixtureModel(filtered_sce)
+filtered_sce <- miQC::filterCells(filtered_sce, model, posterior_cutoff = 1, verbose = FALSE)
```
jashapiro, line 51: Just realized that we should probably save the model in the metadata here as well. I guess that will have to go in a new PR, or maybe just happen when we decide we need it.

allyhawkins: I thought we decided that we don't need to save the model in the metadata here? But yes, if we realize we need it later on (I'm thinking maybe for the qc report) then we can go back and add it in here.

PR #31, opened 2021-09-16
Add generation of QC report to pipeline

### LINKED ISSUE #1: Add generation of QC report to scpca nextflow pipeline
After the counts matrices are obtained for each sample, we will want to generate a QC report. We will want to add a process to the nextflow pipeline that takes the counts matrices (or `SingleCellExperiment` objects) as input and outputs a QC report.

jashapiro:
This PR adds a final step to the pipeline to run the QC generation function from [`scpcaTools`](https://github.com/AlexsLemonade/scpcatools). The process and pipeline are pretty simple, but there were a few changes I made along the way, some for clarity and some for bug fixes and some just because I was inspired and/or delusional

* In testing, I ran into a few outstanding bugs in the pipeline (mostly related to alt experiments) that should now be fixed. (In addition to the bugs in `scpcaTools` itself that were fixed in that repo).
* I altered the container we are using to track the `edge` release for now, so we will always be using the latest scpcaTools as we update things like the QC report. We will want to freeze this to a release when we start processing.
* In a fit of late night hubris, I realized that since we now have to use an R image with R4.1 we could use native pipes instead of `magrittr`, so I did that, just to see if I could. Still works as before, but I could change it back... Or we could change all of `scpcaTools` to use it!

Closes #1

### DIFF
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -1,0 +1,53 @@
+#!/usr/bin/env Rscript
+
+# This script generates a QC report using scpcaTools from a pair of filtered and unfiltered SCE objects
+
+# import libraries
+library(magrittr)
+library(optparse)
+
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = c("-u", "--unfiltered_sce"),
+    type = "character",
+    help = "path to rds file with unfiltered sce object"
+  ),
+  make_option(
+    opt_str = c("-f", "--filtered_sce"),
+    type = "character",
+    help = "path to rds file with filtered sce object"
+  ),
+  make_option(
+    opt_str = c("-s", "--sample_id"),
+    type = "character",
+    help = "Sample identifier for report"
+  ),
+  make_option(
+    opt_str = c("-o", "--output_file"),
+    default = "qc_report.html",
+    type = "character",
+    help = "path to output file"
+  )
+)
+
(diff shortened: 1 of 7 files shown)


bin/sce_qc_report.R
```diff
@@ -0,0 +1,52 @@
...
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# check that input files exists
+if(!file.exists(opt$unfiltered_sce)){
+  stop("Missing unfiltered .rds file")
+}
+if(!file.exists(opt$filtered_sce)){
+  stop("Missing filtered .rds file")
+}
+
```
allyhawkins, line 42: I think we also need a check here to make sure that the sample Id isn't null? I don't think we want a default value for that so we probably want to add that in just in case.

allyhawkins (commented):
This looks pretty good to me. I think the only thing missing is the check for a missing sample ID in the `sce_qc_report.R` script. I feel like that is worthwhile to have. I'm fine with the other changes to the workflow and good catch on the bugs!

bin/sce_qc_report.R
jashapiro: For some reason, I thought that optparse would throw an error if no default was specified and no value given, but it turns out I am wrong about that. So I added the check you suggested.

allyhawkins (approved):
Looks good!

PR #35, opened 2021-09-22
Remove zero genes from filtered sce

allyhawkins:
This PR closes AlexsLemonade/scpcaTools#36. Here, I added in a step during the generation of the filtered sce objects to remove any genes that are not detected prior to writing out the final `filtered.rds` file. I chose to do that here rather than within the `filter_counts()` function within scpcaTools for transparency purposes, rather than having it performed within the function. It seemed like it was an additional filtering step that we were choosing to add, rather than something that we wanted nested within another function, but I'm open to other opinions on this. 

In terms of implementation, I just added an additional step to remove zero count genes as the last step before writing out the file. I chose to not remove any zero counts from the alternative experiments that could be present as I think knowing that an antibody wasn't detected in CITE-seq actually could be useful information for someone and didn't want to remove that from the object.

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -61,6 +61,12 @@ for (alt in alt_names) {
   altExp(filtered_sce, alt) <- scater::addPerFeatureQC(altExp(filtered_sce, alt))
 }
 
+# remove any genes with zero counts 
+detected_genes <- rowData(filtered_sce)$mean > 0 & 
+  rowData(filtered_sce)$detected > 0
+
+filtered_sce <- filtered_sce[detected_genes,]
+
 # write filtered sce to output
 readr::write_rds(filtered_sce, opt$filtered_file, compress = "gz")
 


jashapiro:
This code looks fine, but I meant to comment on this issue earlier... I think we should not filter the genes. The reason is that if people want to combine multiple experiments, the default behavior of most joining algorithms (including the one described first in [OSCA](http://bioconductor.org/books/3.13/OSCA.multisample/integrating-datasets.html#slower-setup) `batchelor::quickCorrect()`) is to use the intersection of the gene sets from the samples being joined. This can result in loss of data where a gene that is present in one sample but absent in another will be discarded, rather than filling in the sample with missing data as zeros. 

Removing the genes with zero counts may falsely suggest that those genes were not assayed in a particular sample, which is not the case. 

Finally, I suspect that removing these rows will result in only the smallest of reduction in object/download size, as the sparse matrix will only have stored the row (gene) names, but no data for those rows.

allyhawkins:
Okay I see where you are coming from and I think that makes sense. In testing this there also was minimal loss of genes so I would agree it's not really saving much room. I think we just want to make sure we specify in the docs that all genes will be included in the final count matrix, even those that may not have been detected in that particular sample. 

If we agree I'll just close out this PR and then close the issue as well.

PR #36, opened 2021-09-23
Implement use of lower=200 during generation of filtered sce

### LINKED ISSUE #30: Modify filtering to use lower=200 for emptyDrops
Based on testing of `emptyDrops` with various thresholds, we have decided to use `emptyDrops` with `lower=200` rather than the default `lower=100`. Currently we have a script that performs filtering using the `filter_counts` function in `scpcaTools`. That function allows us to pass any of the options that you can use with `emptyDrops` through as extra options. 

One option to do this would be to pass in `lower=200` as a parameter through the workflow and then make it an argument in the Rscript used for filtering. We could also set it within the R script without making it a parameter, but that might be less transparent. 

One thing to note is that the `emptyDropsCellRanger()` function from `DropletUtils` will be available on the next release of Bioconductor on October 22. In testing, that function showed slightly better consistency with Cell Ranger so we may want to consider re doing the filtering later on after the soft launch and modifying the filtering to use `emptyDropsCellRanger`.

allyhawkins:
Closes #30. This PR adds the ability to pass through an option for the `lower` parameter as part of `DropletUtils::emptyDrops()` which we call inside the `scpcaTools::filter_counts()`. To do this, I chose to add a parameter for lower to our workflow in the config file and then pass that parameter through to the process `filter_sce` rather than hard code the use of `lower=200` within the Rscript that generates the filtered sce. By doing this, we can directly alter this parameter in the config file if we ever need to rather than going into the R script that is run by the workflow and changing it, making it a bit more transparent. 

In doing this, I also had to remove the pipe between the `make_unfiltered_sce` and `filter_sce` processes so that I could pass through both the `make_unfiltered_sce.out` and `params.lower`, but I am open to other suggestions on how to do this.

### DIFF
--- a/modules/generate-rds.nf
+++ b/modules/generate-rds.nf
@@ -61,7 +61,7 @@ process filter_sce{
         filter_sce_rds.R \
           --unfiltered_file ${unfiltered_rds} \
           --filtered_file ${filtered_rds} \
-          --lower ${lower}
+          --lower ${params.emptydrops_lower}
         """
 }
 
--- a/modules/generate-rds.nf
+++ b/modules/generate-rds.nf
@@ -52,6 +52,7 @@ process filter_sce{
     publishDir "${params.outdir}/${meta.sample_id}"
     input: 
         tuple val(meta), path(unfiltered_rds)
+        val(lower)
     output:
         tuple val(meta), path(unfiltered_rds), path(filtered_rds)
     script:
@@ -59,7 +60,8 @@ process filter_sce{
         """
         filter_sce_rds.R \
           --unfiltered_file ${unfiltered_rds} \
-          --filtered_file ${filtered_rds}
+          --filtered_file ${filtered_rds} \
+          --lower ${lower}
         """
 }
 
@@ -67,8 +69,8 @@ workflow generate_sce {
   // generate rds files for RNA-only samples
   take: quant_channel
   main:
-    make_unfiltered_sce(quant_channel, params.mito_file) \
-      | filter_sce
+    make_unfiltered_sce(quant_channel, params.mito_file)
+    filter_sce(make_unfiltered_sce.out, params.lower)
 
   emit: filter_sce.out
   // a tuple of meta and the filtered and unfiltered rds files
@@ -79,8 +81,8 @@ workflow generate_merged_sce {
   // input is a channel with feature_meta, feature_quantdir, rna_meta, rna_quantdir
   take: feature_quant_channel
   main:
-    make_merged_unfiltered_sce(feature_quant_channel, params.mito_file) \
-      | filter_sce
+    make_merged_unfiltered_sce(feature_quant_channel, params.mito_file)
+    filter_sce(make_merged_unfiltered_sce.out, params.lower)
 
   emit: filter_sce.out
   // a tuple of meta and the filtered and unfiltered rds files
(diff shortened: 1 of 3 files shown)


nextflow.config
```diff
@@ -11,6 +11,7 @@ params{
   fasta         = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
   gtf           = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
   index_path    = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/salmon_index/Homo_sapiens.GRCh38.104.spliced_intron.txome'
+  lower         = 200
   mito_file     = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.mitogenes.txt'
   t2g_3col_path = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv'
```
jashapiro, line 16: Two parts to this suggestion; one is just the organization and an additional comment. The other is to make the parameter name more informative, as `lower` could refer to lots of things!
```suggestion
  mito_file     = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.mitogenes.txt'
  t2g_3col_path = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv'
 
  // Filtering constants
  emptydrops_lower = 200
```
```

File: modules/generate-rds.nf [hunk 2]
```diff
@@ -52,23 +52,25 @@ process filter_sce{
...
     output:
         tuple val(meta), path(unfiltered_rds), path(filtered_rds)
     script:
         filtered_rds = "${meta.library_id}_filtered.rds"
         """
         filter_sce_rds.R \
           --unfiltered_file ${unfiltered_rds} \
-          --filtered_file ${filtered_rds}
+          --filtered_file ${filtered_rds} \
+          --lower ${lower}
```
jashapiro, line 64: Since this parameter value is used only in this one step, I think it makes sense to just call out to it directly, rather than passing it in as an input.  

This change would allow us to keep the pipe, but  that was not really my motivation; I was more just thinking about keeping the inputs more consistent where we could. I guess I can imagine a place where we might use the lower bound as an input and filter an SCE by a different standard, but if we come to that we could reimplement this arrangement.
 
```suggestion
          --lower ${params.emptydrops_lower}
```

jashapiro (commented):
Looks good, with a few suggestions for changes to the workflow. One is mostly more clarity in naming, the other to help (re)streamline the pipeline. 

For the second, I am suggesting using the `param` directly in the process, which makes me wonder if we should doing that for `make_unfiltered_sce` and `make_merged_unfiltered_sce` as well with respect to the mitochondrial file. My inclination is to leave that the way it is, but I can't tell you why I feel differently. Oh, actually, I can! It is because that has to come in as a file path, not just a value, so nextflow requires it to be an input, or the file won't be there (to my understanding).

modules/generate-rds.nf
allyhawkins: I completely forgot you could do this and this makes so much more sense, thanks for catching this!

allyhawkins:
I added the pipe back in and removed passing in the parameter through the workflow since we are using it directly in the process now.

jashapiro (approved):
LGTM!

PR #37, opened 2021-09-24
Export cell statistics & other metadata

### LINKED ISSUE #34: Export cell statistics & other metadata in a table/json
We need to export cell counts for the front end, so it makes sense to have that output be part of the workflow. The most logical place for this seems to me to be with the QC report, so we should modify the [`sce_qc_report`](https://github.com/AlexsLemonade/scpca-nf/blob/main/modules/qc-report.nf) process (and `sce_qc_report.R`(https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/sce_qc_report.R) script)  to also export either a table or json file with all necessary stats for the front end (probably including project ID, library ID, cell count, (unfiltered cell count?), and maybe things like nextflow workflow version (we can get a lot  from [nextflow workflow introspection variables](https://www.nextflow.io/docs/latest/metadata.html)).

Whether we do json or a csv is just a matter of output, as we would build as a single row table in R.

jashapiro:
In this PR I am adding to the QC process/script to output information at the library level, including some basic statistics that we were thinking might go into the portal front end. 

Closes #34 
Also addresses https://github.com/AlexsLemonade/ScPCA-admin/issues/210

I am attaching here an example of the output csv file, [SCPCL000001_metadata.csv](https://github.com/AlexsLemonade/scpca-nf/files/7228171/SCPCL000001_metadata.csv).
The equivalent json (which is also exported, but gh won't let me attach), appears below.
```
{
  "library_id": "SCPCL000001",
  "sample_id": "SCPCS000001",
  "filtered_cells": 1642,
  "unfiltered_cells": 61980,
  "total_reads": 121894873,
  "mapped_reads": 90729577,
  "genome_assembly": "Homo_sapiens.GRCh38.104",
  "mapping_index": "Homo_sapiens.GRCh38.104.spliced_intron.txome",
  "transcript_type": "spliced",
  "has_citeseq": false,
  "workflow": "https://github.com/AlexsLemonade/scpca-nf",
  "workflow_version": "jashapiro/34-export-cell-stats",
  "workflow_commit": "839d5412b47064eb2953f841745e3074f4839d08"
}
```

Are there other statistics and information that we should include in these files? (Do we want both formats?) I'm thinking only the filtered cells are likely to really make it to the front end, but getting the other stats wasn't hard. I could add a processing date?

Worth noting: you only get the workflow version & commit if you start the workflow with:
```
nextflow run AlexsLemonade/scpca-nf -r <branch/tag/commit> 
```
If running from a local file, you will still get the workflow url, because I hard coded that as a fallback.
(I anticipate in production runs we would use a release tag for the version, so it would look like `"v1.0"`, but branches work, so that is what I have in the example here.)

### DIFF
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -4,6 +4,7 @@
 
 # import libraries
 library(optparse)
+suppressPackageStartupMessages(library(SingleCellExperiment))
 
 # set up arguments
 option_list <- list(
@@ -18,23 +19,58 @@ option_list <- list(
     help = "path to rds file with filtered sce object"
   ),
   make_option(
-    opt_str = c("-s", "--sample_id"),
+    opt_str = c("-l", "--library_id"),
     type = "character",
-    help = "Sample identifier for report"
+    help = "Library identifier for report"
   ),
   make_option(
-    opt_str = c("-o", "--output_file"),
+    opt_str = c("-s", "--sample_id"),
+    type = "character",
+    help = "Sample identifier for metadata file"
+  ),
+  make_option(
+    opt_str = c("-q", "--qc_report_file"),
     default = "qc_report.html",
     type = "character",
-    help = "path to output file"
+    help = "path to QC report output file"
+  ),
+  make_option(
+    opt_str = "--metadata_csv",
+    default = "metadata.csv",
+    type = "character",
+    help = "path to metadata csv output file"
+  ),
+  make_option(
+    opt_str = "--metadata_json",
+    default = "metadata.json",
+    type = "character",
+    help = "path to metadata json output file"
+  ),
+  make_option(
+    opt_str = "--genome_assembly",
+    type = "character",
+    default = NA,
+    help = "workflow github url"
+  ),
+  make_option(
+    opt_str = "--workflow_url",
+    type = "character",
+    default = NA,
+    help = "workflow github url"
+  ),
+  make_option(
+    opt_str = "--workflow_version",
+    type = "character",
+    default = NA,
+    help = "workflow version identifier"
   )
 )
 
 opt <- parse_args(OptionParser(option_list = option_list))
-if(is.null(opt$sample_id)){
-  stop("A `sample_id` is required.")
+if(is.null(opt$library_id)){
+  stop("A `library_id` is required.")
 }
-# check that input files exists
+# check that input files exist
 if(is.null(opt$unfiltered_sce) || !file.exists(opt$unfiltered_sce)){
   stop("Unfiltered .rds file missing or `unfiltered_sce` not specified.")
 }
@@ -48,8 +84,32 @@ unfiltered_sce <- readr::read_rds(opt$unfiltered_sce)
 filtered_sce <- readr::read_rds(opt$filtered_sce)
 
 scpcaTools::generate_qc_report(
-  sample_name = opt$sample_name,
+  sample_name = opt$sample_id,
   unfiltered_sce = unfiltered_sce,
   filtered_sce = filtered_sce,
-  output = opt$output_file
+  output = opt$qc_report_file
 )
+
+# Compile metadata for output files
+sce_meta <- metadata(unfiltered_sce)
+
+metadata_list <- list(
+  library_id = opt$library_id,
+  sample_id = opt$sample_id,
+  filtered_cells = ncol(filtered_sce),
+  unfiltered_cells = ncol(unfiltered_sce),
+  total_reads = sce_meta$total_reads,
+  mapped_reads = sce_meta$mapped_reads,
+  genome_assembly = opt$genome_assembly,
+  mapping_index = sce_meta$reference_index,
+  transcript_type = sce_meta$transcript_type,
+  workflow = opt$workflow_url,
+  workflow_version = opt$workflow_version
+) |>
+  purrr::map(~ifelse(is.null(.), NA, .)) # convert any NULLS to NA
+
+# Output metadata
+readr::write_csv(as.data.frame(metadata_list), file = opt$metadata_csv)
+jsonlite::write_json(metadata_list, path = opt$metadata_json)
+  
+
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -137,7 +137,7 @@ metadata_list <- list(
 jsonlite::write_json(metadata_list, path = opt$metadata_json, auto_unbox = TRUE)
   
 scpcaTools::generate_qc_report(
-  sample_id = metadata_list$library_id,
+  sample_name = metadata_list$library_id,
   unfiltered_sce = unfiltered_sce,
   filtered_sce = filtered_sce,
   output = opt$qc_report_file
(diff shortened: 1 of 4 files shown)


bin/sce_qc_report.R
```diff
@@ -17,39 +18,120 @@ option_list <- list(
...
+  library_id = opt$library_id,
+  sample_id = opt$sample_id,
+  date_processed = lubridate::now(),
+  filtered_cells = ncol(filtered_sce),
+  unfiltered_cells = ncol(unfiltered_sce),
+  total_reads = sce_meta$total_reads,
+  mapped_reads = sce_meta$mapped_reads,
+  genome_assembly = opt$genome_assembly,
+  mapping_index = sce_meta$reference_index,
+  transcript_type = sce_meta$transcript_type,
```
allyhawkins, line 119: Do we want to add in the salmon and alevin fry versions here? I feel like if we are already outputting this file with the index, assembly, and transcript information it would be helpful to include as much information that we have about the sample. It could come in handy and depending on if we include this file with the download or not, I think it could be useful for people to have.  
```suggestion
  transcript_type = sce_meta$transcript_type,
  salmon_version = sce_meta$salmon_version,
  alevin_fry_version = sce_meta$alevinfry_version,
```

allyhawkins (commented):
Generally I think the changes here look good, I just have a question about the overall use of these files. I know we are grabbing the number of cells to output to a file that can then be used for the front end. But is there any other reason we may want this file? To me it seems like most of these items can be portrayed in the QC report so we wouldn't need to print them into a separate file, unless we think its useful for people to have this information in a different format outside of the QC report? (I can see an argument for this too.) Either way, I don't think that we need both the csv and the json files and that if the purpose of this is to only provide information for the front end, I'm going to make an assumption that they might prefer a json and that we don't need the csv. If the idea is that this file would be included in download, then I think we could make one single csv file, because people are scared of jsons, and make sure that the csv would work for the dev team.

jashapiro:
> Generally I think the changes here look good, I just have a question about the overall use of these files. I know we are grabbing the number of cells to output to a file that can then be used for the front end. But is there any other reason we may want this file? To me it seems like most of these items can be portrayed in the QC report so we wouldn't need to print them into a separate file, unless we think its useful for people to have this information in a different format outside of the QC report? (I can see an argument for this too.)

I thought it might be generally useful to have a machine-readable version of some of the information, but mostly I would plan on people looking at the QC file or working with the SCE objects (which have much of the data already). I think I would lean toward keeping just the json file as well if only keeping one, as a single row CSV seems kind of gross to me, but it was so easy to have both... 

I also realized that I did not explicitly reference this earlier question to @kurtwheeler: https://github.com/AlexsLemonade/ScPCA-admin/issues/206#issuecomment-911593042, which this PR may address: Does this kind of JSON file fulfill the per-libary info role that was described there? Or would you prefer a separate compiled table of libraries for each project? (Note: we would still provide a table of samples) Is there additional info in the files here that you would want us to include?

kurtwheeler:
The libraries metadata table currently has 4 fields, two of which are ids: `scpca_sample_id,scpca_library_id,seq_unit,technology`.   I don't think I see anything equivalent to `seq_unit` or `technology` in this JSON. There's a lot of fields we don't currently have in our data model. I think it'll probably require some discussion to decide which additional ones we want to add.

allyhawkins:
> I don't think I see anything equivalent to `seq_unit` or `technology` in this JSON. There's a lot of fields we don't currently have in our data model. I think it'll probably require some discussion to decide which additional ones we want to add.

We can easily add `seq_unit` and `technology` here so that you have everything you need for the libraries table but in this json format instead? I think the only other information that would go to the front end from this table would be the `filtered_cells`. Everything else is just extra information but wouldn't need to be user-facing.

kurtwheeler:
Oh right, I forgot we were adding that field to that table. Yeah it looks like that's the only one we don't already have, so it's the only one we'll need from this. But yeah, JSON is as easy to parse as CSV for me.

jashapiro:
I think we could parse the per-library files back to a single file per project as part of the workflow if we wanted to. It would require a bit of thinking though, and I can see a few ways that it might go wrong if we aren't careful. (Incremental updates & reruns are not necessarily well supported by nextflow). I feel better about the per-library file being easier to keep updated, with a script outside the workflow doing any combining. 

I think my thought with leaving it as a separate file per library was that @kurtwheeler would probably be parsing through the workflow output files to some degree no matter what. 

I also can add all of the info from the library table to the JSON files; that is pretty trivial. A question for later discussion is if you would rather just work from a set of JSON files, one per library, with _all_ information you require at the library level, or two kinds of files, one file per project with pre-analysis library info as a table, plus one file per library with post-analysis info (cell counts, basically).

So the general structure would currently be something like this (I will call this **option A**):
```
- Project1
  - Project1_sample_info.csv
  - Project1_library_info.csv (no cell counts)
  - Lib 1
    - Lib1_unfiltered.rds
    - Lib1_filtered.rds
    - Lib1_QC.html
    - Lib1_info.json (or .csv)
 -  Lib 2
    - Lib2_unfiltered.rds
    - Lib2_filtered.rds
    - Lib2_QC.html
    - Lib2_info.json (or .csv)
```

**Option B**: It is really easy to put all the info from `Project1_library_info.csv` into each of the `LibX_info.json` files, which could make `Project1_library_info.csv` redundant. This would require parsing the individual library files into a table later as part of data ingestion.

**Option C**: Putting all the info from `LibX_info.json` into `Project1_library_info.csv` is a bit harder within the workflow, as I discussed (read: subject to unexpected nextflow quirks), but it should be doable.  This would make the individual `.json` files redundant (and we would probably remove them).

So, should we stick with status quo, Option A, with potentially slightly different data choices in the json, but with (partially) redundant files, Or got to option B or C, with less redundancy, but different divisions of labor.

jashapiro:
In discussions, we decided go with option B above, which this PR implements. For posterity (and @kurtwheeler's reference) here are the current JSON file contents:
```
{
  "library_id": "SCPCL000001",
  "sample_id": "SCPCS000001",
  "technology": "10Xv3",
  "seq_unit": "cell",
  "has_citeseq": false,
  "filtered_cells": 1890,
  "unfiltered_cells": 61980,
  "total_reads": 121894873,
  "mapped_reads": 90729577,
  "genome_assembly": "Homo_sapiens.GRCh38.104",
  "mapping_index": "Homo_sapiens.GRCh38.104.spliced_intron.txome",
  "transcript_type": "spliced",
  "date_processed": "2021-10-01T18:53:46+0000",
  "salmon_version": "1.5.2",
  "alevin_fry_version": "0.4.1",
  "workflow": "https://github.com/AlexsLemonade/scpca-nf",
  "workflow_version": "v0.1",
  "workflow_commit": "0cab3346d85ac1846c14a383f2c35f861afb61d7"
}
```

jashapiro:
For future use, I also just added the `miQC` model to the filtered_sce metadata. Kind of unrelated to this PR, but very minor.

allyhawkins (approved):
This looks good to me!

PR #42, opened 2021-09-27
Add gene_symbols to SCE objects

### LINKED ISSUE #38: Add GTF input to sce file generation
When scpcaTools supports adding gene symbols to the output files(https://github.com/AlexsLemonade/scpcaTools/issues/35), it will need a GTF file as input. 

This issue is to track adding that support/input to the workflow and associated scripts, reading the GTF file location from the workflow parameters.

jashapiro:
Closes #38

This update to the workflow adds gene symbols to the row data of the single cell experiment objects that we generate. I do this first as part of the unfiltered_sce script, using the symbol information from the GTF file that is already a parameter defined in the workflow config file. I read the GTF with `rtracklayer::import()` which is pretty efficient once we restrict to reading only genes (not transcripts, cds, etc.).

The filtering script had to be modified a bit, because we no longer want to remove all row data before recalculating gene statistics. Instead, we are removing only the columns of the rowdata table that `addPerFeatureQCMetrics()`adds, then add them back recalculated. There is a bit of oddness here in that we select from the `DataFrame` as if it were a vector (brackets with no comma), as this allows us to keep the column names, which can otherwise be dropped if we are only keeping one column. (I made the same change for the `alt` experiment, even though it should not be affected).

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -101,6 +101,7 @@ if (opt$feature_dir != ""){
 # add per cell and per gene statistics to colData and rowData
 unfiltered_sce <- unfiltered_sce |>
   add_cell_mito_qc(mito = mito_genes) |>
+ # add gene symbols to rowData
   add_gene_symbols(gene_info = gtf) |>
   scuttle::addPerFeatureQCMetrics()
 
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -76,7 +76,7 @@ if(!file.exists(opt$gtf_file)){
 mito_genes <- unique(scan(opt$mito_file, what = "character"))
 
 # read in gtf file (genes only for speed)
-gtf <- rtracklayer::import(opt$gtf_file, feature_type = "gene")
+gtf <- rtracklayer::import(opt$gtf_file, feature.type = "gene")
 
 # convert seq_unit to spliced or unspliced to determine which types of transcripts to include in final counts matrix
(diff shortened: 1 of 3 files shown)


bin/generate_unfiltered_sce.R
```diff
@@ -81,13 +95,14 @@ if (opt$feature_dir != ""){
...
   # add alt experiment features stats
-  altExp(unfiltered_sce, opt$feature_name) <- scater::addPerFeatureQC(altExp(unfiltered_sce, opt$feature_name))
+  altExp(unfiltered_sce, opt$feature_name) <- scuttle::addPerFeatureQCMetrics(altExp(unfiltered_sce, opt$feature_name))
 }
 
 # add per cell and per gene statistics to colData and rowData
 unfiltered_sce <- unfiltered_sce |>
   add_cell_mito_qc(mito = mito_genes) |>
-  scater::addPerFeatureQC()
+  add_gene_symbols(gene_info = gtf) |>
```
allyhawkins, line 104:
```suggestion
 # add gene symbols to rowData
  add_gene_symbols(gene_info = gtf) |>
```
Just adding in a small clarifying comment here to remind us (mostly for myself) where the gene symbols are going and point out when they are being added.

allyhawkins (approved):
I like what you did here and I think it works well. I just added in a comment in one place but other than that looks good!

bin/generate_unfiltered_sce.R
jashapiro: I'll do it, but I tried to make the function name clear!

Issue #41, opened 2021-09-27
Allow explicit restarting from RAD files (or other points?)


jashapiro:
While the `-resume` option in nextflow is great for when a workflow is repeated shortly after running and/or failures, most intermediate files are not stored long-term, so a later rerun with changes in parameters or other code could well result in rerunning the full workflow. 

Since we are storing intermediate files (in particular RAD files), it might be nice to have an option in the workflow that would skip mapping (the most time-consuming step!) if the RAD files are available (after #40 is implemented).

I am not exactly sure what the implementation of this would look like at this stage... We will have to do some investigation of the best way to test for the presence of an alevin output directory and run a different section of the workflow in that case.

PR #44, opened 2021-09-29
Incorporate project info into channels and output

### LINKED ISSUE #17: Change publishDir for alevin output to be different from publishDir for rds files

### LINKED ISSUE #26: Incorporate scpca_project_id into workflow metadata
We are introducing another identifier for the project level in https://github.com/AlexsLemonade/ScPCA-admin/issues/213. This `project_id` should be incorporated into the `meta` object, and probably as the top level of the file structure: i.e.

We can also update #24 to use the project ID as an alternative to submitter ID for running by project

```
SCPCP00001
└── SCPCS00001
   ├── SCPCL00001_unfiltered.rds
   └── SCPCL00001_filtered.rds
```

### LINKED ISSUE #40: Store RAD files & alevin folders after initial mapping
With the knowledge that RAD files do not contain any trace of sequence info, we will likely want to save those as well or instead of the alevin output files that are discussed in #17.

In #17 we are discussing saving the alevin output folder _after_  quantification, but the steps between RAD file generation by `salmon alevin -rad` and the RDS file are quite efficient; we are unlikely to need those intermediates, but we might want the first (uncollated) RAD file in case of future processing updates.

To do this, we would want to add publishDir directives to the [`alevin_rad`](https://github.com/AlexsLemonade/scpca-nf/blob/2a6a35249ab3576e36100fe26ed885b206f57e44/modules/af-rna.nf#L4) and [`alevin_feature`](https://github.com/AlexsLemonade/scpca-nf/blob/2a6a35249ab3576e36100fe26ed885b206f57e44/modules/af-features.nf#L23) processes.

jashapiro:
Here I am incorporating the `scpca_project_id` column into the `meta` passed along with each channel, and using that information to further subdivide the final output. 

In changing the output files, I also split outputs like alevin results & initial rad files (which we will now keep) into a separate `internal` prefix, with the elements that will go to the front end under the `publish` prefix. Some notes:
- This is slightly different from the approach in https://github.com/AlexsLemonade/scpca-nf/pull/28, which created separate `alevin_dir` and `final_output` parameters, and used those for output. 
- In doing this, I also decided to remove the `sample_id` prefix for the internal outputs, as it seemed like for internal things we might not need that level of nesting (which is usually redundant). 
- the internal saved files are separated into `rad` and `af` outputs; the former has the initial mapping results (in order to satisfy #40), and the latter the final alevin-fry folder, with the`.rad` files removed (after filter, collate and quant). 
  - For the `feature` processes, this resulted in a redundant `feature_index` folder being saved for every library due to the way we were passing those files around. It was small, but annoying, so I simplified things a bit by moving the only file we need in the next step to the alevin output directory. 
- I moved more parameters out to the config file, for consolidation, leaving only the choice of runs/projects in `main.nf`

**All of these decisions are open for debate!**

closes #17 
closes #26 
closes #40

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -58,6 +58,7 @@ workflow {
       run_id: it.scpca_run_id,
       library_id: it.scpca_library_id,
       sample_id: it.scpca_sample_id,
+      project_id: it.scpca_project_id?: "SCPCP000000",
       submitter: it.submitter,
       technology: it.technology,
       seq_unit: it.seq_unit,
@@ -68,11 +69,13 @@ workflow {
     // only technologies we know how to process
     .filter{it.technology in tech_list} 
     // use only the rows in the run_id list (run, library, or sample can match)
+    // or run by project or submitter if the project parameter is set
     .filter{run_all 
              || (it.run_id in run_ids) 
              || (it.library_id in run_ids)
              || (it.sample_id in run_ids)
              || (it.submitter == params.project)
+             || (it.project_id == params.project)
             }
   
   // generate lists of library ids for feature libraries & RNA-only
--- a/main.nf
+++ b/main.nf
@@ -1,14 +1,6 @@
 #!/usr/bin/env nextflow
 nextflow.enable.dsl=2
 
-// run parameters
-params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
-params.outdir = "s3://nextflow-ccdl-results/scpca/processed"
-
-params.resolution = 'cr-like' //default resolution is cr-like, can also use full, cr-like-em, parsimony, and trivial
-params.barcode_dir = 's3://nextflow-ccdl-data/reference/10X/barcodes' 
-
-
 // run_ids are comma separated list to be parsed into 
 // a list of run ids, library ids, and or sample_ids
 // or "All" to process all samples in the metadata file
@@ -58,7 +50,7 @@ workflow {
       run_id: it.scpca_run_id,
       library_id: it.scpca_library_id,
       sample_id: it.scpca_sample_id,
-      project_id: it.scpca_project_id?: "SCPCP000000",
+      project_id: it.scpca_project_id?: "no_project",
       submitter: it.submitter,
       technology: it.technology,
       seq_unit: it.seq_unit,
--- a/modules/af-features.nf
+++ b/modules/af-features.nf
@@ -31,7 +31,7 @@ process alevin_feature{
           path(feature_index)
   output:
     tuple val(meta),
-          path(run_dir), path(feature_index)
+          path(run_dir)
   script:
     // label the run directory by id
     run_dir = "${meta.run_id}-features"
@@ -53,7 +53,9 @@ process alevin_feature{
       --umi-geometry ${umi_geom} \
       --rad \
       -o ${run_dir} \
-      -p ${task.cpus} 
+      -p ${task.cpus}
+
+    cp ${feature_index}/t2g.tsv ${run_dir}/t2g.tsv
     """
 }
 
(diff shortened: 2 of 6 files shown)


allyhawkins (commented):
I just have a few follow up questions based on the changes that you made: 
1. Do we want to start building in the structure that Kurt has outlined and can be found in [the Google doc meeting notes](https://docs.google.com/document/d/1hdYOXWWPOldnyvHDn_wjLw72g09BTs5xqOAsf6bRUzs/edit) that has the `.rds` and qc report nested in a `files` directory? 
2. Do we still need to keep the `.rad` files? I thought we had decided on keeping them so that we could restart in the future? If so we should remove the statement that removes those files here.
3. I'm not sure I'm 100% following why we need to separate the RAD and af files the way that you did? I think we would still be able to pick up the run after running `salmon alevin -rad` if the files were stored together in one folder, but perhaps I'm overlooking something.

jashapiro:
> I just have a few follow up questions based on the changes that you made:
> 
> 1. Do we want to start building in the structure that Kurt has outlined and can be found in [the Google doc meeting notes](https://docs.google.com/document/d/1hdYOXWWPOldnyvHDn_wjLw72g09BTs5xqOAsf6bRUzs/edit) that has the `.rds` and qc report nested in a `files` directory?

I hadn't seen this latest version! Yes, we should match that, but I think we will need wait for discussion tomorrow, because I am not sure that is what we will end up with. 

> 2. Do we still need to keep the `.rad` files? I thought we had decided on keeping them so that we could restart in the future? If so we should remove the statement that removes those files here.
> 3. I'm not sure I'm 100% following why we need to separate the RAD and af files the way that you did? I think we would still be able to pick up the run after running `salmon alevin -rad` if the files were stored together in one folder, but perhaps I'm overlooking something.

The reasons for 2 & 3 is that we don't really want to pass the RAD files on to the rest of the pipeline. Moving large files is slow, & storing them redundantly is expensive, so we improve efficiency by removing them at the alevin-fry step (otherwise they would be duplicated in storage). If we kept them all the in the same folder, we would be able to restart, yes, but the folder wouldn't be "clean", which might make detecting changes/errors more trouble. 

If anything, we don't need to keep the alevin folder, as that is quick to recreate, but I thought keeping it for now was fine.

allyhawkins:
> I hadn't seen this latest version! Yes, we should match that, but I think we will need wait for discussion tomorrow, because I am not sure that is what we will end up with.

Okay I'll wait until decisions are made post that discussion to approve this then. 

> The reasons for 2 & 3 is that we don't really want to pass the RAD files on to the rest of the pipeline. Moving large files is slow, & storing them redundantly is expensive, so we improve efficiency by removing them at the alevin-fry step (otherwise they would be duplicated in storage). If we kept them all the in the same folder, we would be able to restart, yes, but the folder wouldn't be "clean", which might make detecting changes/errors more trouble.

The moving large files around part definitely makes sense so I understand the decisions made on that part. In terms of keeping them in the same folder and making sure we are able to restart, I am fine with the separation as long as we know that with using `-resume` we will still be able to run later steps without having to re-run alignment. I think that was our discussion the other day in the scpca meeting (and the general reason behind making sure we publish the rad output folders). I believe as long as we publish them then we should be able to restart since nextflow will know where to look for them and it shouldn't matter that they are separate?

jashapiro:
> I believe as long as we publish them then we should be able to restart since nextflow will know where to look for them and it shouldn't matter that they are separate?

Actually, no. The `-resume` flag will only use the cached files, not published ones. We will have to manually add workflow changes to support starting from mapped files if they exist, which is why that is in as a separate issue (#41). But if we are storing the files, we can work on that later (post-launch).

jaclyn-taroni (approved):
@jashapiro and I talked through this and I think it can get merged.

PR #49, opened 2021-10-06
Add technology to stored metadata

jashapiro:
While we added the ability to store metadata in the SCE objects a while back, we never implemented passing that data in via the workflow. This PR adds  the flags and arguments to do that.


allyhawkins (approved):
LGTM, thanks for catching this!

PR #50, opened 2021-10-06
Add random seed options

jashapiro:
This PR adds the option to set a random seed for the workflow, in particular the `filter_sce_rds.R` script. 

I set a default parameter value of `0` for the workflow, which will cause the workflow to skip the `random_seed` argument to the filter script, but any other value will be passed along to set the seed at the top of the script.

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -24,10 +24,18 @@ option_list <- list(
     opt_str = c("-l", "--lower"),
     type = "integer",
     help = "Value specifying the lower bound on total UMI count used in filtering with DropletUtils::emptyDrops."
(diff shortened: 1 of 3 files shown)


allyhawkins:
I'm curious as to the rationale for setting it up this way? I was envisioning this as setting a default seed, i.e. 2021 for either the argument used in `filter_sce_rds.R` or as the parameter in nextflow. But here it looks like there isn't a default and everytime we run it if we want to set the seed we would have to input our own seed? I'm just wondering if there is a particular reason behind setting it up like that vs. having a default in the workflow.

allyhawkins (approved):
This change looks good to me!

PR #51, opened 2021-10-07
Use v0.1.0 of scpcaTools & memory/retry tweaks

### LINKED ISSUE #47: Create tagged release 0.1.0 of scpca-nf
When we are ready to kick off production jobs with this workflow, we should tag a release. 

This will allow us to reproducibly run the workflow with the following command or something similar:

```
nextflow run AlexsLemonade/scpca-nf -r v0.1.0 -profile batch --project SCPCP000001
```

(this command would run all samples for the `00001` project, on AWS)

jashapiro:
Closes #47 

This updates the tag on the docker image to use v0.1.0 of scpcaTools, freezing that last image in place (for now).

In testing a separate step, I ran into some out of memory errors, so I added another process label with higher memory requirements (but still only one CPU, though I might later reconsider this to bump the CPU a bit too). I also tweaked the nextflow retry settings because those seemed to not be working as expected (i.e. I was clearly getting error 137 but the job was not being retried). With the modifications I made and the test sample I was working with, I still get an OOM error for one of the jobs (filtering), but this then correctly reruns with double the memory.

### DIFF
--- a/modules/generate-rds.nf
+++ b/modules/generate-rds.nf
@@ -1,9 +1,9 @@
-
 // generate unfiltered and filtered RDS files using scpcaTools
 
 // RNA only libraries
 process make_unfiltered_sce{
     container params.SCPCATOOLS_CONTAINER
+    label 'mem_8'
     publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
     input: 
         tuple val(meta), path(alevin_dir)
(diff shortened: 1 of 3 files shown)


jashapiro:
> I hope that for these steps this is enough memory.

I hope so too! I have some ideas that might allow us to reduce memory usage though if we start to run into frequent errors.

PR #53, opened 2021-10-08
Fix miQC failures

### LINKED ISSUE #52: filter_sce_rds fails when miQC fails to fit
When running sample SCPCL000018, we encountered the following error:
```
Error executing process > 'generate_sce:filter_sce (17)'

Caused by:
  Process `generate_sce:filter_sce (17)` terminated with an error exit status (1)

Command executed:

  filter_sce_rds.R           --unfiltered_file SCPCL000018_unfiltered.rds           --filtered_file SCPCL000018_filtered.rds           --lower 200           --random_seed 2021

Command exit status:
  1

Command output:
  (empty)

Command error:
  Warning message:
  In miQC::mixtureModel(filtered_sce) :
    Unable to identify two distributions. Use plotMetrics function
                  to confirm assumptions of miQC are met.
  Error in unlist(x[[m]]@parameters) :
    trying to get slot "parameters" from an object of a basic class ("NULL") with no slots
  Calls: <Anonymous> ... sapply -> lapply -> FUN -> sapply -> lapply -> FUN -> unlist
  Execution halted
```

This seems to be due to a failure to fit the model, which produced a warning at first, but then an error at the following line: 

https://github.com/AlexsLemonade/scpca-nf/blob/0ec92f5a7523bb39a8035c10b0be02c39eeee02f/bin/filter_sce_rds.R#L67

To address this, we should look at this sample (the unfiltered RDs is at `s3://nextflow-ccdl-results/scpca-prod/publish/SCPCP000001/SCPCS000018`), but we will likely need to update the filtering script to fail gracefully for future cases when the miQC model fails to fit. Since such failures can occur stochastically, in my experience, we might first try a second attempt at the fit (using the next seed), but if that fails we would presumably want to fill in the `probability_compromised` column with `NA_real_`.

jashapiro:
This PR applies the new `add_QC()` function from `scpcaTools` implemented in https://github.com/AlexsLemonade/scpcaTools/pull/67.

Filing as a draft until the next release of scpcaTools & the docker image is built.

Closes #52

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -65,7 +65,7 @@ filtered_sce <- filtered_sce |>
   scuttle::addPerFeatureQCMetrics()
 
 # add prob_compromised to colData and miQC model to metadata
-filtered_sce <- scpcaTools::add_miQC()
+filtered_sce <- scpcaTools::add_miQC(filtered_sce)
 
 # grab names of altExp, if any
 alt_names <- altExpNames(filtered_sce)


bin/filter_sce_rds.R
```diff
@@ -62,10 +64,8 @@ rowData(filtered_sce) <- rowData(filtered_sce)[!drop_cols]
 filtered_sce <- filtered_sce |>
   scuttle::addPerFeatureQCMetrics()
 
-# add prob_compromised to colData from miQC::mixtureModel and add model to metadata
-model <- miQC::mixtureModel(filtered_sce)
-filtered_sce <- miQC::filterCells(filtered_sce, model, posterior_cutoff = 1, verbose = FALSE)
-metadata(filtered_sce)$miQC_model <- model
+# add prob_compromised and miQC model to metadata
+filtered_sce <- add_miQC()
```
allyhawkins, line 68:
```suggestion
# add prob_compromised to colData and miQC model to metadata
filtered_sce <- add_miQC()
```

allyhawkins (commented):
Just added a small comment change and then we may need to change the release number after the discussion about changing the DESCRIPTION for `scpcaTools`.

nextflow.config
```diff
@@ -6,7 +6,7 @@ params{
   // Docker container images
   SALMON_CONTAINER = 'quay.io/biocontainers/salmon:1.5.2--h84f40af_0'
   ALEVINFRY_CONTAINER = 'quay.io/biocontainers/alevin-fry:0.4.1--h7d875b9_0'
-  SCPCATOOLS_CONTAINER = 'ghcr.io/alexslemonade/scpca-tools:v0.1.0'
+  SCPCATOOLS_CONTAINER = 'ghcr.io/alexslemonade/scpca-tools:v0.1.1'
```
allyhawkins, line 9: Wouldn't we need to change this tag for the docker image to be v0.1.2 if the release was updated to be v0.1.2?

jashapiro: Oh, yes. But I wasn't planning to make that change now. I don't think it is worth the effort at this stage. I am sure we will be bumping up that up to 0.1.2 soon enough!

PR #54, opened 2021-10-12
Reduce memory requirements & change resolution strategy

jashapiro:
In running the first set of samples, I noticed that we were using quite a bit less memory during mapping than we were requesting. So here I am making the memory request for those jobs smaller. Now that we have a working retry strategy, this should still work even in the rare event we exceed the lower limit. (This may also allow us to recover from some disk space limits we had run into, as the second run will probably require a larger portion of each instance, leaving more disk space available on shared instances.)

I also here am updating to use cr-like-em for processing, which we said we would do, but didn't actually implement.

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -27,7 +27,7 @@ params{
   barcode_dir   = 's3://nextflow-ccdl-data/reference/10X/barcodes' 
  
   // Processing options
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
Curious to see how many times it retries now with this memory, but LGTM.

PR #58, opened 2021-10-20
Add module for bulk RNA-seq to workflow

allyhawkins:
In this PR, I started adding the ability to process bulk RNA sequencing samples to the main workflow. To do this, I first modified the list of allowed tech lists to include the bulk technologies, i.e. single_end and paired_end.

I then created a separate module for the bulk processing that includes running fastp and then salmon. It takes as input the same metadata map that the other modules use and will create a tuple of the metadata adn then the R1 and R2 files. That is then used as input to the fastp process which will output the paths to the trimmed fastq files as a tuple.

Then I am use the trimmed fastq files as input for salmon in a separate process. I was testing this and that's when I had the realization that I needed to use a different index for bulk samples then we are using for the single cell samples, since we don't want the splici index here. So I am filing this as a draft for now and will file the PR with the index as a separate PR stacked on this one.

If people do happen to look at this, the main question that I have is about the options that I chose to use for fastp and salmon and if those are what we would like?

Additionally, is there an alternative approach to dealing with processing both single_end and paired_end samples at the same time? Here, I am passing the path to the folder where the trimmed files live rather than the individual paths to the trimmed R1/R2 files so that it can work with either single-end or paired-end in the same process. I went back and forth on whether or not to separate them somehow, but thought that this would be a good way to approach it.

### DIFF
--- a/modules/bulk-salmon.nf
+++ b/modules/bulk-salmon.nf
@@ -1,0 +1,70 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+process fastp{
+    container params.FASTP_CONTAINER
+    label 'cpus_8'
+    tag "${meta.run_id}-bulk"
+    publishDir "${params.outdir}/internal/fastp/${meta.library_id}"
+    input: 
+        tuple val(meta),path(read1),path(read2)
+    output: 
+        tuple val(meta),path(trimmed_read1),path(trimmed_read2)
+    script: 
+        trimmed_data = "${params.outdir}/data/trimmed"
+        fastp_reports = "${params.outdir}/reports/fastp"
+        """
+        mkdir -p data/trimmed
+        mkdir -p reports/fastp
+
+        fastp --in1 {input.r1} --in2 {input.r2} /
+        --out1 "${trimmed_data}/${id}_R1_001_fastq.gz" /
+        --out2 "${trimmed_data}/${id}_R2_001_fastq.gz" /
+        --html "${fastp_reports}/${id}_fastp.html" /
+        --json "${fastp_reports}/${id}_fastp.json" /
+        --trim_poly_g /
+        --report_title '${id} report'
+        """
+
+}
+
+process salmon{
+    container 'quay.io/biocontainers/salmon:1.4.0--hf69c8f4_0'
+    label 'cpus_8'
+    tag "${id}-${index}"
+    publishDir "${params.outdir}"
+    input: 
+        tuple val(meta),path(trimmed_read1),path(trimmed_read2)
+    output: 
+        tuple val(meta),path()
+    script:
+        salmon_results = "${params.outdir}/salmon-quant"
+        """
+        salmon quant -i ${index} /
+        -l A /
+        -1 "${trimmed_data}/${id}_R1_001_fastq.gz" /
+        -2 "${trimmed_data}/${id}_R2_001_fastq.gz" /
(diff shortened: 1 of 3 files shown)


modules/bulk-salmon.nf
```diff
@@ -0,0 +1,70 @@
...
+}
+
+workflow bulk_quant_rna {
+    take: bulk_channel 
+    // a channel with a map of metadata for each rna library to process
+    main: 
+    // create tuple of (metadata map, [Read 1 files], [Read 2 files])
+        bulk_reads_ch = bulk_channel
+          .map{meta -> tuple(meta,
+                             file("s3://${meta.s3_prefix}/*_R1_*.fastq.gz"),
```
jashapiro, line 63: Are there any cases where `read1` will be more than one file? Does `fastp` handle that properly?

modules/bulk-salmon.nf
```diff
@@ -0,0 +1,70 @@
...
+    output: 
+        tuple val(meta),path(trimmed_reads)
+    script: 
+        trimmed_reads = "${meta.library_id}"
+        """
+        mkdir -p ${meta.library_id}
+
+        fastp --in1 ${read1} \
+        ${meta.technology == 'paired_end' ? "--in2 ${read2}":""} \
+        --out1 ${trimmed_reads}/${meta.library_id}-trimmed-R1.fastq.gz \
```
jashapiro, line 20: Do we need the library_id in the output here? I don't think it much matters, since the directory is named, but uf we don't need it we could simplify this a bit.

nextflow.config
```diff
@@ -22,6 +23,7 @@ params{
   fasta         = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
   gtf           = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
```
jashapiro, line 24: Thinking we should give these more descriptive names (this would involve changing these throughout the project)... maybe a separate PR in a bit.
```suggestion
  ref_fasta         = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
  ref_gtf           = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
```

jashapiro (commented):
This looks like a good start, and most of the functionality is there.  I left a number of little comments for formatting things, but the big thoughts are really about inputs and outputs:

- Are there cases where the input reads are split into multiple R1 and R2 files? If so, how does `fastp` handle that? 
- Should the output from the fastp process just be individual files? I would tend to say yes, but part of me is worried about the case where output files are blank. I also worry about the inputs being blank, in the case of single end reads. It may end up that we want to simply have separate processes for SE and PE, rather than trying to accommodate both. 
- If nextflow doesn't like passing blank files, that could cause trouble for SE samples anyway: maybe we should make the input to the fastp process just the directory with all the fastq files, and then do different things in the script based on `meta.technology`? This would basically move the `*_R1_*.fastq.gz` glob to the process rather than the workflow.
- Do we want to customize fastp behavior at all, or spell out any of the options, even if we are using defaults?

PR #61, opened 2021-11-02
Add process to create user facing salmon output files with tximport 

### LINKED ISSUE #60: Add to workflow generation of output files for bulk RNA-sequencing
Currently the bulk workflow stops at running `salmon` for each individual sample and writing the entire `salmon` folder to the `internal` directory that we don't intend to publish.

It was my interpretation based on discussions in Scpca meetings that we will want the bulk RNA-sequencing to be available as one matrix for each project with all samples for that project included in one file. This would mean that we will probably want to use `tximeta()` to import all of the samples that are processed for each project and output a `tsv` file with with all samples for a given project. This would be the only file we actually publish and include in the portal for each project with bulk RNA-sequencing. 

I think the approach to do this would be to add another module that runs an Rscript that imports all the samples and then outputs the `tsv` file.

allyhawkins:
Closes #60. This PR adds in the next steps for processing the bulk RNA-sequencing data and produce the output file that will be included for each project that contains bulk RNA-sequencing data. For each project, we only want one compiled output file that contains the raw counts for all samples with bulk RNA-seq in the project. To produce this file, I added in a process `group_tximport` that should run an Rscript, `merge-counts-tximport.R` to import all salmon `quant.sf` files for a given project using `tximport` and output the counts as a tsv file. 

I attempted to do this using a `groupTuple` and grouping the salmon output by the `project_id`. I then used the `project_id` and the path to all salmon outputs as the input to the process. Within the process, I am creating a temporary file with the list of all salmon directories and then using that as input to the Rscript, but I think there might be a better way to do this and I'm worried that this may be part of the errors that are currently being produced. 

Currently, I have been testing by running two samples that are within the same project and this will run and produce one output file that contains only one column, when it should produce only one output file with two columns. However the column name has `SCPCL2` rather than the correct library id. After producing that output file there is an error that it cannot get a property on a null `project_id`.
Screen Shot 2021-11-02 at 8 45 04 AM

One option I did think of was to only use only the `project_id` as input and nest the salmon output by `project_id` and then read in the library ID's based on what directories are present inside the project directory within the R script using `list.dirs`, but this seemed less transparent so I wasn't sure if we wanted to go that route. 

Putting this in a draft state right now and tagging @jashapiro just to get some general thoughts and ideas on how to improve this.


modules/bulk-salmon.nf
```diff
@@ -49,6 +49,27 @@ process salmon{
...
+    publishDir "${params.outdir}/publish/${meta.project_id}"
+    input:
+        tuple val(project_id), path('SCPCL*')
+        path(tx2gene)
+    output:
+        path(tximport_file)
+    script:
+        tximport_file = "${project_id}_bulk_quant.tsv"
+        """
+        ls -d SCPCL* > salmon_directories.tsv
```
jashapiro, line 63: This is one directory per line, no tabs, so let's just call it `.txt`
```suggestion
        ls -d SCPCL* > salmon_directories.txt
```

modules/bulk-salmon.nf
```diff
@@ -49,6 +49,27 @@ process salmon{
...
+
+        merge-counts-tximport.R \
+          --project_id ${project_id} \
+          --salmon_dir salmon_directories.tsv \
```
jashapiro, line 67:
```suggestion
          --salmon_dir salmon_directories.txt \
```

bin/merge-counts-tximport.R
```diff
@@ -0,0 +1,58 @@
...
+)
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# read in tx2gene
+tx2gene <- readr::read_tsv(opt$tx2gene, col_names = c("transcript_id", "gene_id"))
+
+# list of paths to salmon files 
+library_ids <- readr::read_tsv(opt$salmon_dir)
```
jashapiro, line 44: I think the problem may be that this read function is assuming that the first row is the column name. Since the `ls` output is simple, we can just read the files more simply to get a vector of all values.
```suggestion
library_ids <- readLines(opt$salmon_dir)
```

modules/bulk-salmon.nf
```diff
@@ -49,6 +49,27 @@ process salmon{
 
 }
 
+process group_tximport {
+    container params.SCPCATOOLS_CONTAINER
+    publishDir "${params.outdir}/publish/${meta.project_id}"
+    input:
+        tuple val(project_id), path('SCPCL*')
```
jashapiro, line 56: I know I suggested this option, but I think it would be better to not use `SCPCL` as the expected value here, just in case we get some other ids somehow that we want to use. I think a `*_salmonquant` or similar would be a better option (though you would have to parse off the ids later)

jashapiro (commented):
I had a quick thought on the error you were running into (and one other thought). If it works, that would be great, but I have done no testing, so I might be totally wrong. If I am, I will have to take a bit more time and thought...

modules/bulk-salmon.nf
```diff
@@ -49,6 +49,27 @@ process salmon{
 
 }
 
+process group_tximport {
+    container params.SCPCATOOLS_CONTAINER
+    publishDir "${params.outdir}/publish/${meta.project_id}"
```
jashapiro, line 54: I think this is the error you are getting with the null object: there is no `meta` passed in so it can't find `project_id` inside it.
```suggestion
    publishDir "${params.outdir}/publish/${project_id}"
```

allyhawkins:
Thanks for the suggestions @jashapiro! It looks like we actually can just use the path to the salmon directories that were output from the salmon process as input to the tximport process that way rather than trying something like `path(*quant.sf)` or `path(SCPCL*)`. When you do that apparently it populates the `*` with numbers rather than treating it like a regular wildcard and then when I use the `ls` command to write the paths to a file to read into the R script there were some wonky things happening. When I just passed the salmon output as input in everything worked as it should and the resulting file contained the columns with the counts and were labeled with the library ID's. 

My only remaining question here is whether or not we want to export the file as an RDS file, or keep it as is and take the gene x sample matrix as a tsv file.

jashapiro:
> Thanks for the suggestions @jashapiro! It looks like we actually can just use the path to the salmon directories that were output from the salmon process as input to the tximport process that way rather than trying something like `path(*quant.sf)` or `path(SCPCL*)`. When you do that apparently it populates the `*` with numbers rather than treating it like a regular wildcard and then when I use the `ls` command to write the paths to a file to read into the R script there were some wonky things happening. When I just passed the salmon output as input in everything worked as it should and the resulting file contained the columns with the counts and were labeled with the library ID's.
>

Oh, that is much simpler! The wildcard behavior is super strange, but we'll just set that aside and go with the fact that it works this way!

> My only remaining question here is whether or not we want to export the file as an RDS file, or keep it as is and take the gene x sample matrix as a tsv file.

I think we probably should stick with the tsv file. That is what we use for refine.bio, and it is the most accessible. Any precision loss should not matter when most values are integers, as they are for count data.

bin/merge_counts_tximport.R
```diff
@@ -0,0 +1,55 @@
...
+library_ids <- readLines(opt$salmon_dir)
+salmon_files <- file.path(library_ids, "quant.sf")
+names(salmon_files) <- library_ids
+
+# import using tximport
+txi_salmon <- tximport(salmon_files, type = "salmon", tx2gene = tx2gene)
+
+# write counts matrix to txt file
+txi_salmon$counts %>%
+  as.data.frame %>%
```
jashapiro, line 53: The parenthesis is not strictly required, but would be if we move to R 4.1 pipes, which I was using in other scripts in this repo, and might very slightly prefer here for consistency (& removing a dependency on magrittr). Absolutely not required though.  
```suggestion
  as.data.frame() %>%
```

jashapiro (approved):
This looks good! I'm really glad you figured out the set of paths issue. 

I have only very minor comments below.

Issue #62, opened 2021-11-29
Add module for spatial transcriptomics quantification


allyhawkins:
Because we have spatial transcriptomics libraries as part of scpca, we will need to allow for processing of these libraries and add in a module that quantifies spatial transcriptomic libraries to the workflow. 

To do this, we will for sure need a process that quantifies the libraries using spaceranger from 10X. We can follow a similar approach that is used in [`alsf-scpca/workflows`](https://github.com/AlexsLemonade/alsf-scpca/blob/ac309d80c6ee40a2049da82c6f4fda0d8ecf82db/workflows/cellranger-quant/run-cellranger.nf#L49-L75). 

To maintain consistencies with other libraries where we used Alevin-fry for quantification, we may also want to process these libraries through Alevin-fry. But we should start with adding in the spaceranger workflow and then add in processing using the Alevin-fry module as a next step if we choose to go that route based on results in AlexsLemonade/alsf-scpca#151.

allyhawkins:
Closing this since we have now added in the module for ST using Spaceranger and we don't anticipate adding in Alevin-fry quantification at this time.

PR #64, opened 2021-12-02
Add spaceranger module to workflow

allyhawkins:
Related to #62. This PR adds in a module for running Spaceranger to quantify spatial transcriptomics libraries to the main scpca workflow. To do this I added in a list of spatial techs into the main workflow and then on those that are spatial, they will be run through the separate module that is the spaceranger module. 

I set up the spaceranger module very similarly to how we had it in [`alsf-scpca/workflows`](https://github.com/AlexsLemonade/alsf-scpca/blob/69f167b4f8b7d34437efd832f1dd6c7d426960e6/workflows/cellranger-quant/run-cellranger.nf#L49-L128). I did also add the use of the bigdisk queue from AWS since spaceranger is greedy, but also didn't test it without it, so maybe we could test that to see if we can remove that label. 

For right now, I am publishing all of the output from this process in an internal folder labeled `spaceranger` and then my plan was to have the output from this process be read into R and converted to a `SpatialExperiment` and exported to RDS files as a separate process. The output of that process would then be published to the publish directory along with the single-cell and single-nuclei files that are currently getting published there. 

If we decide to run Alevin-fry on these samples, we can later add that in.


nextflow.config
```diff
@@ -24,6 +25,7 @@ params{
   gtf           = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
   index_path    = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/salmon_index/Homo_sapiens.GRCh38.104.spliced_intron.txome'
   bulk_index    = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/salmon_index/Homo_sapiens.GRCh38.104.spliced_cdna.txome'
+  spaceranger_index = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/cellranger_index/cdna'
```
jashapiro, line 28: One minor thing, one more major:

Should this be `ensembl104`? I don't know if you used that for the index or if we might need to make another index, but we should be on the same version for everything.

```suggestion
  spaceranger_index = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/cellranger_index/cdna'
```

Separately, we should probably file an issue to add this index creation to `build-index.nf` 

The more minor thing is just that things were so nicely aligned before, it would be nice to maintain that. It will mean changing more lines, but that is okay...

main.nf
```diff
@@ -84,6 +89,10 @@ workflow {
   bulk_ch = runs_ch.filter{it.technology in bulk_techs}
   bulk_quant_rna(bulk_ch)
 
+  // **** Process Spatial Transcriptomics data ****
+  spatial_ch = runs_ch.filter{it.technology in spatial_techs}
+  spaceranger_quant(spatial_ch)
+
   // **** Process RNA-seq data ****
```
jashapiro, line 96: Minor, but can we move this to the end (maybe before the QC reports, but after the single cell data? The order doesn't really matter too much, but it feels odd to put this in between processes that use salmon.

modules/spaceranger.nf
```diff
@@ -0,0 +1,69 @@
...
+  tag "${meta.run_id}-spatial" 
+  label 'cpus_12'
+  label 'bigdisk'
+  input:
+    tuple val(meta), path(fastq_dir), file(image_file)
+    path index
+  output:
+    tuple val(meta), path(output_id)
+  script:
+    output_id = "${meta.run_id}-spatial"
```
jashapiro, line 16: I don't think we want to output the whole folder here. In particular, this will include BAM files, which we will not want to retain long term (they contain individual data), so we should exclude at least those before output. As long as we are doing that, we may want to be a bit more discriminating about what we retain. I am not familiar with the full output folder, but I would assume we will want to keep the folder structure expected by the read10Xvisium() function, and perhaps other report files.

jashapiro (commented):
Thanks for adding this. In addition to my earlier (accidentally separate) comment, I had a coupe things to add here. 

The most important is that we do not want to retain the BAM files long term, so we need to be a bit more discriminating about the outputs from the `spaceranger` process. 

The other is just a bit of organization, which I think may be more important if we decide we are going to do more processing/ reformatting. I am not yet fully convinced that there will be value in separately quantifying these data with alevin-fry, but if we are going that route, we will have some more steps to add, and it makes sense to me to have that chunk of steps after the single cell steps.  We might use the same QC report step and expand that with options for spatial data, but I hadn't thought too much about that yet!

nextflow.config
allyhawkins: We need to make another index that uses ensembl 104 first before we change this, so I will add a separate issue that adds building the cellranger index to `build-index.nf` and then when we do that we can update this.

modules/spaceranger.nf
allyhawkins: So I think the only thing we need to do here is remove the BAM files, so I added in a step at the end of the process to do that. The other thing I thought of that we could do is specifically only write out the `outs` folder (after removing the BAM folder), but then we wouldn't maintain any of the logs, which I don't think we want but we may want for pulling version information in the QC report. I briefly looked through and the only thing that seemed like it might be worth keeping was the `_versions` file. I think we should be okay to just remove the BAM files unless you see other files here that should be removed? 
Screen Shot 2021-12-03 at 10 47 28 AM

jashapiro: Overall, it is hard to answer what parts need to be kept until we know what the next steps will be. I have some scattered thoughts below, but I think we may need more concrete plans before we are really ready to fully decide what we are keeping.

Removing the bam files was the most important. I don't _love_ all the other files being in there, but it may be better to mostly leave them. Or maybe just keep the `outs` folder (renamed, probably?) with the `_versions` file moved in (and maybe renamed to `spaceranger_versions.json`)

There is a fair amount of duplicate data in there:  h5 files and matrix folders are essentially duplicates. If we are not using both, we may not need to keep both.

I suspect the `SPATIAL_RNA_COUNTER_CS` directory is not needed, but had to know without looking in carefully. 

Do you know what the `.mri.gz` file is?

allyhawkins: After doing some more digging around about the output files, I think what we mainly need is the `outs` folder since that will contain the majority of the information. I also like the idea of copying over the `_versions` file and renaming it. 

Here is what I was able to find on the [contents of `SPATIAL_RNA_COUNTER_CS`](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/advanced/pipestance-structure). It appears to contain information about the individual processes that are run when you run `spaceranger_count`. I don't believe we need that and I'm not sure it will be particularly useful to us. 

The [`.mri.tgz` file](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/troubleshooting) is a file that contains information about the run that can be used for diagnostic purposes. I would say we should keep this as this is something that could hypothetically come in handy if we see errors in something and ever need to resolve them. 

> There is a fair amount of duplicate data in there: h5 files and matrix folders are essentially duplicates. If we are not using both, we may not need to keep both.

In regards to the duplicate data, I don't think we know for sure which way we will be reading in the data (Using `read10XVisium` alone appears to only need the HDF5 files, but if combining with Alevin-fry we would need the matrix files). I think for now I would like to keep both of them and then we can go back in later and remove what we don't want.

allyhawkins:
@jashapiro I went ahead and altered the workflow to remove the bam files and then copy over the `.mri.tgz` and `_versions` file to the `outs` folder before only publishing the `outs` folder. However, I'm unable to remove the bam files and I see that permission is denied. I know with the 10X outputs I've seen this before where I can't remove files in the `outs` folder, so I wonder if you have any suggestions here? I'm including a screenshot of the successfully completed process and then the error showing permission denied. 
Screen Shot 2021-12-09 at 10 11 21 AM

jashapiro:
If https://github.com/AlexsLemonade/scpca-nf/pull/64/commits/ff0a25d8c766f60eeea203f622eb31c46fdc9111 does not fix the problem, then probably adding `sudo` to the `rm` commands might work.

I actually doubt we will need the `mri.tgz` file for downstream workflow; it seems like it is only for 10X internal debugging, so we should know if we need it and can dive into the S3 work bucket to get it.

allyhawkins:
I was able to fix it without having to add `sudo` and then I also removed the copying of the `.mri.tgz` file so now it just writes the `outs` folder after removing the bam files and copying over the `_versions` as `spaceranger_versions.json`.

jashapiro (approved):
LGTM!

PR #66, opened 2021-12-08
Add process for building cellranger index to build-index.nf

### LINKED ISSUE #65: Include cellranger index in build-index.nf
Right now we only include the indices for Alevin-fry in `build-index.nf`, however, we will be using Spaceranger for spatial transcriptomics libraries and will need to build a separate index for that. In doing this we should make sure that the ensembl versions that we are building from is consistent across all of our indices that we are using. We should use a similar setup to how we were previously building the [cellranger index in `alsf-scpca`](https://github.com/AlexsLemonade/alsf-scpca/blob/main/workflows/rnaseq-ref-index/build-cellranger-index.nf).

allyhawkins:
Closes #65. This PR adds the `cellranger_index` process to the workflow for building the indices, `build-index.nf` ensuring that when we build all of our indices they will use the same assembly. I followed the same setup that was previously used in generating the cellranger index in [`alsf-scpca/workflows/rnaseq-ref-index/build-cellranger-index.nf`](https://github.com/AlexsLemonade/alsf-scpca/blob/544673eb008843548ac5c8663da65d4732d1f158/workflows/rnaseq-ref-index/build-cellranger-index.nf). 

This was pretty straight forward and I was able to test it and everything ran successfully, producing a new index with ensembl 104. The only question I had was about how we wanted to name this index. I included the assembly name that we use in the splici index and then appended `_cdna_cellranger`, but let me know if there is a different naming convention that would be preferred.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -83,8 +83,7 @@ process cellranger_index{
     gunzip -c ${gtf} > genome.gtf
     cellranger mkgtf \
       genome.gtf \
-      filtered.gtf \
-      --attribute=gene_biotype:protein_coding
+      filtered.gtf
     
     cellranger mkref \
       --genome=${cellranger_index} \
@@ -98,9 +97,9 @@ process cellranger_index{
 
 workflow {
   // generate splici and spliced cDNA reference fasta
-  generate_reference(params.gtf, params.fasta, params.assembly)
+  //generate_reference(params.gtf, params.fasta, params.assembly)
   // create index using reference fastas
-  salmon_index(generate_reference.out.fasta_files, params.fasta)
+  //salmon_index(generate_reference.out.fasta_files, params.fasta)
   // create cellranger index 
   cellranger_index(params.fasta, params.gtf, params.assembly)
 }
--- a/build-index.nf
+++ b/build-index.nf
@@ -81,14 +81,11 @@ process cellranger_index{
     """
     gunzip -c ${fasta} > genome.fasta
     gunzip -c ${gtf} > genome.gtf
-    cellranger mkgtf \
-      genome.gtf \
-      filtered.gtf
     
     cellranger mkref \
       --genome=${cellranger_index} \
       --fasta=genome.fasta \
-      --genes=filtered.gtf \
+      --genes=genome.gtf \
       --nthreads=${task.cpus}
     """
 }
--- a/build-index.nf
+++ b/build-index.nf
@@ -94,9 +94,9 @@ process cellranger_index{
 
 workflow {
   // generate splici and spliced cDNA reference fasta
-  //generate_reference(params.gtf, params.fasta, params.assembly)
+  generate_reference(params.gtf, params.fasta, params.assembly)
   // create index using reference fastas
-  //salmon_index(generate_reference.out.fasta_files, params.fasta)
+  salmon_index(generate_reference.out.fasta_files, params.fasta)
   // create cellranger index 
   cellranger_index(params.fasta, params.gtf, params.assembly)
 }
--- a/build-index.nf
+++ b/build-index.nf
@@ -77,7 +77,7 @@ process cellranger_index{
   output:
     path(cellranger_index)
   script:
-    cellranger_index = "${assembly}_cdna_cellranger"
+    cellranger_index = "${assembly}_cellranger_full"
     """
     gunzip -c ${fasta} > genome.fasta
     gunzip -c ${gtf} > genome.gtf
--- a/build-index.nf
+++ b/build-index.nf
@@ -66,10 +66,41 @@ process salmon_index{
     """
 }
 
+process cellranger_index{
+  container params.CELLRANGER_CONTAINER
+  publishDir "${params.ref_dir}/cellranger_index", mode: 'copy'
+  label 'cpus_8'
+  input:
+    path(fasta)
+    path(gtf)
+    val(assembly)
+  output:
+    path(cellranger_index)
+  script:
+    cellranger_index = "${assembly}_cdna_cellranger"
+    """
+    gunzip -c ${fasta} > genome.fasta
+    gunzip -c ${gtf} > genome.gtf
+    cellranger mkgtf \
+      genome.gtf \
+      filtered.gtf \
+      --attribute=gene_biotype:protein_coding
+    
+    cellranger mkref \
+      --genome=${cellranger_index} \
+      --fasta=genome.fasta \
+      --genes=filtered.gtf \
+      --nthreads=${task.cpus}
+    """
+}
+
+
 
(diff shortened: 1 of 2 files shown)


build-index.nf
```diff
@@ -66,10 +66,41 @@ process salmon_index{
...
+    path(cellranger_index)
+  script:
+    cellranger_index = "${assembly}_cdna_cellranger"
+    """
+    gunzip -c ${fasta} > genome.fasta
+    gunzip -c ${gtf} > genome.gtf
+    cellranger mkgtf \
+      genome.gtf \
+      filtered.gtf \
+      --attribute=gene_biotype:protein_coding
```
jashapiro, line 87: Is this going to be too stringent? I would think we should include ncRNA and such: The default cellranger index includes much more: https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/advanced/references#mkgtf

(I just looked at the full list of `gene_biotypes` in the reference we are using, which appears below (generated with `grep -o 'gene_biotype [^;]*' Homo_sapiens.GRCh38.104.gtf |sort | uniq`) and noted that the code used by 10X might miss `lncRNA` as it is looking for `lincRNA`, according to that page.) 

Alternatively, we could probably just not filter the gtf file? This would probably be more comparable to what we have for the salmon index, which I do not think we filter at all (hence 60K "genes").


```
gene_biotype "IG_C_gene"
gene_biotype "IG_C_pseudogene"
gene_biotype "IG_D_gene"
gene_biotype "IG_J_gene"
gene_biotype "IG_J_pseudogene"
gene_biotype "IG_V_gene"
gene_biotype "IG_V_pseudogene"
gene_biotype "IG_pseudogene"
gene_biotype "Mt_rRNA"
gene_biotype "Mt_tRNA"
gene_biotype "TEC"
gene_biotype "TR_C_gene"
gene_biotype "TR_D_gene"
gene_biotype "TR_J_gene"
gene_biotype "TR_J_pseudogene"
gene_biotype "TR_V_gene"
gene_biotype "TR_V_pseudogene"
gene_biotype "lncRNA"
gene_biotype "miRNA"
gene_biotype "misc_RNA"
gene_biotype "polymorphic_pseudogene"
gene_biotype "processed_pseudogene"
gene_biotype "protein_coding"
gene_biotype "pseudogene"
gene_biotype "rRNA"
gene_biotype "rRNA_pseudogene"
gene_biotype "ribozyme"
gene_biotype "sRNA"
gene_biotype "scRNA"
gene_biotype "scaRNA"
gene_biotype "snRNA"
gene_biotype "snoRNA"
gene_biotype "transcribed_processed_pseudogene"
gene_biotype "transcribed_unitary_pseudogene"
gene_biotype "transcribed_unprocessed_pseudogene"
gene_biotype "translated_processed_pseudogene"
gene_biotype "translated_unprocessed_pseudogene"
gene_biotype "unitary_pseudogene"
gene_biotype "unprocessed_pseudogene"
gene_biotype "vault_RNA"
```

build-index.nf
```diff
@@ -66,10 +66,41 @@ process salmon_index{
...
+  publishDir "${params.ref_dir}/cellranger_index", mode: 'copy'
+  label 'cpus_8'
+  input:
+    path(fasta)
+    path(gtf)
+    val(assembly)
+  output:
+    path(cellranger_index)
+  script:
+    cellranger_index = "${assembly}_cdna_cellranger"
```
jashapiro, line 80: If we are not filtering, I would probably shorten this to:
```suggestion
    cellranger_index = "${assembly}_cellranger"
```
or perhaps 
```suggestion
    cellranger_index = "${assembly}_cellranger_full"
```

Either way, this should probably be added to the `nextflow.config` `params` block.

jashapiro (commented):
This looks good. I made a couple of suggestions below, which might make this index more comparable to the one we use for salmon.

I was going to make an additional suggestion about naming params, but ended up putting it in a separate issue: https://github.com/AlexsLemonade/scpca-nf/issues/67

build-index.nf
allyhawkins: We definitely don't do any filtering like this for the salmon index, so I can remove this to make it more comparable.

allyhawkins:
I went ahead and removed filtering and it worked fine, creating a larger index this time.

build-index.nf
```diff
@@ -66,10 +66,40 @@ process salmon_index{
...
+      --nthreads=${task.cpus}
+    """
+}
+
+
 
 workflow {
   // generate splici and spliced cDNA reference fasta
-  generate_reference(params.gtf, params.fasta, params.assembly)
+  //generate_reference(params.gtf, params.fasta, params.assembly)
```
jashapiro, line 100: Uncommenting lines we want to keep
```suggestion
  generate_reference(params.gtf, params.fasta, params.assembly)
```

build-index.nf
```diff
@@ -66,10 +66,40 @@ process salmon_index{
...
   // create index using reference fastas
-  salmon_index(generate_reference.out.fasta_files, params.fasta)
+  //salmon_index(generate_reference.out.fasta_files, params.fasta)
```
jashapiro, line 102: ditto
```suggestion
  salmon_index(generate_reference.out.fasta_files, params.fasta)
```

jashapiro (commented):
Looks good, with two small suggestions I made, and one that I can't actually put in, which is to add 
```
  cellranger_index = "s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/cellranger_index/Homo_sapiens.GRCh38.104_cellranger_full"
```
to the params in `nextflow.config`

build-index.nf
```diff
@@ -66,10 +66,40 @@ process salmon_index{
...
+    gunzip -c ${fasta} > genome.fasta
+    gunzip -c ${gtf} > genome.gtf
+    cellranger mkgtf \
+      genome.gtf \
+      filtered.gtf
+    
+    cellranger mkref \
+      --genome=${cellranger_index} \
+      --fasta=genome.fasta \
+      --genes=filtered.gtf \
```
jashapiro, line 91: I think you can actually completely skip `mkgtf` as I don't think it is doing anything here.
```suggestion
    cellranger mkref \
      --genome=${cellranger_index} \
      --fasta=genome.fasta \
      --genes=genome.gtf \
```

allyhawkins:
@jashapiro I went ahead and made the edits that you suggested and then also had to increase the memory that the process is using. I noticed that it was only making some of the index files but not all of them even though it was completing the process successfully without error. It wasn't until I attempted to run spaceranger using the index later on that I realized there was missing files and only a partial index so had to bump up the memory required too.

jashapiro (approved):
LGTM!

Issue #68, opened 2021-12-08
Decide what output files to include for Spatial Transcriptomics


allyhawkins:
Based on a discussion with @jashapiro in slack, before doing #63, we should decide if we would like to create `rds` files for the spatial libraries for users, or if we would prefer to create a `tar.gz` of the `outs` folder after running spaceranger and provide that to users. 
Part of the reasoning here is that we would only be importing the spaceranger output into R as a `SpatialExperiment` and then outputting that as an `rds` file, without adding any additional analyses. Would we want to add the extra step of creating the `rds` file if we are not going to do anything extra to the actual `SpatialExperiment` object and would people prefer to have the `.cloupe` file or `.mtx.gz` files to load into R themselves. 

The pro for creating an `.rds` file is that the import scripts for the portal are already written to accept `.rds` files so we could keep everything in the same format and consistent across all samples. So I guess the question is how difficult would it be to change the importing for only a subset of samples that fall under the spatial category? Tagging @kurtwheeler for any thoughts he might have on that. 

We also provide QC reports for the other libraries, however here we are running spaceranger which generates its own summary `html` file. Would it be sufficient to use this report or is there any reason to create our own report? (I think the only case for this would be if we were using Alevin-fry + Spaceranger together). 

Based on the decisions made here, we may or may not need to complete #63.

allyhawkins:
After our discussion today in the ST benchmarking meeting, we have decided to use Spaceranger and provide the outputs from Spaceranger as a zip file. We should include the unfiltered and filtered output from Spaceranger, the web summary (equivalent to our qc report), the spatial folder, and a metadata file that we would add in with version information. This is approximately what the contents of the download for each library would then look like for the ST libraries. 

```
├── SCPCL000000_filtered_files
│   ├── barcodes.tsv.gz
│   ├── features.tsv.gz
│   └── matrix.mtx.gz
├── SCPCL000000_spaceranger_summary.html
├── SCPCL000000_spatial
│   ├── aligned_fiducials.jpg
│   ├── detected_tissue_image.jpg
│   ├── scalefactors_json.json
│   ├── tissue_hires_image.png
│   ├── tissue_lowres_image.png
│   └── tissue_positions_list.csv
├── SCPCL000000_unfiltered_files
│   ├── barcodes.tsv.gz
│   ├── features.tsv.gz
│   └── matrix.mtx.gz
└── SCPCL00000_metadata.json
```

jashapiro:
[...]

PR #72, opened 2021-12-22
Restructure output files for spatial module

### LINKED ISSUE #71: Adjust output of ST module to reflect desired output files
After quantification with Spaceranger, we will want to reorganize the output files to only publish those that we are going to be user facing. Right now the workflow includes publishing the entirety of the `outs` folder from Spaceranger, which we will not need. We can adjust the workflow to only include the output files mentioned in https://github.com/AlexsLemonade/scpca-nf/issues/68#issuecomment-998926142. We should then include all of the files below as one zip file before publishing the files as part of the process in Nextflow. 

Originally posted in https://github.com/AlexsLemonade/scpca-nf/issues/68#issuecomment-998926142. 

> After our discussion today in the ST benchmarking meeting, we have decided to use Spaceranger and provide the outputs from Spaceranger as a zip file. We should include the unfiltered and filtered output from Spaceranger, the web summary (equivalent to our qc report), the spatial folder, and a metadata file that we would add in with version information. This is approximately what the contents of the download for each library would then look like for the ST libraries.

```
├── SCPCL000000_filtered_files
│   ├── barcodes.tsv.gz
│   ├── features.tsv.gz
│   └── matrix.mtx.gz
├── SCPCL000000_spaceranger_summary.html
├── SCPCL000000_spatial
│   ├── aligned_fiducials.jpg
│   ├── detected_tissue_image.jpg
│   ├── scalefactors_json.json
│   ├── tissue_hires_image.png
│   ├── tissue_lowres_image.png
│   └── tissue_positions_list.csv
├── SCPCL000000_unfiltered_files
│   ├── barcodes.tsv.gz
│   ├── features.tsv.gz
│   └── matrix.mtx.gz
└── SCPCL00000_metadata.json
```

allyhawkins:
Closes #71. Here I am modifying the workflow for quantifying spatial data to publish only the desired output files in the structure that was discussed in https://github.com/AlexsLemonade/scpca-nf/issues/68#issuecomment-998940420. 

Here, the approach that I took was to move the desired files from the output of spaceranger to a new folder that is named by the library ID and then publish only that folder and it's contents, rather than publishing the entirety of the spaceranger output folder. Right now the way I have it set up is to perform both `spaceranger count` and then the file moving within the same process before only publishing the desired files. An alternative approach would be to keep the spaceranger output folder intact (other than removing the bam files) and publishing it to the `internal` folder as we were doing previously and then create a separate process to copy over the desired files, but that seemed like over kill and I don't believe there's any reason to keep the other intermediate files. 

One thing to note is that I am keeping the `_versions` file and copying it over as `spaceranger_versions.json` for now as I believe we will want to add a secondary process that grabs the version information and adds the associated metadata to create a `metadata.json` file. In doing that we will need this versions file to be saved in our output. 

The last step that is still missing is the actual compression of the output files. I was unsure if we wanted to compress all files together or keep out the json and/or html files. Until we resolve that, and are sure about our output structure, I am keeping this in draft form. 

** I also noticed that in the last merge somehow the `cellranger_index` did not get merged in to the main config file so I am adding that in here as well as the workflow fails without it.

### DIFF
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -37,7 +37,7 @@ process spaceranger{
     mv ${outs_dir}/web_summary.html ${spatial_out}/${meta.library_id}_spaceranger_summary.html
 
     # move over versions file temporarily to be passed to metadata.json
-    mv ${meta.run_id}/_versions ${outs_dir}/spaceranger_versions.json
+    mv ${meta.run_id}/_versions ${spatial_out}/spaceranger_versions.json
 
     """
 }
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -14,7 +14,7 @@ process spaceranger{
     tuple val(meta), path(spatial_out)
   script:
     spatial_out = "${meta.library_id}"
-    outs_dir = "${meta.run_id}-spatial/outs"
+    out_id = "${meta.run_id}-spatial"
     """
     spaceranger count \
       --id=${meta.run_id}-spatial \
@@ -31,13 +31,13 @@ process spaceranger{
     mkdir ${spatial_out}
 
     # move over needed files to outs directory 
-    mv ${outs_dir}/filtered_feature_bc_matrix ${spatial_out}
-    mv ${outs_dir}/raw_feature_bc_matrix ${spatial_out}
-    mv ${outs_dir}/spatial ${spatial_out}
-    mv ${outs_dir}/web_summary.html ${spatial_out}/${meta.library_id}_spaceranger_summary.html
+    mv ${out_id}/outs/filtered_feature_bc_matrix ${spatial_out}
+    mv ${out_id}/outs/raw_feature_bc_matrix ${spatial_out}
+    mv ${out_id}/outs/spatial ${spatial_out}
+    mv ${out_id}/outs/web_summary.html ${spatial_out}/${meta.library_id}_spaceranger_summary.html
 
     # move over versions file temporarily to be passed to metadata.json
-    mv ${meta.run_id}-spatial/_versions ${spatial_out}/spaceranger_versions.json
+    mv ${out_id}/_versions ${spatial_out}/spaceranger_versions.json
 
     """
 }
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -14,10 +14,10 @@ process spaceranger{
     tuple val(meta), path(spatial_out)
   script:
     spatial_out = "${meta.library_id}"
-    outs_dir = "${meta.run_id}/outs"
+    outs_dir = "${meta.run_id}-spatial/outs"
     """
     spaceranger count \
-      --id=${meta.run_id} \
+      --id=${meta.run_id}-spatial \
       --transcriptome=${index} \
       --fastqs=${fastq_dir} \
       --sample=${meta.cr_samples} \
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -17,7 +17,7 @@ process spaceranger{
     out_id = "${meta.run_id}-spatial"
     """
     spaceranger count \
-      --id=${meta.run_id}-spatial \
+      --id=${out_id} \
       --transcriptome=${index} \
       --fastqs=${fastq_dir} \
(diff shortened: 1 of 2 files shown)


modules/spaceranger.nf
```diff
@@ -3,34 +3,41 @@ nextflow.enable.dsl=2
...
-    # copy over needed files to outs directory 
-    mv ${meta.run_id}-spatial/_versions ${outs_dir}/spaceranger_versions.json
+    # move over needed files to outs directory 
+    mv ${outs_dir}/filtered_feature_barcode_bc_matrix ${spatial_out}
+    mv ${outs_dir}/raw_feature_bc_matrix ${spatial_out}
+    mv ${outs_dir}/spatial ${spatial_out}
+    mv ${outs_dir}/web_summary.html ${spatial_out}/${meta.library_id}_spaceranger_summary.html
+
+    # move over versions file temporarily to be passed to metadata.json
+    mv ${meta.run_id}/_versions ${outs_dir}/spaceranger_versions.json
```
jashapiro, line 40: I think?
```suggestion
    mv ${meta.run_id}/_versions ${spatial_out}/spaceranger_versions.json
```

But also, how do we plan to incorporate this into `metadata.json`? Should that be part of this PR?

allyhawkins:
There were a few minor changes I had to make to this after testing. Notably, spaceranger complains if you just use the `library_id` as the output folder, claiming that it already exists, so when I added back the spatial tag it no longer complains. This shouldn't be an issue since we don't write out that folder anyway and copy over the desired contents to a different output folder that ultimately gets published in the `publishDir`. Additionally, there was a minor typo in the paths when moving the files after completing spaceranger that now should be solved. I am doing another test run now, but it takes over 8 hours to complete. In the meantime, I am taking this out of the draft stage and will add any updates if the test run fails.

modules/spaceranger.nf
```diff
@@ -3,16 +3,17 @@ nextflow.enable.dsl=2
...
   label 'disk_big'
   input:
     tuple val(meta), path(fastq_dir), file(image_file)
     path index
   output:
-    tuple val(meta), path(outs_dir)
+    tuple val(meta), path(spatial_out)
   script:
+    spatial_out = "${meta.library_id}"
     outs_dir = "${meta.run_id}-spatial/outs"
```
jashapiro, line 17: I made this suggestion in https://github.com/AlexsLemonade/scpca-nf/pull/74#discussion_r782254307, because I was reviewing in the wrong order, but I can make it here as well for easier application...

```suggestion
    out_id = "${meta.run_id}-spatial"
```

This should be coupled with  changing line 20 to
```
 --id=${out_id} \
```
and the additional changes below

modules/spaceranger.nf
```diff
@@ -24,13 +25,19 @@ process spaceranger{
...
-    # copy over needed files to outs directory 
-    mv ${meta.run_id}-spatial/_versions ${outs_dir}/spaceranger_versions.json
+    # move over needed files to outs directory 
+    mv ${outs_dir}/filtered_feature_bc_matrix ${spatial_out}
+    mv ${outs_dir}/raw_feature_bc_matrix ${spatial_out}
+    mv ${outs_dir}/spatial ${spatial_out}
+    mv ${outs_dir}/web_summary.html ${spatial_out}/${meta.library_id}_spaceranger_summary.html
+
+    # move over versions file temporarily to be passed to metadata.json
+    mv ${meta.run_id}-spatial/_versions ${spatial_out}/spaceranger_versions.json
```
jashapiro, line 40:
```suggestion
    mv ${out_id}/outs/filtered_feature_bc_matrix ${spatial_out}
    mv ${out_id}/outs/raw_feature_bc_matrix ${spatial_out}
    mv ${out_id}/outs/spatial ${spatial_out}
    mv ${out_id}/outs/web_summary.html ${spatial_out}/${meta.library_id}_spaceranger_summary.html

    # move over versions file temporarily to be passed to metadata.json
    mv ${out_id}/_versions ${spatial_out}/spaceranger_versions.json
```

allyhawkins:
I incorporated the directory changes that were suggested and tested this and all looks good so going ahead and merging.

PR #73, opened 2022-01-06
Adjust memory requirements for alevin rad process

allyhawkins:
In processing the Mullighan samples, there was one sample in particular with FASTQ files totaling 116 GB in size. Previously we had been using the default queue on batch which was not big enough to hold these files and the only way to successfully process this sample (or any other sample with FASTQ files this large) is to use the `bigdisk` queue. 

However, we do not want to use the `bigdisk` queue for all of the samples, so I have added a new label that uses the default queue on the first task attempt and then the `bigdisk` queue on the second task attempt. I have tested this and it works successfully for the Mullighan sample that had issues previously and switches to the `bigdisk` queue on the second attempt. 

Additionally, the `cellranger_index` was left out of the params previously and an error had been thrown so I added that in here so that the workflow would run successfully.

### DIFF
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -3,8 +3,7 @@
 // generates RAD file using alevin
 process alevin_rad{
   container params.SALMON_CONTAINER
-  label 'cpus_12'
-  label 'bigdisk'
+  label 'dynamic_disk'
   tag "${meta.run_id}-rna"
   publishDir "${params.outdir}/internal/rad/${meta.library_id}"
   input:
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -4,7 +4,7 @@
 process alevin_rad{
   container params.SALMON_CONTAINER
   label 'cpus_12'
-  label 'dynamic_disk'
+  label 'disk_dynamic'
   tag "${meta.run_id}-rna"
   publishDir "${params.outdir}/internal/rad/${meta.library_id}"
   input:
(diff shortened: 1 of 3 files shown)


nextflow.config
```diff
@@ -73,6 +74,11 @@ profiles{
       withLabel: bigdisk {
         queue = 'nextflow-batch-bigdisk-queue'
       }
+      withLabel: dynamic_disk {
+        cpus = 12
+        memory = { 24.GB * task.attempt }
```
jashapiro, line 79: I hadn't remembered that we can use more than one label (as we do for spaceranger), so I think that is a better strategy to use here as well. We can keep the memory/cpus separate from disk space for less redundancy/more flexibility.

```suggestion
```

modules/af-rna.nf
```diff
@@ -3,7 +3,7 @@
 // generates RAD file using alevin
 process alevin_rad{
   container params.SALMON_CONTAINER
-  label 'cpus_12'
+  label 'dynamic_disk'
```
jashapiro, line 6: Using two labels.
```suggestion
  label 'cpus_12'
  label 'dynamic_disk'
```

nextflow.config
```diff
@@ -73,6 +74,11 @@ profiles{
       withLabel: bigdisk {
```
jashapiro, line 74: So very minor, but do we want to update this to `big_disk` for consistency?

jashapiro (approved):
Great that this worked! I had a suggestion to make the way we handle labels a bit more consistent; one set for cpu and one for disk space.

Looking at naming again as I write this: maybe `disk_big` and `disk_dynamic` to keep the format of the `cpu_*` and `mem_*` labels? (Just make sure these get changed in all the processes they are used for as well!)

These are all minor tweaks, so I am approving with no need for followup review.

PR #74, opened 2022-01-07
Add process to export metadata.json for spatial libraries

allyhawkins:
Stacked on #72. 

This PR adds another process to the spatial workflow to create a `library_id_metadata.json` for spatial libraries, similar to the `json` file we create for single-cell/single-nuclei libraries. After quantifying the spatial libraries using Spaceranger, the metadata and directory containing the quantified output are passed to a process to generate a `metadata.json` file.  Inside that process, an Rscript to read in the specifics about each library and write them out in `json` format is run. 

In general, this follows a similar format to creating the other `metadata.json `which is done inside the `sce_qc_report.R` script. The main differences are as follows: 

- Here we do not have `RDS` files that contain the output, but only the direct output files from spaceranger. In order to get the number of filtered and unfiltered spots, I am reading in the `filtered_feature_bc_matrix/barcodes.tsv.gz` and `raw_feature_bc_matrix/barcodes.tsv.gz` and obtaining the number of rows, corresponding to the number of spot barcodes identified. 
- Previously, we had been storing metadata inside of the SCE object and grabbing some of the metadata from that object (like the number of reads), but here we do not have an SCE object. To get the number of reads, there is a `metrics_summary.csv` file that is output by spaceranger. I added that to be included in the spaceranger process and then am reading that in as part of the Rscript and grabbing the desired stats. [You can find the table of all metrics included in that summary on the 10X website](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/output/gex-metrics). I included total reads and total reads mapped confidently to genome, are there any additional metrics that you think we should include here? 
- The name of the index with Alevin-fry was previously obtained from one of the log files output by Alevin-fry. It doesn't appear that can be identified from a log file here, so to get the index name I am reading in the entire path and then using the `basename()` function to grab just the index name. Let me know if you think there is a better approach to doing this.
- To obtain the version of spaceranger, I am grabbing that from the `_versions` file from spaceranger count (which we have renamed as `spaceranger_versions.json`. In that file there is also a martian version. Do we want to also include that here? 

The only other question I had for reviewers was if there were additional things that should be output as part of the `metadata.json` file or things you think are unnecessary and should be removed? 

This is currently being tested, which takes a long time, but will add any updates based on if things run successfully.

### DIFF
--- a/bin/generate_spaceranger_metadata.R
+++ b/bin/generate_spaceranger_metadata.R
@@ -99,7 +99,7 @@ if(is.null(opt$sample_id)){
 }
 
 # check that barcode files exist
-if(is.null(opt$unfiltered_barcodes_file) || !dir.exists(opt$unfiltered_barcodes_file)){
+if(is.null(opt$unfiltered_barcodes_file) || !file.exists(opt$unfiltered_barcodes_file)){
   stop("Unfiltered barcodes file missing or `unfiltered_barcodes_file` not specified.")
 }
 if(is.null(opt$filtered_barcodes_file) || !file.exists(opt$filtered_barcodes_file)){
--- a/bin/generate_spaceranger_metadata.R
+++ b/bin/generate_spaceranger_metadata.R
@@ -1,0 +1,161 @@
+#!/usr/bin/env Rscript
+
+# This script generates the metadata.json file for Spatial Transcriptomics libraries. 
+
+# import libraries
+library(optparse)
+suppressPackageStartupMessages(library(SingleCellExperiment))
+
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = c("-l", "--library_id"),
+    type = "character",
+    help = "Library identifier for report and metadata file"
+  ),
+  make_option(
+    opt_str = c("-s", "--sample_id"),
+    type = "character",
+    help = "Sample identifier for metadata file"
+  ),
+  make_option(
+    opt_str = c("-u", "--unfiltered_barcodes_file"),
+    type = "character",
+    help = "path to file containing list of all spot barcodes identified in library with no filtering"
+  ),
+  make_option(
+    opt_str = c("-f", "--filtered_barcodes_file"),
+    type = "character",
+    help = "path to file containing list of spot barcodes identified in library after filtering"
+  ),
+  make_option(
+    opt_str = c("--metrics_summary_file"),
+    type = "character",
+    help = "path to file containing summary of run metrics from spaceranger count, the metrics_summary.csv file output by spaceranger count"
+  ),
+  make_option(
+    opt_str = c("--spaceranger_versions_file"),
+    type = "character",
+    help = "path to json file containing spaceranger version information, the _versions file output by spaceranger count"
+  ),
+  make_option(
+    opt_str = "--metadata_json",
+    default = "metadata.json",
+    type = "character",
+    help = "path to metadata json output file"
+  ),
+  make_option(
+    opt_str = "--technology",
+    type = "character",
+    default = NA,
+    help = "sequencing technology"
+  ),
+  make_option(
+    opt_str = "--seq_unit",
+    type = "character",
+    default = NA,
+    help = "sequencing unit"
+  ),
+  make_option(
+    opt_str = "--genome_assembly",
+    type = "character",
+    default = NA,
+    help = "genome assembly used for mapping"
+  ),
+  make_option(
+    opt_str = "--index_path",
+    type = "character",
+    default = NA,
+    help = "path to index used for mapping"
+  ),
+  make_option(
+    opt_str = "--workflow_url",
+    type = "character",
+    default = NA,
+    help = "workflow github url"
+  ),
+  make_option(
+    opt_str = "--workflow_version",
+    type = "character",
+    default = NA,
+    help = "workflow version identifier"
+  ),
+  make_option(
+    opt_str = "--workflow_commit",
+    type = "character",
+    default = NA,
+    help = "workflow commit hash"
+  )
+)
+
+opt <- parse_args(OptionParser(option_list = option_list))
+if(is.null(opt$library_id)){
+  stop("A `library_id` is required.")
+}
+# check that barcode files exist
+if(is.null(opt$unfiltered_barcodes_file) || !dir.exists(opt$unfiltered_barcodes_file)){
+  stop("Unfiltered barcodes file missing or `unfiltered_barcodes_file` not specified.")
+}
+if(is.null(opt$filtered_barcodes_file) || !file.exists(opt$filtered_barcodes_file)){
+  stop("Unfiltered barcodes file missing or `unfiltered_barcodes_file` not specified.")
+}
+
+# check that metrics summary file exists 
+if(is.null(opt$metrics_summary_file) || !file.exists(opt$metrics_summary_file)){
+  stop("Metrics summary file missing or `metrics_summary_file` not specified.")
+}
+
+# check that version file exists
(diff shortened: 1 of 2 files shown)


modules/spaceranger.nf
```diff
@@ -36,12 +36,43 @@ process spaceranger{
...
 }
 
+process make_metadata{
+  container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
+  input:
+    tuple val(meta), path(spatial_out)
+    path (index)
+  output:
+    tuple val(meta), path(spatial_out), path(metadata_json)
```
jashapiro, line 53: I don't think we want to re-emit the output folder here, as we are not modifying it, and this will result in re-copying the files. They are not too large, but I think we should avoid it anyway.

modules/spaceranger.nf
```diff
@@ -36,12 +36,43 @@ process spaceranger{
...
 
     """
 }
 
+process make_metadata{
+  container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
+  input:
+    tuple val(meta), path(spatial_out)
+    path (index)
```
jashapiro, line 51: The index is fairly large, so including it here seems like not the best way to proceed. I think a better approach might be to add the index to the `meta` object as part of the spaceranger process, so then we can access it directly. 

I think you can just do something like:
```
meta.spaceranger_index = index
```
at the top of the `spaceranger` process script (before the `"""`), which will then be passed along in the `meta` object that is output.

This would be similar to what I did here: https://github.com/AlexsLemonade/alsf-scpca/blob/3f9fead95da61908f695ab2b5b0a8137c90273ae/workflows/genetic-demux/cellsnp.nf#L15-L17

Alternatively, I think you can probably just use `val(index)` as the input here, which would pass the value along to the `make_metadata` process, but not force nextflow to actually copy the files when setting up the process. This may be simpler, but there is the small advantage of the previous option that the index is only specified once, so it is guaranteed to be the one that was used to generate the output (no way to put two different values in by mistake). It is a pretty small risk, but worth a mention.

modules/spaceranger.nf
```diff
@@ -36,12 +36,43 @@ process spaceranger{
...
 
-    # move over versions file temporarily to be passed to metadata.json
+    # move over versions and metrics summary file temporarily to be passed to metadata.json
     mv ${meta.run_id}/_versions ${spatial_out}/spaceranger_versions.json
+    mv ${outs_dir}/metrics_summary.csv ${spatial_out}/spaceranger_metrics_summary.csv
 
     """
 }
 
+process make_metadata{
```
jashapiro, line 46: I'm suggesting here changing the name to be a bit more specific.

```suggestion
process spaceranger_metadata{
```

bin/generate_spaceranger_metadata.R
```diff
@@ -0,0 +1,168 @@
...
+# check for sample and library id
+if(is.null(opt$library_id)){
+  stop("A `library_id` is required.")
+}
+if(is.null(opt$sample_id)){
+  stop("A `sample_id` is required.")
+}
+
+# check that barcode files exist
+if(is.null(opt$unfiltered_barcodes_file) || !dir.exists(opt$unfiltered_barcodes_file)){
```
jashapiro, line 102:
```suggestion
if(is.null(opt$unfiltered_barcodes_file) || !file.exists(opt$unfiltered_barcodes_file)){
```

jashapiro (commented):
This looks good. I have a few suggestions for efficiency so that we are not passing around largish files unnecessarily.

allyhawkins:
Thanks for the comments @jashapiro. I incorporated your suggestions and I believe I addressed the concern about the index as well. I went ahead and tested this and everything works as expected. Sending over to you for another round of review.

modules/spaceranger.nf
```diff
@@ -36,9 +37,38 @@ process spaceranger{
     mv ${outs_dir}/spatial ${spatial_out}
     mv ${outs_dir}/web_summary.html ${spatial_out}/${meta.library_id}_spaceranger_summary.html
 
-    # move over versions file temporarily to be passed to metadata.json
+    # move over versions and metrics summary file temporarily to be passed to metadata.json
     mv ${meta.run_id}-spatial/_versions ${spatial_out}/spaceranger_versions.json
+    mv ${outs_dir}/metrics_summary.csv ${spatial_out}/spaceranger_metrics_summary.csv
```
jashapiro, line 42: This is a pretty minor thing for now, but I thought I would put the comment in anyway...

Since the `outs_dir` is entirely internal to the process, it seems like it probably isn't necessary to have its name dependent on the `run_id` at all. So if I were redoing this, I might just hard code the `--id` as something like `spaceranger_result` and then use that repeatedly. But using the `run_id` is fine, and shouldn't generally cause problems. 

However, since `outs_dir` _is_ defined above, we should probably use it for all output paths, so that we don't have to remember how it is defined. There are two ways to do this, and I think I like one better than the other.

The first is to leave things as they are, and make the following single change to direct `mv` to use `_versions` in the parent directory:
```suggestion
    mv ${outs_dir}/../_versions ${spatial_out}/spaceranger_versions.json
    mv ${outs_dir}/metrics_summary.csv ${spatial_out}/spaceranger_metrics_summary.csv
```

But I don't love that, partly because it still leaves us to remember that `--id` and `$outs_dir` have to be defined the with the same root.

An alternative is to change `outs_dir = "${meta.run_id}-spatial/outs"` to something like `out_id = "${meta.run_id}-spatial"`, change to `--id=${out_id}` and then explicitly reference the `outs` subdirectory when needed (this requires changes outside this suggestion as well, of course).

```suggestion
    mv ${out_id}/_versions ${spatial_out}/spaceranger_versions.json
    mv ${out_id}/outs/metrics_summary.csv ${spatial_out}/spaceranger_metrics_summary.csv
```

This does add some redundancy, but it is a bit more explicit about the file organization expected.

modules/spaceranger.nf
```diff
@@ -15,10 +15,11 @@ process spaceranger{
   script:
     spatial_out = "${meta.library_id}"
     outs_dir = "${meta.run_id}-spatial/outs"
+    meta.cellranger_index = "${index}"
```
jashapiro, line 18: I think this may include ugly `/tmp` directory paths, which is not terrible given the way we are using it now (with `basename` in the R script), but might be ugly and a bit unstable.

I think we can get the basename here for clarity like this:
```suggestion
    meta.cellranger_index = index.getBaseName()
```
This will mean we can't use `meta.cellranger_index` in the script itself as it won't have the full path info, but I don't see a problem with that... we can just revert to using `${index}` there.

modules/spaceranger.nf
```diff
@@ -15,10 +15,11 @@ process spaceranger{
...
     """
     spaceranger count \
       --id=${meta.run_id}-spatial \
-      --transcriptome=${index} \
+      --transcriptome=${meta.cellranger_index} \
```
jashapiro, line 22: See earlier comment.
```suggestion
      --transcriptome=${index} \
```

jashapiro (approved):
Looks good. I have a few relatively minor additional thoughts here, which unfortunately will require testing, but I don't think I should need to rereview.

modules/spaceranger.nf
jashapiro: Following up on this; if it is working as expected, I think we might want to delete the `basename()` call in the R script:  In the event that the directory name contains a `.`; `basename()` will strip everything from that onward which would not be desirable.

PR #75, opened 2022-01-10
Reorganize configuration file and rename ambiguous parameters

### LINKED ISSUE #67: Reorganize config and rename ambiguous params
The section `nextflow.config` with index file locations is getting pretty big, and with multiple indices, some of the parameter names are getting a bit ambiguous. For example, `fasta` and `gtf` should probably be something like `ref_fasta` and `ref_gtf`. Changing these will require changing their use throughout the modules.

We can also reorganize somewhat, breaking up the config file into smaller ones that define the processes and parameters separately and taking advantage of the `includeConfig` directive. I also realized that we _can_ use variables in the config files , so we can cut a lot of redundant path info. More info on this here: https://www.nextflow.io/docs/latest/config.html?highlight=configuration#config-syntax

allyhawkins:
Closes #67. Here I am reorganizing the main configuration file for the workflow and breaking out the parameters into its own configuration file and the `process` section including the batch specific configuration into its own file as well. I'm then using the `includeConfig` directive to include these files in the main configuration file. 

While doing this, I also renamed a few of the ambiguous parameters such as `gtf` and `fasta` to `ref_gtf` and `ref_fasta` and `index_path` to `splici_index` to be more consistent with the names of the other indices. I also added in the use of variables in creating the paths to the index and annotation files so we aren't re-writing the same thing multiple times when they are in the same base directory. I then propagated the name changes to the rest of the files that referenced these parameters and ran a test on both the main workflow and index workflow and the configuration file appears to work as expected. 

For reviewers, was this what was envisioned in terms of breaking out the parameters and processes into their own files as described in #67? Or was there a different organization that is preferred?


nextflow.config
```diff
@@ -51,32 +13,7 @@ profiles{
...
-      }
-      withLabel: disk_big {
-        queue = 'nextflow-batch-bigdisk-queue'
-      }
-      withLabel: disk_dynamic {
-        queue = { task.attempt < 2 ? 'nextflow-batch-default-queue' : 'nextflow-batch-bigdisk-queue'}
-      }
-    }
+    // batch parameters
+    includeConfig 'config/process.config'
```
jashapiro, line 17: Suggesting renaming this file, as the config in it is specific to aws. As we are going to separate this out, then I think it should include all of the aws settings, from `bucketDir = 's3://nextflow-ccdl-data/work'` down, not just the process settings, as many of the settings don't make sense outside of aws.

We could probably figure out a way to pull out the common parts that might still apply outside of AWS, but I don't think that is worth the effort at this time.

We could also move _all_ of the `profiles` block to a separate file, but I kind of like having the available options visible in the main `nextflow.config` file. 

```suggestion
    includeConfig 'config/profile_awsbatch.config'
```

jashapiro (commented):
Thank you for tackling this! I like a lot of the changes here, especially cleaning up some of the paramter names and using variables to simplify changes later. I have a few thoughts about further organization, which I think are mostly in the line by line comments.

Of those, I definitely think we should separate containers to their own file, and keep the aws settings in a single place.

I feel less strongly about whether we have a separate parameters file. Maybe the processing options should be in the main file, and file paths in a separate file?

config/parameters.config
```diff
@@ -0,0 +1,43 @@
...
+  emptydrops_lower = 200 // emptydrops lower bound for cell UMI count
+  seed = 2021   // random number seed for filtering (0 means use system seed)
+
+  // Docker container images
+  SALMON_CONTAINER = 'quay.io/biocontainers/salmon:1.5.2--h84f40af_0'
+  ALEVINFRY_CONTAINER = 'quay.io/biocontainers/alevin-fry:0.4.2--h7d875b9_0'
+  FASTP_CONTAINER = 'quay.io/biocontainers/fastp:0.23.0--h79da9fb_0'
+  SCPCATOOLS_CONTAINER = 'ghcr.io/alexslemonade/scpca-tools:v0.1.1'
+  CELLRANGER_CONTAINER = '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:6.1.2'
+  SPACERANGER_CONTAINER = '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-spaceranger:1.3.1'
```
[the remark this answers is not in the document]
allyhawkins: I had played around with this at one point where I had the reference files and annotation files in their own file and then the containers in their own file, but had ended back at this (and now can't remember why I ended back here). I definitely like the idea of pulling out the containers, so let me play around with another option and see what you think.

allyhawkins:
I went ahead and reorganized this so that the parameters related to the reference files are in their own config file and the containers are in their own config file, leaving any of the other basic parameters in the main config file. Let me know what you think of that organization. 

I also moved the entire batch profile related items to their own config file like you had suggested. I did a quick test with both workflows in this repo and things are working as expected.

nextflow.config
```diff
@@ -1,42 +1,26 @@
...
   emptydrops_lower = 200 // emptydrops lower bound for cell UMI count
   seed = 2021   // random number seed for filtering (0 means use system seed)
-}
 
+  // Docker container images
+  includeConfig 'config/parameters_containers.config'
+
+  // reference and annotation parameters
+  includeConfig 'config/parameters_reference.config'
+
```
jashapiro, line 22: Do we want to simplify these filenames a bit? Not a strong feeling, but `parameters_` seems redundant. 
```suggestion
  includeConfig 'config/containers.config'

  // reference and annotation parameters
  includeConfig 'config/reference_paths.config'

```

jashapiro (approved):
LGTM! 

One nitpicky naming preference that you can ignore if you want.

PR #76, opened 2022-01-11
Update workflow to use new scpcaTools (0.1.2)

### LINKED ISSUE #69: Update workflow to use new scpcaTools (0.1.2)
When scpcaTools 0.1.2 is completed (https://github.com/AlexsLemonade/scpcaTools/issues/78), we will need to update this workflow to use it. 

This should mostly require updating the scpcatools docker image to the latest version, but it is worth noting that this will also change the default filtering to `emptydrops_CR` & may alter `miQC` results. 

We will want to make sure we are testing with updated versions before merging. (and note that downstream docs will require changes).

jashapiro:
The primary purpose of this PR is to move to scpcaTools 0.1.2
Closes #69 

This change also changes the action of `scpcaTools::filter_counts()` such that the `lower` argument causes an error, as that is not an argument to the underlying `emptyDropsCellRanger()` function. 

I solved this in the simplest way by eliminating the argument from the `filter_sce_rds.R` script & the calls within it, but if we think there is a reason why we might want to keep it in this workflow, we could get more clever about how we specify it. This would probably involve explicitly adding options for using the original `emptyDrops` function, so it will add some complexity throughout.

One other change: the references file didn't match it's `includeConfig` statement, so I renamed the file.

I have tested the workflow with these changes, and it seems to run as expected.


allyhawkins (approved):
LGTM! Thanks for making the changes to remove the `lower` argument as well.

PR #77, opened 2022-01-24
Allow for skipping of `salmon alevin` process and proceed directly to alevin-fry quantification

### LINKED ISSUE #41: Allow explicit restarting from RAD files (or other points?)

allyhawkins:
Closes #41. This PR, adds in flexibility to the workflow to skip the first step of running `salmon alevin` in cases where the RAD file already exists and a new parameter, `rad_skip` is set to true. 

Here we are specifically only skipping the first step in the `af-rna.nf` module as this is the step that is the most costly in both time and memory. This update will enable us to re-run samples that have already been processed and incorporate changes in the later parts of the workflow, such as the filtering steps, or future changes to the QC report, without having to re-run the mapping step. 

To do this, we first add the paths to the rad directory to the meta object. Then we create two channels within the `map_quant_rna` workflow. The first channel includes libraries that are missing the rad directory and then becomes the input to the `alevin_rad` process. The second channel contains libraries that already have a rad directory, and becomes the input to the next step, `fry_quant_rna`. We then combine these two channels, pull out the `barcode_file` to create a tuple of `[barcode_file, rad_dir, meta]` to be used as input to `fry_quant_rna`. 

A few things to note. The first is that in order to properly combine the output channel from `alevin_rad` with the `rna_rad_ch`, I had to rename the output of the process as `rad_dir` to match with how we had named it in the meta. I could rename everything here, `run_dir`, but I thought `rad_dir` was a better fit for everything related to looking for the rad files and then I am still using `run_dir` for the downstream `fry_quant_rna` process. 

I also am not quite sure why, but nextflow didn't like `.map{it + [it[0].barcode_file]}` to pull out the barcode file, but preferred `.map{[it[0].barcode_file] + it}`. So I changed it to have the barcode file as the first item in the tuple and it stopped complaining. 

I tested the workflow and everything worked as plan, skipping the RAD step and taking only 7 minutes to process a single RNA sample! 

We also added here the parameter, `rad_skip` which I set to true. By having this parameter here, we can force the rad step to still run and set it to false at the command line by using `--rad_skip false`. I tested this out and when set to false it no longer skips the `alevin_rad` step as planned. 
Also when adding in this parameter, I moved all of the other remaining parameters to the main configuration file rather than keeping some of them within `main.nf`.

### DIFF
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -6,7 +6,7 @@ process alevin_rad{
   label 'cpus_12'
   label 'disk_dynamic'
   tag "${meta.run_id}-rna"
-  publishDir "${params.outdir}/internal/rad/${meta.library_id}"
+  publishDir "${meta.rad_publish_dir}"
   input:
     tuple val(meta), 
           path(read1), path(read2)
@@ -14,7 +14,7 @@ process alevin_rad{
   output:
     tuple val(meta), path(rad_dir) 
   script:
-// define the local location for the rad output
+    // define the local location for the rad output
     rad_dir = file(meta.rad_dir).name
     // choose flag by technology
     tech_flag = ['10Xv2': '--chromium',
@@ -46,7 +46,7 @@ process fry_quant_rna{
   publishDir "${params.outdir}/internal/af/${meta.library_id}"
 
   input:
-    tuple path(barcode_file), val(meta), path(run_dir)
+    tuple val(meta), path(run_dir), path(barcode_file)
     path tx2gene_3col
   output:
     tuple val(meta), path(run_dir)
@@ -91,11 +91,11 @@ workflow map_quant_rna {
        // split based in whether rad_skip is true and a previous dir exists
       .branch{
           has_rad: params.rad_skip && file(it.rad_dir).exists()    
-          do_rad: true
+          make_rad: true
        }     
     
     // If We need to create rad files, create a new channel with tuple of (metadata map, [Read1 files], [Read2 files])   
-    rna_reads_ch = rna_channel.do_rad
+    rna_reads_ch = rna_channel.make_rad
       .map{meta -> tuple(meta,
                          file("s3://${meta.s3_prefix}/*_R1_*.fastq.gz"),
                          file("s3://${meta.s3_prefix}/*_R2_*.fastq.gz")
@@ -104,7 +104,6 @@ workflow map_quant_rna {
     // if the rad directory has been created and rad_skip is set to true
     // create tuple of (metdata map, rad_directory) to be used directly as input to alevin-fry quantification
     rna_rad_ch = rna_channel.has_rad
-      .filter{params.rad_skip && file(it.rad_dir).exists()}
       .map{meta -> tuple(meta, 
                          file(meta.rad_dir)
                          )}
@@ -114,7 +113,7 @@ workflow map_quant_rna {
 
     // combine ouput from running alevin step with channel containing libraries that skipped creating a RAD file 
     all_rad_ch = alevin_rad.out.mix(rna_rad_ch)
-      .map{it.asList() + file(it[0].barcode_file)} // add barcode file to channel to use in fry_quant_rna process
+      .map{it.toList() + it[0].barcode_file} // add barcode file to channel to use in fry_quant_rna process
 
     // quantify with alevin-fry
     fry_quant_rna(all_rad_ch, params.t2g_3col_path)
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -12,7 +12,7 @@ process alevin_rad{
           path(read1), path(read2)
     path index
   output:
-    tuple val(meta), path(run_dir)
+    tuple val(meta), path(run_dir) 
   script:
     // label the run-dir
     run_dir = "${meta.run_id}-rna"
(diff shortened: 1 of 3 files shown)


jashapiro:
👏🏼 👏🏼 👏🏼 !!!

I will review this tomorrow, but I wanted to say congrats on getting it done!

I also had a quick thought about why `it + [x]` pattern didn't work, which I thought was worth mentioning now. I'm not sure now whether we will want to do anything about it, but I have an idea of how we would if we wanted to. 

My explanation is that the elements in a channel are stored by nextflow as a tuple, and tuples are immutable. So we can't add to a tuple, but we can add a tuple to a list (which is mutable) we create with `[]`, which explains why the other direction works. A potential option if we want to keep the order would be to do `it.toList() + [x]`*. Again, not sure whether that is worth it, but it might be nice to keep the metadata first for consistency.

\* I'm actually not sure that the square brackets are needed; when I was looking this up and testing in the console, it seemed to work as just `it.toList() + x`, but I feel like I must have done it before for a reason? Probably because I was putting things first for sorting reasons, so the first element had to be a list. There is also apparently the option to use `it.toList() << x` which is a groovy-specific way to add to a list. I don't love it.

jashapiro (commented):
This looks good, though I haven't tested it yet, which I will do before I approve it.

I had a few of minor-ish suggestions, but the most important is to use the `meta.rad_dir` within the workflow rather than redefining part of it so we do not have to worry about the file locations getting out of sync (same for the `publishDir`)

modules/af-rna.nf
```diff
@@ -83,21 +82,39 @@ workflow map_quant_rna {
...
+    rna_channel = rna_channel
+      .map{it.rad_publish_dir = "${params.outdir}/internal/rad/${it.library_id}";
+           it.rad_dir = "${it.rad_publish_dir}/${it.run_id}-rna"; 
+           it.barcode_file = "${params.barcode_dir}/${params.cell_barcodes[it.technology]}";
+           it}
+
+    // if the rad directory has not been created or rad_skip is set to false
+    // create a new channel with tuple of (metadata map, [Read1 files], [Read2 files])
     rna_reads_ch = rna_channel
+      .filter{!(params.rad_skip && file(it.rad_dir).exists())}
```
jashapiro, line 95: I just had a somewhat more radical idea to use the nextflow branch command to pull the logic here and at line 104 together: rather than `filter` we could use [`branch`]( https://www.nextflow.io/docs/latest/operator.html#branch), which create two sub-channels we can address separately. This ensures that we don't accidentally get the logic of the two filter statements out of sync.
```suggestion
      .map{it.rad_publish_dir = "${params.outdir}/internal/rad/${it.library_id}";
           it.rad_dir = "${it.rad_publish_dir}/${it.run_id}-rna"; 
           it.barcode_file = "${params.barcode_dir}/${params.cell_barcodes[it.technology]}";
           it}
       // split based in whether rad_skip is true and a previous dir exists
      .branch{
          has_rad: params.rad_skip && file(it.rad_dir).exists()    
          do_rad: true
       }     
    
    // If We need to create rad files, create a new channel with tuple of (metadata map, [Read1 files], [Read2 files])   
    rna_reads_ch = rna_channel.do_rad
```

modules/af-rna.nf
```diff
@@ -83,21 +82,39 @@ workflow map_quant_rna {
...
       .map{meta -> tuple(meta,
                          file("s3://${meta.s3_prefix}/*_R1_*.fastq.gz"),
                          file("s3://${meta.s3_prefix}/*_R2_*.fastq.gz")
                         )}
 
-    cellbarcodes_ch = rna_channel
-      .map{file("${params.barcode_dir}/${params.cell_barcodes[it.technology]}")}
+    // if the rad directory has been created and rad_skip is set to true
+    // create tuple of (metdata map, rad_directory) to be used directly as input to alevin-fry quantification
+    rna_rad_ch = rna_channel
```
jashapiro, line 103: Use the other side of the branch:
```suggestion
    rna_rad_ch = rna_channel.has_rad
```

jashapiro (commented):
One more thought I had, which cleans up the splitting logic/alternatives to simplify a bit and avoid redundancy.

allyhawkins:
Thanks for the helpful tips on why adding to the tuple wasn't working @jashapiro, and I like the use of `branch` here as well! I went ahead and implemented the changes you suggested and tested them. I just changed the name of the `branch` to `make_rad` instead of `do_rad`. It was minor but something about it wasn't sitting well with me. 

And then when testing, there was some issues with using `.map{it.toList() + file(it[0].barcode_file)}` and nextflow yelled at me saying that the input tuple didn't match the input set cardinality declared by process. When removing `file` it worked as expected.

jashapiro (approved):
Looks great!  I did testing with a number of potential combinations of options/rad files present or not, etc. and everything seemed to work as expected!

I did have a couple of minor additional comments below (and I tested with the `<<` version to be sure it worked), but I'm excited to have this optimization in, and I might consider this worthy of a version bump to 0.2

allyhawkins:
> I did have a couple of minor additional comments below (and I tested with the << version to be sure it worked), but I'm excited to have this optimization in, and I might consider this worthy of a version bump to 0.2

I would agree with this! I think before we process the next set of projects it would be great to have this release as an updated version. That would be especially great for the projects where we were planning to process the new modalities and now we could process the whole project instead of just the new modalities and incorporate all of the updates for single-cell/nuclei samples too that have happened downstream of `alevin-fry`.

PR #83, opened 2022-02-02
Remove explicit dependence on S3

### LINKED ISSUE #78: Remove explicit dependence on S3
As discussed in https://github.com/AlexsLemonade/ScPCA-admin/issues/298, the current metadata file that we use as the control list for this workflow, defined in `params.run_metafile`, assumes that certain files are on S3. When we make that explicit (allowing for files to be stored on other services or locally), we should adjust the workflow to use the urls as specified, not adding `s3://` as we currently do, for example at https://github.com/AlexsLemonade/scpca-nf/blob/bc587b58d392ee700e3fd9dd46d0aedac253907d/modules/af-rna.nf#L99-L102

Note that if we changed the name of the `s3_prefix` field while accomplishing https://github.com/AlexsLemonade/ScPCA-admin/issues/298, this will also need to be updated.

jashapiro:
This PR follows https://github.com/AlexsLemonade/ScPCA-admin/pull/299 to make the changes to the nf workflow to remove explicit dependence on s3.

Closes #78

I think I got everything, and I tested using the libarary info file from https://github.com/AlexsLemonade/ScPCA-admin/pull/299 to make sure there were not immediate errors, using samples from bulk, single cell and spatial to test all parts of the workflow. I did not wait for final success on some (looking at spaceranger), but all types started fine, so I believe that all paths were created as expected.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -55,7 +55,7 @@ workflow {
       seq_unit: it.seq_unit,
       feature_barcode_file: it.feature_barcode_file,
       feature_barcode_geom: it.feature_barcode_geom,
-      s3_prefix: it.s3_prefix,
+      files_directory: it.files_directory,
       slide_serial_number: it.slide_serial_number,
       slide_section: it.slide_section,
       files: it.files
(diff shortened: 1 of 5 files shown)


allyhawkins (approved):
LGTM!

PR #84, opened 2022-02-03
Add process to generate bulk metadata file for each project 

allyhawkins:
This PR adds a process to the bulk workflow to generate one metadata file containing information about each bulk library that has been processed in that project. This results in one `bulk_quant.tsv` and one matching `bulk_metadata.tsv` where the rows of `bulk_metadata.tsv` contain the library ID's that are present in the columns of `bulk_quant.tsv`. 

To do this, I used the same grouped salmon channel that we group by project ID after performing mapping with salmon. This contains the project ID and all associated directories to salmon outputs for that project. I also am inputting the metadata file that contains the metadata information for each library like seq unit and technology. 

Inside the process I am running an Rscript that reads in the library metadata file and list of salmon directories, subsets the library metadata file to only those library ids that have been processed and are part of this project and selects the metadata columns that we would like to retain in the output. I then followed a similar format to the other metadata files that we have for single-cell libraries and added columns containing the `date_processed`, `geneome_assembly`, `workflow`, `workflow_version`, and `workflow_commit`. 

Finally, for each salmon output directory, I am grabbing the processing information from the json files in that directory that contain the `salmon_version`, `index`, and then number of reads processed and mapped. 

I then write out the file as a `tsv` file as was discussed in our multi-team planning meeting. I have attached an example of what this file would look like. 
[SCPCP000001_bulk_metadata.tsv.zip](https://github.com/AlexsLemonade/scpca-nf/files/7995535/SCPCP000001_bulk_metadata.tsv.zip)

An alternative to this, would be creating individual `json` files for each library that is included in bulk, but was unsure we wanted to have that many individual metadata files when we only have one output file for bulk.

### DIFF
--- a/bin/generate_bulk_metadata.R
+++ b/bin/generate_bulk_metadata.R
@@ -1,0 +1,134 @@
+#!/usr/bin/env Rscript
+
+# This script generates the metadata.tsv file for bulk RNA-seq libraries.
+# It outputs one metadata file per project.
+
+# import libraries
+library(optparse)
+suppressPackageStartupMessages(library(SingleCellExperiment))
+
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = c("-p", "--project_id"),
+    type = "character",
+    help = "scpca project ID",
+  ),
+  make_option(
+    opt_str = c("-s", "--salmon_dirs"),
+    type = "character",
+    help = "Path to text file containing salmon output directories, one per line."
+  ),
+  make_option(
+    opt_str = c("--library_metadata_file"),
+    type = "character",
+    help = "path to metadata file containing scpca_library_id, scpca_sample_id and associated metadata"
+  ),
+  make_option(
+    opt_str = "--metadata_output",
+    default = "metadata.tsv",
+    type = "character",
+    help = "path to metadata tsv output file"
+  ),
+  make_option(
+    opt_str = "--genome_assembly",
+    type = "character",
+    default = NA,
+    help = "genome assembly used for mapping"
+  ),
+  make_option(
+    opt_str = "--workflow_url",
+    type = "character",
+    default = NA,
+    help = "workflow github url"
+  ),
+  make_option(
+    opt_str = "--workflow_version",
+    type = "character",
+    default = NA,
+    help = "workflow version identifier"
+  ),
+  make_option(
+    opt_str = "--workflow_commit",
+    type = "character",
+    default = NA,
+    help = "workflow commit hash"
+  )
+)
+
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# checks 
+
+# check for project id
+if(is.null(opt$project_id)){
+  stop("A `project_id` is required.")
+}
+
+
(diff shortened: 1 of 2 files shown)


bin/generate_bulk_metadata.R
```diff
@@ -0,0 +1,132 @@
...
+                technology, seq_unit) |> 
+  # rename column names to match format of metadata files from other modalities
+  dplyr::rename(
+    sample_id = scpca_sample_id,
+    library_id = scpca_library_id,
+    project_id = scpca_project_id
+  ) |>
+  # add columns with processing information and date processed (same for all libraries )
+  dplyr::mutate(date_processed = lubridate::format_ISO8601(lubridate::now(tzone = "UTC"), usetz = TRUE),
+                genome_assembly = opt$genome_assembly,
```
jashapiro, line 99: I don't love having the date set here, as this may not reflect when the actual processing happened. (I probably have the same complaint for the SC data, but that is set when the RDS is made, as I recall, so that may be closer to the actual processing time). Here I am anticipating that we might add skipping of the salmon mapping to bulk as well, so individual samples might have different processing times. I will put a suggestion in the `processing_info` section for how we might get what we want out.
```suggestion
  dplyr::mutate(genome_assembly = opt$genome_assembly, 
```

allyhawkins:
Thanks for taking a look at this! I went ahead and merged the two processes that create the bulk counts file and the bulk metadata file together. Let me know if you are okay with the new process name of `group_output`.  Another option I had thought of was `group_counts_metadata`. 

I also addressed the concern about the date by grabbing it from the `meta_info.json` file from the salmon output. I tested the changes and everything worked as expected.

modules/bulk-salmon.nf
```diff
@@ -49,16 +49,20 @@ process salmon{
 
 }
 
-process group_tximport {
+process group_output {
```
jashapiro, line 52: Since you asked about this, how about `merge_bulk_quants` or something like that? `group_output` seems too generic to me, and the grouping is happening in the workflow, whereas this is taking groups and merging them.

bin/generate_bulk_metadata.R
```diff
@@ -0,0 +1,138 @@
...
+  
+  # read in json files for that library containing individual process information
+  cmd_info_file <- file.path(library_id, "cmd_info.json")
+  meta_info_file <- file.path(library_id, "aux_info", "meta_info.json")
+  
+  cmd_info <- jsonlite::read_json(cmd_info_file)
+  meta_info <- jsonlite::read_json(meta_info_file)
+  
+  date_processed <- lubridate::as_datetime(meta_info$end_time, 
+                                           format = "%a %b %d %T %Y")
```
jashapiro, line 115: This is obscure enough to be worth a comment:
```suggestion
                                           format = "%a %b %d %T %Y") // salmon date format: like "Mon Jan 03 15:13:14 2022"
```

jashapiro (approved):
This looks good to me. I had a few minor comments, and a suggestion on naming, since you asked about it!

PR #85, opened 2022-02-03
Change index and barcodes paths to public reference bucket 

### LINKED ISSUE #79: Use public indexes for workflow
Following on https://github.com/AlexsLemonade/ScPCA-admin/issues/304, we will want to update the index locations used by the workflow to be the public bucket locations.

I _think_ this will require only a single update for both using the workflows in `main.nf` and `build-index.nf`

allyhawkins:
Closes #79. Here I'm changing the directories for the `ref_dir` and the `barcode_dir` to be the public S3 bucket, `s3://scpca-references` that contain the necessary index and barcode files for the workflow. These two spots should be the only place that need to be changed, since I kept the remaining structure within the reference directory the same. All of the parameters for the indexes use the `ref_dir` as a variable in their path, so changing that one parameter changes all of them. 

I tested that this works and when running the workflow the updated index locations are shown in the configuration in tower and when looking at the command that was executed. Additionally there were no errors for missing index files and everything completed successfully.

### DIFF
--- a/config/reference_paths.config
+++ b/config/reference_paths.config
@@ -1,6 +1,6 @@
 // reference params and files 
 assembly         = 'Homo_sapiens.GRCh38.104'
-ref_dir          = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104'
+ref_dir          = 's3://scpca-references/homo_sapiens/ensembl-104'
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM! (and glad to see the earlier change to use `$ref_dir` paying off!)

PR #87, opened 2022-02-04
Skip bulk mapping step

### LINKED ISSUE #86: Allow skipping of salmon mapping in bulk
Similar to https://github.com/AlexsLemonade/scpca-nf/issues/41 and #81, we should continue our efforts to make expensive slow steps skippable by adding this functionality to the bulk workflow. Since we are unlikely to want to skip bulk but not single-cell or vice versa (an update to salmon would presumably affect both) we can consolidate the current `--rad-skip` option with this one to make a single `--skip-mapping` option. 

Alternatively, we could rename the option `--repeat-mapping`, which has a nice little benefit that we can set it to `false` by default, and then nextflow will set the value to `true` when invoked without an argument. Which is to say we would not need to do something like `nextflow run ... --rad-skip false` as we do now, but rather just `nextflow run ... --repeat-mapping`, which is a bit cleaner.

allyhawkins:
Closes #86. This PR adds the flexibility to skip the salmon mapping step in the bulk workflow, similar to skipping the `salmon alevin` mapping in the alevin-fry workflow. In doing this, I implemented the suggestion in #86 to use a parameter `--repeat_mapping` and set the default value to `false`, meaning that mapping will be skipped by default if the necessary salmon output files are present and will be repeated if using the `--repeat_mapping` flag at the command line. 

I followed a very similar structure to the steps we take to skip mapping in the fry workflow, adding the salmon directory to the metadata and then using `branch` to split the channel based on having or not having the expected output file from salmon. The branch that doesn't have the salmon files done yet are then run through the `fastp` and `salmon` processes, while the other branch skips those processes. The output from the `salmon` process is then mixed with the channel that skipped mapping before creating the final output files. 

The main difference here is that instead of simply looking for the salmon output directory to be created (like we do in the fry workflow), I am looking for the creation of the `quant.sf` file which is explicitly needed in the next steps. Do reviewers agree with this decision or prefer to just look for the output directory? 

Additionally, here I am only looking for completion of the salmon step and then if that step is completed skipping both `fastp` and `salmon`. I thought it was fairly safe to assume that if the `quant.sf` files existed, then both processes had been previously completed and we did not need a separate check for if trimming had completed but not salmon since trimming is not very time consuming. 

I tested that this works both with skipping and not skipping mapping and all looked good. I also tested the fry workflow since I had changed the use of the parameter there and that also worked as expected.

### DIFF
--- a/modules/bulk-salmon.nf
+++ b/modules/bulk-salmon.nf
@@ -24,7 +24,7 @@ process fastp{
 
 process salmon{
     container params.SALMON_CONTAINER
-    label 'cpus_8'
+    label 'cpus_12'
     tag "${meta.library_id}-bulk"
     publishDir "${meta.salmon_publish_dir}"
     input: 
@@ -110,8 +110,8 @@ workflow bulk_quant_rna {
         // create tuple of metadata map, salmon output directory to use as input to merge_bulk_quants
         quants_ch = bulk_channel.has_quants
           .map{meta -> tuple(meta,
-                       file(meta.salmon_results_dir)
-                       )}
+                             file(meta.salmon_results_dir)
+                             )}
 
         
         // run fastp and salmon for libraries that are not skipping salmon
--- a/modules/bulk-salmon.nf
+++ b/modules/bulk-salmon.nf
@@ -26,12 +26,12 @@ process salmon{
     container params.SALMON_CONTAINER
     label 'cpus_8'
     tag "${meta.library_id}-bulk"
-    publishDir "${params.outdir}/internal/salmon/"
+    publishDir "${meta.salmon_publish_dir}"
     input: 
         tuple val(meta), path(read_dir)
         path (index)
     output: 
-        tuple val(meta), path(salmon_results)
+        tuple val(meta), path(salmon_results_dir)
     script:
         salmon_results = "${meta.library_id}"
         """
@@ -39,7 +39,7 @@ process salmon{
         -l A \
         ${meta.technology == 'paired_end' ? "-1": "-r"} ${read_dir}/*_R1_*.fastq.gz \
         ${meta.technology == 'paired_end' ? "-2 ${read_dir}/*_R2_*.fastq.gz" : "" } \
-        -o ${salmon_results} \
+        -o ${salmon_results_dir} \
         --validateMappings \
         --rangeFactorizationBins 4 \
         --gcBias \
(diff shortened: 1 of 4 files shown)


modules/bulk-salmon.nf
```diff
@@ -88,23 +88,46 @@ workflow bulk_quant_rna {
     take: bulk_channel 
     // a channel with a map of metadata for each rna library to process
     main: 
-    // create tuple of (metadata map, [Read 1 files], [Read 2 files])
-        bulk_reads_ch = bulk_channel
+        bulk_channel = bulk_channel
+          // add salmon directory and salmon file location to meta 
+          .map{it.salmon_publish_dir = "${params.outdir}/internal/salmon";
+               it.salmon_results_dir = "${it.salmon_publish_dir}/${it.library_id}";
+               it.salmon_quant_file = "${it.salmon_results_dir}/quant.sf";
```
jashapiro, line 95: I don't think we should need to check for the specific file, unless we are going to check for _all_ the files we will need, which would include the `.json` files as well. In most cases I would expect the directory should be sufficient. 

I did notice though that may be "leaking" the trimmed fastq files here, which should be deleted. I think this is because at line 13 the trimmed files are going into the same directory that we are using for the output. We should update that to maybe something like the below to avoid this.
```
trimmed_reads = "${meta.library_id}_trimmed"
```

modules/bulk-salmon.nf
```diff
@@ -88,23 +88,46 @@ workflow bulk_quant_rna {
...
+        grouped_salmon_ch = salmon.out.mix(quants_ch)
             .map{[it[0]["project_id"], it[1]]}
             .groupTuple(by: 0)
 
         // create tsv file and combined metadata for each project containing all libraries
         merge_bulk_quants(grouped_salmon_ch, params.t2g_bulk_path, params.run_metafile)
 
     emit: 
-        merge_bulk_quants.out
+        merge_bulk_quants.out.bulk_counts
```
jashapiro, line 131: Is there a reason you went back to separating these outputs? Is it required? If so, I suppose we probably should rename them as you had before, even if it looks redundant!

allyhawkins: Oh yes, I forgot to mention that in the PR. I got an error that it couldn't emit a multi-channel output so had to change this. I can go back to naming them.

allyhawkins:
I went ahead and updated the changes so they only look for creation of the the salmon directory and added the `_trimmed` ending to the folder for the trimmed fastq files like you suggested. Only the salmon results should be included in that directory now.

jashapiro (approved):
LGTM! I did some testing and things seemed to work as expected. I did run into a failure for memory limits, so I thought we might bump up the config for the salmon step to match alevin. I don't think we should need to worry about larger disks, as I hope nobody is doing bulk to depths where it would matter!

Issue #80, opened 2022-01-31
Add genetic demultiplexing workflow


jashapiro:
The genetic demultiplexing workflow currently implemented in https://github.com/AlexsLemonade/alsf-scpca/tree/main/workflows/genetic-demux should be integrated into the main scpca-nf workflow. 

If possible, this may reuse the STAR index created for Cell Ranger, rather than requiring a separate addition to `build-index.nf`.

allyhawkins:
Closed by #93.

PR #92, opened 2022-02-08
Split up CPU and memory labels

jashapiro:
In preparing #80, I needed to make some tweaks to memory requests, and decided that the easiest way to accomplish that was to split up some of the coupling that we had between cpu and memory in the process labels. This may allow us to make some processing more efficient as well (though we are still subject to whims of batch scheduling to a certain extent), as we can use much less RAM for some processes like alevin-fry quant that don't require much memory, while retaining multiprocessing.

So here I have split the CPU and memory requests, and added extra labels to processes as best I thought made sense. 

I also decreased the increments by which we expand memory requests on the logic that we probably don't need to double on failure, but adding 50% memory might make more sense.

**Important questions for review!**

Do  the memory choices I made look reasonable? I did my best to look through recent runs to see that I was covering the memory usage I saw, but I know that Ally has seen some things, so I expect she may have better insights for some of the larger data sets. If you let me know which ones were most problematic, I can also test those specifically.

I left `alevin_rad` with 24GB memory, but I can imagine we might be able to get away with less? Similarly, `fry_quant` might be able to do even less than 8GB, but I'm not sure.

I expect we can decrease the `alevin_feature` memory quite a bit, as I have never seen those use much memory (the index is tiny!), but I left it with 8GB for now.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -6,7 +6,7 @@ process generate_reference{
   container params.SCPCATOOLS_CONTAINER
   // publish fasta and annotation files within reference directory 
   publishDir params.ref_dir
-  memory { 28.GB * task.attempt}
+  label 'mem_32'
   errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
   maxRetries 1
   input:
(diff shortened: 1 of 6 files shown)


allyhawkins (approved):
I think this looks good and I like splitting out the `cpu` label. I am okay with the decrease in the `alevin_rad` process to 24 gb and looked at a few of the more recent runs with the larger samples and it should be okay. The bigger problem was the disk space which we solved with the dynamic label. I also noticed that for `fry_quant_rna` the memory usage is typically around 2-4 gb so we could probably lower, but I've never tested anything that low so it makes me nervous to go to low. I would leave it at 8gb like you have it.

PR #101, opened 2022-02-15
Update to scpcaTools:v0.1.4

allyhawkins:
Updates to use the latest `scpcaTools`, v0.1.4 which incorporates the new `emptyDrops` error handling. I tested this with the default run ID's and with the run ID that I know fails `emptyDropsCellRanger` and all worked as expected.


jashapiro (approved):
LGTM!

PR #102, opened 2022-02-17
Make genetic demultiplexing optional

### LINKED ISSUE #95: Make genetic demultiplexing optional
Genetic demultiplexing is a slow and expensive process, as it requires full alignment and SNP calling in multiple bulk samples and each multiplexed library. So we should have the ability to skip it if it has already been completed for a particular library.

To do this we can follow the patterns of https://github.com/AlexsLemonade/scpca-nf/pull/77 and https://github.com/AlexsLemonade/scpca-nf/pull/87. 

While it might be nice to save the intermediate alignments to save the most costly step, unfortunately that is not a good option, as those alignments will contain patient genetic data that we do not want to keep long term. So we will be limited to repeating the whole genetic demultiplexing process in an all-or-none fashion. I expect we will want to make that a separate flag from the normal mapping: I can imagine we might want to remap RNA with a new salmon version, but don't want to do a full realignment with STAR.

jashapiro:
Here I am adding the ability to skip genetic demultiplexing if we already have results.

This seemed bit more complex than some of the earlier skipping steps, because there is a fair amount of `meta` manipulation that happens along the way through the workflow, as we are combining multiple runs and samples in ways that previous skipping steps did not. I didn't want to lose whatever information might have been added to `meta` along the way. My initial attempt was to repeat all of the manipulations along the "skip" pathway, but that seemed  redundant and error-prone, and I ran into some roadblocks along the way.

My second idea, implemented here, is to simply write out the `meta` object as a JSON file in the output directory for vireo, then read that in for any potential downstream steps. In the end I think I like this approach, and I hope you do too!

I'm doing things with a combination of groovy functions to create the JSON string and then writing the file itself in the shell script. I originally explored doing everything in groovy, but then I realized I would have to use the `publishDir` value within the workflow script itself, and that would mean that any changes to publishing location might have to happen in two places, which I didn't love.

This approach of writing the `meta` object to output directories might even be something we could consider adding at other points in the workflow to improve logging/reproducibility. Let me know what you think!

closes #95


allyhawkins (commented):
I really like how you did this with outputting the metadata and then reading it back in for the libraries that had already had mapping performed! 
The only question I had here was about why you chose to change the `publishDir` and the output directory? In the previous version of the workflow you were storing everything in `internal/vireo` and then each output directory was `meta.library_id-vireo` but here you are now adding another layer and using `internal/vireo/meta.library_id` with the `run_id` as the output id? I was just wondering if there was rationale for doing that maybe for use downstream? I wasn't sure why we would need the extra nesting there.

jashapiro:
> The only question I had here was about why you chose to change the `publishDir` and the output directory? In the previous version of the workflow you were storing everything in `internal/vireo` and then each output directory was `meta.library_id-vireo` but here you are now adding another layer and using `internal/vireo/meta.library_id` with the `run_id` as the output id? I was just wondering if there was rationale for doing that maybe for use downstream? I wasn't sure why we would need the extra nesting there.

The reason I did this was just for consistency with the other outputs. We follow this structure for the RAD files, so I was copying that. Though I notice that we do not do the same for bulk RNA... I guess the difference is that we have more than one set of reads (RNA + feature) that we generate RAD files from for those libraries, but only one set of bulk RNA reads per library, and the genetic demux should only be one per library. So I could go either way.

allyhawkins:
> The reason I did this was just for consistency with the other outputs. We follow this structure for the RAD files, so I was copying that. Though I notice that we do not do the same for bulk RNA... I guess the difference is that we have more than one set of reads (RNA + feature) that we generate RAD files from for those libraries, but only one set of bulk RNA reads per library, and the genetic demux should only be one per library. So I could go either way.

Yea I think that was part of why I was asking is because it seemed like something we didn't need to do unless there was a potential reason where we would have multiple runs per library, which doesn't seem the case here. I think I would prefer to not have that extra step of nesting the run id in the library folders since that doesn't seem to benefit us here and would be consistent with what is in bulk, another scenario where there is only going to be one run id associated per library.

allyhawkins (approved):
This looks good to me and I like the change to unnest the output folders.

PR #103, opened 2022-02-17
Increase memory for bulk workflow

allyhawkins:
In running the samples for the Murphy project, many of them are between 30-40 GB of FASTQ files for bulk samples (compared to previously accommodating for processing ~6 GB files). A few of the samples were failing the `fastp` process and retrying and taking multiple hours and I noticed an increase in their memory usage of 2-3GB up to 6-8 GB. Because `fastp` is normally a much faster process for most samples, I increased the memory to 8 GB for that process and it appears to be working better for all of the difficult samples I had found. 

I also was having issues where `salmon` was failing even after two tries using 32 GB. Without making too many drastic changes, I just increased the number of tries that were allowed from 2-3 and noticed that this also helped. I am running two more libraries right now that had previously had difficulties to make sure they can make it through before we tag a release and process the project as a whole, but wanted to get this up in a PR so @jashapiro could comment before he leaves.

### DIFF
--- a/modules/bulk-salmon.nf
+++ b/modules/bulk-salmon.nf
@@ -9,7 +9,7 @@ process fastp{
     input: 
         tuple val(meta), path(read1), path(read2)
     output: 
-        tuple val(meta), path(trimmed_reads), path(fastp_report)
+        tuple val(meta), path(trimmed_reads)
     script: 
         trimmed_reads = "${meta.library_id}_trimmed"
         fastp_report = "${meta.library_id}_fastp.html"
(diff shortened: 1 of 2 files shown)


jashapiro:
This looks fine, but maybe we should just abandon my attempts to get cute with memory usage at rerun in https://github.com/AlexsLemonade/scpca-nf/pull/92: https://github.com/AlexsLemonade/scpca-nf/blob/96cbdfd9c052f367d7a8ca86b93c2faf13a55589/config/profile_awsbatch.config#L18-L25 and just go back to doubling on failure. In other words, for `mem_24`, just `memory = {24.GB * task.attempt }`

jashapiro:
> I thought about that too... but tried the three times first and it seemed to work. Are you saying go back to `memory = {24.GB * task.attempt }` with retry limit of 2? or keep it at 3?

I would leave it at 3. It should be rare that we get there, so there is little cost to leaving it.

modules/bulk-salmon.nf
```diff
@@ -25,7 +26,7 @@ process fastp{
 process salmon{
     container params.SALMON_CONTAINER
     label 'cpus_12'
-    label 'mem_24'
+    label 'mem_32'
```
jashapiro, line 29: With the change in https://github.com/AlexsLemonade/scpca-nf/pull/103/commits/1c35785e1f8344ef8ea88c7d611a0d1fbd11f0c2, I think we can go back to  `mem_24` here. The max with 3 retries then is 72GB, which is greater than the previous max of `mem_32` with a 16 GB step, whcih was only 64GB.
```suggestion
    label 'mem_24'
```

jashapiro (approved):
LGTM! I would just bump the default salmon back down, as discussed below.

PR #104, opened 2022-02-18
Increase the number of retries allowed to 2

allyhawkins:
In #103, we changed the ternary operator to finish if the number of retries was less than 3 instead of the previous 2. However, we missed changing the number of `maxRetries` to be 2 rather than 1. This meant that it never actually made it to the third try. There were 3 samples that seemed to have slipped through that needed the the third try and were unable to complete successfully. 

I tested this change for the 3 remaining samples that had failed and with the 3 retries the workflow does complete.


jaclyn-taroni (approved):
Let's go!

Issue #88, opened 2022-02-04
Update README to include instructions on how to run pipeline for external users 


allyhawkins:
Currently the main [README file](https://github.com/AlexsLemonade/scpca-nf#readme) includes instructions on how to run the pipeline either locally or with `batch` on AWS. In an effort to transition the workflow to be usable by external users, we should start by updating the main README file by adding instructions on how to set up and run the pipeline on your own. 

Some things that should be included in these instructions are: 

- Creating the run metadata file 
- What parameters need to/ can be adjusted 
- How to set up the profile that you need based on your system*
- Creating your own cellranger and spaceranger containers if planning to run the spatial workflow  

*When looking at how [`nf-core` sets their options for profiles](https://nf-co.re/scrnaseq/usage#profile), I noticed that they have a set of basic profiles and then a series of [custom configs that are set up for different institutions](https://github.com/nf-core/configs/tree/master/conf) and all loaded when running. The user then inputs the profile they want at the command line, or at least that is what I am gathering when looking at the documentation and the repo. I think we could start with documentation on how to set up a profile to start and then go from there if we want to add any pre-set profiles based on what user's want. 

We also may want to break out this issue into smaller steps once we have a better idea of what else we think would be important to include in the instructions, but just wanted to get our thoughts and what we've discussed so far written down.

jaclyn-taroni:
Let's outline this by the end of this sprint, because we expect there might be gaps once we do that.

allyhawkins:
Closed by #106.

PR #106, opened 2022-02-24
Start addition of instructions for external users to README

allyhawkins:
Begins to address #88. 

Here I have started to adjust the main README file to include an outline for the instructions we will want to include for external users for `scpca-nf`. I decided to break this up into two main sections, the first section includes running the workflow with the default configurations and the second is running the workflow for external users. The first section takes much of the original contents of the README and then adds in how to run the workflow for production runs. 

In the second section, I included sections on preparing the metadata file, setting up the configuration for AWS, adding an additional profile if not using AWS or local, choosing an index, adjusting optional parameters, and then special considerations for running spatial libraries. I only partially filled in the sections to get an idea for how things would flow together and if it made sense in this order since right now our focus is on getting an outline. Once we decide which sections we would like to keep, we can file issues to fill in each section separately. 

The area that I think could use the most feedback here is how much information we want to include on the index files. Do we want to include instructions on how to create your own index files? I think if we do, we also need to include the structure of the reference directory considering all of the reference files use the `ref_dir` parameter to define the paths to all the reference files. 

Additionally, do we also want to include a section on using `build-index.nf`? 

The other thing that I want to note is that currently, we have all three index files required to run the workflow, the splici index, bulk index, and cellranger index. If any of those are not able to be found, the workflow will fail. This might be a future thought for a scenario in which someone wants to provide their own index, but I think it might be reasonable to make those parameters optional and we add in a parameter for which type of sequencing you are analyzing (single-cell/bulk/spatial/hashed).


jashapiro:
> The other thing that I want to note is that currently, we have all three index files required to run the workflow, the splici index, bulk index, and cellranger index. If any of those are not able to be found, the workflow will fail.

I would not expect the cellranger index to be required unless we get to the spaceranger part of the workflow, and the same for the bulk index? If this is not the case, we should be able to modify the workflow to avoid this.

allyhawkins:
> I would not expect the cellranger index to be required unless we get to the spaceranger part of the workflow, and the same for the bulk index? If this is not the case, we should be able to modify the workflow to avoid this.

Nextflow looks for any files indicated in the config files when setting up the run. So if those files aren't there when they run the workflow it will fail, even if they aren't running the bulk or spatial workflows.

jashapiro (commented):
This looks like a good start, so thanks!

Some initial thoughts and comments:

- Do  we want external use to be a separate doc? I tend to think that might be a bit more straightforward, as these docs may get more complex and it may be easier to have those instructions separate from the README where somebody might come in just looking at the repo to see what we did.

- I think we want to encourage people to create a `.config` file to override our defaults, which is where they can and should incorporate their own profile settings, but they can also include the `run_metafile` and `outdir` there if they like.
  - The reason for this is that we want people to use the released versions of the workflow proper, so we want people to use them direct from github rather than downloading and modifying
  - I had some thought that we could tell people to fork the repo for their changes, but I think that will introduce more trouble, and it becomes harder for us to ensure they did not change settings in the repo itself.

- I think we do NOT want people to change indices, at least not for the case where we are hoping to incorporate samples into ScPCA, so I would not include those instructions for now. 
  - If people want to use the repo for their own purposes, then that may come up, but that is a separate use case than the one I think we should be targeting to start. 

Other comments below...

jaclyn-taroni:
I'm going to review this _after_ the response to @jashapiro's review. Presumably I'm most useful reviewing this from the perspective of whether or not I think I'd be able to set something up.

allyhawkins:
Thanks for the feedback on this @jashapiro. I went ahead and separated out the external user instructions into its own document but incorporated a link to that second doc in the main readme. I also deleted the index section and cleaned up the configuration section to talk about creating your own configuration file. 

One thought I had was that I think we may want to create a template config file that external users could use? Because we still want them to be able to use the containers config file and the references config file, it could be helpful to have a template where they fill in the parameters like `outdir` and `run_metafile` and then fill in a profile but that config file still includes the reference and container configs for them to use. 

My other question was how to break up the section on config files in the instructions. Right now I included sections for creating profiles, adding in parameters, and then a section on AWS because I thought that would take some extra explanation. I don't feel strongly about where the AWS information should go though so if you have other suggestions please let me know.

jashapiro:
> One thought I had was that I think we may want to create a template config file that external users could use? Because we still want them to be able to use the containers config file and the references config file, it could be helpful to have a template where they fill in the parameters like `outdir` and `run_metafile` and then fill in a profile but that config file still includes the reference and container configs for them to use.

Mentioned in the meeting, but to record here, I like the idea of a template config file. I _think_ that we should be able to get away with this being a minimal file, with only the params that are likely to be changed by a user. I just confirmed that if run with a command like `nextflow run AlexsLemonade/scpca-nf -r "v0.2.3" -c test.config`, nextflow will correctly fall back to the parameters defined in the repository `nextflow.config` file for undefined parameters (like cellranger images).

README.md
```diff
@@ -1,8 +1,11 @@
-## scpca-nf
+## Running scpca-nf for the ScPCA portal 
 
-This repo holds the scripts to run the nextflow workflow to process samples as part of the ScPCA project. Fastq files for each sample are processed using [alevin-fry](https://alevin-fry.readthedocs.io/en/latest/). All samples are aligned, using selective alignment, to an index with transcripts corresponding to spliced cDNA and to intronic regions, denoted by alevin-fry as `splici`. 
+This repo holds the scripts to run the nextflow workflow to process samples as part of the ScPCA project.
+Fastq files for single-cell and single-nuclei RNA-seq samples are processed using [alevin-fry](https://alevin-fry.readthedocs.io/en/latest/). All samples are aligned, using selective alignment, to an index with transcripts corresponding to spliced cDNA and to intronic regions, denoted by alevin-fry as `splici`.
```
jashapiro, line 4: Formatting: sne sentence per line 
```suggestion
Fastq files for single-cell and single-nuclei RNA-seq samples are processed using [alevin-fry](https://alevin-fry.readthedocs.io/en/latest/).
All samples are aligned, using selective alignment, to an index with transcripts corresponding to spliced cDNA and to intronic regions, denoted by alevin-fry as `splici`. 
```

README.md
```diff
@@ -15,4 +18,12 @@ You can then run the same workflow with the `batch` profile, which has been conf
...
\ No newline at end of file
+```
+
+When running the workflow for a project or group of samples that is ready to be released on ScPCA portal, please use the tag for the latest release: 
+
+```
+nextflow run AlexsLemonade/scpca-nf -r v0.2.3 -profile batch --project SCPCP000000
+```
+
+To process samples that are not part of the ScPCA project, please see the [instructions on using `scpca-nf` with external data](external-data-instructions.md).
```
jashapiro, line 29: I think we may want to emphasize that it is not just the samples but the infrastructure...
```suggestion
Note that the default settings of the workflow is configured for the ALSF Childhood Cancer Data Lab computational infrastructure. 
To process samples that are not part of the ScPCA project, please see the [instructions on using `scpca-nf` with external data](external-data-instructions.md).
```

external-data-instructions.md
```diff
@@ -0,0 +1,70 @@
...
+In order to use `scpca-nf` to process your own data, you will need to make sure you have the following installed: 
+
+- [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
+- [Docker](https://docs.docker.com/get-started/#download-and-install-docker) (installed either locally or on an HPC)
+
+Once you have setup your environment and completed the below steps you will be able to setup your run as follows, adding any additional optional parameters that you may choose: 
+
+```
+nextflow run AlexsLemonade/scpca-nf \
+ -r 0.2.3 \
```
jashapiro, line 27: It is probably worth adding a comment that mentions that the `-r` flag is required (or highly recommended).

external-data-instructions.md
```diff
@@ -0,0 +1,70 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## How to use `scpca-nf` as an external user 
+
+In order to use `scpca-nf` to process your own data, you will need to make sure you have the following installed: 
+
+- [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
+- [Docker](https://docs.docker.com/get-started/#download-and-install-docker) (installed either locally or on an HPC)
+
+Once you have setup your environment and completed the below steps you will be able to setup your run as follows, adding any additional optional parameters that you may choose:
```
jashapiro, line 23: I want to give a sense of what the setup referred to "below" is to reduce the cognitive load. I think the wording still needs work. We might also not want to talk about parameters at this stage, as we might expect (encourage) people to set those in the config file?

```suggestion
You will also need to create a metadata file and a nextflow configuration file (see below).
Once you have set up your environment and created these files you will be able to start your run as follows, adding any additional optional parameters that you may choose: 
```

external-data-instructions.md
```diff
@@ -0,0 +1,70 @@
...
+You will need the following columns to be present in your metadata file... 
+
+### Configuring `scpca-nf` for your environment
+
+By default, the workflow is set up to run in a local environment. 
+If choosing to run the workflow locally, you will only need to provide the path to the metadata file and output directory where you would like the results to be stored. 
+
+```
+nextflow run AlexsLemonade/scpca-nf \
+ -r 0.2.3 \
```
jashapiro, line 45: I think this code snippet may be redundant?

jashapiro (commented):
This organization makes a lot more sense to me, thanks! I am not approving not because I think there is anything major left in the outline, but simply to prevent merging before we do the additional parts (as stacked PRs)

allyhawkins:
We have incorporated all of the individual sections now so that this document is complete. 
This should be ready for a look from @jaclyn-taroni.

external-data-instructions.md
```diff
@@ -0,0 +1,226 @@
...
+# How to use `scpca-nf` as an external user 
+
+In order to use `scpca-nf` to process your own data, you will need to make sure you have the following installed: 
+
+- [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
+- [Docker](https://www.nextflow.io/docs/latest/docker.html) or [Singlularity](https://www.nextflow.io/docs/latest/singularity.html) (installed either locally or on an HPC)
+
+You will also need to have files organized so that all the sequencing files for each run are in their own directory or folder.
+Any sequencing runs that contain multiple samples must be demultiplexed and FASTQ files must be placed into separate distinct folders. 
+If a run has been re-sequenced for any reason (e.g. to increase coverage), all FASTQ files should be combined into the same folder.
```
jaclyn-taroni, line 26:
```suggestion
If a run has been re-sequenced for any reason (e.g., to increase coverage), all FASTQ files should be combined into the same folder. 
```

external-data-instructions.md
```diff
@@ -0,0 +1,226 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# How to use `scpca-nf` as an external user 
+
+In order to use `scpca-nf` to process your own data, you will need to make sure you have the following installed: 
+
+- [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
+- [Docker](https://www.nextflow.io/docs/latest/docker.html) or [Singlularity](https://www.nextflow.io/docs/latest/singularity.html) (installed either locally or on an HPC)
+
+You will also need to have files organized so that all the sequencing files for each run are in their own directory or folder.
```
jaclyn-taroni, line 24: I don't think we will have defined what a run is by the time I've gotten here, right? I think we mean file/set of files.

external-data-instructions.md
jaclyn-taroni: But I'm fairly sure we mean library here?

external-data-instructions.md
```diff
@@ -0,0 +1,226 @@
...
+**Note:** `scpca-nf` is under active development.
+We strongly encourage you to use a release tagged version of the workflow, set here with the `-r` flag.
+Released versions can be found on the [`scpca-nf` repo releases page](https://github.com/AlexsLemonade/scpca-nf/releases).
+
+## Prepare the metadata file 
+
+Using `scpca-nf` requires a metadata file as a TSV (tab separated values) file, where each sequencing run to be processed is a row and columns contain associated information about that run. 
+
+For each run, you will need to provide a Run ID (`scpca_run_id`), library ID (`scpca_library_id`), and sample ID (`scpca_sample_id`). 
+Each row should contain a unique run ID, corresponding to a sequencing run or set of FASTQ files.
```
jaclyn-taroni, line 49: The information about set of files needs to get moved up.

external-data-instructions.md
```diff
@@ -0,0 +1,226 @@
...
+
+In order to use `scpca-nf` to process your own data, you will need to make sure you have the following installed: 
+
+- [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
+- [Docker](https://www.nextflow.io/docs/latest/docker.html) or [Singlularity](https://www.nextflow.io/docs/latest/singularity.html) (installed either locally or on an HPC)
+
+You will also need to have files organized so that all the sequencing files for each run are in their own directory or folder.
+Any sequencing runs that contain multiple samples must be demultiplexed and FASTQ files must be placed into separate distinct folders. 
+If a run has been re-sequenced for any reason (e.g. to increase coverage), all FASTQ files should be combined into the same folder. 
+
```
jaclyn-taroni, line 27: Should we instead just say "you have to organize your files" and then add a section (after metadata, perhaps) about preparing your file organization?

external-data-instructions.md
```diff
@@ -0,0 +1,226 @@
...
+In most Batch queue setups, each AWS compute node has a fixed amount of disk space. 
+We found it useful to have two queues: one for general use and one for jobs that may require larger amounts of disk space. 
+The two compute environments use the same AMI, but use [Launch Templates](https://docs.aws.amazon.com/batch/latest/userguide/launch-templates.html) to configure the nodes on launch with different amounts of disk space. 
+Currently, our default queue is configured with a disk size of 128 GB for each node, and our "bigdisk" queue has 1000 GB of disk space.
+The queue used by each process is determined by Nextflow labels and associated profile settings.
+
+The Data Lab's [AWS Batch config file](https://github.com/AlexsLemonade/scpca-nf/blob/main/config/profile_awsbatch.config) may be helpful as a reference for creating a profile for use with AWS, but note that the queues and file locations listed there are not publicly available, so these will need to be set to different values your own profile.
+
+## Repeating the mapping step
```
jaclyn-taroni, line 169:
```suggestion
## Repeating mapping steps
```

jaclyn-taroni (commented):
My main concern at the moment is that some of the context that would probably be helpful for file set up is under the metadata section. I would also naively expect to see a section explicitly on file organization if I am trying to run a workflow for the first time. So even though it's not particularly complicated, and is mostly governed by the metadata file as I understand it... it still seems like it is missing?

external-data-instructions.md
jaclyn-taroni: Actually I guess I mean take the first part of the metadata section and split it out into its own thing.

allyhawkins:
I did a bit of reorganization based on your comments @jaclyn-taroni. Mainly I pulled out the description of the files into a file organization section, where I attempted to describe how you would deal with different scenarios of combining vs. separating files into folders and then labeling that folder with what becomes the run id. I then linked that section to the metadata section where we describe how to actually fill in the tables with `run_id`, `library_id`, and `sample_id`. I was having trouble fully separating file organization from the metadata table since they are in part kind of dependent on each other. Let me know if this is more clear or if you have any thoughts on other ways to separate them or reorder the sections.

external-data-instructions.md
```diff
@@ -0,0 +1,231 @@
...
+
+See the [Nextflow documentation](https://www.nextflow.io/docs/latest/config.html) and the below sections for more detail on creating your own configuration file.
+
+### Setting up a profile in the configuration file
+
+Local running may be sufficient for small jobs or testing, but you will most likely want to run your workflow in a high performance computing environment (HPC), such as an institutional computing cluster or on a cloud service like AWS.
+To do this, we recommend using [Nextflow profiles](https://www.nextflow.io/docs/latest/config.html#config-profiles) to encapsulate settings like the [`executor`](https://www.nextflow.io/docs/latest/executor.html) that will be used to run each process and associated details that may be required, such as queue names or the container engine (i.e. [Docker](https://www.nextflow.io/docs/latest/docker.html) or [Singularity](https://www.nextflow.io/docs/latest/singularity.html)) your system uses.
```
jaclyn-taroni, line 142:
```suggestion
To do this, we recommend using [Nextflow profiles](https://www.nextflow.io/docs/latest/config.html#config-profiles) to encapsulate settings like the [`executor`](https://www.nextflow.io/docs/latest/executor.html) that will be used to run each process and associated details that may be required, such as queue names or the container engine (i.e., [Docker](https://www.nextflow.io/docs/latest/docker.html) or [Singularity](https://www.nextflow.io/docs/latest/singularity.html)) your system uses.
```

external-data-instructions.md
```diff
@@ -0,0 +1,231 @@
...
+The library ID will be unique for each set of cells that have been isolated from a sample and have undergone droplet generation. 
+For single-cell/single-nuclei RNA-seq runs, the library ID should be unique for each sequencing run.
+For libraries that have corresponding CITE-seq, they should share the same library ID as the associated single-cell/single-nuclei RNA-seq run, indicating that the sequencing data has been generated from the same group of cells. 
+
+Finally, the sample ID will indicate the unique tissue or source from which a sample was collected. 
+If you have two libraries that have been generated from the same original tissue, then they will share the same sample ID. 
+
+For more information on understanding the difference between library and sample IDs, see the [FAQ on library and sample IDs in the ScPCA portal documentation](https://scpca.readthedocs.io/en/latest/faq.html#what-is-the-difference-between-samples-and-libraries). 
+
+Before using the workflow with data that you might plan to submit to ScPCA, please be sure to obtain a list of sample identifiers to use for your samples from the Data Lab.
```
jaclyn-taroni, line 69:
```suggestion
**Before using the workflow with data that you might plan to submit to ScPCA, please be sure to obtain a list of sample identifiers to use for your samples from the Data Lab.**
```

jaclyn-taroni (approved):
The updated organization is an improvement in my opinion! Returning some minor style comments.

PR #108, opened 2022-02-25
Split process settings from AWS profile

### LINKED ISSUE #107: Pull default process settings from profile
We currently have the memory & cpu requirements set within the aws profile, but this means that anybody who needs to set a new profile will have to define all of those on their own. My initial reading of a warning in the nextflow docs (the !Danger box [here](https://www.nextflow.io/docs/latest/config.html#config-profiles)) had me thinking it had to be that way, but nf-core sets `process` scope elements outside and within profiles just fine, so we ought to be able to follow their lead. If this works, it should be 

We may want to add `max_cpus` and `max_memory` flags as well as part of this, again echoing nf-core, but I am not sure if that is really needed. 

For reference: https://nf-co.re/usage/configuration

jashapiro:
Here I am pulling the `process` directives, including settings for most of the labels we use, out to a separate file, and unnesting them from the `batch` profile.

This allows them to be reused if others add a separate profile (for slurm, say), without having to recreate all of the various levels of cpu and memory requirements. 

I left the disk labels in the AWS config (adding a bit of commentary), as those seem to be more specific to AWS, as what they do is set queues. Other systems may not require modifying disk sizes, and if they do they may specify it differently.

Also added a change from `bucketDir` to `workDir`, which is the only option in the docs. In the other direction, I added a `docker.userEmulation` option, copied from nfcore, for local running. This should solve potential problems with files getting set as owned by root when running locally).

closes #107

### DIFF
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -1,0 +1,24 @@
+process{
+  memory = { 4.GB * task.attempt }
+  
+  errorStrategy = { task.attempt < 3 ? 'retry' : 'finish' }
+  maxRetries = 2
+  maxErrors = '-1'
+
+  withLabel: mem_8 {
+    memory = { 8.GB * task.attempt }
+  }
+  withLabel: mem_16 {
+    memory = { 16.GB * task.attempt }
+  }
+  withLabel: mem_24 {
+    memory = { 24.GB * task.attempt }
+  }
+  withLabel: mem_32 {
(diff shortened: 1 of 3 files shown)


allyhawkins (approved):
This looks good to me and I tested it and didn't see any issues. Thanks for doing this!

Issue #110, opened 2022-02-28
scpca-nf user instructions: metadata file section


allyhawkins:
As part of the user instructions for `scpca-nf`, we will need to update the section on creating a metadata file. This section should detail the necessary columns needed in the metadata file and instructions on what to include in each column. 

This should also include instructions on how to obtain `scpca_run_id`, `scpca_sample_id`, and `scpca_library_id`. As discussed in our planning meeting, right now we plan on telling users to use a set of ID's that we provide and based on user feedback and how frequently others are using this we may adjust this approach. 

These instructions should be included as a branch on the branch `allyhawkins/outline-scpca-nf-instructions` from #106 so we can merge all the instructions together.

jashapiro:
Just a note that this will not need to have every field in our library file. The only fields the workflow looks at are defined in main.nf:
https://github.com/AlexsLemonade/scpca-nf/blob/3c1004dee87df56618f57308c5a3c0868a1b72bc/main.nf#L45-L62

Notably absent are things like the submitter_id, alt_ids and md5 files.

jashapiro:
Looking at this again. We don't actually use the `files` field in the pipeline anymore. I'm going to remove it from that initial step, so that we should be able to leave it out in the example metadata file.

allyhawkins:
Just noting that we do use it in the spaceranger workflow so we should address that you will need to add it when we add the spaceranger section in #114.

jashapiro:
> Just noting that we do use it in the spaceranger workflow so we should address that you will need to add it when we add the spaceranger section in #114.

Oops, I missed that. Never mind!
We might say that filling it in isn't required unless using spaceranger. Longer term, we might see if we can remove that dependency and just get the file list programmatically as we do elsewhere.

jashapiro:
closed by #118

PR #115, opened 2022-03-01
Add template config for external users 

### LINKED ISSUE #111: Create template config file for external users
Originally mentioned in https://github.com/AlexsLemonade/scpca-nf/pull/106#issuecomment-1054594618, we should have a simple template config file that contains the minimal content that would need to be updated for users to run in another environment that is not a local environment. This would include a section with the parameters that we would suggest editing (`run_metafile` and `outdir`) and  a section to add their own profile. 
We might also consider adding a slurm profile to this config as a baseline for people to add to if we anticipate most users to be using slurm.

allyhawkins:
Closes #111. 

Here I am adding in a template config file that can be used as a guide for external users to set up their own configuration settings. I am leaving this file in the root directory, but I had also thought about moving it to the `config` folder. If anyone has opinions on that, I'm happy to move it. 

The first section includes the parameters that they would need to alter, specifically the `run_metafile` and the `outdir`. I did not include other parameters as those are not something that everyone would need to change. 
I did include the parameter for the spatial containers with a note that the spaceranger container would only need to be updated in the event that the user is using the spatial transcriptomics workflow. Are we okay with including this here? 

The second section contains the `profiles` section which will need to be updated depending on how they plan to run the workflow. Here, we are showing an example of running with slurm. Only the executor and queue are included as the allocation of resources (like mem and cpu) are loaded through `process_base.config`. Is there anything here that you think I am missing that we should be sure to include? 

The last section is to enable singularity in the event that the environment they would like to run the workflow on uses singularity rather than docker. There are instructions in the comments to remove that section if they are going to be using docker. 

I believe I covered all of the necessary items that would need to be changed for users trying to run the workflow externally in a different setting, but if there are any obvious things that I am missing please let me know. Also please take a look at the instructions or comments in the file to make sure that they are clear.

### DIFF
--- a/examples/external_user_template.config
+++ b/examples/external_user_template.config
@@ -4,37 +4,42 @@
 params {
 
   // Input data table
-  run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+  run_metafile = 'example_metadata.tsv'
 
-  // Output base directory
-  outdir = "s3://nextflow-ccdl-results/scpca/processed"
+  // Output base directory for results
+  outdir = "scpca_out"
 
   // which run ids to process
   // by default this is set to all, but to change to a specific subset input a comma separated list e.g., "SCPCR000001,SCPCR000002"
   run_ids = "All"
 
-  // Spaceranger container, only needs to be updated if using spatial transcriptomics workflow
-  SPACERANGER_CONTAINER = '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-spaceranger:1.3.1'
+  // Spaceranger container, only required if using spatial transcriptomics workflow
+  SPACERANGER_CONTAINER = ''
 
 }
 
-// profile settings for executing the workflow, these are specific to your setup 
-// this profile should be adjusted based on where you plan to run this workflow, here we provide an example of using slurm 
-// for more information on setting up profiles see https://www.nextflow.io/docs/latest/config.html#config-profiles
+// profile settings for executing the workflow
+// Here we provide an example using slurm as the executor system. 
+// For information about available process executors,  and their options, see https://www.nextflow.io/docs/latest/executor.html
+// This profile should be adjusted based on the details of your system and can be invoked at runtime with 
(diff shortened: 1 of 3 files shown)


examples/external_user_template.config
```diff
@@ -0,0 +1,40 @@
...
+
+params {
+
+  // Input data table
+  run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+
+  // Output base directory
+  outdir = "s3://nextflow-ccdl-results/scpca/processed"
+
+  // which run ids to process
```
jashapiro, line 12: A note that while this parameter is `run_ids`, it can be used to refer samples or libraries as well. We might want to add a comment to that effect.

jashapiro (commented):
Thanks for starting this! I think you got the major settings that we want people to set. There are a number of comments/suggestions below about organization and what we might want to include as example values (which can mostly be summarized as "not our internal settings").

The one other thing I think we should include at the top is a usage example, which I think can be as simple as:
```
nextflow run AlexsLemonade/scpca-nf -c user_template.config -profile slurm
```

(I simplified the file name above too; I think the current name is a bit too long, but I don't like what I chose either very much.)

allyhawkins:
Thanks for these really helpful suggestions @jashapiro. I changes the file name per your suggestion. I can't really think of anything that's better and also still short. I also added in an example at the top of using this config file with the workflow and made some minor wording adjustments throughout, adding a description of using either run_ids, library_ids, or sample_ids for `params.run_ids`. 

Thanks for adding in the use of the `$SCRATCH` directory. I was on the fence on adding that in, but I think it makes sense to have with a note. 

As for the example metadata file, I was planning on adding that in as part of working on #110 and will be sure that the file name matches the one that we used here.

jashapiro (approved):
Looks good! I expect this will require updates in the future, but this seems like a good place to start!

jashapiro:
Merging this so I can refer to it in docs more easily... I expect changes, but we know it works, at least partially (slurm mysteries remaining)

PR #118, opened 2022-03-02
Add instructions for metadata section of user instructions

### LINKED ISSUE #110: scpca-nf user instructions: metadata file section

allyhawkins:
Closes #110. 

Here I am adding the content for the metadata instructions section to the user instructions. I included a table that describes all of the columns that are needed to run the workflow with notes about columns that are only required for either CITE or spatial. 
- Should we separate these out in a different way so that they aren't all in one giant table? 
- Are the descriptions clear on what should be included in each column? 

I also included a note that the user should obtain SCPCX ids from us rather than provide them on their own. In doing this I also thought we might want to add contact information but wasn't quite sure what contact information we should add, if any? 

Also, I know in the discussion in #110, there was talk of removing the `submitter_id` column, however right now we are having them use ids that we provide so I was thinking there needs to be some way to keep track of their sample identifiers vs. the SCPCX sample identifiers. 

Are there any other things from this section that you think are missing from the description and instructions? 

I also included an example metadata file which I filled in with dummy data right now. I will file an issue for actually populating the example file with the real data since I think deciding what data we want to use is a separate issue. I did include more than just a single-cell example since I thought that could be useful. I added an example of having CITE-seq and then also spatial so users could use that as reference. I'm happy to change any of the dummy file names or paths, I just filled in something so we would have somewhere to start. 

As a final note, this PR is stacked on #106.


external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
 ### Prepare the metadata file 
 
 Using `scpca-nf` requires a metadata file where each library to be processed is a row and columns contain associated information about that library. 
-You will need the following columns to be present in your metadata file... 
+You will need to create a tab separated file with the following columns: 
+
+| column_id       | contents                                                       |
+|-----------------|----------------------------------------------------------------|
+| `scpca_run_id`    | Run ID in the form `SCPCR000000`                               |
```
jashapiro, line 41: I think we may want to separate which parts are required for running the workflow from the parts that are required for submission to ScPCA. What I mean here is that in this table, there is no requirement that the `scpca_run_id` be a number of that form; that is just for us.  People can use whatever id they want to start testing. 

What should be here (or above) is maybe a bit more information about what each of these ids is referring to. What is a "run" in our parlance? What is a "library", what is a "sample"? These should probably be defined before the table itself.

In that section before the table, we can also include the note about providing ids.
```suggestion
| `scpca_run_id`    | A unique run ID                           |
```

external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
...
+| `seq_unit`        | Sequencing unit (likely one of: `cell`, `nucleus`, `bulk`, or `spot`)           |
+| `feature_barcode_file` | path/uri to directory containing the feature barcode sequences (only applicable for CITE-seq)  |	
+| `feature_barcode_geom` | A salmon `--read-geometry` layout string/ See https://github.com/COMBINE-lab/salmon/releases for details (only applicable for CITE-seq) |
+| `slide_section`   | The slide section for spatial transcriptomics samples               |
+| `slide_serial_number` | The slide serial number for spatial transcriptomics samples     |
+| `files_directory` | path/uri to directory containing fastq files from run (unique per run) |
+| `files`           | All sequencing files in the run folder, `;`-separated (only applicable for spatial transcriptomics)  |
+
+The ids that should be used for `scpca_run_id`, `scpca_library_id`, and `scpca_sample_id` will be provided to each user based on the number and types of samples that are being processed as to avoid overlap with existing sample identifiers. 
+Before using the workflow, please be sure to obtain the list of sample identifiers to use for your samples from the Data Lab.
```
jashapiro, line 56: I think we might encourage people to test a bit before we assign them ids.
```suggestion
Before using the workflow with data that you might plan to submit to ScPCA, please be sure to obtain a list of sample identifiers to use for your samples from the Data Lab. 
```

external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
...
+|-----------------|----------------------------------------------------------------|
+| `scpca_run_id`    | Run ID in the form `SCPCR000000`                               |
+| `scpca_library_id` | Library ID in the form `SCPCL000000` <br> The numeric portion will equal the `scpca_run_id` for RNA-seq data, but for associated feature data (CITE-seq) this will be shared with the associated RNA-seq data that it is paired with.|
+| `scpca_sample_id` | Sample ID in the form `SCPCS000000`                            |
+| `submitter_id`    | Original sample identifier defined by user                     |
+| `submitter`       | Name of user submitting name/id                                 |
+| `technology`      | Sequencing/library technology used <br> For single-cell/single-nuclei libraries use either `10Xv2`, `10Xv2_5prime`, `10Xv3`, or `10Xv31`. <br> For CITE-seq libraries use either `CITEseq_10Xv2`, `CITEseq_10Xv3`, or `CITEseq_10Xv3.1` <br> For bulk RNA-seq use either `single_end` or `paired_end`. <br> For spatial transcriptomics use either `visium_v1` or `visium_v2`      |
+| `seq_unit`        | Sequencing unit (likely one of: `cell`, `nucleus`, `bulk`, or `spot`)           |
```
jashapiro, line 47: I think the "likely" was there for our internal use. We don't want people putting anything else in there with the current workflow!
```suggestion
| `seq_unit`        | Sequencing unit (one of: `cell`, `nucleus`, `bulk`, or `spot`)           |
```

external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
...
+| `feature_barcode_file` | path/uri to directory containing the feature barcode sequences (only applicable for CITE-seq)  |	
+| `feature_barcode_geom` | A salmon `--read-geometry` layout string/ See https://github.com/COMBINE-lab/salmon/releases for details (only applicable for CITE-seq) |
+| `slide_section`   | The slide section for spatial transcriptomics samples               |
+| `slide_serial_number` | The slide serial number for spatial transcriptomics samples     |
+| `files_directory` | path/uri to directory containing fastq files from run (unique per run) |
+| `files`           | All sequencing files in the run folder, `;`-separated (only applicable for spatial transcriptomics)  |
```
jashapiro, line 53:
```suggestion
| `files`           | All sequencing files in the run folder, `;`-separated (only required for spatial transcriptomics)  |
```

external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
...
-You will need the following columns to be present in your metadata file... 
+You will need to create a tab separated file with the following columns: 
+
+| column_id       | contents                                                       |
+|-----------------|----------------------------------------------------------------|
+| `scpca_run_id`    | Run ID in the form `SCPCR000000`                               |
+| `scpca_library_id` | Library ID in the form `SCPCL000000` <br> The numeric portion will equal the `scpca_run_id` for RNA-seq data, but for associated feature data (CITE-seq) this will be shared with the associated RNA-seq data that it is paired with.|
+| `scpca_sample_id` | Sample ID in the form `SCPCS000000`                            |
+| `submitter_id`    | Original sample identifier defined by user                     |
+| `submitter`       | Name of user submitting name/id                                 |
```
jashapiro, line 45: I believe these are both optional, functionally. I'm not sure whether we still use submitter at all, except to run all samples from a given submitter.
```suggestion
| `submitter_id`    | Original sample identifier defined by user (for reference only; optional)    |
| `submitter`       | Name of user submitting name/id           (optional)                  |
```

external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
...
+| `technology`      | Sequencing/library technology used <br> For single-cell/single-nuclei libraries use either `10Xv2`, `10Xv2_5prime`, `10Xv3`, or `10Xv31`. <br> For CITE-seq libraries use either `CITEseq_10Xv2`, `CITEseq_10Xv3`, or `CITEseq_10Xv3.1` <br> For bulk RNA-seq use either `single_end` or `paired_end`. <br> For spatial transcriptomics use either `visium_v1` or `visium_v2`      |
+| `seq_unit`        | Sequencing unit (likely one of: `cell`, `nucleus`, `bulk`, or `spot`)           |
+| `feature_barcode_file` | path/uri to directory containing the feature barcode sequences (only applicable for CITE-seq)  |	
+| `feature_barcode_geom` | A salmon `--read-geometry` layout string/ See https://github.com/COMBINE-lab/salmon/releases for details (only applicable for CITE-seq) |
+| `slide_section`   | The slide section for spatial transcriptomics samples               |
+| `slide_serial_number` | The slide serial number for spatial transcriptomics samples     |
```
jashapiro, line 51:
```suggestion
| `slide_section`   | The slide section for spatial transcriptomics samples     (spatial transcriptomics only)       |
| `slide_serial_number` | The slide serial number for spatial transcriptomics samples  (spatial transcriptomics only)   |
```

external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
...
+| `files_directory` | path/uri to directory containing fastq files from run (unique per run) |
```
jashapiro, line 52: I feel like we should move this up: the core set of required fields first, followed by type-specific or optional fields.

external-data-instructions.md
```diff
@@ -34,7 +34,28 @@ nextflow run AlexsLemonade/scpca-nf \
 ### Prepare the metadata file 
 
 Using `scpca-nf` requires a metadata file where each library to be processed is a row and columns contain associated information about that library. 
-You will need the following columns to be present in your metadata file... 
+You will need to create a tab separated file with the following columns: 
+
+| column_id       | contents                                                       |
+|-----------------|----------------------------------------------------------------|
+| `scpca_run_id`    | Run ID in the form `SCPCR000000`                               |
+| `scpca_library_id` | Library ID in the form `SCPCL000000` <br> The numeric portion will equal the `scpca_run_id` for RNA-seq data, but for associated feature data (CITE-seq) this will be shared with the associated RNA-seq data that it is paired with.|
```
jashapiro, line 42: I don't love the wording here for external use. I think _we_ know what it means, but I'm not sure others will... I don't have something better right now though. This detail fit better in the preamble too?

jashapiro (commented):
Thanks for getting started on this! I think there are a few things that we may want to rearrange and reword for external consumption, as I am worried that a few things that seem obvious to us now will not be so obvious to others. 

I think we can probably punt for now about what the right contact info is: we don't expect people using this for ScPCA to do so without prompting, and people who are using it for other purposes can probably go ahead and run it without our IDs. Which is related to a point I made about removing the statements about what the format of the ID is: We might hope that people would do their testing before they request IDs from us, and we should be sure they know that they can do that with whatever ids they feel comfortable with.

allyhawkins:
Thanks for taking a look at this @jashapiro. I believe I've addressed all of your comments, but let me know if I missed anything. The main changes that I made was splitting up the table of what columns you need into a required table and a other modalities/optional table. I thought about further splitting it up so that we had three tables or even taking out the `submitter` and `submitter_id` and adding in a sentence on that instead of including those in the tables. If you like one of those ideas better let me know. 
I also included the introduction of run ID's, library ID's, and sample ID's to be before the table along with a link to the FAQ we have on the difference between them in the scpca portal docs. Let me know if anything is still unclear in that explanation.

jashapiro (commented):
I had some minor wording updates here, but I also realized that we probably want to add something (maybe to the initial requirements section) to address the fact that the files themselves need to be organized in a particular way. In particular, each run needs to be in its own folder with all of the fastqs for that run (and whatever else). I can imagine people having multiplexed (at the sequencer level) files all in the same folder, which would have to be split up, and top-up runs in separate folders, which would need to be combined.

allyhawkins:
> I had some minor wording updates here, but I also realized that we probably want to add something (maybe to the initial requirements section) to address the fact that the files themselves need to be organized in a particular way. In particular, each run needs to be in its own folder with all of the fastqs for that run (and whatever else). I can imagine people having multiplexed (at the sequencer level) files all in the same folder, which would have to be split up, and top-up runs in separate folders, which would need to be combined.

I went ahead and added a few sections to the beginning section about how to organize files. I think it is helpful to have that section and keep the edits to the metadata file section about including FASTQ files in a unique directory since it could be helpful to have the information in two different places. The first one tells you how to set up your files, while the second tells you how to add the file information to the metadata based on the setup. Let me know if you are okay with that or if you had a different idea for adding in the file organization?

jashapiro (approved):
Looks good. I had some minor wording thoughts, including one sentence that I think we can maybe just drop for redundancy, but I don't think I need to look again.

(I actually thought I had submitted this hours ago, sorry!)

PR #120, opened 2022-03-03
Add instructions/notes for AWS

### LINKED ISSUE #113: scpca-nf user instructions: AWS Section
We should include some guidance on how to use `scpca-nf` with AWS. 
This will probably only include some basic information on how you could use AWS with some guidance to reach out to someone for help as this can be complicated to setup. 

These instructions should be included as a branch on the branch `allyhawkins/outline-scpca-nf-instructions` from https://github.com/AlexsLemonade/scpca-nf/pull/106 so we can merge all the instructions together.

jashapiro:
Closes #113

This PR adds the sections about AWS settings. I struggled a bit with how much detail to include, so please let me know if this seems about right, or if there are places where I should include more or less detail. 

I will say that the audience I was aiming for is the one with some experience with AWS, as I don't necessarily want people who have not used it at all to make this their first project. So if it seems somewhat opaque, that might be partly by design, though I tried not to make it _too_scary. I mostly wanted to hit the things that didn't seem obvious to me at first, and the places where we deviated most from the instructions in the Nextflow documentation, which is why the stuff about launch templates and queues is there.

I debated linking to the Terraform files that we used do define the infrastructure, but ultimately decided that was just another level of confusion. If people ask questions, we might add that, or provide it on request. I also thought about providing access to our AMI, but I think that is a bigger discussion, and I really just directly followed the nextflow instructions, so I don't think there is not a lot of value there at this stage.

### DIFF
--- a/external-data-instructions.md
+++ b/external-data-instructions.md
@@ -6,8 +6,7 @@
   - [Prepare the metadata file](#prepare-the-metadata-file)
   - [Configuring `scpca-nf` for your environment](#configuring-scpca-nf-for-your-environment)
     - [Setting up a profile in the config file](#setting-up-a-profile-in-the-config-file)
-    - [Adding parameters to the config file](#adding-parameters-to-the-config-file)
-    - [Using `scpca-nf` with AWS](#using-scpca-nf-with-aws)
+    - [Using `scpca-nf` with AWS Batch](#using-scpca-nf-with-aws-batch)
   - [Adjust optional parameters](#adjust-optional-parameters)
   - [Special considerations for using `scpca-nf` with spatial transcriptomics libraries](#special-considerations-for-using-scpca-nf-with-spatial-transcriptomics-libraries)
 
@@ -20,7 +19,7 @@ In order to use `scpca-nf` to process your own data, you will need to make sure 
 - [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
 - [Docker](https://docs.docker.com/get-started/#download-and-install-docker) (installed either locally or on an HPC)
 
-You will also need to create a metadata file and a nextflow configuration file (see below).
+You will also need to create a metadata file and a Nextflow configuration file (see below).
 Once you have set up your environment and created these files you will be able to start your run as follows, adding any additional optional parameters that you may choose: 
 
 ```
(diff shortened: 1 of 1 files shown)


external-data-instructions.md
```diff
@@ -94,7 +94,23 @@ nextflow run AlexsLemonade/scpca-nf \
...
+
+To run `scpca-nf`, you will need to set up at least one batch queue and an associated compute environment configured with a custom Amazon Machine Image (AMI) prepared according to the [Nextflow instructions](https://www.nextflow.io/docs/latest/awscloud.html#custom-ami). 
+You will also need an [S3 bucket](https://aws.amazon.com/s3/) path to use as the Nextflow `work` directory for intermediate files.
+As the intermediate files can get quite large, you will likely want to set up a [lifecycle rule](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html) to delete files from this location after a fixed period of time (e.g., 30 days).
+
+Our [AWS Batch config file](https://github.com/AlexsLemonade/scpca-nf/blob/main/config/profile_awsbatch.config) may be helpful as a starting reference, but note that the queues and file locations listed there are not publicly available, so these will need to be set in your own profile.
```
allyhawkins, line 106: I feel like this is out of place here. Maybe a better spot for it is right after mentioning that you need to set up a batch queue and compute environment above? It seems like it's directly related to that since it's providing an example of what you would need to set up.

allyhawkins (approved):
Generally this looks good to me. I think having the key information for AWS is all we need, as I also agree we probably don't want to encourage people who haven't used it before to try and set this up on their own. I also don't think we need to link to terraform just yet since that is confusing and something we can add if we notice that people will find it useful.

I just had one wording comment and then one comment about where to place the link to our batch profile. Feel free to take the suggestions or not, but I don't think I need to take another look.

external-data-instructions.md
jashapiro: I agree that this is oddly placed, but I think after the batch queue and compute environment is probably not quite right either, as the bucket is also required... Maybe _before_ both of those? But it might be odd to talk about queues and file locations before saying you need them. So I am kind of torn. Maybe at the end?

Issue #105, opened 2022-02-23
Add parameter checks to main workflow


jashapiro:
To aid in usability as we prepare for external users, we could add some parameter checks to the main workflow to give more informative errors if certain parameters are improperly set.

For many of these, we might check that files exists/are accessible. That might look something like the code below:

```nextflow
param_error = false

// repeat blocks like this for each file to check, or create a loop to check a series of expected files
if(!file(params.infile).exists()){
  param_error = true
  log.error "The file '${params.infile}' can not be found."
}

// after all checks
if(param_error){
  System.exit(1)
}
```

For parameters which have specific values, we might add other checks. I have some concern about separating the default values from the check, but I think it likely worth the effort.

jaclyn-taroni:
We'll set this for this sprint, but it's a nice to have not a must have. It should be the last thing we pick up!

jashapiro:
One thought on this is that we may want to think about having some parameters, such as  `run_metafile` and `outdir` set as blank by default, to prevent external users from getting cryptic errors if they do not set them and their run tries to access our files/locations for which they do not have permission. 

We could then set these values within a `ccdl` profile block, such that we would invoke them with: `-profile ccdl`, or more likely with `-profile ccdl,batch`, as you can include multiple profiles, comma separated.

Not quite as easy as using them by default, but not bad, and could save external user confusion.

PR #121, opened 2022-03-04
Add config file instructions and reorganize a bit

### LINKED ISSUE #112: scpca-nf user instructions: setting up a config file
The user instructions for `scpca-nf` should include a section on how to set up your own config file. This is currently divided into two sections in #106, setting up a profile and adjusting parameters. Both of these sections should be addressed so users are able to set up their config file. 
We will also provide a template config file, #111, that can be referenced and instructions on what should be edited in this template file should be included in this section. 

These instructions should be included as a branch on the branch `allyhawkins/outline-scpca-nf-instructions` from #106 so we can merge all the instructions together.

jashapiro:
Closes #112 

This PR adds more detail about config files and profiles to the external user documentation. 

⚠️ As I was writing, I made some organizational changes to that section and the section about AWS, so this PR should be evaluated before we get too far into revisions, so that those organizational changes don't cause too many conflicts! ⚠️
(I retargeted #120 to this branch, as I think it makes sense to finish this one first.)

The organizational changes are mostly about hierarchy: I moved most everything up a level, but nested the "profile" discussion within the more general configuration discussion. It seemed like talking about workflow parameters in a config file was the easiest transition in, so I did that first, then moved on to the discussion of profiles.

I also moved the two required parameters up to the very top of the "Configuring `scpca-nf`" section, as I want to make sure that people don't miss those. (I had some thoughts about how to make it more clear that those are required at the command line, but I think that will be part of #105, and I will put those thoughts there.)

I didn't include too much detail about profiles as it seems like there were a number of ways one could go on that front, and most would not apply to most people. If you think there are places that I should expand a bit more though, I would be happy to do so!

### DIFF
--- a/examples/user_template.config
+++ b/examples/user_template.config
@@ -5,7 +5,7 @@
 /*
 Global parameters for scpca-nf
 ------------------------------
-The values presented here are examples, and should be replaced in with 
+The values presented here are examples, and should be replaced with 
 the correct paths to any indicated files or other values that need to be set.
 */
 
@@ -32,7 +32,7 @@ Profile settings for the workflow execution environment
 Here we provide an example using slurm as the executor system. 
 For information about available process executors and their options, see https://www.nextflow.io/docs/latest/executor.html
 This profile should be adjusted based on the details of your system and can be invoked at runtime with 
-`--profile cluster` (or whatever name you have chosen)
+`-profile cluster` (or whatever name you have chosen)
 
 For more information on setting up profiles see https://www.nextflow.io/docs/latest/config.html#config-profiles
 */
--- a/examples/user_template.config
+++ b/examples/user_template.config
@@ -31,7 +31,7 @@ params {
 // For more information on setting up profiles see https://www.nextflow.io/docs/latest/config.html#config-profiles
 
 profiles{
- slurm {
+ cluster {
     process {
       executor = 'slurm'
       // queue name for your cluster
--- a/examples/user_template.config
+++ b/examples/user_template.config
@@ -1,34 +1,41 @@
-// To use this config file with the workflow use the following command:
-// `nextflow run AlexsLemonade/scpca-nf -c user_template.config -profile slurm`
-// Note that if you change the profile name you will have to use the name you chose in place of slurm.
+// user_template.config
+// An example template for use with the scpca-nf Nextflow workflow.
 
-// global parameters for workflows
-// parameters should be filled in with the correct paths to these files, here we provide examples
-params {
 
-  // Input data table
-  run_metafile = 'example_metadata.tsv'
+/*
+Global parameters for scpca-nf
+------------------------------
+The values presented here are examples, and should be replaced in with 
+the correct paths to any indicated files or other values that need to be set.
+*/
 
-  // Output base directory for results
-  outdir = "scpca_out"
+params.run_metafile = 'example_metadata.tsv' // Input data table
 
-  // which samples or libraries to process
-  // although the name of the parameter is `run_ids`, any ids found in `scpca_run_id`, `scpca_library_id`, or `scpca_sample_id` can be used as input
-  // by default this is set to all, but to change to a specific subset input a comma separated list e.g., "SCPCR000001,SCPCR000002"
-  // optionally, run_ids can be changed at the command line by using `--run_ids SCPCR000001`
-  run_ids = "All"
+params.outdir = "scpca_out" // Output base directory for results
 
-  // Spaceranger container, only required if using spatial transcriptomics workflow
-  SPACERANGER_CONTAINER = ''
+params.run_ids = "All"  // Which samples or libraries to process
+/*
+Note: Although the name of this parameter is `run_ids`, any ids found in 
+  `scpca_run_id`, `scpca_library_id`, or `scpca_sample_id` can be used.
+  By default this is set to "All", but to change to a specific subset,
+  use a comma separated list e.g., "SCPCR000001,SCPCR000002"
+  Optionally, `run_ids` can be set at the command line with `--run_ids SCPCR000003`
+*/
 
-}
+// Spaceranger container, only required if using spatial transcriptomics workflow
+params.SPACERANGER_CONTAINER = ''
 
-// profile settings for executing the workflow
-// Here we provide an example using slurm as the executor system. 
-// For information about available process executors,  and their options, see https://www.nextflow.io/docs/latest/executor.html
-// This profile should be adjusted based on the details of your system and can be invoked at runtime with 
-// `--profile slurm` (or whatever name you have chosen)
-// For more information on setting up profiles see https://www.nextflow.io/docs/latest/config.html#config-profiles
+
+/*
+Profile settings for the workflow execution environment
+-------------------------------------------------------
+Here we provide an example using slurm as the executor system. 
+For information about available process executors and their options, see https://www.nextflow.io/docs/latest/executor.html
+This profile should be adjusted based on the details of your system and can be invoked at runtime with 
+`--profile cluster` (or whatever name you have chosen)
+
+For more information on setting up profiles see https://www.nextflow.io/docs/latest/config.html#config-profiles
+*/
 
 profiles{
  cluster {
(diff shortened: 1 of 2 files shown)


external-data-instructions.md
```diff
@@ -2,70 +2,105 @@
...
-  --outdir <path/to/output>
+  -r v0.2.4 \
+  -config my_config.config \
+  --run_metafile <path/to/metadata_file>
 ```
 
-### Prepare the metadata file 
+This will pull the `scpca-nf` workflow directly from Github, using the `v0.2.4` version, and run it based on the settings in the local configuration file `my_config.config`.
+
+**Note:** `spca-nf` is under active development.
```
allyhawkins, line 34:
```suggestion
**Note:** `scpca-nf` is under active development.
```
Minor typo

external-data-instructions.md
```diff
@@ -2,70 +2,105 @@
...
+
+See the [Nextflow documentation](https://www.nextflow.io/docs/latest/config.html) and the below sections for more detail on creating your own configuration file.
+
+### Setting up a profile in the configuration file
+
+Local running may be sufficient for small jobs or testing, but you will most likely want to run your workflow in a high performance computing environment (HPC), such as an institutional computing cluster or on a cloud service like AWS.
+To do this, we recommend using [Nextflow profiles](https://www.nextflow.io/docs/latest/config.html#config-profiles) to encapsulate settings like the [`executor`](https://www.nextflow.io/docs/latest/executor.html) that will be used to run each process and associated details that may be required, such as queue names or the container engine your system uses. 
+
+In our example template file [user_template.config](https://github.com/AlexsLemonade/scpca-nf/blob/main/examples/user_template.config), we define a profile named `cluster` which could be invoked with the following command:
```
allyhawkins, line 88: I think it would be worth adding a sentence about using singularity here. Our setup assumes usage of docker, so I think it would be useful to state that if your system uses singularity and does not use docker, then that will need to be reflected in the config file and maybe even show an example or link to the [nextflow docs on using singularity](https://www.nextflow.io/docs/latest/singularity.html). I think this will be a common enough use that it would be helpful information to have.

external-data-instructions.md
```diff
@@ -2,70 +2,105 @@
...
 
-However, if you need to run your workflow on an HPC, on AWS, or other workspace we encourage you to create your own config file.  
-The config file will allow you to define the system where you would like to execute nextflow, allocate resources, and define parameters for the workflow. 
-After creating a config file, `my_config.config`, that file can be used in conjunction with `scpca-nf` at the command line by using `-c my_config.config`. 
-See the [Nextflow documentation](https://www.nextflow.io/docs/latest/config.html) and the below sections for more detail on creating your own config file. 
+Note that *workflow* parameters such as `--run_metafile` and `--outdir` are denoted at the command line with double hyphen prefix, while options that affect Nextflow itself have only a single hyphen. 
+
+### Configuration files
+
+Workflow parameters can also be set in a configuration file by setting the values `params.run_metafile` and `params.outdir` as follows:
```
allyhawkins, line 63: I keep going back and forth on this, but I wonder if it would be helpful to have an intro into what a configuration file is and also state that they require the `.config` extension. I feel like right now we just jump into adding parameters to the config file without a lot of explanation, but then again maybe it doesn't need explanation and the links you provided in the later section are enough.

allyhawkins (commented):
I think this looks pretty good overall. I'm okay with the changes to the organization and they make sense. My biggest comment is about adding in some information about using singularity, since I think that will be fairly common for people using an HPC and is also present in the config file. 

I had one other comment about adding an intro sentence before you jump into adding parameters to the config files, but I could go either way on that one.

external-data-instructions.md
jashapiro: I hinted at this in the sentence before "associated details... such as container engine" but I will expand that a bit, and also add Singularity to the prereq options.

I didn't want to specifically call out too much about the example file here, to preserve our ability to modify it if we wanted later. That was also part of the logic behind changing the name to the generic "cluster". My thought was that people are very much likely to need to read the full documentation, so I don't want to provide the illusion of completeness, that they might be able to just take the version we provide and have it "just work".

external-data-instructions.md
jashapiro: I added a link to the documetation for the config file here, but the minimal version really is just a list of parameters, so I wanted to just show that example first. I wasn't clear if the way I did it here was the best; I can try to play with wording. 

A note though... the file doesn't need to have a `.config` extension; that is only if is going to be found automatically as a `nextflow.config` file. I do want every example to have that extension, as I think it is a good idea, but it isn't a requirement.

external-data-instructions.md
```diff
@@ -2,70 +2,105 @@
...
-See the [Nextflow documentation](https://www.nextflow.io/docs/latest/config.html) and the below sections for more detail on creating your own config file. 
+Note that *workflow* parameters such as `--run_metafile` and `--outdir` are denoted at the command line with double hyphen prefix, while options that affect Nextflow itself have only a single hyphen. 
+
+### Configuration files
+
+Workflow parameters can also be set in a configuration file by setting the values `params.run_metafile` and `params.outdir` as follows: 
+
+```groovy
+// my_config.config
+params.run_metafile = '<path/to/metadata_file>'
```
jashapiro, line 67: @allyhawkins I used the `params.variable` convention here, because it seemed more straightforward in this simple example. But I was wondering if you thought we should perhaps use that also for `user_template.config` for consistency? I feel like the blocks/bracket syntax is good when things get more complex, but it may be simpler to reserve that for the profile section, and have the "universal" settings with the simpler (but more redundant?) syntax.

external-data-instructions.md
allyhawkins: I am good with the change to `cluster` and understand about not wanting to use the example too much. I think linking to the nextflow docs on singularity would be helpful though or at least a statement saying that our default container settings are to use docker so if you are using something else, like singularity, that needs to be reflected in your config file.

external-data-instructions.md
allyhawkins: I was thinking about this and wondering if it was going to be confusing to have two separate version of it, but agree that here it makes sense to use `params.variable`. I am good with the change to the template to be inline with using `params.variable` syntax and be consistent then.

jashapiro:
@allyhawkins I think I addrssed most of your comments, and now link to the Singularity instructions in two places. (I also removed the direct docker link, instead linking to the nextflow instructions there as well, for consistency. I was originally going to link directly to singularity, but their docs are terrible for initial users.) 

I also updated the template file a bit, trying to format it to make the organization a bit more clear, and removing the `params{}` block in favor of `params.variable`. 

I hope this addresses the suggestions you had made, but let me know if I missed anything!

examples/user_template.config
```diff
@@ -1,37 +1,44 @@
...
-// parameters should be filled in with the correct paths to these files, here we provide examples
-params {
 
-  // Input data table
-  run_metafile = 'example_metadata.tsv'
+/*
+Global parameters for scpca-nf
+------------------------------
+The values presented here are examples, and should be replaced in with 
+the correct paths to any indicated files or other values that need to be set.
```
allyhawkins, line 9:
```suggestion
The values presented here are examples, and should be replaced with 
the correct paths to any indicated files or other values that need to be set.
```
Minor typo.

examples/user_template.config
```diff
@@ -1,37 +1,44 @@
...
+params.SPACERANGER_CONTAINER = ''
+
+
+/*
+Profile settings for the workflow execution environment
+-------------------------------------------------------
+Here we provide an example using slurm as the executor system. 
+For information about available process executors and their options, see https://www.nextflow.io/docs/latest/executor.html
+This profile should be adjusted based on the details of your system and can be invoked at runtime with 
+`--profile cluster` (or whatever name you have chosen)
```
allyhawkins, line 35:
```suggestion
`-profile cluster` (or whatever name you have chosen)
```
I think this should be a single dash.

allyhawkins (approved):
These changes look good to me! I just picked up on two minor typos that need to be fixed and then its good to go.

PR #122, opened 2022-03-08
Add parameter checks

### LINKED ISSUE #105: Add parameter checks to main workflow

### LINKED ISSUE #109: Add manifest to config
For better metadata of the nextflow workflow itself, we can add a `manifest` block to the config. I'm not 100% sure how this gets used in every case, and whether it is available internally to the workflow, but it seems like an easy thing to add.

Docs are here: https://www.nextflow.io/docs/latest/config.html#scope-manifest

jashapiro:
This PR adds some parameter checks to the main workflow to catch what I thought were the most likely errors for now. It turned out in my mind that there really wasn't too much to check, but having the framework there was still worth doing.

Closes #105 (probably)

The most important check is for the presence of the metadata file, I think checking the format of that file is going to be a whole different can of worms, so I didn't do that at this point. My groovy skills aren't there yet.

We could check all of the reference files as well, but I didn't do that at this stage. I think I would want to figure out some loop for that, rather than a long series of `if`s, but I didn't get there yet. If we think that should be part of closing #105, I could do that as part of this PR, or add it as a separate one. 

If there are other checks that you think this PR should include, please let me know!

I also added a manifest to close #109, since that seemed pretty trivial.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -1,6 +1,16 @@
 #!/usr/bin/env nextflow
 nextflow.enable.dsl=2
 
+manifest{
+  name = 'scpca-nf'
+  description = 'A nextflow workflow for processing single-cell RNA-seq data as part of the ScPCA project.'
+  homepage = 'https://github.com/AlexsLemonade/scpca-nf'
+  author = "Childhood Cancer Data Lab, Alex's Lemonade Stand Foundation"
+  mainScript = 'main.nf'
+  version = 'v0.2.4'
+}
+
+
 // 10X barcode files
 cell_barcodes = [
   '10Xv2': '737K-august-2016.txt',
--- a/main.nf
+++ b/main.nf
@@ -16,7 +16,7 @@ cell_barcodes = [
   ]
 
 // supported technologies
-single_cell_techs= cell_barcodes.keySet()
+single_cell_techs = cell_barcodes.keySet()
 bulk_techs = ['single_end', 'paired_end']
 spatial_techs = ["spatial", "visium_v1", "visium_v2"]
 all_techs = single_cell_techs + bulk_techs + spatial_techs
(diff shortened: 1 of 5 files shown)


jashapiro:
I found that you _can_ access the manifest values from within processes, so I have also updated the places where we used to use `params.workflow_url` to use that, and removed that parameter completely, which feels much better.

allyhawkins (approved):
This looks pretty good to me. I just had a few clarification questions on the manifest that may be moot points and a reminder that we need an issue for a new release before we announce/ after #106 gets merged. 

I'm going to agree that we should check for the reference files and make a separate issue for that. For the purposes of external use, the most important check is the metadata file. I don't think there are any other checks that are missing.

jashapiro (commented):
Thanks for the comments @allyhawkins! 

The question of the default branch is something I was considering as a separate question, and you can see some of my thoughts below. On second thought, I wonder if `defaultBranch` allows a tag instead of an actual branch. I will test that in a bit, because it might be a nice way to set it to the latest for people who do remember the `-r` flag.

allyhawkins:
> This does raise a question I was considering though, which is whether we want to make a separate release branch and set that as the default, so that somebody running the script will automatically get that, and it automatically contains the latest "stable" version, not including any features we are still working on.

I like the idea of this! I think that would be helpful so that someone doesn't accidentally run something while we are developing and get errors that they don't know how to handle.

jashapiro:
> > This does raise a question I was considering though, which is whether we want to make a separate release branch and set that as the default, so that somebody running the script will automatically get that, and it automatically contains the latest "stable" version, not including any features we are still working on.
> 
> I like the idea of this! I think that would be helpful so that someone doesn't accidentally run something while we are developing and get errors that they don't know how to handle.

The challenge here is that this is hard to test... I think we need the manifest to be in `main` so it knows to grab a different branch. I realized that the tag version is even worse (maybe not possible) as we could end up setting a tag that doesn't yet exist.

PR #124, opened 2022-03-10
Add description of output files to scpca-nf instructions

allyhawkins:
Closes #116. 

Here I added a section at the end of the instructions to describe the output files that should be expected after running the workflow. I noted that there should be two folders published to the `outdir`, the `internal` and the `publish` folder. I then explained the contents of each folder and showed the expected structure of each folder if running single-cell/single-nuclei samples. I also included a link to the docs to explain the contents of the files and focused here on the actual structure of the files. 

I added a note for each folder on if you were to run bulk samples, what the folders would look like. Would it be preferred to add this to the tree diagram rather than just adding a note? I was trying to avoid things getting really long with the `internal` folder specifically, but if we think it's important to have I can add it in. 

One thing to note is that I did not include the output structure for spatial yet as that is subject to change as part of #82. 

The main question I had was if we are okay with separating the structure into `publish` and `internal` or if it's preferred to put them all together and show one tree of the output directory? 
Additionally, should we show all of the intermediate files in the `internal` folder like I have it right now? Or is it better to just show the overall structure?

PR #125, opened 2022-03-10
Add special considerations to scpca-nf user instructions

### LINKED ISSUE #114: scpca-nf user instructions: special considerations for using the spatial workflow
If users would like to analyze spatial transcriptomics data with `scpca-nf`, they will need to have created their own docker container that contains `spaceranger` and updated the parameter for the spaceranger container. We should provide some cautionary guidance that the workflow will not run out of the box for these samples in case users are interested in using that part of the workflow. 

These instructions should be included as a branch on the branch `allyhawkins/outline-scpca-nf-instructions` from https://github.com/AlexsLemonade/scpca-nf/pull/106 so we can merge all the instructions together.

allyhawkins:
Closes #114. 

Here I added in special instructions for using the spatial workflow. I specifically noted that the `.jpg` file must be included in the directory with the FASTQ files and noted the columns that must be added to the `run_metafile`. 

I also included brief instructions on creating a docker image that includes Space Ranger. I wasn't quite sure how much information we wanted to include on actually using Docker itself and creating the image. One thing I thought of was including the link to our Dockerfile in `alsf-scpca` but I wasn't 100% sure if that was okay so didn't include that. Are there things that I should include more detail on? 

In doing this, I also decided to wrap up the special instructions and added a short section at the bottom for using the `repeat_mapping` parameter as was discussed in slack with @jashapiro. Rather than having a section on all of the parameters, I deleted that section and included a short section at the bottom specifically for if users would want to repeat a mapping step they can use the flag at the command line. 

I stacked this on #124 as to minimize merge conflicts.


jashapiro (commented):
I had some smallish updates, mostly wording and organization.

I wanted to add a bit more about docker too, with our example. I did realize though that there could be a limitation with docker where I am not sure if/how nextflow supports private repositories. I don't have any solutions, so I don't even know what to put in the docs...

allyhawkins:
> I wanted to add a bit more about docker too, with our example. I did realize though that there could be a limitation with docker where I am not sure if/how nextflow supports private repositories. I don't have any solutions, so I don't even know what to put in the docs...

I also was really unsure about this... which is why I originally had aired on the side of including less information. I think what you added is fine and then if we find that users are having trouble accessing private repositories we can go from there?

allyhawkins:
I also moved the section on repeat the mapping step and added in the link to our Docker container for Space Ranger like you had suggested. This should be ready for another look.

external-data-instructions.md
```diff
@@ -208,3 +224,5 @@ If bulk libraries are processed, there will be an additional `salmon` folder tha
 
 All files pertaining to a specific library will be nested within a folder labeled with the library ID.
 Additionally, for each run, all files related to that run will be inside a folder labeled with the run ID followed by the type of run (i.e. `rna` or `features` for CITE-seq) and nested within the library ID folder.
+
+If bulk libraries are processed, there will be an additional `salmon` folder that contains the output from running [`salmon`](https://salmon.readthedocs.io/en/latest/file_formats.html#fileformats) on each library processed.
```
jashapiro, line 228: This snuck back in after moving in the other branch, I think.
```suggestion

```

jashapiro (approved):
LGTM, less a duplicate line noted below

jashapiro:
Make sure you delete the `allyhawkins/outline-scpca-nf-instructions` branch or manually retarget this before merging!

PR #129, opened 2022-03-17
Merge v0.2.4 release into latest branch

### LINKED ISSUE #128: Create 0.2.4 release
Most of the docs now refer to version 0.2.4, as there were some updates in the workflow related to that.

I have created draft release notes in https://github.com/AlexsLemonade/scpca-nf/releases/edit/untagged-b3cfb48ee17dede68328, which I think cover the changes that have been made.

After that release is done, the 0.2.4 draft should be merged into the `latest` branch so that the `defaultBranch` tag we added in the manifest works as expected. 

Tagging @allyhawkins in case there are any additional tasks that she can think of before we complete this release.

jashapiro:
Before the creation of v0.2.4 tag/release, this should make the `latest` branch have the same contents as main.

I realized that the better order of operations was to merge into `latest` _first_ and then tag _there_, so I actually just deleted the 0.2.4 tag, and we will merge this first, then tag and release from that branch.

Please review #130 first though, to catch a little error!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,12 @@
-## scpca-nf
+## Running scpca-nf for the ScPCA portal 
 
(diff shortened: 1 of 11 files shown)


allyhawkins (approved):
🚀  (post typo fix of course)

PR #134, opened 2022-03-23
Add option to skip spaceranger quantification

### LINKED ISSUE #81: Make spaceranger process optional
Because the spaceranger step is very costly and takes a long time to run, we should incorporate the option to skip this step (similar to how we skip the production of the rad file in alevin-fry). In doing this, we should break out the file reorganization into it's own process so the first process is running spaceranger and the second process is file re-organization and creating the `metadata.json`.

### LINKED ISSUE #82: Re-organize spatial outputs to reflect desired hand-off structure
Based on our conversation in the multi-team planning meeting, we have made small revisions to the download structure of the spatial outputs. We will need to adjust the current output of the process accordingly to reflect those changes. This should be done after addressing #81, so testing is less time consuming. 

This is the structure that will be expected for hand-off. 

```
└── SCPCP00006
    ├── samples_metadata.csv
    └── SCPCS000203
        ├── SCPCL000372_metadata.json 
        └── SCPCL000372_spatial
	    ├── SCPCL000372_spaceranger_summary.html
	    └──filtered_feature_bc_matrix
	       ├── barcodes.tsv.gz
	       ├── features.tsv.gz
	       ├── matrix.mtx.gz
	    └──raw_feature_bc_matrix
	       ├── barcodes.tsv.gz
	       ├── features.tsv.gz
	       ├── matrix.mtx.gz
	    └──spatial
	       ├── aligned_fiducials.jpg
	       ├── detected_tissue_image.jpg
	       ├── scalefactors_json.json
	       ├── tissue_hires_image.png
	       ├── tissue_lowres_image.png
	       ├── tissue_positions_list.csv

```

This includes removing `spaceranger_metrics_summary.csv` and `spaceranger_versions.html` and attaching `_spatial` to the library folder name. 

Downloads will be similar to this but will contain a `libraries_metadata.csv` instead of `samples_metadata.csv`, will not include `metadata.json`, and will also have a `README.md`. 
Like so: 
```
└── SCPCP00006
    ├── libraries_metadata.csv
    ├── README.md
    └── SCPCS000203
        └── SCPCL000372_spatial
	    ├── SCPCL000372_spaceranger_summary.html
	    └──filtered_feature_bc_matrix
	       ├── barcodes.tsv.gz
	       ├── features.tsv.gz
	       ├── matrix.mtx.gz
	    └──raw_feature_bc_matrix
	       ├── barcodes.tsv.gz
	       ├── features.tsv.gz
	       ├── matrix.mtx.gz
	    └──spatial
	       ├── aligned_fiducials.jpg
	       ├── detected_tissue_image.jpg
	       ├── scalefactors_json.json
	       ├── tissue_hires_image.png
	       ├── tissue_lowres_image.png
	       ├── tissue_positions_list.csv
```

allyhawkins:
Here, I added in the option to skip the spaceranger quantification, if the spaceranger output folder already exists. This was pretty similar to the previous steps that we skipped where as long as the `--repeat_mapping` flag is set to false and the output directory exists, then it will be included in a separate branch on the spatial channel that will not go through the `spaceranger` process. 

As part of this PR, I also addressed the reorganization of the output files, removing the `spaceranger_versions.json` and `spaceranger_metrics_summary.csv` files from the final output. However, we need those files to create the `metadata.json` file that we create in the `spaceranger_metadata` process. 

So here, what I did was break out the file reorganization that was originally in the `spaceranger` process and put it in the `spaceranger_metadata` process. This means that we now are going to store the outs folder from cellranger as is (with removal of the large bam files) in the `internal/spaceranger` folder, and then actually include the files we would like to include on the portal in the `publish` directory. I'm torn on this because we don't really need to have all of these smaller files, but it seemed like the clean way to do it. This way we have the original folder that remains untouched and if it exists we skip spaceranger. Then we output the final files in a separate output folder in the file structure we would like without disturbing the original files. 

Lastly, to finish addressing the reorganization, I appended the `_spatial` flag to the library folder that gets published as described in #82. 

I tested these changes and the spaceranger process is successfully skipped and the output files look as expected. 

Closes #81.
Closes #82.

### DIFF
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -12,9 +12,8 @@ process spaceranger{
     tuple val(meta), path(fastq_dir), file(image_file)
     path index
   output:
-    tuple val(meta), path(spatial_out)
+    tuple val(meta), path(out_id)
   script:
-    spatial_out = "${meta.library_id}_spatial"
     out_id = "${meta.run_id}-spatial"
     meta.cellranger_index = index.fileName
     """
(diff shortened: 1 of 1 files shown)


modules/spaceranger.nf
```diff
@@ -95,21 +94,39 @@ workflow spaceranger_quant{
...
     main: 
         // create tuple of (metadata map, [])
-        spaceranger_reads = spatial_channel
-          // add sample names to metadata
-          .map{it.cr_samples =  getCRsamples(it.files); it}
+        spatial_channel = spatial_channel
+          // add sample names and spatial output directory to metadata
+          .map{it.cr_samples =  getCRsamples(it.files);
+               it.spaceranger_publish_dir =  "${params.outdir}/internal/spaceranger/${it.library_id}";
+               it.spaceranger_results_dir = "${it.spaceranger_publish_dir}/${it.run_id}-spatial";
```
jashapiro, line 101: super minor, but if we are using `_` elsewhere, we might keep it the same here?
```suggestion
               it.spaceranger_results_dir = "${it.spaceranger_publish_dir}/${it.run_id}_spatial";
```

modules/spaceranger.nf
```diff
@@ -12,9 +12,8 @@ process spaceranger{
     tuple val(meta), path(fastq_dir), file(image_file)
     path index
   output:
-    tuple val(meta), path(spatial_out)
+    tuple val(meta), path(out_id)
   script:
-    spatial_out = "${meta.library_id}"
     out_id = "${meta.run_id}-spatial"
```
jashapiro, line 17: This is to match my previous comment about `-` vs `_` consistency.
```suggestion
    out_id = "${meta.run_id}_spatial"
```

modules/spaceranger.nf
```diff
@@ -95,21 +94,39 @@ workflow spaceranger_quant{
     // a channel with a map of metadata for each spatial library to process 
     main: 
         // create tuple of (metadata map, [])
-        spaceranger_reads = spatial_channel
-          // add sample names to metadata
-          .map{it.cr_samples =  getCRsamples(it.files); it}
+        spatial_channel = spatial_channel
+          // add sample names and spatial output directory to metadata
+          .map{it.cr_samples =  getCRsamples(it.files);
+               it.spaceranger_publish_dir =  "${params.outdir}/internal/spaceranger/${it.library_id}";
```
jashapiro, line 100: I managed to get myself confused here, so I'm wondering if we want to change this to `it. spaceranger_internal_dir`? I may be overthinking this, and whatever we do should be consistent with the other "skip" settings, so probably not worth changing.

modules/spaceranger.nf
```diff
@@ -95,21 +94,39 @@ workflow spaceranger_quant{
...
+        spaceranger_quants_ch = spatial_channel.has_spatial
+          .map{meta -> tuple(meta,
+                             file("${meta.spaceranger_results_dir}")
+                             )}
+
+        grouped_spaceranger_ch = spaceranger.out.mix(spaceranger_quants_ch)
+
           // generate metadata.json
-          | spaceranger_metadata
+        spaceranger_metadata(grouped_spaceranger_ch)
```
jashapiro, line 127: Suggesting here, though the main change is higher, that we should rename this process, as it is now not just creating the metadata. Perhaps `spaceranger_publish`?

jashapiro (approved):
This looks good to me! I actually kind of like that we have moved the rearranging and reformatting to the second process, as that makes the initial process cleaner and allows us to keep the "spaceranger format" for a bit longer, which may be beneficial down the line.

My only suggestions are quite minor about the naming of things (a difficult matter), so I don't think I need to review again.

modules/spaceranger.nf
allyhawkins: I was trying to match the format of the other folders that are present in the `internal` folder that have `-rna` appended to the folder. Whereas in the `publish` directory all of our filenames use underscores instead.

jashapiro: Okay, makes sense!

PR #136, opened 2022-03-23
Move ccdl-specific parameters to a profile

### LINKED ISSUE #123: Move ccdl-specific parameters to a profile
We may want to think about having some parameters, such as  `run_metafile` and `outdir`, set as blank by default, to prevent external users from getting cryptic errors if they do not set them and their run tries to access our files/locations for which they do not have permission. 

We could then set these values within a `ccdl` profile block, such that we would invoke them with: `-profile ccdl`, or more likely with `-profile ccdl,batch`, as you can include multiple profiles, comma separated.

Not quite as easy as using them by default, but not bad, and could save external user confusion.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/105#issuecomment-1059494399_

jashapiro:
Closes #123

This PR pulls the CCDL-specific settings out to a profile that can be invoked with `nextflow run AlexsLemonade/scpca-nf -profile ccdl` (once this is all merged to main, that is). 

I added default settings for the metadata file and output directory, and updated the docs to reflect that, but I am not sure whether that is better than leaving them blank by default. I welcome feedback on that decision.

I did set the defaults for the 10X containers to be blank, so that we only invoke our containers with the `-profile ccdl` argument.

Are there any other settings that I should move to the `ccdl` profile?

Please also look at the the updates to the documentation that I made to be sure those are clear and understandable. One notable change was to move the warning about who the instructions are for to the top.

Finally, I added a github action that will run `nextflow config` to just check that the config files are well-formatted. The main reason for this is that we can use it later to enforce that branches are up to date before merging to `development` or `main`, which we have found can eliminate some accidental reversions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ nextflow run AlexsLemonade/scpca-nf -profile ccdl
 ```
 Although running workflows locally can be done, we recommend using AWS batch for this workflow. 
 The first step in running the workflow is ensuring that your AWS credentials are configured. 
-()
 
 You can then run the same workflow with the `batch` profile, which has been configured in the `nextflow.config` file. 
 Note that you will still need the `ccdl` profile, and you can specify both with by separating them with a comma. 
--- a/README.md
+++ b/README.md
@@ -10,26 +10,30 @@ For more information on the processing of other modalities, please see the [ScPC
 
 ## Running scpca-nf for the ScPCA portal 
 
-To run `scpca-nf` in its default configuration: 
+The instructions below assume that you are a member of the CCDL with access to AWS.
+Most of the workflow settings described are configured for the ALSF Childhood Cancer Data Lab computational infrastructure. 
+To process samples that are not part of the ScPCA project, please see the [instructions on using `scpca-nf` with external data](external-data-instructions.md). 
+
+To run `scpca-nf` in its default configuration for the CCDL, you can use the following command: 
 
 ```
-nextflow run AlexsLemonade/scpca-nf 
+nextflow run AlexsLemonade/scpca-nf -profile ccdl
 ```
-
 Although running workflows locally can be done, we recommend using AWS batch for this workflow. 
 The first step in running the workflow is ensuring that your AWS credentials are configured. 
+()
 
 You can then run the same workflow with the `batch` profile, which has been configured in the `nextflow.config` file. 
+Note that you will still need the `ccdl` profile, and you can specify both with by separating them with a comma. 
 
 ```
-nextflow run AlexsLemonade/scpca-nf -profile batch
+nextflow run AlexsLemonade/scpca-nf -profile ccdl,batch
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -10,26 +10,30 @@ For more information on the processing of other modalities, please see the [ScPC
...
+To run `scpca-nf` in its default configuration for the CCDL, you can use the following command: 
 
 ```
-nextflow run AlexsLemonade/scpca-nf 
+nextflow run AlexsLemonade/scpca-nf -profile ccdl
 ```
-
 Although running workflows locally can be done, we recommend using AWS batch for this workflow. 
 The first step in running the workflow is ensuring that your AWS credentials are configured. 
+()
```
allyhawkins, line 24:
```suggestion
```
I think this is an accident?

nextflow.config
```diff
@@ -5,22 +5,22 @@ manifest{
...
-  version = 'v0.2.4'
+  version = 'v0.2.5'
 }
 
 
 // global parameters for workflows
 params {
   // Input data table
-  run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+  run_metafile = "run_metadata.tsv"
```
allyhawkins, line 15: Hmm... I'm torn on if we want to have a default here that's different from the `user_template.config` file? I feel like if we are going to put in a dummy default, then maybe we want to use our example metadata file that we have just for consistency?

nextflow.config
```diff
@@ -5,22 +5,22 @@ manifest{
...
 
   // Output base directory
-  outdir = "s3://nextflow-ccdl-results/scpca/processed"
+  outdir = "scpca_output"
```
allyhawkins, line 18: Same thing here. In the example config, we have this as `scpca_out` so should we be consistent?

allyhawkins (commented):
Generally all the changes look good to me. And it doesn't look like you missed any parameters. I'm fine with having default values for the `run_metafile` and `outdir`, however, I wonder if we want to use the same example defaults that we use in `user_template.config`? I could be convinced either way, just like I can also see an argument for leaving them blank, but I think for at least the output directory they should match.

nextflow.config
jashapiro: That's a good point, and I hadn't really thought about it. At first, I thought I agreed that consistency would be good here, and was thinking about which I would want to change. But then I thought maybe we should have different values to justify the fact that you might change it with a config file (If it is the same value, why bother specifying it?).

I feel like the very generic version here works as a default, whereas `example_metadata.tsv` works when people look into the examples folder. So I think I lean toward leaving it this way?

nextflow.config
jashapiro: Contrary to my previous comment, this difference is minor enough that I think it makes sense to change one or the other to be consistent. I will update it.

allyhawkins (approved):
LGTM!

PR #137, opened 2022-03-24
Integrate vireo output with SCE objects

### LINKED ISSUE #96: Integrate vireo output with SCE objects
Similarly to how we integrate feature data in the `generate_merged_sce` sub-workflow, we will want to integrate the results from the genetic demultiplexing workflow with the gene expression results from alevin-fry. This workflow step will take advantage of the functions developed as part of https://github.com/AlexsLemonade/scpcaTools/issues/93. 

This workflow output should integrate with the previous outputs to result in a channel with a single unfiltered SCE object per library, which can then be passed along to filtering & QC without changes.

jashapiro:
Closes #96 (stacked on #102, which is stacked on #80)

⚠️ This depends on `scpcaTools:edge`, so should not be merged until a new release of scpcaTools has been created.

This PR started with the intent of simply adding the `merge_vireo_sce.R` script and associated process in order to allow integration of genetic demux results in the form of `vireo` output into the existing SCE results. 

~~Doing that ended up requiring a bit more rejiggering of the main workflow than expected. The main change there was splitting the SCE creation and filtering out of subworkflows, because it seemed to make sense to keep everything as unfiltered data sets and do all the filtering at once, just before the QC report.~~
Update: undid this, and now using the filtered data only for adding the genetic demultiplexing results (and in the future the other cellhash results). Logic here is partly to avoid annoying limitations in Seurat, but a post-hoc justification is that that unfiltered data is probably more there for "expert" use, in which case the user can pull extra annotations as needed. (Empty drops are extremely unlikely to get genotype-based demux calls anyway.)

The other change is the use of `.branch` to split up the various data types right at the start, rather than filtering at each analysis stage. I think this cleans things up a bit.

In the other direction though, the merging of the SCE and vireo results is a bit messy, but hopefully not too bad, as it follows the same pattern as other parts of the workflow, generally speaking, of pulling out the library id from `meta`, joining on that, then dropping it again before continuing with the next step.

I also reconsidered using `_` as the sample separator, based on the fact that it seemed quite possible that a `sample_id` from an external user could have an underscore. I do still change it, now to `,`, because a semicolon is more likely to result in errors at the command line, but I will note that the R script which incorporates the `sample_ids` and `library_id` into the SCE metadata will accommodate either commas or semicolons.


jashapiro:
I have finished testing on this, so I think it is ready for review.  However, I am leaving it in draft form, due to the `scpcaTools` issue discussed earlier. I will start the process for a new `scpcaTools` release in that repository.

modules/sce-processing.nf
```diff
@@ -4,7 +4,7 @@
 process make_unfiltered_sce{
     container params.SCPCATOOLS_CONTAINER
     label 'mem_8'
-    publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
```
allyhawkins, line 7: Why did you remove publishing the unfiltered RDS files from here? Am I missing something? I think we still need this to be published because we want to provide these output files to users.

allyhawkins: Ah I'm seeing now, we have it being published when we perform filtering so we don't need it here.

modules/sce-processing.nf
```diff
@@ -72,6 +77,27 @@ process filter_sce{
...
+  publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
+  input:
+    tuple val(demux_meta), path(vireo_dir),
+          val(meta), path(unfiltered_rds), path(filtered_rds)
+  output:
+    tuple val(meta), path(unfiltered_rds), path(filtered_rds)
+  script:
+    // output will be same as input, with replacement of the filtered_rds file
+    """
+    mv ${filtered_rds} input.rds
```
allyhawkins, line 93:
```suggestion
    mv ${filtered_rds} no_demux_filtered.rds
```
This is minor, but I think we should change this to a more informative name than `input`, so that it's clear that this is the same as the filtered data but without the demux results. And the output will now have the demux results. You don't necessarily need to choose this name, but I just was trying to come up with an option.

modules/sce-processing.nf
```diff
@@ -72,6 +77,27 @@ process filter_sce{
...
+  label 'mem_8'
+  tag "${meta.library_id}"
+  publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
+  input:
+    tuple val(demux_meta), path(vireo_dir),
+          val(meta), path(unfiltered_rds), path(filtered_rds)
+  output:
+    tuple val(meta), path(unfiltered_rds), path(filtered_rds)
+  script:
+    // output will be same as input, with replacement of the filtered_rds file
```
allyhawkins, line 91:
```suggestion
    // output will be same as input, with replacement of the filtered_rds file with the same filtered SCE object but with demultiplexed results appended to the colData
```
Maybe some more description to explain that it's not exactly the same file that's coming out but a similar file with new information added would be helpful here.

allyhawkins (commented):
I think this looks good and I like the change to using branch. I had some minor comments about naming and comments but I'll wait to approve until this is on a new release of `scpcaTools`.

jashapiro:
Updated the tag on the scpcaTools container, so this should be ready to go!

allyhawkins (approved):
LGTM 🚀

PR #139, opened 2022-03-30
No longer require list of files for spaceranger

### LINKED ISSUE #117: Remove `files` as required input field for spaceranger
In the interests of making the workflow easier to use and more robust, we might not want to require a list of input files in the library metadata file. For most modules, we get the lists of files programmatically from the contents of `files_directory`, and we _ought_ to be able to do that for spaceranger too.

>> Just noting that we do use it in the spaceranger workflow so we should address that you will need to add it when we add the spaceranger section in #114.

> Oops, I missed that. Never mind!
> We might say that filling it in isn't required unless using spaceranger. Longer term, we might see if we can remove that dependency and just get the file list programmatically as we do elsewhere.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/110#issuecomment-1057268369_

allyhawkins:
Closes #117. 

This PR removes the need to have a separate column in the metadata file with the list of files, specifically for external users using the spaceranger workflow. We had previously been grabbing the sample names needed to run spaceranger from the list of file names, but here we pass the directory through to the `getCRsamples` function, list all the files present in the directory and then grab the sample names that way. In addition, I removed the addition of files to the `unfiltered_runs_ch` in the main workflow, as we no longer need that column for any of the workflow to work. 

I also updated the example metadata to remove that column and updated the external instructions to remove any mention of a `files` column. I did not actually remove the `files` column from our metadata, as we use the list of files to perform data integrity checks in our internal workflows. I tested this with a spatial sample and it successfully populates the `--sample` option when running spaceranger with the names from the files as it was previously.


jashapiro (approved):
LGTM!

Issue #141, opened 2022-03-30
Prepare for scpca-nf release v0.2.5


allyhawkins:
## Steps for a new release of `scpca-nf`
We need a new release in order to process the spatial samples to hand over to devs and see if there is any more modifications that we will need to be made. 
This release would also incorporate the genetic demultiplexing workflow up through changes in #137. Do we also want to address #100 and #138 for this release or incorporate those for the next release which I assume will not be far off? Tagging @jashapiro for any thoughts on this. 

### Preparing for the release

- [x] Are all of the issues planned for this release resolved? If there are any issues that are unresolved, mark this issue as blocked by those on ZenHub.
- [x] Update code and documentation with the latest version number in the `development` branch:
  - [x] [nextflow.config](https://github.com/AlexsLemonade/scpca-nf/blob/main/nextflow.config) 
  - [x] [README.md](https://github.com/AlexsLemonade/scpca-nf/blob/main/README.md)
  - [x] [external-data-instructions.md](https://github.com/AlexsLemonade/scpca-nf/blob/main/external-data-instructions.md)
- [x] Test that the workflow is in good working order with `nextflow run alexslemonade/scpca-nf -latest -r development`
- [x] File a PR from the `development` branch to the `main` branch. This should include all of the changes that will be associated with the next release.

### Creating a release
- [x] On the [releases page](https://github.com/AlexsLemonade/scpca-nf/releases), choose `Draft a new release`.
- [x] In `Choose a tag`, type a new release number using semantic versioning (vX.X.X) (you did update the title of this issue to match, right?), then click `Create a new tag: vX.X.X on publish`.
- [x] Write a description of the major changes in this release. You may want to start with the auto-generated release notes to save time.
- [ ] Optional: If not all issues have been addressed, save a draft to return to later.
- [x] Publish the release!

jashapiro:
I think #100 can wait, but I think some progress on #138 would be a good idea, if we can wait for it. It should probably include a note about the limitations described in #100. I just filed #140, but that probably doesn't need to be part of this for now.

allyhawkins:
Noting that this release is complete and includes updates for spatial and the changes to the config files, but does not include anything related to genetic demultiplexing. Changes that incorporate that workflow will come in a later release.

PR #142, opened 2022-03-31
Merge v0.2.5 release changes into main

allyhawkins:
This PR incorporates the changes needed for the release v0.2.5 as mentioned in #141. Here, we are not including any changes related to genetic demultiplexing and we will wait until those changes are complete and documentation has been added before merging into main and incorporating into a release. 

This includes changes that were merged in #134 and #139 to complete the spaceranger workflow. 
I also added the changes from #136 to add in the ccdl specific profile as well (Let me know if I should remove these changes for any reason, but there were no conflicts when merging these). 
I then updated all the release numbers from `0.2.4` to `0.2.5`. 

I did make one minor change in the workflow in `main.nf` related to  https://github.com/AlexsLemonade/ScPCA-admin/issues/341 and that is to remove `visium_v1` and `visium_v2` from the list of technologies and allow for `visium` to be an accepted technology in anticipation of changing all libraries to this tech. I also updated the line in the external instructions that lists the options for tech to reflect this change and added a note at the end of the spatial section that states that only visium libraries have been tested with the workflow and no other spatial transcriptomics methods have been used. I wasn't sure if this was completely necessary, but it seemed like something important to have if we are using `visium` as the technology listed.

### DIFF
--- a/.github/workflows/nextflow-config-check.yaml
+++ b/.github/workflows/nextflow-config-check.yaml
@@ -1,0 +1,19 @@
+
+name: Check nextflow config
+
+on:
+  pull_request:
+    branches:
+      - main
+      - development
+
+
+jobs:
+  nf-config-check:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check nextflow params
+        uses: docker://nextflow/nextflow:21.10.6
+        with: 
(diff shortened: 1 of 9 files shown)


jashapiro:
When this is done, please file a followup PR merging `main` -> `development` to keep things in sync.

PR #143, opened 2022-03-31
Sync changes from v0.2.5 in main to development

allyhawkins:
Here I am merging any changes that were made in #142 back to the `development` branch. To do this I just opened a PR comparing the two branches from each other so let me know if I did something wrong because I'm seeing diffs for files that should already be in `development` and were unmodified between `development` and `main`, which was confusing me but I didn't want to mess around with it too much and break something so trying this first...


jashapiro:
Okay, that is odd, and there is a conflict that should not be there too... So I think we will have to make a branch off `development`, merge `main` into that, then resolve conflicts and merge back to development.

PR #144, opened 2022-03-31
Actually merge changes from v0.2.5 in main to development

allyhawkins:
Okay this PR actually takes the changes from main and merges them into a separate branch off `development` and then is merging those changes into `development`. As expected, here there are only the changes to the spatial workflow and spatial instructions that we would expect. Note that the version in `development` had already been increased to `v0.2.5`. I will close out the previous PR #143.


jashapiro (approved):
LGTM!

PR #148, opened 2022-04-05
Move addition of cellranger index name to `spaceranger_publish` process to avoid null value when spaceranger is skipped

allyhawkins:
In processing the spatial libraries, I noticed that some of the `metadata.json` values had a null value in the slot for the `mapping_index`, which should correspond to the base filename of the index that was used for mapping. Interestingly, this was only a problem for libraries where the spaceranger process had completed successfully once and the process was being skipped upon re-running the entire project, but not on samples where they whole workflow was run from start to finish. This is because we were adding in the name of the index to the metadata in the spaceranger process, so when that process is skipped, it doesn't get added and then that value is null in the metadata. To accommodate for times when we skip spaceranger but still want to re-generate the metadata file, I moved the addition of the file name to the metadata to the `spaceranger_publish` process instead. I tested this and it does produce a metadata file that looks as expected when skipping the spaceranger process. 

I did think about also directly adding the filename to the metadata at the same time we are adding the publish directories, but I was getting some errors upon doing that so the method I implemented here seemed more straight forward. 

I made this PR to `main` here, as I plan on then creating a PR to `development` with the exact same changes. We would then need a new release in order to process the spatial project to hand over but want to make sure these changes get in before we incorporate any demultiplexing changes.

### DIFF
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -15,7 +15,6 @@ process spaceranger{
     tuple val(meta), path(out_id)
   script:
     out_id = "${meta.run_id}-spatial"
-    meta.cellranger_index = index.fileName
     """
     spaceranger count \
       --id=${out_id} \
@@ -38,10 +37,12 @@ process spaceranger_publish{
   publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
   input:
     tuple val(meta), path(spatial_out)
+    path index
   output:
     tuple val(meta), path(spatial_publish_dir), path(metadata_json)
   script:
     spatial_publish_dir = "${meta.library_id}_spatial"
+    meta.cellranger_index = index.fileName
     metadata_json = "${meta.library_id}_metadata.json" 
     workflow_url = workflow.repository ?: workflow.manifest.homePage
     """
@@ -123,7 +124,7 @@ workflow spaceranger_quant{
         grouped_spaceranger_ch = spaceranger.out.mix(spaceranger_quants_ch)
 
           // generate metadata.json
-        spaceranger_publish(grouped_spaceranger_ch)
+        spaceranger_publish(grouped_spaceranger_ch, params.cellranger_index)
 
     // tuple of metadata, path to spaceranger output directory, and path to metadata json file
     emit: spaceranger_publish.out
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -37,12 +37,12 @@ process spaceranger_publish{
   publishDir "${params.outdir}/publish/${meta.project_id}/${meta.sample_id}"
   input:
     tuple val(meta), path(spatial_out)
-    path index
+    val index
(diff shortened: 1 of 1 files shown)


jashapiro (commented):
There are a couple of reasons I don't really like this solution, but it might be worth doing as a temporary measure.

The first is that passing the index to the publish step requires passing a fairly large directory, and I would rather not do that if we don't have to. This can be fixed by passing it as a value rather than a path, so I have added that suggestion in a way that I think will work, but that requires testing.

The second thing is that what we have done here is to separate the actual index use from where we record what file was used, which is not great (they could get out of sync if we created a new index). I don't think this is the only place we do this, and it was something I was thinking about fixing more generally with the kind of approach I took with genetic demultiplexing: for steps that we might skip, we can write out the metadata as a json file, and then read it back in before the next step.  Example code: 

https://github.com/AlexsLemonade/scpca-nf/blob/1bb59c6ca848341f7bc4f47a45a4323db62b851d/modules/cellsnp.nf#L45-L55
https://github.com/AlexsLemonade/scpca-nf/blob/1bb59c6ca848341f7bc4f47a45a4323db62b851d/modules/genetic-demux.nf#L58-L61

I _do not_ think this is the time/PR to fix that issue, but it was in my mind, so I wanted to get it written down somewhere (I will make this into an issue separately).

Finally, a note on process for merging this to `development`. The easiest thing _should_ be to to merge this into `main`, but rather than deleting the branch when that is done, you can then merge `development` into this branch and file a PR to `development` directly from the same branch.

allyhawkins:
> The second thing is that what we have done here is to separate the actual index use from where we record what file was used, which is not great (they could get out of sync if we created a new index). I don't think this is the only place we do this, and it was something I was thinking about fixing more generally with the kind of approach I took with genetic demultiplexing: for steps that we might skip, we can write out the metadata as a json file, and then read it back in before the next step.

I think this is a really good point and I like the idea of implementing this. It would mean re-running all of the samples and repeating mapping so I agree that for the interest of time this should be a later implementation. 

I also wanted to note that I tested out use of `val index` as the input and that doesn't work... Nextflow complains with an uninformative error and I was sure to change the line to `file(index).fileName` and that still doesn't work. I changed it back to reading in the index which I know is not ideal but since we are planning on changing it, I think it is okay for now.

jashapiro (approved):
> Nextflow complains with an uninformative error and I was sure to change the line to `file(index).fileName` and that still doesn't work.

:sighs at nextflow errors:

I may have gotten the code wrong, I noticed in another place we used this pattern, but with `.name`.

https://github.com/AlexsLemonade/scpca-nf/blob/1bb59c6ca848341f7bc4f47a45a4323db62b851d/modules/cellsnp.nf#L44 

If you want to test that, you can, but I'll approve it either way.

allyhawkins:
Using `.name` works and is much faster! Thanks for the suggestion!

PR #150, opened 2022-04-06
Update version in config file and docs for new release 

allyhawkins:
Since incorporating #148, we need to create a new release to be able to run the spatial samples. This PR just updates the version numbers in the documents specified in the release checklist to be v0.2.6. I am merging these to `main` and then I can create a PR to `development` after merging these changes. I tested opening a PR with the new changes incorporated into `main` in #148 and it showed only the one file that was modified as expected so hypothetically we shouldn't have any issues after these changes are merged.


jashapiro (approved):
LGTM, though I wonder if we might want to reduce the number of places that we include the revision in the future?

PR #164, opened 2022-04-12
Move spatial metadata.json to be nested inside spatial folder

### LINKED ISSUE #160: Move metadata file in spaceranger process
We would like to be able to provide the `_metadata.json` file for each spatial library with each download. Right now that file is not included in the `SCPCL000000_spatial` folder which contains all of the output for each library but it is outside of that folder. This means that it is harder to distinguish it from the `_metadata.json` files that are for single cell libraries vs. spatial libraries. The single cell libraries will not be included in the download, while spatial ones will. Additionally, it would make sense that for each `_spatial` folder that is included in a user's download, it would contain all the files related to that library. 

To improve the file organization and ensure that all files related to each library are nested within the `_spatial` folder, we should move the `_metadata.json` file that is currently outside the folder to inside the folder. This would mean a change in the [line that defines the path to the `_metadata.json` file in the spaceranger workflow](https://github.com/AlexsLemonade/scpca-nf/blob/0112aac248f0015d3b87602d7ceb387bc593f15b/modules/spaceranger.nf#L47). 

**Additional context**
We may want to pair this change with changes proposed in #149 so that we can make all changes related to the metadata together before creating any new releases and re-running the spatial project.

allyhawkins:
Closes #160. Here I updated the line in the spatial workflow that defines the path to the `metadata.json` file to now output the file to be inside the `SCPCL000000_spatial` directory as discussed in our design meeting. I tested this change and the output folder now looks as expected and the metadata file itself still contains the contents it should have. 

I'm also updating the relevant version numbers here as the next steps after filing this PR is to file a PR to `main` with the same exact changes that are included here and then create a new release so that we can process the spatial project again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ nextflow run AlexsLemonade/scpca-nf -profile ccdl,batch
 When running the workflow for a project or group of samples that is ready to be released on ScPCA portal, please use the tag for the latest release: 
 
 ```
-nextflow run AlexsLemonade/scpca-nf -r v0.2.6 -profile ccdl,batch --project SCPCP000000
+nextflow run AlexsLemonade/scpca-nf -r v0.2.7 -profile ccdl,batch --project SCPCP000000
 ```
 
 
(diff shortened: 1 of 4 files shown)


jaclyn-taroni (approved):
👍🏻 the change to `spaceranger.nf` seems reasonable and I checked locally to see if all the instances where `0.2.6` were present in the repo were updated here.

Issue #157, opened 2022-04-08
Restructure instructions for external users so it doesn't seem like it's quick


dvenprasad:
In the intro, the amount of work involved before they can run the command is kinda minimized. I think we should:
- In `How to use scpca-nf as an external user`
   - Expand on what file organization means and then link to File organization section
   - Expand on what creating a metadata file means and link to metadata file section
   - Expand on what config file means and link to config file section
   - Then show them the command. I also think there were different variations of running the commands, which went over my head. but I think we should link them here.
   - Talk about the output files briefly here-  maybe the show the folder structure and then link to output file structure
 

@sjspielman suggested having the different variations of commands together. (please tell me if I misunderstood)
From the last session with  @jaclyn-taroni and @jashapiro,  we wanted to remove that users could run them locally.

allyhawkins:
Closed by #168.

Issue #153, opened 2022-04-07
Update documentation with resource requirements


jashapiro:
We should provide an initial strong caution in the docs about the resources that are needed for running the pipeline. Currently for alevin-fry we need at least 12 cpus & 24 GB of RAM (though 2X that if a run fails for any reason). 

This means that running locally is _possible_ but only on a fairly well-specced machine (not most laptops).

We should probably remove any references to running the workflow _without_ setting up a profile, as that is just setting people up for failure.

As we address questions in #151, we might be able to rescope a bit (fewer cpus is possible, but probably not less RAM?), but the warning should remain.

allyhawkins:
Closed by #171

PR #167, opened 2022-04-12
Reorganize main readme file to include section for external users

### LINKED ISSUE #156: Restructure README so link to setup externally is easy to spot
The link to setup externally is currently under setting up. In order to make it clear that there are instructions for two types of users (external and Data Staff), I propose the following structure

**scpca-nf**
[same content]

**Running scpca-nf in your HPC environment** (want to be a bit more descriptive than external user, what qualifies someone as an external user?)
Say a bit more about what that entails and link to the doc. I'm not sure if we should put the command here (in the interest of not misleading users to think that this is a quick and easy setup)

**Running scpca-nf for the Data Lab Staff**
same content, without external stuff under Running scpca-nf for the ScPCA portal

allyhawkins:
Closes #156. 

Here I reorganized the main `README.md` file so that we have included a section geared towards outside users looking to process their samples with `scpca-nf`. Following the description in #156 and some feed back in the pilot sessions, I kept the beginning intro section unchanged but had the next header be specific to running outside samples for those outside of CCDL. I then changed the next header to make it clear that that section is specifically for CCDL staff. 

In the section for other users, I added the link to the external instructions and also included a brief outline of the steps that they will need to take with links to the sections that have each of the relevant information. I was back and forth on how much information is too much vs. how much information is helpful to get an idea of what you will need to setup. 

I didn't put the command to actually run here, because I want people to actually go to the link and read about how to set up the files. I also anticipate we will do some further reorganization of the external data instructions so that we might have a more detailed intro at the beginning of that (based on #157) where I think commands would be appropriate. 

The other thing I added here was a note about needed specific resources as to begin to address #153. We will also need to make changes throughout to remove the reference to running locally and also will want to put these resource requirements in the main external instructions which is a separate PR, but I thought it would be helpful to have here on the front page as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,14 @@ This repository holds a [Nextflow](https://www.nextflow.io) workflow to process 
 Fastq files for single-cell and single-nuclei RNA-seq samples are processed using [alevin-fry](https://alevin-fry.readthedocs.io/en/latest/).
 All samples are aligned, using selective alignment, to an index with transcripts corresponding to spliced cDNA and to intronic regions, denoted by alevin-fry as `splici`. 
 `scpca-nf` can also process CITE-seq, bulk RNA-seq, and spatial transcriptomics. 
-For more information on the processing of all types of modalities, please see the [ScPCA Portal docs](https://scpca.readthedocs.io/en/latest/). 
+For more information on the processing of all modalities, please see the [ScPCA Portal docs](https://scpca.readthedocs.io/en/latest/). 
 
-## Using scpca-nf to process samples in your HPC
+## Using scpca-nf to process your samples
 
-The `scpca-nf` workflow is currently set up to process samples as part of the ScPCA portal and requires access to AWS through the CCDL.  
+The `scpca-nf` workflow is currently set up to process samples as part of [the ScPCA portal](https://scpca.alexslemonade.org/) and requires access to AWS through the Data Lab.  
 For all other users, `scpca-nf` can be set up to process your samples in your computing environment by following the [instructions for using `scpca-nf` with external data](external-data-instructions.md). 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,10 +5,24 @@ This repository holds a [Nextflow](https://www.nextflow.io) workflow to process
...
+
+:warning: Please note that processing single-cell and single-nuclei RNA-seq samples, requires access to a high performance computing (HPC) environment that can accomodate up to 24 gb of RAM and 12 cpus. 
+
+To run `scpca-nf` on your own samples, you will need to complete the following steps: 
+
+1. [Organize your files](./external-data-instructions.md#file-organization) so that each folder contains fastq files relevant to a single sequencing run. 
+2. [Prepare a metadata file](./external-data-instructions.md#prepare-the-metadata-file) with one row per library containing all information needed to process your samples.
+3. Set up a [configuration file](./external-data-instructions.md#configuration-files), including the [definition of a profile](./external-data-instructions.md#setting-up-a-profile-in-the-configuration-file), dictating where nextflow should execute the workflow. 
+
+If you would like, after you created your configuration file, you may also [test your set up using example data](examples/README.md).
```
jaclyn-taroni, line 23: Pretty sure they need to create a metadata file too?

```suggestion
You may also [test your set up using example data](examples/README.md) after completing steps 2 and 3.
```

jaclyn-taroni (approved):
Looking forward to having these changes make it in. I had a few suggestions, but no need for me to re-review.

README.md
allyhawkins: They need to create a metadata file that specifically has the information for the example data, not their own sample information. Those instructions are in the readme for the example data so it really is just about testing your configuration file. 

How about just this? 
`You may also [test your configuration file using example data](examples/README.md).`

PR #169, opened 2022-04-14
Make publishing alevin-fry output optional

### LINKED ISSUE #154: Consider removing publishing from alevin-fry step
Since we do not use the output from `alevin-fry` as a checkpoint, we might not want to include it in the `interal` outputs as we proceed toward wider use.

To make this change, we would remove the following line from the module: 
 https://github.com/AlexsLemonade/scpca-nf/blob/0112aac248f0015d3b87602d7ceb387bc593f15b/modules/af-rna.nf#L48

Similarly, we should probably remove `publishDir` directives from the feature mapping: https://github.com/AlexsLemonade/scpca-nf/blob/main/modules/af-features.nf We don't currently checkpoint feature mapping at all (though we could)

If we need to look at any of these files for development or debugging, the `publishDir` directive can always be added back temporarily.

One further thought I had was also to rename `internal` to `checkpoints` or similar to make it more clear what the function of that folder is.

allyhawkins:
Closes #154. 

Here the main change I made was to publishing the alevin-fry output files for both the single-cell and CITE-seq alevin fry processes. I utilized the [`enabled` option within Nextflow `publishDir`](https://www.nextflow.io/docs/latest/process.html#publishdir) to only output the files when specified at the command line. I created the parameter  `--publish_fry_outs` (open to new names if anyone has ideas...) and stored it as `false` so that by default we aren't writing out the files. Then if we want to write out the files (or a user wants to write out the files, then you can just use `--publish_fry_outs` at the command line when setting up the run. I tested these changes and it worked as expected, only writing out the files when using `--publish_fry_outs`. 

The other change that was discussed in #154 was using a more meaningful name for the `internal` folder. Here, I switched the names of the folders: 
`internal` --> `checkpoints` 
`publish` --> `results` 
I'm not crazy about results and also thought maybe using `expression_data` or `counts_data`, but was torn so went with `results` and am open to new ideas. 

A few things I thought about when making this change: 
- If I were to just change every place we have `internal` to `checkpoints` then when we went back to re-process any of our internal projects we would have to start from scratch, because the location of the rad files would change. Although we could just re-name the folder on S3 there aren't really folders on S3 so I think we have to copy it to a new name. It's probably feasible but not ideal. 
-  I was noticing that I was making the change in a lot of places... You'll notice that there are a lot of files in this PR but that's mostly because there are files with one line changes from `internal` to `checkpoints`. It seemed not ideal to be making so many of the same changes to multiple files. 

The solution that I came up with to address this was to turn the `checkpoints_dir` and `results_dir` into parameters in the main config file, utilizing the `outdir` as a reference to build the paths to those directories. I did this in both the main `nextflow.config` as well as in the config that holds the `ccdl` profile. The reason for this is twofold - one because if you don't put it in the `ccdl` profile then the only one present is the one in the main config file and then it uses the `outdir` mentioned in the main config file and writes it to a folder `scpca_out` on your local computer rather to the location on S3 where it should be (even when specifying on the command line where to send the outs too). The second reason is this allows us to keep our internal structure for our use if we want and also use an external structure with more meaningful names for them. 

Lastly, I updated the external instructions to reflect the name change and the addition of the optional flag to write out the alevin-fry results. In doing this, I included a tree of the `checkpoints` folder that only shows a limited view of the files since there was a lot of discussion of trees being helpful during the pilots last week.

### DIFF
--- a/external-data-instructions.md
+++ b/external-data-instructions.md
@@ -268,36 +268,58 @@ After building the docker image, you will need to push it to a [private docker r
 ## Output files 
 
 Upon completion of the `scpca-nf` workflow, the results will be published to the specified `outdir`. 
-Within the `outdir`, two folders will be present, `publish` and `internal`. 
+Within the `outdir`, two folders will be present, `results` and `checkpoints`. 
 
-The `publish` folder will contain the final output files produced by the workflow and the files that are typically available for download on the ScPCA portal. 
+The `results` folder will contain the final output files produced by the workflow and the files that are typically available for download on the ScPCA portal. 
 
-Within the `publish` folder, all files pertaining to a specific sample will be nested within a folder labeled with the sample ID.
+Within the `results` folder, all files pertaining to a specific sample will be nested within a folder labeled with the sample ID.
 All files in that folder will be prefixed by library ID, with the following suffixes:  `_unfiltered.rds`, `_filtered.rds`, `_metadata.json`, and `_qc.html`. 
 The `_unfiltered.rds` and `_filtered.rds` files contain the quantified gene expression data as a [`SingleCellExperiment` object](https://bioconductor.org/packages/release/bioc/html/SingleCellExperiment.html).
 For more information on the contents of these files, see the [ScPCA portal docs section on single cell gene expression file contents](https://scpca.readthedocs.io/en/latest/sce_file_contents.html).
 
-See below for the expected structure of the `publish` folder: 
+See below for the expected structure of the `results` folder: 
 
 ```
 publish
-└── SCPCS999991
-    ├── SCPCL999991_filtered.rds
-    ├── SCPCL999991_metadata.json
-    ├── SCPCL999991_qc.html
-    └── SCPCL999991_unfiltered.rds
+└── sample_id
+    ├── library_id_filtered.rds
+    ├── library_id_metadata.json
+    ├── library_id_qc.html
+    └── library_id_unfiltered.rds
 ```
 
 If bulk libraries were processed, a `bulk_quant.tsv` and `bulk_metadata.tsv` summarizing the counts data and metadata across all libraries will also be present in the `publish` directory. 
 
-The `internal` folder will contain intermediate files that are produced by individual steps of the workflow, including mapping with `salmon` and quantification with `alevin-fry`. 
-The contents of this folder are used to allow restarting the workflow from internal checkpoints (in particular so the initial read mapping does not need to be repeated), and may contain log files and other outputs useful for troubleshooting or alternative analysis.
+The `checkpoints` folder will contain intermediate files that are produced by individual steps of the workflow, including mapping with `salmon`. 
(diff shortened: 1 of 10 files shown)


nextflow.config
```diff
@@ -17,6 +17,10 @@ params {
   // Output base directory
   outdir = "scpca_out"
 
+  // create paths to output folders
+  checkpoints_dir = "$outdir/checkpoints"
+  results_dir = "$outdir/results"
```
jashapiro, line 22: `results` seems good to me. I had the thought of maybe `final`, but that seems like a bad assumption to ever make.

nextflow.config
```diff
@@ -31,7 +35,8 @@ params {
   af_resolution = 'cr-like-em' // alevin-fry quant resolution method: default is cr-like, can also use full, cr-like-em, parsimony, and trivial
   repeat_mapping = false // if alevin or salmon mapping has already been performed and output files exist, mapping is skipped by default. Use `--repeat_mapping` to perform mapping again
   repeat_gdemux = false // if genetic demultiplexing has been performed and output files exist, genetic demux is skipped by default. Use `--repeat_gdemux` to run these steps. 
-  
+  publish_fry_outs = false // alevin-fry outputs are not published by default. Use `--write_fry_outs` to publish these files to the `checkpoints` folder.
```
jashapiro, line 38: Making the comment match.
```suggestion
  publish_fry_outs = false // alevin-fry outputs are not published by default. Use `--publish_fry_outs` to publish these files to the `checkpoints` folder. 
```

external-data-instructions.md
```diff
@@ -268,36 +268,58 @@ After building the docker image, you will need to push it to a [private docker r
...
 Additionally, for each run, all files related to that run will be inside a folder labeled with the run ID followed by the type of run (i.e. `rna` or `features` for CITE-seq) and nested within the library ID folder.
+
+See below for the expected structure of the `checkpoints` folder: 
+
+```
+checkpoints
+├── rad
+│   └── library_id
+│       └── run_id-rna
+└── salmon
```
jashapiro, line 309: I think we might want to have a bit more of an example here, with maybe 2 libraries, one with feature data? We could even use values from example data, if we wanted.

```suggestion
checkpoints
├── rad
│   ├── library01
│   │   ├── run01-rna
│   │   └── run02-features
│   └── library02
│       └── run03-rna
└── salmon
```

jashapiro (approved):
This looks good to me overall. 

I don't think moving the files on S3 will be that big of a deal; `aws s3 mv --recursive` is pretty easy, but I also think using variables here is a good idea, so I can go either way. I could even imagine some people ignoring the outdir completely, and just directly specifying the directories for checkpoints and results (maybe just `scpca_results` and `scpca_checkpoints` with no nesting), but I don't think we need to get into those kinds of details. 

My only substantive comment was about a change to how much detail in the tree you show. I'll leave it to you to decide how much we really want to show. The example I made doesn't fully match the example data, which could be a bit confusing, but I liked showing a bit more...

Issue #163, opened 2022-04-12
Modify the sample and library ID's in the example metadata file to remove `SCPCX` formatting


allyhawkins:
As originally posted in #155 below, we should modify the current example metadata to remove the `SCPCX` formatting. This may be less confusing to users as the formatting is not required for them to run the pipeline. 

I've been thinking about whether we really want to use SCPCX prefixes here. I think maybe not, as I think they may be confusing in this context. There is no requirement that they follow our ID format if they are just running on their own. 

I think we might want to change the example files as well, but that might warrant some more discussion.

I also made this look like a full path.

```suggestion
| run01 | library01 | sample01 | project01 | 10Xv3.1 | cell | /path/to/example_fastq_files | 
```

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/155#discussion_r846373214_

allyhawkins:
Closed by #170.

PR #170, opened 2022-04-15
Reorganize example metadata

### LINKED ISSUE #158: Group required and optional fields in the example metadata file
I think we should reorder the fields so that all the required fields are shown first and then the optional fields. It would be best if it reflected the order that is shown in the documentation.

allyhawkins:
Closes #158 and #163. 

Here I am updating the `example_metadata.tsv` file that we provide to mirror the external user instructions and to have run/library/sample ID's that are easier to understand for external users and not of the format `SCPCX`. I started by moving all of the optional fields to be in the second half of the columns so that the first columns are all of the required fields. 

In doing this I made a few changes to the actual columns: 

- I added `scpca_project_id` to the list of required columns. Previously it wasn't listed at all although we had it in the example file. The reasoning for this is because right now the workflow is set up to nest the output by project ID. If a project ID is not present then it will nest the output into a folder with `no project` as the name. I thought the simpler way to go about this was to add it to the required input rather than to explain that it's optional and that you will get results in a `no project` folder. I also was thinking that maybe we should figure out a way to make nesting by project ID optional in general... but that's a different discussion entirely.
- I also removed `submitter` from the example metadata and changed it to `submitter_id`. We had listed both of these columns in the optional columns list, but I'm not sure what value `submitter` is giving external users since they are the ones trying to run their own data. I thought that `submitter_id` made more sense as that allows them to map back to any original ID's that they use, especially in a case where we might provide them run/sample/library IDs they would be able to use that column to map back to their own personal sample Ids. 

In addition to rearranging the columns I modified the formatting of the IDs, using the format of `run01` and `library01` rather than `SCPCX`. This is consistent with what we landed on when writing the instructions for testing the example data. I updated the ID's and updated the example paths to the fastq files, using the format of `example_fastqs/run01`. Hopefully that makes things a little more clear that each folder should hold an individual sequencing run. If others have other suggestions on how to format these ids please let me know and I'm happy to discuss them.

### DIFF
--- a/examples/example_metadata.tsv
+++ b/examples/example_metadata.tsv
@@ -1,7 +1,7 @@
-scpca_run_id	scpca_library_id	scpca_sample_id	scpca_project_id	technology	seq_unit	files_directory	submitter_id	feature_barcode_file	feature_barcode_geom	slide_section	slide_serial_number
-run01	library01	sample01	project01	10Xv3.1	cell	example_fastqs/run01	example01				
-run02	library01	sample01	project01	CITEseq_10Xv3	cell	example_fastqs/run02	example02	example_barcode_files/cite_barcodes.tsv	2[1-15]		
-run03	library02	sample02	project01	visium	spot	example_fastqs/run03	example03			A1	VXXXX-XXX
-run04	library03	sample03;sample04	project01	10Xv3.1	cell	example_fastqs/run04	example04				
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
This looks good to me, but I think I might want to just leave out the `submitter_id` column in the example table. I would expect most people to use their own ids except when we provide SCPCX ids, so for most of these cases I would not expect people to use that column in testing. The provided `example00` values might also be confusing. 

Other than that, I think this looks good, so I am going to approve it.

PR #171, opened 2022-04-21
Include resource requirements and remove local running instructions

### LINKED ISSUE #153: Update documentation with resource requirements

allyhawkins:
Closes #153. This PR removes any remaining areas of both the main readme and then the external instructions that mentioned running the workflow locally. I also added in statements throughout that mentioned the resource requirements to hopefully drive the point home. 

On the main readme page, I updated the instructions for Data Lab staff to only include running with batch, removing the command that runs locally. 

For the external instructions document, I removed any comments that referred to running the workflow locally and added in the resource requirements in a few places. When referring to setting up parameters I also removed the command that showed setting them at the command line as that didn't have a config file or profile there so if people copy and paste that it would run locally. Instead I included the flag in a sentence of how you could invoke them at the command line but that we encourage people to set them in the config file and linking to the below section. 

In the config file section, I added in a note about how the example command doesn't invoke a profile name and that if using that command as is, it would attempt to run locally which is not advised. Do we think that note is sufficient or do we want to maybe remove that command alltogether (or modify to include a profile)? I was trying to think of the least confusing way to say don't just copy this but this is how to use a config file. 

The other thing I did while going through these instructions was remove the number of times we have the version tag and replaced it with `-r <tagged version number>` throughout. I kept it in the first time we refer to it and then explain what the `-r` means but removed it for the remainder of the command examples to minimize the amount of times we need to change it for a release.

### DIFF
--- a/external-data-instructions.md
+++ b/external-data-instructions.md
@@ -148,9 +148,10 @@ nextflow run AlexsLemonade/scpca-nf \
 
 For reference, we provide an example template configuration file, [`user_template.config`](examples/user_template.config), which includes some other workflow parameters that may be useful, as well as an example of configuring a profile for executing the workflow on a cluster, discussed below.
 
-**Note:** This example tells Nextflow to use the configuration set up in the configuration file, but it does not invoke a specific profile, which is necessary to tell Nextflow where to execute the workflow.
-Without a profile name, Nextflow will attempt to run the workflow locally which will not work without access to the minimum computing requirements, 24 GB of RAM and 12 CPUs.
-For more on [creating and using a profile](#setting-up-a-profile-in-the-configuration-file) see the below section. 
+**Note:** This example tells Nextflow to use the configuration set up in the configuration file, but it does not invoke a specific profile, and will use the `standard` profile.
+Under the `standard` profile, Nextflow will attempt to run the workflow locally using Docker. 
+This will most likely result in an error unless the minimum computing requirements (24 GB of RAM and 12 CPUs) are met on the local machine.
+For more on [creating and using a profile](#setting-up-a-profile-in-the-configuration-file) see the section below. 
 
 See the [Nextflow documentation](https://www.nextflow.io/docs/latest/config.html) and the below sections for more detail on creating your own configuration file.
 
--- a/external-data-instructions.md
+++ b/external-data-instructions.md
@@ -47,7 +47,7 @@ Once you have set up your environment and created these files you will be able t
 
 ```bash
 nextflow run AlexsLemonade/scpca-nf \
-  -r v0.2.7
+  -r v0.2.7 \
   -config <path to config file>  \
   -profile <name of profile>
 ```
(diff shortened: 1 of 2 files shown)


allyhawkins:
> I also wonder if rather than replacing the coded versions with <tagged version number>, we might just delete those lines? Now that the latest version is in main, I think we can probably remove this as a "standard" instruction, and rather let people know that this is what is going on. We might add a note about versions, with something like "If you want to be sure you are using a consistent version, you can use the -r option and specify a version tag, such as -r v0.2.7." I don't know if we want to add instructions for nextflow run -latest or nextflow pull to get the latest version as well, but maybe?

Hmm... Have you tested this recently? I wanted to play around with this before I edited the instructions and it doesn't seem to work. I would have also thought that we should be able to run the workflow without specifying a revision, but I'm consistently getting the error, `Project `AlexsLemonade/scpca-nf` currently is sticked on revision: v0.2.7 -- you need to specify explicitly a revision with the option `-r` to use it`. 
I tried to use `nextflow run -latest` and also do `nextflow pull` before running the workflow and the pull seemed to work successfully... Just wondering if you have come across this @jashapiro. 
Screen Shot 2022-04-25 at 9 00 48 AM

jashapiro:
> Hmm... Have you tested this recently? I wanted to play around with this before I edited the instructions and it doesn't seem to work. I would have also thought that we should be able to run the workflow without specifying a revision, but I'm consistently getting the error, `Project `AlexsLemonade/scpca-nf`currently is sticked on revision: v0.2.7 -- you need to specify explicitly a revision with the option`-r` to use it`. I tried to use `nextflow run -latest` and also do `nextflow pull` before running the workflow and the pull seemed to work successfully... Just wondering if you have come across this @jashapiro. Screen Shot 2022-04-25 at 9 00 48 AM

You may need to run `nextflow pull AlexsLemonade/scpca-nf -r main` one time to make sure it has checked out the `main` branch if it never has before.  Alternatively, start fresh with `nextflow drop AlexsLemonade/scpca-nf`. Then `nextflow run AlexsLemonade/scpca-nf` with no revision specified should work. If somebody never specifies a revision, it should use `main` by default and stay up to date `pull` or a `-latest` flag. (Nextflow needs to fix this `sticked` error, which is super unhelpful, and should be easy to fix.)

allyhawkins:
I have gone ahead and updated the instructions to remove the requirements of using a release tag, but added that as an option. I updated the note that we had on release tags to reflect that running without a release should use the latest version while running with the release can be helpful for ensuring the version you use to process is consistent and then provided an example. 
I did not insert explicit instructions on using `nextflow -r latest` or `nextflow pull` as when I was writing those the instructions seemed to get confusing and lengthy, keeping in mind this is currently in the overview section. @jashapiro if you have other ideas on where to incorporate this I'm open to those, but also think it might be okay to keep them out for now.

external-data-instructions.md
```diff
@@ -56,9 +55,17 @@ Where `<path to config file>` is the **relative** path to the [configuration fil
 This command will pull the `scpca-nf` workflow directly from Github, using the `v0.2.7` version, and run it based on the settings in the configuration file that you have defined.  
 
 **Note:** `scpca-nf` is under active development.
-We strongly encourage you to use a release tagged version of the workflow, set here with the `-r` flag.
+Using the above command will run the latest version of the workflow.
```
jashapiro, line 58: Since it might not always be the latest version, I changed the wording a bit, but then added a sentence with how to update to the latest version. 
 
```suggestion
Using the above command will run the workflow from the `main` branch of the workflow repository. 
To update to the latest released version you can run `nextflow pull AlexsLemonade/scpca-nf` before the `nextflow run` command. 
 
```

jashapiro (approved):
This looks good; thank you for incorporating those changes. 

I added one more sentence about `nextflow pull` because I wanted to note how you would update to the latest version if that is desired, in addition to running on a specific version. I don't think it is too confusing, but let me know your thoughts.  I could imagine somebody thinking that they were always on the latest version based on the wording that was there, but it would really only be the latest version from the _first_ time they ran the workflow, as it doesn't pull new versions unless asked (I think?).

Issue #173, opened 2022-04-28
QC report additions for multiplexed samples


jashapiro:
**Is your feature request related to a problem? Please describe.**

The current QC report doesn't include any analysis for multiplexed libraries. We should probably have _something_ there

**Describe the solution you'd like**

For multiplexed libraries, we should add a section to the QC report, perhaps with some summary of the cell calls from cellhash demultiplexing and genetic demultiplexing.

It would probably make sense to also have a section at the top that warns when a library is multiplexed.

Similarly to the cite-seq section, this content should probably mostly live in a separate document that is included dynamically when multiplex results are present.

PR #179, opened 2022-05-04
Add metadata for multiplex counts to json

### LINKED ISSUE #175: Add demux estimates to metadata
The metadata file that we currently output has cell count estimates only for the library as a whole. For multiplexed data, it would be helpful to include estimates of the number of cells per sample. 

While we don't really trust these values, they are a starting point that somebody evaluating a library might want to look at. For example, if there are only a few cells at best from the sample of interest, it might not be worth using the library in an analysis of that sample.

We can add extra fields in the `_metadata.json` file generated by `sce_qc_report.R` that report the counts for each sample. WE will add `sample_estimated_cells`, which will contain a dictionary of the following form:
```
{
  "sample_01": 100,
  "sample_02": 200,
  "sample_03": 150
}
```

The contents can be based on the demultiplexing method with the largest number of calls, or we could always use genetic demultiplexing if available? (I'm leaning toward the latter, after a bit of thought).

In addition, we will add `is_multiplexed` (true or false) for all cells and,`demux_method` for multiplexed libraries.

jashapiro:
Closes #175

This PR is primarily about adding a sampling of the demultiplexing results to the `_metadata.json` file. I decided to make teh demultiplexing method an option, rather than having it vary by sample. Currently it uses `vireo` by default, and I don't specify the option in the workflow itself, but it is present in the script. I could add it there if we think it would clarify anything, or as a pointer for sometime in the future when we might reconsider.

### DIFF
--- a/bin/add_demux_sce.R
+++ b/bin/add_demux_sce.R
@@ -91,7 +91,6 @@ if(is.null(opt$cellhash_pool_file)){
 
 # read in sce rds file
 sce <- readRDS(opt$sce_file)
-metadata(sce)
 
 # check for cellhash altExp if we will use it
 if( opt$hash_demux || opt$seurat_demux ){
--- a/bin/add_demux_sce.R
+++ b/bin/add_demux_sce.R
@@ -1,5 +1,10 @@
 #!/usr/bin/env Rscript
 
+# Script for adding demultiplexing information to a SingleCellExperiment object
+# If provided, it will add genetic demultiplexing information from vireo,
+# and/or demultiplexing based on cellhash data using Seurat::HTOdemux
+# and/or DropletUtils::HashedDrops
+
 # import libraries
 suppressPackageStartupMessages({
   library(optparse)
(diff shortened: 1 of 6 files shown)


bin/add_demux_sce.R
```diff
@@ -86,6 +91,7 @@ if(is.null(opt$cellhash_pool_file)){
 
 # read in sce rds file
 sce <- readRDS(opt$sce_file)
+metadata(sce)
```
allyhawkins, line 94: Is this supposed to be here?

allyhawkins (approved):
This looks pretty good to me. I just had one minor comment. 

My other comment is about how we want to control which demultiplexing method is used for a library. I know right now with just the one project setting the default of vireo makes sense, but I wonder if for future proofing purposes we want to include that in the nextflow workflow as an argument and set the default to vireo there. If we ever do get another project that doesn't have bulk RNA-seq we can add a column to the metadata file that indicates the method to use and then pass that through, but at least if we add an argument now its indicated in the nextflow workflow that the choice of genetic demultiplexing to use for the metadata is happening during that step.

bin/add_demux_sce.R
jashapiro: Nope!

jashapiro:
After some consideration, I think I am going to leave this as is, without modifying the workflow itself. If we get to the point where we need to support libraries without information for genetic demultiplexing, then we will have a fair amount of work to do anyway, and I don't want to make assumptions about what that will look like at this stage. I might reconsider when looking at #173, but I think I will merge this as is for now.

PR #182, opened 2022-05-09
Try handleing miQC errors

jashapiro:
This PR is designed as a temporary fix for miQC errors that we have run into. If add_miQC fails, it attempts to reapply the function 2 more times to handle stochastic errors. 

In the event of failure after 3 tries, it fills in prob_compromised with NA values, as we would have done if a model fit failure had been handled internal to the function.

related to https://github.com/AlexsLemonade/scpcaTools/issues/101, and the final fix may be happening in https://github.com/AlexsLemonade/scpcaTools/pull/102, pending testing


allyhawkins (approved):
Thanks for making these changes! I tested this with one of the libraries was failing and all looked good.

PR #184, opened 2022-05-11
Fix demux test check

jashapiro:
I somehow missed this when I was testing the last PR; I must have added this error message at the last minute and missed testing it somehow.

fail if *not* in the list, not in the list.


allyhawkins (approved):
This might explain why my tests are failing right now, thanks for catching this!

PR #187, opened 2022-05-11
Remove miQC try/while loop and update scpcaTools version 

allyhawkins:
Since the new changes that have been incorporated into `scpcaTools` fix the errors seen in miQC we no longer need to have the try/while loop around adding the miQC results to the filtered sce object here. I've removed it and restored the line back to what it was previously, which was just calling `scpcaTools::add_miQC()`. I tested that the changes in `scpcaTools` were sufficient to catch the miQC error and fix it by using the `edge` tag for `scpcaTools` first and testing the workflow on some of the libraries that were previously failing. I confirmed that they do indeed pass now and complete the workflow. I also updated the tag and am using the tag for what will be the next release of `scpcaTools`.

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -61,12 +61,6 @@ filtered_sce <- filtered_sce |>
 # add prob_compromised to colData and miQC model to metadata
 filtered_sce <- scpcaTools::add_miQC(filtered_sce)
 
-# set prob_compromised to NA if miQC failed
-if (is.null(miqc_sce)){
-  filtered_sce$prob_compromised <- NA_real_
-}
-
-
 # grab names of altExp, if any
 alt_names <- altExpNames(filtered_sce)
 
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -59,7 +59,18 @@ filtered_sce <- filtered_sce |>
   scuttle::addPerFeatureQCMetrics()
 
 # add prob_compromised to colData and miQC model to metadata
-filtered_sce <- scpcaTools::add_miQC(filtered_sce)
+# since this can fail, we will check for success
+miQC_worked <- FALSE
+try({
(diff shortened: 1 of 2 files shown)


bin/filter_sce_rds.R
```diff
@@ -59,21 +59,7 @@ filtered_sce <- filtered_sce |>
...
-    miqc_sce <- scpcaTools::add_miQC(filtered_sce)
-    filtered_sce <- miqc_sce
-    })
-}
-# set prob_compromised to NA if miQC failed
-if (is.null(miqc_sce)){
-  filtered_sce$prob_compromised <- NA_real_
-}
-
+filtered_sce <- scpcaTools::add_miQC(filtered_sce)
```
jashapiro, line 62: Just for safety, I wonder if we shouldn't keep this wrapped in a trycatch block

```suggestion

tryCatch(
  filtered_sce <- scpcaTools::add_miQC(filtered_sce),
  error = function(e) {
    warning("miQC failed. Setting `prob_compromised` to NA.")
    filtered_sce$prob_compromised <<- NA_real_
  }
)

```

There is a slightly odd thing here where I am using `<<-` so that I can assign to a value _outside_ the error function call. 

Another option might be something like:
```suggestion
myQC_pass = FALSE
try({
  filtered_sce <- scpcaTools::add_miQC(filtered_sce)
  myQC_pass = TRUE
 })
 
# set prob_compromised to NA if miQC failed
if (!miQC_pass){
  filtered_sce$prob_compromised <- NA_real_
}
```
Which avoids the strange `<<-` and might be a bit easier to read.

allyhawkins:
I went ahead and incorporated the second suggestion you had since I thought that was a little more straight forward personally, and then also tested it on the library that had previously failed and all worked as expected. So this is ready for another look.

jashapiro (approved):
LGTM!

PR #188, opened 2022-05-24
update param for merge_bulk_quants

jashapiro:
In testing autoscaling, we ran into a warning about the publishDir, which resulted from an error in the param setting for the `merge_bulk_quants` process. This should the one-line fix. Not sure why we didn't see this earlier?


allyhawkins (approved):
Thanks for catching this! 🚀

PR #190, opened 2022-05-25
Add autoscaling ebs as a new profile

jashapiro:
Rather than replacing the original batch configuration, I thought we might want to keep a separate config for some more extensive testing, while preserving the old settings. 

This profile could be invoked with `nextflow run main.nf -profile autobatch,ccdl` etc.

Replaces #189


nextflow.config
```diff
@@ -59,6 +59,10 @@ profiles{
   batch {
     includeConfig 'config/profile_awsbatch.config'
   }
+  // AWS batch profile with autoscaling disk
+  autobatch {
+    includeConfig 'config/profile_awsbatch-auto.config'
```
arkid15r, line 64: I'd use "profile_awsbatch_auto.config" for consistency

arkid15r (approved):
Otherwise looks good!

allyhawkins (approved):
This looks good to me. Thanks to both of you guys for working on this and making this change!

Issue #174, opened 2022-04-28
Alevin-fry reporting 0 cells detected for some libraries


allyhawkins:
**Describe the bug**
In running the samples for the Mullighan and Teachey/Tan projects, many of the libraries are failing downstream steps after completing Alevin-fry, such as the steps that involve merging with CITE-seq data or filtering the object. This is because there are no cells being identified by Alevin-fry for these libraries. 

For the Teachey/tan project this was the error being reported for a handful of samples during the `make_merged_unfiltered_sce` object. I then went back to the work directory of the previous steps which was quantifying the CITE and RNA libraries with alevin-fry and saw that nothing appeared off with the CITE-seq data, however in the RNA output there was no cell barcodes identified and an empty matrix with just gene names was being produced. 
<img width="990" alt="Screen Shot 2022-04-28 at 8 57 15 AM" src="https://user-images.githubusercontent.com/54039191/165769003-fdd16a95-5d97-48c8-8c09-004e0222fe7e.png">

**To reproduce**

This occurred with multiple samples, but I particularly investigated SCPCR000692, SCPCR000064, SCPCR000704, SCPCR000602, and SCPCR000163 noting that they were all failing at the step after alevin-fry quantification because of 0 cells. I also ran SCPCR000692 through cellranger and noticed that there were no issues with the sequencing library and it successfully identified 16,000 cells. 

What command did you use to invoke the `scpca-nf` workflow?

```
nextflow run AlexsLemonade/scpca-nf \
		 -r v0.2.7 \
		 -profile batch,ccdl \
		 --project teachey_tan \
		 --outdir 's3://nextflow-ccdl-results/scpca-prod' \
		 -with-tower
```

**Expected behavior**
We would typically expect Alevin-fry to quantify the number of cells identified if the sequencing was okay, which in this case it seems like the sequencing doesn't seem to be the problem. 

**Version information**
What version of `scpca-nf` are you using, and what version of Nextflow?
v0.2.7 and Nextflow version 21.04.1

**Additional context**
I also had tried running alevin-fry with different technology thinking that may be the issue, but using 10Xv2 and/or 5 prime still resulted in 0 cells identified. In running cellranger, I confirmed that the technology version was correctly reported in the metadata.

rob-p:
Hi @allyhawkins,

[...]

PR #191, opened 2022-05-25
Use dynamic disk label for alevin-fry and downstream processes

allyhawkins:
Related to #174. After identifying that the reason Alevin-fry was identifying 0 cells, we narrowed it down to writing out incomplete RAD files which is due to not enough disk space. Because of that, here I increased the disk used for the `salmon alevin` process to `disk_big`. Additionally, there were more errors that were occurring in downstream processes (all the way through to generating the QC report) where there was not enough space left on the disk. To get around this, I added the `dynamic_disk` label to all downstream processes starting with `alevin-fry`. I tested this with a large project and everything seemed to work and the job finally completed when using `dynamic_disk`.

### DIFF
--- a/config/profile_awsbatch.config
+++ b/config/profile_awsbatch.config
@@ -8,12 +8,9 @@ aws{
 process{
   executor = 'awsbatch'
   // default queue has 128GB local disks
-  queue = 'nextflow-batch-default-queue' 
+  queue = { task.attempt < 2 ? 'nextflow-batch-default-queue' : 'nextflow-batch-bigdisk-queue'}
   withLabel: disk_big {
     // bigdisk queue has 1000GB local disks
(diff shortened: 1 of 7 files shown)


jashapiro:
This is fine, though I think the approach I would have taken would have been to just change the default queue in the code below to: 
```
queue = { task.attempt < 2 ? 'nextflow-batch-default-queue' : 'nextflow-batch-bigdisk-queue'} 
```
https://github.com/AlexsLemonade/scpca-nf/blob/16e5fa55c78f70bc6a5054fd564a9f20c571cf08/config/profile_awsbatch.config#L9-L12

i.e. just use dynamic everywhere without requiring a label, rather than labelling everything.

I really hope this is all moot with the autoscaling though!

PR #196, opened 2022-06-03
Change generate_qc_report argument name and update version numbers

### LINKED ISSUE #194: Change argument name in `generate_qc_report` to `library_id`
After the changes were made in `scpcaTools` to the `generate_qc_report` function to have `library_id` as the argument name rather than `sample_name`, that change needs to be implemented here as well when calling that function. 

Note that I tested this before creating the new `scpcaTools` release but forgot to file the issue sooner, so just wanted to make sure I did file it for tracking purposes.

allyhawkins:
Closes #194. 

Here I changed the name for the argument used when calling `scpcaTools::generate_qc_report()` to reflect the new argument, `library_id`. I tested this and all looked good and produced a QC report as expected for a multiplexed library. 

I also am changing the version of `scpcaTools` that we are using to v0.1.7 and then updating the version for `scpca-nf` to v0.3.2 to prepare for the new release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ nextflow run AlexsLemonade/scpca-nf -profile ccdl,batch
 When running the workflow for a project or group of samples that is ready to be released on ScPCA portal, please use the tag for the latest release: 
 
 ```
-nextflow run AlexsLemonade/scpca-nf -r v0.3.1 -profile ccdl,batch --project SCPCP000000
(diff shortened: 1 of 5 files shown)


jashapiro (approved):
Trusting your testing (which I assume was done with `:edge`), but noting that this should not go in until the scpcatools 0.1.7 docker image final finishes building: https://github.com/AlexsLemonade/scpcaTools/actions/runs/2436196085

PR #198, opened 2022-06-07
Skip demultiplexing if only one cellhash ID 

allyhawkins:
When running the library that had only one HTO, but was classified as demultiplexing, Seurat was failing to perform demultiplexing. To avoid this, we discussed skipping demultiplexing altogether if there is only one HTO and not reporting sample calls. 

Here I check for the number of cellhash ID's and if that's > 1, then we continue with add the demultiplexing results. We are also making changes in scpcaTools to the QC report to adjust for samples like this (AlexsLemonade/scpcaTools#111). I will wait until that gets merged and I can then test those changes with`scpcaTools::edge` and these changes before wrapping up this PR, so I'm going to keep it in a draft state for now. I did test these changes and things worked successfully, but I would like to test in combination with the changes to `scpcaTools`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ nextflow run AlexsLemonade/scpca-nf -profile ccdl,batch
 When running the workflow for a project or group of samples that is ready to be released on ScPCA portal, please use the tag for the latest release: 
 
 ```
-nextflow run AlexsLemonade/scpca-nf -r v0.3.2 -profile ccdl,batch --project SCPCP000000
+nextflow run AlexsLemonade/scpca-nf -r v0.3.3 -profile ccdl,batch --project SCPCP000000
 ```
 
 ### Processing example data 
--- a/bin/add_demux_sce.R
+++ b/bin/add_demux_sce.R
@@ -103,22 +103,25 @@ if( opt$hash_demux || opt$seurat_demux ){
   }
 }
 
+cellhash_ids <- rownames(altExp(sce))
 
-
-# add HashedDrops results
-if(opt$hash_demux){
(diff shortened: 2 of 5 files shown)


allyhawkins:
I have gone ahead and tested this using the most recent changes in `scpcaTools` and the QC report looks as expected and the library in question now completes the workflow successfully so it is ready for review. I did update the container to use `scpcaTools::edge` here and am wondering if we want to do a release of `scpcaTools` first and up date the tag for the container in this PR? Or if we want to incorporate these changes and then do the release and separate PR to update the container tag.

jashapiro (commented):
LGTM!

>I did update the container to use scpcaTools::edge here and am wondering if we want to do a release of scpcaTools first and up date the tag for the container in this PR? Or if we want to incorporate these changes and then do the release and separate PR to update the container tag.

I don't have a strong opinion on this right now. If we are doing a release of scpca-nf, we definitely want to have that tied to a specific version of scpcaTools, but if we think there might be more changes before the next release, I would be okay with leaving this pointing at edge for a bit longer.

allyhawkins:
> I don't have a strong opinion on this right now. If we are doing a release of scpca-nf, we definitely want to have that tied to a specific version of scpcaTools, but if we think there might be more changes before the next release, I would be okay with leaving this pointing at edge for a bit longer.

As we discussed in slack, going to go ahead and do a release with these changes so I updated this PR to have the new `scpcaTools` tag for the container and the new version for `scpca-nf`. I will wait to merge this once the docker build is finished, but it is ready for another look @jashapiro.

PR #201, opened 2022-07-27
Jashapiro default multiplex

jashapiro:
This adds a dummy file name to hopefully fix a bug when a real multiplex pool file is not specified.


allyhawkins (approved):
Looks good!

PR #202, opened 2022-07-31
Make bin scripts executable

jashapiro:
Based on a bug report from a tester, we may need to make the R scripts in `bin/` executable for some environments (seems not to matter with AWS?), so this PR does that. (Not sure how Github will show the changes, since there are none in the scripts themselves, just permissions.)


allyhawkins (approved):
I checked out the branch and checked that these are now executable so 👍

Issue #207, opened 2022-08-29
Add process to perform normalization/dimensionality reduction to workflow


allyhawkins:
After addressing #203, we will want to add a process to the nextflow workflow to run the script that normalizes and performs post-processing of single-cell data. We will probably want to add this process to the sce processing module. This will need to happen before creating the process that creates the QC report but after filtering because we will want to include a UMAP with clusters in the QC report eventually.

Issue #203, opened 2022-08-01
Add normalization and dimension reduction to workflow


jashapiro:
We should add a new step to the workflow that takes the filtered SCE object and applies normalization and dimension reduction (UMAP), storing those results in the SCE that is finally published.

This step should immediately precede the QC report.

We should follow the same code steps that we use in https://github.com/AlexsLemonade/scpca-downstream-analyses, but I expect we will want to combine any additional filtering, normalization, and the UMAP calcualtion into a single script so it can run all together.

One question is whether we want to include filtering as specified by miQC, and any additional criteria.  I had the thought that we could do this in a "soft" filter. In other words, we would do so before calculating UMAP, and then merge the calculated UMAP matrix in with the full data set, setting values to NA for cells that were filtered out.  This allows us to keep the QC template as is, but still include UMAP plots for only those cells that pass additional filtering (avoiding a cluster of high-mito cells, for example)

jashapiro:
closed by #207

PR #210, opened 2022-08-31
Script for filtering, normalization, and dim reduction of SCE objects

allyhawkins:
Related to #203, but not quite sure that it closes it. 

This starts the addition of adding in post processing of SCE objects to the workflow by adding a script that filters SCE objects, performs normalization, and then adds PCA and UMAP embeddings. The input to the script should be the output of the `filter_sce_rds.R` script and include an SCE object with empty droplets removed. 

There were a few decisions that I made here that will definitely need some inputs: 

- I did not include any additional options other than to filter using miQC. We add a column to the colData of the SCE objects that contains the probability of cells being compromised as calculated by miQC in the `filter_sce_rds.R` script, so I use that column directly to perform filtering. If miQC fails, that column will contain `NA`, so I check for that and if that's the case no filtering is performed. Do we want to perform a different type of filtering instead and use manual cutoffs in that scenario? 
- Prior to PCA calculation we typically calculate highly variable genes and then use those as inputs. I did include two options, the number of genes to use for HVG selection and then a `--use_hvg` option, indicating whether or not to use HVG as input to the PCA calculation. I don't think we will be using this without inputting HVG, but users may one day appreciate the flexibility, so I thought we might want to include that. Should I change the default so that it always uses HVG, but to use all genes you have to specify an option? 
- Right now I am writing out the filtered, normalized SCE object with the PCA and UMAP embeddings to a new file rather than adding UMAP embeddings to the input file. This was a suggestion in the original issue #203, but I think we also want to have an additional file that contains the processed SCE object that is separate from the `filtered.rds` we provide now (resulting in 3 rds files per library). I can also add the UMAP embeddings to the input SCE file so that we can continue to read that into the qc report generation. 

I am adding this script first and then was planning on tackling #207, the addition of the process to run this script as a separate PR. 

The last thing I wanted to note is that I haven't yet considered accounting for CITE-seq samples in an `altExp`. I think we do want to think about how we want to deal with those since that data will not be normalized. That is definitely outside the scope of this PR, but something we might want to think of exploring in downstream analyses soon...


allyhawkins:
> I think we do want to include a cutoff for a minimum number of reads and/or genes to use in addition to miQC, or as a floor in the case of miQC failure. Even if miQC does not remove a cell, a cell with only 100 or 1000 genes is probably not going to be useful, so removing those is probably a good idea. I would make this threshold a parameter, but we should pick an arbitrary threshold.

I went ahead and added a minimum gene cutoff argument and set it to 200... We could probably comfortably move that up to 500 but the cutoffs really do depend on sequencing depths so I don't want to go too high in hopes that miQC will handle most of the optimal filtering. The minimum cutoff should really be used to get any additional stragglers that shouldn't be there and would impact normalization and dimensionality reduction down the line so I think we want to be fairly lenient but still get the obvious low quality cells out of there. 

> I'm not totally clear here on what you mean here (having not looked at the code yet), but if you are meaning whether to model the biological variance or not, I would say we can just make a decision about whether to do this or not (I say yes). You can also pick HVG from raw variance, which is what the PCA function does, but I think we just want to say that is not what we do for this pipeline, just to keep it simpler. The number of HVGs we use can be a parameter though.

Just wanted to note that I did keep the modeling of gene variance and the parameter to specify number of HVGs to identify. 

> This seems right to me (a separate file). The other thing we could do is to add a column with the filter results to the original file. I am not sure whether we can add the UMAP to the original file, as it will have a different number of cells?

I went ahead and added a `ccdl_filter` column to the original input SCE that indicates whether or not cells should be removed. I also added a note to the metadata on how that column was produced, either through miQC or just through the minimum gene cutoff if miQC failed. I then write out the input SCE file back to that same input file. Let me know if you had a different idea in mind on how to handle this. 

As for the UMAP of the filtered data, I did a test to add the filtered UMAP to the original input SCE and you are correct that it doesn't work because of mismatched dimensions. For now I didn't add in the UMAP as we talked about keeping that out during science pair it and modifying the report instead, but let me know if you've changed your mind and I can play around with adding that in.

jashapiro (commented):
This looks good overall (now that I have had a chance to actually look through the code too!)

I made a few relatively small suggestions (too many about naming), and had a question about the tryCatch block (when is it needed?)

Otherwise, I think this looks close, with the caveat that I have not tested it. 

Looking ahead, I expect we will want to put whatever basic clustering we do into this same script/process: the headroom of writing and reading the objects seems not to be justified when we could just run another few steps in the same R session. 

For now, I think it should be fine to keep the PCA and UMAP results separate. I can imagine a way that we could expand the matrix with the removed cells as NA values... It might be worth exploring so we still only have a single file to return and have to make fewer changes to the portal, but I would see that as a separate issue that we can return to.

allyhawkins:
I went ahead and incorporated the smaller name changes that you suggested and added the argument for `n_pcs`. 

As for the `tryCatch` block around the `qclust`, this is specifically to deal with scenarios we found when developing downstream analyses where there were too few cells to implement `quickCluster` so we were seeing failures as noted in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/29#issuecomment-977227329. Our approach to that was to check if `qclust` was completed successfully (which we originally added in AlexsLemonade/scpca-downstream-analyses#32), but I'm open to alternative solutions to that.

bin/post-process-sce.R
```diff
@@ -0,0 +1,150 @@
...
+  metadata(filtered_sce)$normalization <- "deconvolution"
+})
+
+if (is.null(qclust)) {
+  
+  # Include note in metadata re: failed clustering
+  metadata(filtered_sce)$normalization <- "log-normalization"
+  
+  # Keep positive counts for `logNormCounts()`
+  filtered_sce <- filtered_sce[, colSums(counts(filtered_sce)) > 0]
```
jashapiro, line 119: Now that I understand the reason for the `tryCatch` to be a small number of cells and not remaining cells with zero counts, we should be able to eliminate these lines. I do not think there can be any cells we are keeping that have zero counts.

```suggestion
```

jashapiro (commented):
This looks good to me, but I think we should hold off on merging until we have completed the bugfix release.

bin/post-process-sce.R
```diff
@@ -0,0 +1,150 @@
...
+    sce$prob_compromised < opt$prob_compromised_cutoff & 
+      sce$detected >= opt$gene_cutoff, 
+    "Keep", 
+    "Remove"
+  )
+  metadata(sce)$ccdl_filter_method <- "miQC"
+}
+
+# filter sce using criteria in ccdl_filter 
+filtered_sce <- sce[, sce$ccdl_filter == "Keep"]
```
jashapiro, line 98: Adding a `which` here for the `NA` case, which I think may come up when `prob_compromised` is `NA`
```suggestion
filtered_sce <- sce[, which(sce$ccdl_filter == "Keep")]
```

bin/post-process-sce.R
```diff
@@ -0,0 +1,150 @@
...
+  
+  # Include note in metadata re: clustering before computing sum factors
+  metadata(filtered_sce)$normalization <- "deconvolution"
+})
+
+if (is.null(qclust)) {
+  
+  # Include note in metadata re: failed clustering
+  metadata(filtered_sce)$normalization <- "log-normalization"
+
```
jashapiro, line 117: Just removing some excess whitespace.
```suggestion
  # Include note in metadata re: failed clustering
  metadata(filtered_sce)$normalization <- "log-normalization"
```

bin/post-process-sce.R
jashapiro: I just realized that after this step we need to recalculate `perFeatureQCMetrics` on the filtered_sce, as the detected counts and means for genes will have changed after removing cells.

bin/post-process-sce.R
```diff
@@ -0,0 +1,150 @@
...
+  # Include note in metadata re: failed clustering
+  metadata(filtered_sce)$normalization <- "log-normalization"
+  
+}
+
+# Normalize and log transform
+normalized_sce <- scater::logNormCounts(filtered_sce)
+
+# remove filtered SCE to save space 
+rm(filtered_sce)
```
jashapiro, line 124: for simplicity, we could just write the logNormCounts back to `filtered_sce`.

jashapiro (commented):
I thought I was going to approve this, but then I realized that we need to update stats after filtering, so I would like to see that implemented first. I also suggested a `which` to avoid `NA` issues that I _think_ will come up.

allyhawkins:
I added the recalculation of the feature stats and then also saved everything to the `filtered_sce` instead of creating a separate `normalized_sce`. This is ready for another look.

allyhawkins:
Sorry about that... I changed the name to match the other files in the `bin` directory. I removed the duplicated script so it should be good now.

jashapiro (approved):
LGTM!

PR #220, opened 2022-09-29
Track metadata in checkpoints

### LINKED ISSUE #149: Make step-skipping metadata robust to index changes
There are a couple of reasons I don't really like this solution, but it might be worth doing as a temporary measure.

The second thing is that what we have done here is to separate the actual index use from where we record what file was used, which is not great (they could get out of sync if we created a new index). I don't think this is the only place we do this, and it was something I was thinking about fixing more generally with the kind of approach I took with genetic demultiplexing: for steps that we might skip, we can write out the metadata as a json file, and then read it back in before the next step.  Example code: 

https://github.com/AlexsLemonade/scpca-nf/blob/1bb59c6ca848341f7bc4f47a45a4323db62b851d/modules/cellsnp.nf#L45-L55
https://github.com/AlexsLemonade/scpca-nf/blob/1bb59c6ca848341f7bc4f47a45a4323db62b851d/modules/genetic-demux.nf#L58-L61

I _do not_ think this is the time/PR to fix that issue, but it was in my mind, so I wanted to get it written down somewhere (I will make this into an issue separately).

Finally, a note on process for merging this to `development`. The easiest thing _should_ be to to merge this into `main`, but rather than deleting the branch when that is done, you can then merge `development` into this branch and file a PR to `development` directly from the same branch.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/148#pullrequestreview-933752689_

### LINKED ISSUE #218: Update salmon and alevin versions
In concert with #214 and #149, we should probably update Salmon to version 1.9.0. Looking at the [release notes](https://github.com/COMBINE-lab/salmon/releases/tag/v1.9.0), it looks like the primary update for our purposes is better error handling when running out of space (hopefully not an issue as much with the auto-expanding volumes), and a smaller Docker image.

Updating alevin-fry to 0.7.0 seems likely to be mostly address errors we are unlikely to encounter ([release notes](https://github.com/COMBINE-lab/alevin-fry/releases)), but upgrading in parallel seems prudent.

jashapiro:
The primary change here is to track metadata by writing a json file with the contents of the `meta` object when we write files to a checkpoint. 

I did this primarily by implementing the same system as in the `genetic-demux` workflow, but I moved some of the logic out to a separate file, `lib/Utils.groovy` which is automatically loaded by Nextflow (I can't find the documentation for this behavior, but  it seems to load any groovy files in the lib directory, and `nf-core` relies on it, so I don't expect it to change. The closest to documentation I could find was that there is a `-lib` argument to `nextflow`, which I guess might allow a different library to be loaded (as well? instead?))

A limitation of this approach is that it seems like some of the Nextflow functions and objects are not available when the `Utils` class is compiled, so I couldn't use the `file()` function within class. Instead, we have to use that when calling the function, which is ugly, but it works.

To simplify some tracking, I also wrote the paths of the references and other information to `meta` in the initial setup. Most of this is unused, but I did switch where we used `params.assembly` in the QC report to use the value from the `meta` object, which will now track what was used at the mapping step rather than at any later run. Similarly, we now track the transcript2gene file so that we know it will match the rad file that was used. Workflow version, however, is still set by the version active when compiling the QC report.

I considered outputting the metadata in the final publish directories too, but decided that was not worth doing, partly because it includes a bunch of file paths that may or may not be relevant.

### Bulk mapping

Bulk mapping is treated a bit differently. Since we are combining all of the bulk samples into a single matrix for output, I decided to check that all mappings used the same assembly, and that was equal to the version specified in the `params`.  So now if the assembly version in the stored metadata does not match `params.assembly`, mapping is repeated. I did this because it is easier to check against the params than to check all against one another, but I am not sure this is a good long-term solution, as we probably want to keep old mappings if we can, but I haven't come up with a good solution as yet.

### Additional changes
- Since this change is a breaking change in that we now expect a `scpca-meta.json` file in any checkpoint output, I also changed the output directory paths in the `ccdl` profile to match the defaults of `checkpoints` and `results`. I thought I would be able to just delete those redefinitions, but the order of how `params` are resolved means that these have to be repeated.

- Updated salmon and alevin versions. Again, as long as I am making breaking changes that require us to repeat mappings, this seemed a good time.

- Another change is using `join` in place of `combine` in a few places when merging channels. This should not have any functional change, but allowed me distinguish between cases where we are doing one-to-one joining, as in the case of joining together two components of a library (rna & features) vs one-to-many, as we do when joining a feature reference to the fastq files that use it. In the former case, we can add some extra checks to be sure the files are as expected.

Closes #149 
Closes #218

### DIFF
--- a/modules/qc-report.nf
+++ b/modules/qc-report.nf
@@ -5,7 +5,7 @@ process sce_qc_report{
     container params.SCPCATOOLS_CONTAINER
     tag "${meta.library_id}"
     publishDir "${params.results_dir}/${meta.project_id}/${meta.sample_id}"
-    input: 
+    input:
         tuple val(meta), path(unfiltered_rds), path(filtered_rds)
     output:
         tuple val(meta), path(qc_report), path(metadata_json)
@@ -24,7 +24,7 @@ process sce_qc_report{
           --metadata_json ${metadata_json} \
           --technology "${meta.technology}" \
           --seq_unit "${meta.seq_unit}" \
-          --genome_assembly "${params.assembly}" \
+          --genome_assembly "${meta.ref_assembly}" \
           --workflow_url "${workflow_url}" \
           --workflow_version "${workflow_version}" \
           --workflow_commit "${workflow.commitId}"
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -1,6 +1,6 @@
 #!/usr/bin/env Rscript
 
-# This script takes a SingleCellExperiment stored in a .rds file with empty droplets removed 
+# This script takes a SingleCellExperiment stored in a .rds file with empty droplets removed
 # and removes low quality cells, performs normalization, and dimensionality reduction.
 
 # import libraries
@@ -44,7 +44,7 @@ option_list <- list(
     opt_str = c("--n_pcs"),
     type = "integer",
     default = 50,
-    help = "Number of principal components to retain in the returned SingleCellExperiment object 
+    help = "Number of principal components to retain in the returned SingleCellExperiment object
       when using scater::runPCA; the default is n_pcs = 50",
   ),
   make_option(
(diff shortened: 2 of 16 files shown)


modules/qc-report.nf
```diff
@@ -24,7 +24,7 @@ process sce_qc_report{
           --metadata_json ${metadata_json} \
           --technology "${meta.technology}" \
           --seq_unit "${meta.seq_unit}" \
-          --genome_assembly "${params.assembly}" \
+          --genome_assembly "${meta.ref_assembly}" \
           --workflow_url "${workflow_url}" \
           --workflow_version "${workflow_version}" \
```
allyhawkins, line 29:
```suggestion
          --workflow_version "${meta.scpca_version}" \
```
Super minor but I think you could now technically grab this from the metadata? Unless you specifically want to grab it from when it was run instead?

allyhawkins (approved):
I just had one small minor comment but this looks good to me and the approach here makes sense to me. It does make me sad that we have to repeat mapping for everything, but we were going to re-run everything anyways 😢

modules/qc-report.nf
jashapiro: Yeah, this was a debate in my head, but if we grab it from the metadata it will be the version that was used for mapping, but not quantification, etc. So it seemed like this should reflect the latest version, not the checkpoint version.

allyhawkins: Yea that makes sense to me. We could also have a mapping workflow version as a separate line in the metadata. That might be overkill but it might help track if mapping was done on previous versions.

PR #221, opened 2022-09-29
Add process to nextflow for running post process script

### LINKED ISSUE #207: Add process to perform normalization/dimensionality reduction to workflow

allyhawkins:
Closes #207

Here I added a process to run the script added in #210. I added the process to the `sce-processing.nf` module and run it directly in the main workflow immediately before the qc report process. This means that I had to pass the `unfiltered_rds` file through it even though we don't actually use it in the process. We do that for some of the demuxing processes so I thought that would be okay, but if there are other ideas on how to do this I'm open to that. 

I also added parameters to the main config file for the arguments in the script like number of HVG and number of PCs to use. 

I'm still testing this because I need to wait until the docker image for `scpcaTools` builds to try testing again now that `scran` has been added, but once I finish testing then I'll go ahead and request a review. 
Also apologies for the diff having a bunch of white space removal.


jashapiro:
> This means that I had to pass the `unfiltered_rds` file through it even though we don't actually use it in the process. We do that for some of the demuxing processes so I thought that would be okay, but if there are other ideas on how to do this I'm open to that.

We could do something fancier with joins, but I don't think the cost is worth the effort. One thing I would change as you start testing though is to remove the publishDir steps for the prior outputs, so the only place we publish all of these files is in the new post-processing step. Since the filtered data changes there as well with the addition of the "keep" column, we can hopefully avoid potential errors from overwriting.

jashapiro:
I forgot to check before approving the previous PR, but do make sure that `post_process_sce.R` is executable (`chmod +x`) for this PR.

nextflow.config
```diff
@@ -38,6 +38,12 @@ params {
 
   seed = 2021   // random number seed for filtering (0 means use system seed)
```
jashapiro, line 39:
```suggestion
  seed = 2021   // random number seed for filtering and post-processing (0 means use system seed)
```

jashapiro (commented):
Assuming this passes testing, this looks good. Just the prior comments about removing publishing from earlier steps so we only publish once and making the script file executable.

allyhawkins:
After testing this it now works successfully and should be ready for a formal review.

jashapiro (approved):
This looks good to me, with one very minor comment still to be resolved: https://github.com/AlexsLemonade/scpca-nf/pull/221/files#r983946797

allyhawkins:
> This looks good to me, with one very minor comment still to be resolved: https://github.com/AlexsLemonade/scpca-nf/pull/221/files#r983946797

Ah sorry I missed that! Just fixed it.

jashapiro (approved):
Really didn't need to rereview with just the comment change! 😆

PR #223, opened 2022-09-30
Make the metadata.json look pretty

### LINKED ISSUE #222: Make metadata.json file pretty
When I was reviewing #220 and I was looking at the outputs I was reminded again about our `metadata.json` file being ugly and I think it would be nice to make them pretty like the other `json` files that are now being written to the checkpoints folder. This should be an easy fix because I think we just need to add the `pretty=TRUE` option to `jsonlite::write_json` when we are writing out that file.

allyhawkins:
Closes #222 

Mainly because I was tired of scrolling to look at the metadata file, but also because it looks so much better now, this just changes printing out the metadata file to turn it into pretty format.


jashapiro (approved):
LGTM!

PR #225, opened 2022-10-06
Use vireo image

### LINKED ISSUE #219: Make a Docker image for Vireo
As there was no docker image for Vireo easily available, we had used a generic conda image, and install Vireo on the fly within the image when needed. This is fast enough not to be a major issue, but it presents a problem if a user were to try to use Vireo on a system where nodes did not have access to the internet. To avoid that potential problem, we should create a Docker image with Vireo pre-installed, which could then be pre-staged for use.

jashapiro:
After creating the vireo image in https://github.com/AlexsLemonade/alsf-scpca/pull/170, this PR  implements its use in the workflow.

I also changed how the Nextflow version is stored in `meta`, as I found in testing that the previous code was storing it as an object type that was interfering with the `-resume` option. So now it is stored as just a string, which is easier to interpret anyway.

closes #219


allyhawkins (approved):
LGTM 🚀

PR #228, opened 2022-10-10
Add filtering thresholds to metadata of processed SCE object

### LINKED ISSUE #227: Store min gene cutoff and prob compromised in metadata of processed SCE object
In order to create the table being added in https://github.com/AlexsLemonade/scpcaTools/pull/131 we will want to have any parameters used for filtering low quality cells in the metadata of the processed SCE object. Right now we just store the filtering method but not the cutoffs used. We should also add the `gene_cutoff` argument and the `prob_compromised_cutoff` to the metadata of the processed SCE object in `post_process_sce.R`.

allyhawkins:
Closes #227 

Here I added the minimum genes detected cutoff and the prob compromised cutoff to the metadata of the processed SCE object that is written out by `post_process_sce.R`. This will result in these values being printed in the `Cell Statistics` table found in the qc report being altered in AlexsLemonade/scpcaTools#131.

### DIFF
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -92,7 +92,11 @@ if(all(is.na(sce$prob_compromised))){
     "Remove"
   )
   metadata(sce)$ccdl_filter_method <- "miQC"
+  metadata(sce)$prob_compromised_cutoff <- opt$prob_compromised_cutoff
 }
+
+# add min gene cutoff to metadata
+metadata(sce)$min_gene_cutoff <- opt$gene_cutoff
 
 # filter sce using criteria in ccdl_filter
 filtered_sce <- sce[, which(sce$ccdl_filter == "Keep")]


bin/post_process_sce.R
```diff
@@ -92,8 +92,12 @@ if(all(is.na(sce$prob_compromised))){
     "Remove"
   )
   metadata(sce)$ccdl_filter_method <- "miQC"
+  metadata(sce)$prob_compromised_cutoff <- opt$prob_compromised_cutoff
```
jashapiro, line 95: For completeness, do we want to set this to `NA` for the simpler filter method? I know that just leaving it off is fine with the way we handle things later, but it seems like setting an explicit value is less likely to cause trouble.

jashapiro (approved):
This looks good, I just had a thought about whether to explicitly store an `NA` value for `prob_compromised_cutoff`, but I don't feel too strongly.

bin/post_process_sce.R
allyhawkins: I think that's a good so I did that in https://github.com/AlexsLemonade/scpca-nf/pull/228/commits/fe82df4fa51c06628e2ed04c79781ec4ff6d5012 if you want to look at that before I merge.

bin/post_process_sce.R
```diff
@@ -83,6 +83,7 @@ if(all(is.na(sce$prob_compromised))){
     sce$detected >= opt$gene_cutoff, "Keep", "Remove"
   )
   metadata(sce)$ccdl_filter_method <- "Minimum_gene_cutoff"
+  metadata(sce)$prob_compromised_cutoff <- NA_character_
```
jashapiro, line 86: I'm pretty sure this won't matter, but why did you choose `NA_character_` instead of `NA_real_` or just `NA` here? (Since this is not getting joined into anything but a list, I don't think it should matter at all)

allyhawkins: No particular reason other than I wasn't sure if it would affect how it printed out to the table. I can change it to `NA`.

Issue #211, opened 2022-09-06
Add local reference file instructions


jashapiro:
**Is your feature request related to a problem? Please describe.**

Some users have HPCs with nodes that can not access the internet, so they will need local versions of all reference files downloaded to their system and to update the config paths in their local config file.

**Describe the solution you'd like**

We need instructions for downloading the reference files and the paths that will need to be specified in the config file. In particular, all fields in https://github.com/AlexsLemonade/scpca-nf/blob/main/config/reference_paths.config will need to be updated.

**Describe alternatives you've considered**

We might want to include a script that could do the download to a sepcified location _and_ generate/append to the config file? I think doing this right might be a challenge, but would add to convenience.

**Additional context**

PR #230, opened 2022-10-27
Add script for downloading local reference files and images

jashapiro:
For the case when compute nodes may not have access to the internet, we will need to download all reference files and docker images before the workflow begins, as described in #211. This PR adds a script to perform these downloads, preserving the file structure that would be found on S3, which allows us to change only one parameter for running from these local files.

To make things a bit easier, the script also outputs a file, `local_refs.yaml` by default, which contains the setting that needs updating (`ref_rootdir`, is the only required one, but I also included `assembly` for clarity). I discovered that this setting can not be changed in a user-provided config file that specifies `params.ref_rootdir` unless that config file _also_ contains all of the affected param variables in `reference_paths.config`.  This probably has to do with the order in which config files are read and some opaque precedence rules. So the solution is either to use `--ref_rootdir` on the command line, which will properly override the setting in `reference_paths.config` or to use `-params-file local_refs.yaml` or similar. The latter is more traceable, so that is the direction I took with this download.

Downloading STAR and Cell Ranger indexes is optional (these are large), as is downloading docker or singularity packages to the local cache. I ran into some strange behavior with singularity seeming to not wanting to download files to one cache location if it had already downloaded to another, so I force those pull requests.

One unresolved issue is downloading any Space Ranger image that might be required; it won't get those from the github repo, but that is probably part of a larger issue, if people are going to use that functionality. I will try to address the way to handle this in docs.

For now, the script is in the root of the repository. If there are thoughts on where it should ultimately live, please let me know!

Speaking of docs, those will come in a separate PR.


get_refs.py
```diff
@@ -0,0 +1,225 @@
...
+    "cellranger_mit_license.txt",
+    "visium-v1.txt",
+    "visium-v2.txt"
+]
+
+ref_paths += [barcode_dir / f for f in barcode_files]
+
+## download all the files and put them in the correct locations ##
+print("Downloading reference files...")
+for path in ref_paths[0:2]:
```
allyhawkins, line 150:
```suggestion
for path in ref_paths:
```
I don't think this should be here, otherwise it would only download the first 2

allyhawkins (commented):
Thanks for doing this! My one main ideas is if you think there is a way we could get the paths to the reference files from the config file we have rather than directly declaring them here. I'm just worried about things getting out of sync in the future, but if it's too much of a headache this way works too.

get_refs.py
```diff
@@ -0,0 +1,225 @@
...
+assembly = "Homo_sapiens.GRCh38.104"
+genome_dir = pathlib.Path("homo_sapiens/ensembl-104")
+ref_subdirs =[
+    "fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz",
+    "fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.fai",
+    "annotation/Homo_sapiens.GRCh38.104.gtf.gz",
+    "annotation/Homo_sapiens.GRCh38.104.mitogenes.txt",
+    "annotation/Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv",
+    "annotation/Homo_sapiens.GRCh38.104.spliced_cdna.tx2gene.tsv"
+]
```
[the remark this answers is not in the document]
jashapiro: We might be able to do this... I didn't because I was being a bit lazy, but I can try to figure out how to parse it. I worry it may get out of sync anyway if file formats change.

get_refs.py
```diff
@@ -0,0 +1,225 @@
...
+    "visium-v1.txt",
+    "visium-v2.txt"
+]
+
+ref_paths += [barcode_dir / f for f in barcode_files]
+
+## download all the files and put them in the correct locations ##
+print("Downloading reference files...")
+for path in ref_paths[0:2]:
+    outfile = args.refdir / path
```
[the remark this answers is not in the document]
jashapiro: I am using [`pathlib.Path` objects](https://docs.python.org/3/library/pathlib.html) to store paths, which gives some nice extras, rather than as storing them as strings. In an interesting choice, they overload `/` to join `Path` objects.

get_refs.py
jashapiro: Oh yes, that was for testing. I thought I took it out!

get_refs.py
```diff
@@ -0,0 +1,225 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument("--refdir", type=str,
+                    default="scpca-references",
+                    help = "destination directory for downloaded reference files")
+parser.add_argument("--replace",
+                    action = "store_true",
+                    help = "replace previously downloaded files")
+parser.add_argument("--paramfile", type=str,
+                    default="local_refs.yaml",
+                    help = "nextflow param file to write (default: `local_refs.params`)")
```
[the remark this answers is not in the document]
jashapiro: yes... I had originally named the file `.params`, but it turns out nextflow wants a `.yaml` extension. Might change the name to `localref_params.yaml` for clarity?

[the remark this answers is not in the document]
allyhawkins: I like that!

get_refs.py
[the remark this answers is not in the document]
jashapiro: Yes, I realize that I can repurpose a bunch of the code that I wrote for the containers, though I have to deal with the variable substitution...

jashapiro:
> My one main ideas is if you think there is a way we could get the paths to the reference files from the config file we have rather than directly declaring them here.

Okay, I have now implemented this. I thought I was being clever enough to make it work with the version of `reference_paths.config` that is in `main`, but there were more changes there than I expected, and the logic to handle them does not seem worth it at all. So if you want to test this, you will need to run it with `--revision development`.

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+parser.add_argument("--refdir", type=str,
+                    default="scpca-references",
+                    help = "destination directory for downloaded reference files")
+parser.add_argument("--replace",
+                    action = "store_true",
+                    help = "replace previously downloaded files")
+parser.add_argument("--paramfile", type=str,
+                    default="localref_params.yaml",
+                    help = "nextflow param file to write (default: `localref_params.yaml`)")
+parser.add_argument("--revision", type=str,
```
sjspielman, line 23: I might suggest this be named `--version`, `--workflow_version`, `--release`, etc. or similar. Up to you.

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+containerfile_url = f"https://raw.githubusercontent.com/AlexsLemonade/scpca-nf/{args.revision}/config/containers.config"
+
+# download reference file
+print("Getting list of required reference files")
+refs = {}
+try:
+    ref_file =  urllib.request.urlopen(reffile_url)
+except urllib.error.URLError as e:
+    print(e.reason)
+    print(f"The file download failed for {reffile_url}, please check the URL for errors")
```
sjspielman, line 55:
```suggestion
    print(f"The file download failed for {reffile_url}. Please check the URL for errors.")
```

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+import sys
+import urllib.request
+
+from pathlib import Path
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--refdir", type=str,
+                    default="scpca-references",
+                    help = "destination directory for downloaded reference files")
+parser.add_argument("--replace",
```
sjspielman, line 17: I'd suggest `--overwrite` here instead of `--replace`. Maybe even `overwrite_refs`?

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+                    help = "get STAR index (required for genetic demultiplexing)")
+parser.add_argument("--cellranger_index",
+                    action = "store_true",
+                    help = "get Cell Ranger index (required for spatial data)")
+parser.add_argument("--docker",
+                    action = "store_true",
+                    help = "pull and cache images for docker")
+parser.add_argument("--singularity",
+                    action = "store_true",
+                    help = "pull and cache images for singularity")
```
sjspielman, line 38:
```suggestion
                    help = "whether to pull and cache images for singularity")
```

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+                    help = "tag for a specific workflow version (defaults to latest revision)")
+parser.add_argument("--star_index",
+                    action = "store_true",
+                    help = "get STAR index (required for genetic demultiplexing)")
+parser.add_argument("--cellranger_index",
+                    action = "store_true",
+                    help = "get Cell Ranger index (required for spatial data)")
+parser.add_argument("--docker",
+                    action = "store_true",
+                    help = "pull and cache images for docker")
```
sjspielman, line 35:
```suggestion
                    help = "whether to pull and cache images for docker")
```

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+parser.add_argument("--paramfile", type=str,
+                    default="localref_params.yaml",
+                    help = "nextflow param file to write (default: `localref_params.yaml`)")
+parser.add_argument("--revision", type=str,
+                    default="main",
+                    metavar = "vX.X.X",
+                    help = "tag for a specific workflow version (defaults to latest revision)")
+parser.add_argument("--star_index",
+                    action = "store_true",
+                    help = "get STAR index (required for genetic demultiplexing)")
```
sjspielman, line 29:
```suggestion
                    help = "whether to get STAR index (required for genetic demultiplexing)")
```

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+                    default="scpca-references",
+                    help = "destination directory for downloaded reference files")
+parser.add_argument("--replace",
+                    action = "store_true",
+                    help = "replace previously downloaded files")
+parser.add_argument("--paramfile", type=str,
+                    default="localref_params.yaml",
+                    help = "nextflow param file to write (default: `localref_params.yaml`)")
+parser.add_argument("--revision", type=str,
+                    default="main",
```
sjspielman, line 24: Wondering if there is a reason to set this default to `development` instead, since that matches the current functionality. But, I am disagreeing with myself a good deal even as I write this, because of this that will get messy when this eventually is merged into `main` and until merge one just provides via command line.

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+    "visium-v1.txt",
+    "visium-v2.txt"
+]
+barcode_dir = Path(refs.get("barcode_dir"))
+if root_re.match(barcode_dir.parts[0]):
+    barcode_dir = barcode_dir.relative_to(barcode_dir.parts[0])
+ref_paths += [barcode_dir / f for f in barcode_files]
+
+## download all the files and put them in the correct locations ##
+print("Downloading reference files...")
```
sjspielman, line 199: Might be nice to add a blurb about "this will take a while, why not go take a coffee break?" Any kind of expectation-setting statement about runtime, ☕ or otherwise.

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+        exit(1)
+print("Done with reference file downloads\n"
+      f"Reference files can be found at '{args.refdir}'\n")
+
+# write param file if requested
+if args.paramfile:
+    pfile = Path(args.paramfile)
+    # check if paramfile exists & move old if needed
+    if pfile.exists():
+        print(f"A file already exists at `{pfile}`, renaming previous file to `{pfile.name}.bak`")
```
sjspielman, line 224:
```suggestion
        print(f"A file already exists at `{pfile}`. Renaming existing file to `{pfile.name}.bak` and writing new file to `{pfile}`.")
```

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+    print(f"Getting {path}")
+    # make parents
+    outfile.parent.mkdir(exist_ok=True, parents = True)
+    # download and write
+    file_url = f"{url_root}/{path}"
+    try:
+        urllib.request.urlretrieve(file_url, outfile)
+    except urllib.error.URLError as e:
+        print(e.reason)
+        print(f"The file download failed for {file_url}, please check the URL for errors",
```
sjspielman, line 213:
```suggestion
        print(f"The file download failed for {file_url}. Please check the URL for errors.",
```

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+
+## Get docker containers from workflow
+if args.singularity or args.docker:
+    print("Getting list of required containers")
+    containers = {}
+    try:
+        container_file =  urllib.request.urlopen(containerfile_url)
+    except urllib.error.URLError as e:
+        print(e.reason)
+        print(f"The file download failed for {container_url}, please check the URL for errors")
```
sjspielman, line 244:
```suggestion
        print(f"The file download failed for {container_url}. Please check the URL for errors.")
```

sjspielman (commented):
I mostly left some textual comments here! 

> Speaking of docs, those will come in a separate PR.

Checking that this also will include some additional comments within the script?

This is almost certainly a problem on my end, but just in case it's not! I was running from the `continuumio/miniconda3:4.10.3p0` container, and encountered this excitement:
```
(base) root@b815efb3cc5b:/home# python3 get_refs.py 
Getting list of required reference files
Downloading reference files...
Getting homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz
Getting homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.fai
Getting homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz
Getting homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.mitogenes.txt
Getting homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv
Getting homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.spliced_cdna.tx2gene.tsv
Getting homo_sapiens/ensembl-104/salmon_index/Homo_sapiens.GRCh38.104.spliced_intron.txome/complete_ref_lens.bin
Getting homo_sapiens/ensembl-104/salmon_index/Homo_sapiens.GRCh38.104.spliced_intron.txome/ctable.bin
ERRO[2189] error waiting for container: invalid character 'c' looking for beginning of value 
```

I'm blaming this on weird comms with the docker engine, but wanted to share just in case this happened to someone else and it's not just a me-problem. I'm re-running now in my conda `base` environment since Docker got weirdly borked after this and required a couple restarts for the daemon to get back up and running. So far, so good.

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+                    help = "cache directory for singularity")
+args = parser.parse_args()
+
+# scpca-nf resource urls
+reffile_url = f"https://raw.githubusercontent.com/AlexsLemonade/scpca-nf/{args.revision}/config/reference_paths.config"
+containerfile_url = f"https://raw.githubusercontent.com/AlexsLemonade/scpca-nf/{args.revision}/config/containers.config"
+
+# download reference file
+print("Getting list of required reference files")
+refs = {}
```
allyhawkins, line 50: Should this be here? Or does it need to be before line 61? Really minor, but I noticed it isn't used in this chunk.

get_refs.py
```diff
@@ -0,0 +1,275 @@
...
+        refs[match.group('id')] = match.group('loc')
+
+# regular expressions for parameter expansion
+root_re = re.compile(r'\$\{?(params.)?ref_rootdir\}?$')
+refdir_re = re.compile(r'\$\{?(params.)?ref_dir\}?$')
+
+# get assembly and root location
+assembly = refs.get("assembly", "NA")
+root_parts = refs.get("ref_rootdir").split('://')
+if root_parts[0] == 's3':
```
allyhawkins, line 73: Can you add a comment here that you are grabbing the bucket name? I was a bit confused here on what 0, 1 were

allyhawkins (approved):
Just a few minor comments, but this looks good to me. I was able to test pulling docker and everything seemed to work for me.

get_refs.py
jashapiro: I am using revision because that is the term used by nextflow. I don't love it, but I wanted to be consistent with the terminology used there.

get_refs.py
jashapiro: True, it isn't used for a few lines, but it needs to be initialized somewhere...

get_refs.py
jashapiro: Oh, this isn't quite correct, I realized (or as flexible as it needs to be). I will add comments as I update.

get_refs.py
jashapiro: Yes... in theory it should work with `main` once merged, so this should only be a problem for us and the initial testers, who will get a bit more hand-holding.

jashapiro:
> This is almost certainly a problem on my end, but just in case it's not! I was running from the `continuumio/miniconda3:4.10.3p0` container, and encountered this excitement:
> 
> ```
> (base) root@b815efb3cc5b:/home# python3 get_refs.py 
> ```

I'm not sure what that would be, to be honest... I'm somewhat surprised this worked as written above: without specifying `development` I would have expected failure based on how the input file changed.

Seems related to https://github.com/docker/for-mac/issues/5139... There should be no reason to run this in docker, as long as you have some version of python3 installed; if I am using something not available in an earlier version of python3, I'd like to know about it!

> Checking that this also will include some additional comments within the script?

I added a few more comments, but I was not planning to add much more. Was there something specific you were looking for?

sjspielman:
> I added a few more comments, but I was not planning to add much more. Was there something specific you were looking for?

Mostly looking for a docstring or so at the _top_ of the script with 1-2 sentences about what the script does and why it might be useful.

jashapiro:
> > I added a few more comments, but I was not planning to add much more. Was there something specific you were looking for?
> 
> Mostly looking for a docstring or so at the _top_ of the script with 1-2 sentences about what the script does and why it might be useful.

Yes, that should be there; done!

get_refs.py
sjspielman: Agreed for consistency! But it's a disappointing fact to learn as I start my nextflow learning adventure.

sjspielman (approved):
👍

Issue #213, opened 2022-09-12
[BUG] missing rad files?


jashapiro:
**Describe the bug**
Alevin-fry quant can't find rad files.

**To reproduce**

This bug occurred with an external user, who had the following output:

```
Error executing process > 'map_quant_rna:fry_quant_rna (X_7316-1663_5pr_150PE-rna)'

Caused by:
  Process `map_quant_rna:fry_quant_rna (X_7316-1663_5pr_150PE-rna)` terminated with an error exit status (101)

Command executed:

  alevin-fry generate-permit-list       -i X_7316-1663_5pr_150PE-rna       --expected-ori rc       -o X_7316-1663_5pr_150PE-rna       --unfiltered-pl 737K-august-2016.txt
  
  alevin-fry collate       --input-dir X_7316-1663_5pr_150PE-rna       --rad-dir X_7316-1663_5pr_150PE-rna       -t 8
  
  alevin-fry quant       --input-dir X_7316-1663_5pr_150PE-rna       --tg-map Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv       --resolution cr-like-em       -o X_7316-1663_5pr_150PE-rna       --use-mtx       -t 8 
  # remove large files
  rm X_7316-1663_5pr_150PE-rna/*.rad X_7316-1663_5pr_150PE-rna/*.bin

Command exit status:
  101

Command output:
  (empty)

Command error:
  2022-09-09 11:27:34 INFO number of unfiltered bcs read = 737,280
  thread 'main' panicked at 'could not open input rad file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/cellfilter.rs:609:52
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

**Expected behavior**
The step should have run...

**Additional context**
Attaching a the command.run file that seems to contain the error: 
[command.run.txt](https://github.com/AlexsLemonade/scpca-nf/files/9551709/command.run.txt)

We would not expect the file to be deleted, but it is possible that tracing the symlinks is causing trouble. If we can force copying of the input files at this stage (or the output of the previous stage) that might be a solution worth trying.

PR #231, opened 2022-11-02
Add instructions for downloading local reference files

### LINKED ISSUE #211: Add local reference file instructions

jashapiro:
Closes #211

Mostly this is documentation, but I also had to make some changes to the `get_refs.py` to give pulled/cached image files the names Nextflow expects them to have (this is mostly undocumented, so yay). 

Please let me know if these docs make sense. We will get another set of eyes when we deploy to testers.

I think this may also address #213, as I seemed to run into that when testing. I think the issue may have been somehow invalid symbolic links created by publish steps when skipping remapping. To address this, I changed all publish steps to include `mode: 'copy'`. This should prevent any deletion of temp or scratch files from removing results in checkpoints or results.

### DIFF
--- a/external-data-instructions.md
+++ b/external-data-instructions.md
@@ -250,6 +250,10 @@ If you will be analyzing spatial expression data, you will also need the Cell Ra
 #### Downloading container images
 
 If your compute nodes do not have internet access, you will likely have to pre-pull the required container images as well.
+When doing this, it is important to be sure that you also specify the revision (version tag) of the `scpca-nf` workflow that you are using.
+For example, if you would run `nextflow run AlexsLemonade/scpca-nf -r v0.3.4`, then you will want to set `-r v0.3.4` for `get_refs.py` as well to be sure you have the correct containers.
+Be default,  `get_refs.py` will download files and images associated with the latest release.
+
 If your system uses Docker, you can add the `--docker` flag:
 
 ```
--- a/external-data-instructions.md
+++ b/external-data-instructions.md
@@ -37,7 +37,7 @@ You will need to make sure you have the following software installed on your HPC
     These usually require installation by system administrators, but most HPC systems have one available (usually Singularity).
     - Other software dependencies are handled by Nextflow, which will download Docker or Singularity images as required
 
-1. **Organize your files.**
+2. **Organize your files.**
 You will need to have your files organized in a particular manner so that each folder contains only the FASTQ files that pertain to a single library.
 See the [section below on file organization](#file-organization) for more information on how to set up your files.
 
--- a/external-data-instructions.md
+++ b/external-data-instructions.md
@@ -31,10 +31,13 @@ Here we provide an overview of the steps you will need to complete:
 
 1. **Install the necessary dependencies.**
 You will need to make sure you have the following software installed on your HPC where you plan to execute the workflow:
-    - [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
-    - [Docker](https://docs.docker.com/get-docker/) or [Singularity](https://sylabs.io/guides/3.0/user-guide/installation.html#installation)
+    - [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation), the main workflow engine that scpca-nf relies on.
+    This can be downloaded and installed by any user, with minimal external requirements.
+    - [Docker](https://docs.docker.com/get-docker/) or [Singularity](https://sylabs.io/guides/3.0/user-guide/installation.html#installation), which allows the use of container images that encapsulate other dependencies used by the workflow reproducibly.
+    These usually require installation by system administrators, but most HPC systems have one available (usually Singularity).
+    - Other software dependencies are handled by Nextflow, which will download Docker or Singularity images as required
 
-2. **Organize your files.**
+1. **Organize your files.**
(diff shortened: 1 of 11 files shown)


external-data-instructions.md
```diff
@@ -200,6 +203,73 @@ The queue used by each process is determined by Nextflow labels and associated p
 
 The Data Lab's [AWS Batch config file](https://github.com/AlexsLemonade/scpca-nf/blob/main/config/profile_awsbatch.config) may be helpful as a reference for creating a profile for use with AWS, but note that the queues and file locations listed there are not publicly available, so these will need to be set to different values your own profile.
 
+### Using `scpca-nf` on nodes without direct internet access
```
allyhawkins, line 206: I think it might be worth here adding a note in the overview that explains that internet is required for Nextflow to pull reference files and docker/singularity containers. And then a link to this section if you are running on a node without internet. I just noticed a lot in my evals that people were confused on how Docker or singularity is being used so I think explaining upfront that we are pulling containers, like you do here, would also be helpful.

external-data-instructions.md
```diff
@@ -200,6 +203,73 @@ The queue used by each process is determined by Nextflow labels and associated p
...
+If your system does not allow direct internet access, you will need to pre-download the required reference files to a local directory and adjust parameters to direct the workflow to use the local files.
+
+We provide the script [`get_refs.py`](get_refs.py) to download these reference files and optionally pull container images to the location of your choice.
+If you have downloaded the full `scpca-nf` repository, this script is included in the base directory.
+Alternatively, you can download and run this script on its own with the following commands from the location of your choice.
+
+<!-- TODO: Update to `main` before merging -->
+```
+curl -O https://raw.githubusercontent.com/AlexsLemonade/scpca-nf/development/get_refs.py
+chmod +x get_refs.py
```
allyhawkins, line 219: I think my preference is to actually break this up and include only these two commands here which refer specifically to how to access the script and then have a separate chunk that shows the line below, which is running the script. I would then move the below line `./get_refs.py` to below the explanation so users are less tempted to run things without reading what they are actually running.

external-data-instructions.md
```diff
@@ -200,6 +203,73 @@ The queue used by each process is determined by Nextflow labels and associated p
...
+./get_refs.py
+```
+
+By default, this will download the files required for mapping gene expression data sets to the subdirectory `scpca-references` at your current location.
+The script will also create a parameter file named `localref_params.yaml` that defines the `ref_rootdir` and `assembly` Nextflow parameter variables required to use these local data files.
```
allyhawkins, line 224:
```suggestion
Once the script is downloaded and executable, running the script will download the files required for mapping gene expression data sets to the subdirectory `scpca-references` at your current location.
The script will also create a parameter file named `localref_params.yaml` that defines the `ref_rootdir` and `assembly` Nextflow parameter variables required to use these local data files.
```
After this is where I would put the line to run the script.

allyhawkins (commented):
This looks pretty good. I just had a few thoughts for some reorganization and adding some more explanation to the section on installing dependencies (in the overview of the external instructions doc) to specify that internet access is required and why.

jashapiro:
Thanks for looking at this @allyhawkins! Sorry it took a while to get back to, but I have made updates in response to your comments, including adding a bit more to the intro, including some updates to clarify requirements related to containers, as we definitely saw some confusion around that as well.

external-data-instructions.md
```diff
@@ -28,24 +31,30 @@ Here we provide an overview of the steps you will need to complete:
...
-    - [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
-    - [Docker](https://docs.docker.com/get-docker/) or [Singularity](https://sylabs.io/guides/3.0/user-guide/installation.html#installation)
+    - [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation), the main workflow engine that scpca-nf relies on.
+    This can be downloaded and installed by any user, with minimal external requirements.
+    - [Docker](https://docs.docker.com/get-docker/) or [Singularity](https://sylabs.io/guides/3.0/user-guide/installation.html#installation), which allows the use of container images that encapsulate other dependencies used by the workflow reproducibly.
+    These usually require installation by system administrators, but most HPC systems have one available (usually Singularity).
+    - Other software dependencies are handled by Nextflow, which will download Docker or Singularity images as required
 
-2. **Organize your files.**
+1. **Organize your files.**
```
allyhawkins, line 40:
```suggestion
2. **Organize your files.**
```
I think this should still be 2?

allyhawkins (approved):
One minor comment, but looks good to me!

external-data-instructions.md
jashapiro: Yes, but it doesn't matter for markdown rendering, interestingly enough.

PR #233, opened 2022-11-03
Pass probability compromised cutoff through filtering script 

### LINKED ISSUE #232: Use prob_compromised in filtering, then use it in post-processing
> One thing I noticed while doing this is that we have some discrepancies between how we tell people to remove low quality cells in our docs vs. how we are now doing it in the post processing script. We include a `miQC_pass` column in the `_filtered.rds` file that is based on the prob compromised and mito content and we tell people to filter based on that in the getting started section. However, we filter on unique genes to get `_processed.rds`. I just wanted to bring this up because we probably need to address this both in the docs and in `scpca-nf` and I was curious if others have opinions. I think we want to make the `miQC_pass` column consistent with how we are actually filtering.

Oh, I missed this in earlier review. I think we should actually use `miQC_pass` + gene count as our filter in the workflow. We should modify `filter_sce_rds.R` to take the `prob_compromised_cutoff` as an argument so that the `miQC_pass` column reflects the full miQC filtering, then use that later.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-docs/issues/97#issuecomment-1302382649_

allyhawkins:
Closes #232 

Here I altered the `filter_sce_rds.R` script to take the probability compromised cutoff param used for filtering in as an argument rather than passing it through the post process script. This change allows us to use it to directly create the `miQC_pass` column that is present in the SCE that gets written to `_filtered.rds`. As a result of this I also stored the cutoff in the metadata of that object, rather than only in the processed object. 

Then in the post processing script I changed the removal of bad cells to directly use the `miQC_pass` column, if miQC was successful, in combination with the minimum gene cutoff. I believe I addressed everything that was mentioned in the issue, but let me know if I missed something. 

The test run is about halfway done, so I'll wait for that to complete successfully before requesting a review.

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -21,6 +21,13 @@ option_list <- list(
     help = "path to output filtered rds file. Must end in .rds"
   ),
   make_option(
+    opt_str = c("--prob_compromised_cutoff"),
+    type = "double",
+    default = 0.75,
+    help = "probability compromised cutoff used for filtering cells with miQC",
+    metavar = "double"
+  ),
+  make_option(
    opt_str = c("-r", "--random_seed"),
    type = "integer",
    help = "A random seed for reproducibility."
@@ -40,6 +47,11 @@ if(!file.exists(opt$unfiltered_file)){
 # check that output file name ends in .rds
 if(!(stringr::str_ends(opt$filtered_file, ".rds"))){
   stop("filtered file name must end in .rds")
+}
+
+# check that prob compromised cutoff is between 0-1
+if(!dplyr::between(opt$prob_compromised_cutoff, 0, 1)){
+  stop("--prob_compromised_cutoff must be a number between 0 to 1")
 }
 
 # read in unfiltered rds file
@@ -62,7 +74,9 @@ filtered_sce <- filtered_sce |>
 # since this can fail, we will check for success
 miQC_worked <- FALSE
 try({
-  filtered_sce <- scpcaTools::add_miQC(filtered_sce)
+  filtered_sce <- scpcaTools::add_miQC(filtered_sce,
+                                       posterior_cutoff = opt$prob_compromised_cutoff)
+  metadata(sce)$prob_compromised_cutoff <- opt$prob_compromised_cutoff
   miQC_worked <- TRUE
  })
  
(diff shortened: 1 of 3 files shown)


bin/post_process_sce.R
```diff
@@ -72,28 +65,20 @@ if(!(stringr::str_ends(opt$output_sce_file, ".rds"))){
 # read in input rds file
 sce <- readr::read_rds(opt$input_sce_file)
 
-# check that prob compromised cutoff is between 0-1
-if(!dplyr::between(opt$prob_compromised_cutoff, 0, 1)){
-  stop("--prob_compromised_cutoff must be a number between 0 to 1")
-}
-
 # create ccdl_filter column
 if(all(is.na(sce$prob_compromised))){
```
jashapiro, line 69: Since we are using `miQC_pass` now, I think we should test on that? (See also prior comment)
```suggestion
if(all(is.na(sce$miQC_pass))){
```

jashapiro (commented):
This looks good, thanks for tackling it!

I have a couple small comments, but the changes would require testing.

allyhawkins:
I added in the change to check using the `miQC_pass` column and then tested that it still works correctly.

bin/filter_sce_rds.R
```diff
@@ -20,6 +20,13 @@ option_list <- list(
     type = "character",
     help = "path to output filtered rds file. Must end in .rds"
   ),
+  make_option(
+    opt_str = c("--prob_compromised_cutoff"),
+    type = "double",
+    default = 0.75,
+    help = "probability compromised cutoff used for filtering cells with miQC",
+    metavar = "double"
```
jashapiro, line 28: Shouldn't need the metavar, but if we do want it, it should probably be something like `PROB`. (This is the text that is displayed with help in place of the value that would be set, like `--prob_compromized_cutoff PROB`, but we can leave it off.
```suggestion
    help = "probability compromised cutoff used for filtering cells with miQC"
```

bin/filter_sce_rds.R
```diff
@@ -62,14 +74,18 @@ filtered_sce <- filtered_sce |>
...
+                                       posterior_cutoff = opt$prob_compromised_cutoff)
+  metadata(filtered_sce)$prob_compromised_cutoff <- opt$prob_compromised_cutoff
   miQC_worked <- TRUE
  })
  
 # set prob_compromised to NA if miQC failed
 if (!miQC_worked){
   warning("miQC failed. Setting `prob_compromised` to NA.")
   filtered_sce$prob_compromised <- NA_real_
+  filtered_sce$miQC_pass <- NA_real_
```
jashapiro, line 87: Since this is a logical value, NA should be fine on its own.
```suggestion
  filtered_sce$miQC_pass <- NA
```

jashapiro (approved):
LGTM... just a couple tiny changes to suggest.

PR #239, opened 2022-11-16
change ccdl -> scpca for filtering and update template

jashapiro:
A bit of a twofer here. 

One part is updating the references to `ccdl_filter` to scpca, which is a bit more specific, and avoids some branding issues. (see https://github.com/AlexsLemonade/scpca-docs/pull/100/files/b23fbc57a2622d1b9a288b89b73602ee04d7c6eb..8b4bf0723a38e6d527e4485136bf55f557d72ed3#r1024292671)

The other is a small change to talk about cacheDir for singularity, which is a thing that has come up with updates to allow pre-pulling images (I should have fixed this in the last PR but missed it)

### DIFF
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -65,27 +65,27 @@ if(!(stringr::str_ends(opt$output_sce_file, ".rds"))){
 # read in input rds file
 sce <- readr::read_rds(opt$input_sce_file)
 
-# create ccdl_filter column
+# create scpca_filter column
 if(all(is.na(sce$miQC_pass))){
-  colData(sce)$ccdl_filter <- ifelse(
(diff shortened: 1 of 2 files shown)


allyhawkins:
Just realized I think we also should update the metadata slot to be `scpca_filter_method` rather than `ccdl_filter_method`.

jashapiro:
> Just realized I think we also should update the metadata slot to be `scpca_filter_method` rather than `ccdl_filter_method`.

I thought I did that here?

jashapiro:
Oh, it also needs to change in scpcaTools? Filed https://github.com/AlexsLemonade/scpcaTools/pull/145 for that just now

PR #240, opened 2022-11-18
Initiate the integration workflow

### LINKED ISSUE #236: Initiate data integration workflow
We would like to start creating a data integration workflow that will live in this repo as a separate workflow that can be run to integrate a set of processed libraries. The input to the workflow should be processed `SingleCellExperiment` objects that contain normalized and filtered data, the `_processed.rds` returned by `scpca-nf`. The output of the workflow will be an integrated `SingleCellExperiment` object and a summary report. 

We will also read in an integration metadata file that will contain a column for library ID and a column with the group assignment for that library. The group assignment (or integration grouping, whatever we end up calling it) will be used to designate which libraries should be grouped together. 

The first step  or process in the workflow should be merging the individual `SingleCellExperiment` objects into one object and adding a column to the `colData` that contains the library ID that each cell originated from. The merged object will then be used as input to downstream steps which will perform integration via `fastMNN` and `harmony`.

allyhawkins:
Closes #236 

Here I am starting the template for the integration workflow so we can start to build in the individual steps. I focused here on the framework of the nextflow workflow and then once we have established what we want, we can move into building the actual script to merge.

For this workflow I started off with the minimum that we needed to get started. I added a param for integration groups and a param for the integration metadata file. For the integration group for now I set it up to be default of `All`, similar to how we have set up the run IDs in the main workflow. The other parameters that we will need for this workflow, the checkpoints directory, the outputs directory, and the library metadata file are all defined in the main config file so I didn't define them here. 

The main thing I worked on was setting up reading in the channels and getting them into the first process, which is the merging process. For the merging, I wanted to have it so that we provide all of the paths to the input SCE files that are going to be merged along with the `integration_group` or name associated with that group of SCE files. To do this I read in the integration metadata file and the library metadata file separately. For the integration group file I filter by the integration groups we want to run. For the library metadata file, I created a new column for the path to the SCE file for each library. I then joined that with the integration metadata and create a grouped tuple with `[integration_group, [scpca_nf_files]]` that become the input channel to the merging process. 

Within the merging process, I am planning to have as input the `integration_group` and a text file with the list of SCE file paths (similar to how we have it set up in the bulk salmon workflow when we read in multiple salmon output directories). I added creation of the text file with one file per line and one file per group as a test right now that Nextflow is set up correctly. 

The only issue that I'm seeing right now is that the text file that's being produced only prints out the file name and not the full path to the file, even though I'm defining it? I included an example of one of the outputs below. 

You should be able to test this by just using `nextflow run integrate.nf -profile batch,ccdl`.

In the meantime, I also think some of the steps we have in merging (like the actual combining of sces into one), might make sense to include as a function in `scpcaTools`. So we may want to make that addition there first before writing the script here. For right now I'm including this as a draft PR, also because we want to get `v0.4.0` of `scpca-nf` out before we start merging this in. 

I'm going to still request review just to get some 👀  on the set up. 
[acute_myeloid_leukemia-gawad_merged.txt](https://github.com/AlexsLemonade/scpca-nf/files/10045073/acute_myeloid_leukemia-gawad_merged.txt)

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -27,6 +27,12 @@ option_list <- list(
     default = 2000,
     help = "number of high variance genes to use for dimension reduction;
             the default is n_hvg = 2000"
+  ),
+  make_option(
+    opt_str = c("t", "--threads"),
+    type = "integer",
+    default = 1,
+    help = "Number of multiprocessing threads to use"
   )
 )
 
@@ -66,15 +72,33 @@ missing_sce_files <- input_sce_files[which(!file.exists(input_sce_files))]
 if(length(missing_sce_files) > 0){
   stop(
     glue::glue(
-      "\nCannot find input file(s): {missing_sce_files}."
+      "\nCannot find input file: {missing_sce_files}."
     )
   )
 }
+
+# set up multiprocessing params
+if(opt$threads > 1){
+  bp_param = BiocParallel::MulticoreParam(opt$threads)
+} else {
+  bp_param = BiocParallel::SerialParam()
+}
+
 # Merge SCEs -------------------------------------------------------------------
 
 # get list of sces
 sce_list <- purrr::map(input_sce_files, readr::read_rds)
-names(sce_list) <- input_library_ids
+
+
+# check that all input RDS files contain SCE objects
+sce_checks <- purrr::map(sce_list,
+                         \(sce)
+                         is(sce, "SingleCellExperiment"))
+if(!any(sce_checks)){
+  stop(
+    "All input files must contain a `SingleCellExperiment` object."
+  )
+}
 
 # create combined SCE object
 merged_sce <- scpcaTools::merge_sce_list(sce_list,
@@ -90,7 +114,8 @@ batch_column <- merged_sce$library_id
 
 # model gene variance
 gene_var_block <- scran::modelGeneVar(merged_sce,
-                                      block = batch_column)
+                                      block = batch_column,
+                                      BPPARAM = bp_param)
 # identify subset of variable genes
 hvg_list <- scran::getTopHVGs(gene_var_block,
                                n = opt$n_hvg)
@@ -103,16 +128,20 @@ metadata(merged_sce)$merged_hvgs <- hvg_list
 multi_pca <- batchelor::multiBatchPCA(merged_sce,
                                       subset.row = hvg_list,
                                       batch = batch_column,
-                                      preserve.single = TRUE)
+                                      preserve.single = TRUE,
+                                      BPPARAM = bp_param)
 
 # add PCA results to merged SCE object 
 reducedDim(merged_sce, "PCA") <- multi_pca[[1]]
 
 # add UMAP 
-merged_sce <- scater::runUMAP(merged_sce, dimred = "PCA")
+merged_sce <- scater::runUMAP(merged_sce,
+                              dimred = "PCA",
+                              BPPARAM = bp_param)
 
 # write out merged sce file 
 readr::write_rds(merged_sce, 
-                 opt$output_sce_file, 
+                 opt$output_sce_file,
                  compress = "gz",
                  compression = 2)
+
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -26,7 +26,7 @@ option_list <- list(
     type = "integer",
     default = 2000,
     help = "number of high variance genes to use for dimension reduction;
-            the default is n_hvg = 2000",
+            the default is n_hvg = 2000"
   )
 )
 
@@ -95,7 +95,7 @@ gene_var_block <- scran::modelGeneVar(merged_sce,
 hvg_list <- scran::getTopHVGs(gene_var_block,
                                n = opt$n_hvg)
 
-metadata(merged_sce)$merged_highly_variable_genes <- hvg_list
+metadata(merged_sce)$merged_hvgs <- hvg_list
 
 # Dim Reduction PCA and UMAP----------------------------------------------------
 
@@ -106,10 +106,13 @@ multi_pca <- batchelor::multiBatchPCA(merged_sce,
                                       preserve.single = TRUE)
 
 # add PCA results to merged SCE object 
-reducedDim(merged_sce, "PCA") <- multi_pca@listData[[1]]
+reducedDim(merged_sce, "PCA") <- multi_pca[[1]]
 
 # add UMAP 
 merged_sce <- scater::runUMAP(merged_sce, dimred = "PCA")
 
 # write out merged sce file 
-readr::write_rds(merged_sce, opt$output_sce_file)
+readr::write_rds(merged_sce, 
+                 opt$output_sce_file, 
+                 compress = "gz",
+                 compression = 2)
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -38,7 +38,5 @@ input_library_ids <- stringr::str_split(opt$library_ids, ',')
 sce_list <- purrr::map(input_sce_files, readr::read_rds)
 names(sce_list) <- input_library_ids
 
-merged_sce <- sce_list[[1]]
-
 # write out merged sce file 
-readr::write_rds(merged_sce, opt$output_sce_file)
+readr::write_rds(sce_list, opt$output_sce_file)
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -31,9 +31,9 @@ opt <- parse_args(OptionParser(option_list = option_list))
 
 # check that file extension for output file is correct 
 
-# list of paths to salmon files 
+# list of paths to sce files 
 input_sce_files <- unlist(stringr::str_split(opt$input_sce_files, ','))
-# pull library ids from list of 
+# pull library ids from list 
 input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ','))
 
 # check that input files exist 
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -32,9 +32,9 @@ opt <- parse_args(OptionParser(option_list = option_list))
 # check that file extension for output file is correct 
 
 # list of paths to salmon files 
-input_sce_files <- stringr::str_split(opt$input_sce_files, ',')
+input_sce_files <- unlist(stringr::str_split(opt$input_sce_files, ','))
 # pull library ids from list of 
-input_library_ids <- stringr::str_split(opt$library_ids, ',')
+input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ','))
 
 # check that input files exist 
 
@@ -49,4 +49,4 @@ names(sce_list) <- input_library_ids
 # Dim Reduction PCA and UMAP 
 
 # write out merged sce file 
-readr::write_rds(sce_list, opt$output_sce_file)
+readr::write_rds(sce_list[1], opt$output_sce_file)
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -1,0 +1,44 @@
+#!/usr/bin/env Rscript
+
+# import libraries
+library(magrittr)
+library(optparse)
+library(SingleCellExperiment)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("--input_library_ids"),
+    type = "character",
+    help = "Comma separated list of library IDs corresponding to the libraries being integrated."
+  ),
+  make_option(
+    opt_str = c("--input_sce_files"),
+    type = "character",
+    help = "Comma separated list of input sce file paths corresponding to the sces being integrated."
+  ),
+  make_option(
+    opt_str = c("-o", "--output_sce_file"),
+    type = "character",
+    help = "Path to output RDS file, must end in .rds"
+  )
+)
+
+# Setup ------------------------------------------------------------------------
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# list of paths to salmon files 
+input_sce_files <- stringr::str_split(opt$infiles, ',')
+# pull library ids from list of 
+input_library_ids <- stringr::str_split(opt$library_ids, ',')
+
+# get list of sces
+sce_list <- purrr::map(input_sce_files, readr::read_rds)
+names(sce_list) <- input_library_ids
+
+merged_sce <- sce_list[[1]]
+
+# write out merged sce file 
+readr::write_rds(merged_sce, opt$output_sce_file)
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -1,7 +1,6 @@
 #!/usr/bin/env Rscript
 
 # import libraries
-library(magrittr)
 library(optparse)
 library(SingleCellExperiment)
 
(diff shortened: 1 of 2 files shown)


integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+if(param_error){
+  System.exit(1)
+}
+
+// merge individual SCE objects into one SCE object
+process merge_sce {
+  container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.checkpoints_dir}/merged_sces"
+  input:
+    tuple val(int_group), path(scpca_nf_file)
```
jashapiro, line 30: This is not doing what you expect because nextflow does not pass full paths to processes as files will be moved or aliased to the working directory for use. So if you wanted to do this, you would need to pas a `val` here, but you don't really need that, as the paths you were are getting _are_ the paths that script will need. That said, I'm not sure we need the text file? You could pass the list of paths as a comma (or semicolon) -separated list with something like:
```
in_sces = scpca_nf_file.join(',')
```

(Of course, that is a different `join` from the one that works on channels, just for jollies; this is a string join)

Then you could have your R script read in something like 
```
merge_sce.R --group ${int_group} --infiles "${in_sces}" --outfile $merged_sce_file 
```
(side note, `int_group` makes me think it is an integer)

with an early step being `stringr::split(args.infiles, ',')` etc.

integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+        scpca_nf_file: "${params.results_dir}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
+      ]}
+      .map{meta -> tuple(meta,
+                         scpca_nf_file: file(meta.scpca_nf_file)
+                         )}
+
+    all_meta_ch = integration_meta_ch
+      .map{[it["library_id"]] + it }
+      // pull out library_id from meta and use to join
+      .join(runs_ch.map{[it[1]["library_id"]] + it }, by: 0)
```
jashapiro, line 78: I think this  should be `.combine`, because `.join` will only use the first match, and it is possible that library would be in more than one integration group. That might not be a problem, since the groups are first, but I am never confident about what will happen with `.join`.

integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+      // create tuple of just integration group and output file from scpca_nf
+      .map{[
+        it[1].integration_group,
+        file(it[2].scpca_nf_file)
+        ]}
```
jashapiro, line 83: I would probably maintain the `library_ids` here for potential future use. This might be a place to make a new meta object (dict), or you could just add back in the library field, which is probably simpler for now.
```suggestion
      .map{[
        it[1].integration_group,
        it[0],
        file(it[2].scpca_nf_file)
        ]}
```

You could then build the `meta` object in `merge_sce()`

integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+      ]}
+      .filter{int_groups_all  || (it.integration_group in int_groups)}
+
+    // channel with run metadata, keeping only the columns we need
+    runs_ch = Channel.fromPath(params.run_metafile)
+      .splitCsv(header: true, sep: '\t')
+      .map{[
+        library_id: it.scpca_library_id,
+        run_id: it.scpca_run_id,
+        sample_id: it.scpca_sample_id.split(";").sort().join(","),
```
jashapiro, line 68: Do you expect to use this at all? Merging multiplexed libraries is going to be a whole set of headaches that I am not sure we want to be adding here...

jashapiro (commented):
The nextflow logic here looks good, with a couple suggestions for my own comfort and past errors. Your question about the path not looking like what you expect is, I think a nextflow thing that is fine. You can expect the given paths to work within the context of a process, but you should not expect the paths to be stable _between_ processes. That said, I don't think you need to make the file at all? My suggestion is in comments below;

integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+        scpca_nf_file: "${params.results_dir}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
+      ]}
+      .map{meta -> tuple(meta,
```
sjspielman, line 71: What I've experienced is that this always prepends to meta (self) regardless of order provided to tuple. Confirming this is expected behavior?

integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+                         scpca_nf_file: file(meta.scpca_nf_file)
+                         )}
+
+    all_meta_ch = integration_meta_ch
+      .map{[it["library_id"]] + it }
```
sjspielman, line 76: We can add to `.append()` to lists (or whatever this kind of object is called)? I'm a little confused about the structure that comes out of this.

sjspielman (approved):
My review here is actually some in-line questions making sure I understand what's happening, mostly with some of the data structures here. I can run it all and it makes sense which means it "passes my review" :tada: (in other words, probably/definitely/obviously focus on @jashapiro comments for final approval..).

integrate.nf
jashapiro: This creates a list with all of the previous contents of `it` (an element of the channel) prepended by `library_id`. In this case, since `it` is a single map(dict), this is equivalent to `[it["library_id"], it]`, but if `it` was a tuple, as we often have in our channels, then when we do something like:
```
[[it[0]["library_id"]] + it]
```

That results in:
```
[it[0]["library_id"], it[0], it[1], ...]
```
While 
```
[it[0]["library_id"], it]
```
 would give 
```
[it[0]["library_id"], [it[0], it[1], ...] ]
```

Why not use an `add()`or `addAll()` (the append I think you are looking for?) or something to insert the value earlier?
Because `it` (if a tuple) is immutable, so we have to make a mutable list to be able to add to it. And more importantly, those methods don't return the list, so we would have to do something like:
```
.map{out = [it[0]["library_id"]]; out.addAll(it); out}`
```
to actually get the modified list out.

I should say, this is all based on my understanding of groovy, which is not deep!

integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+    runs_ch = Channel.fromPath(params.run_metafile)
+      .splitCsv(header: true, sep: '\t')
+      .map{[
+        library_id: it.scpca_library_id,
+        run_id: it.scpca_run_id,
+        sample_id: it.scpca_sample_id.split(";").sort().join(","),
+        scpca_nf_file: "${params.results_dir}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
+      ]}
+      .map{meta -> tuple(meta,
+                         scpca_nf_file: file(meta.scpca_nf_file)
```
jashapiro, line 72: Since this is a tuple, we just need the value, with no key... I'm actually not sure what it does in this case?
```suggestion
                         file(meta.scpca_nf_file)
```

integrate.nf
allyhawkins: Hmm...so using `.combine` causes a problem because it grabs all matches. So if a library ID is present in the metadata file multiple times, so for CITE-seq samples, it will show all matches and then cause a file collision. So I think what I'm going to do is just filter to only certain technology right now to exclude CITE-seq and then combine.

jashapiro: Yes, filtering first is a good idea. This is good to catch, because otherwise you could get into trouble if the cite-seq run was first!

allyhawkins:
Thanks for taking a look at this @jashapiro and for the helpful feedback. I incorporated your suggestions and was able to simplify some things before the grouping. I also removed making the text file and directly read in the sce files as a comma separated list of files. I also added a simple script right now that reads in the files and then writes out the first SCE object. Since we are keeping this in draft state right now anyways, I'll work on building up the subsequent steps in other PRs.

bin/merge_sces.R
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env Rscript
+
+# import libraries
+library(magrittr)
```
jashapiro, line 4: Okay, but I really do want to convert everyone to `|>`

bin/merge_sces.R
```diff
@@ -0,0 +1,52 @@
...
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# check that file extension for output file is correct 
+
+# list of paths to salmon files 
+input_sce_files <- unlist(stringr::str_split(opt$input_sce_files, ','))
+# pull library ids from list of 
+input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ','))
```
jashapiro, line 37: Do we want this to be optional? We could get the library ids from sce files themselves.

bin/merge_sces.R
```diff
@@ -0,0 +1,52 @@
...
+
+# check that input files exist 
+
+# get list of sces
+sce_list <- purrr::map(input_sce_files, readr::read_rds)
+names(sce_list) <- input_library_ids
```
jashapiro, line 43: If `library_ids` are brought in independently, we should still check that they match, or throw an error.

bin/merge_sces.R
```diff
@@ -0,0 +1,52 @@
...
+
+# create combined SCE object with function in scpcaTools 
+
+# HVG calculation (from perform_hvg_selection in integration-helpers)
+
+# Dim Reduction PCA and UMAP 
+
+# write out merged sce file 
+readr::write_rds(sce_list[1], opt$output_sce_file)
```
jashapiro, line 52: I probably would have written the full list for now, but it really doesn't matter!

integrate.nf
```diff
@@ -0,0 +1,88 @@
...
+    path merged_sce_file
+  script:
+    input_sces = scpca_nf_file.join(',')
+    input_library_ids = library_ids.join(',')
+    merged_sce_file = "${integration_group}_merged.rds"
+    """
+    merge_sces.R \
+      --input_library_ids "${input_library_ids}" \
+      --input_sce_files "${input_sces}" \
+      --output_sce_file ${merged_sce_file}
```
jashapiro, line 41: May as well put them all in quotes.
```suggestion
      --output_sce_file "${merged_sce_file}"
```

integrate.nf
```diff
@@ -0,0 +1,88 @@
...
+process merge_sce {
+  container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.checkpoints_dir}/merged_sces"
+  input:
+    tuple val(integration_group), val(library_ids), path(scpca_nf_file)
+  output:
+    path merged_sce_file
+  script:
+    input_sces = scpca_nf_file.join(',')
+    input_library_ids = library_ids.join(',')
```
jashapiro, line 35: Keeping order consistent 
```suggestion
    input_library_ids = library_ids.join(',')
    input_sces = scpca_nf_file.join(',')
```

integrate.nf
```diff
@@ -0,0 +1,88 @@
...
+      .splitCsv(header: true, sep: '\t')
+      // only include single-cell/single-nuclei and make sure no CITE-seq/ hashing libraries
+      .filter{it.technology in ['10Xv2', '10Xv2_5prime', '10Xv3', '10Xv3.1']}
+      .map{[
+        library_id: it.scpca_library_id,
+        scpca_nf_file: "${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
+      ]}
+
+    all_meta_ch = integration_meta_ch
+      .map{[it["library_id"]] + it }
```
jashapiro, line 73: thinking on this, I think it is a bit simpler to read and more consistent with later use if we do this as:
```suggestion
      .map{[it.library_id, it]}
```

integrate.nf
```diff
@@ -0,0 +1,88 @@
...
+      // pull out library_id from meta and use to join
+      .combine(runs_ch.map{[it["library_id"]] + it }, by: 0)
```
jashapiro, line 75:
```suggestion
      .combine(runs_ch.map{[it.library_id, it]}, by: 0)
```

integrate.nf
```diff
@@ -0,0 +1,88 @@
...
+      // create tuple of integration group, library ID, and output file from scpca_nf
+      .map{[
+        it[1].integration_group,
+        it[0], // library_id
+        file(it[2].scpca_nf_file)
```
jashapiro, line 80: A couple of clarifying comments.
```suggestion
        it[1].integration_group, // from integration_meta_ch
        it[0], // library_id
        file(it[2].scpca_nf_file) // from runs_ch
```

integrate.nf
```diff
@@ -0,0 +1,88 @@
...
+        ]}
+      // grouped tuple of [integration_group, [file1, file2, file3, ...]]
```
jashapiro, line 82: Updating comment
```suggestion
      // grouped tuple of [integration_group, [library_id1, library_id2, ...], [sce_file1, sce_file2, ...]]
```

bin/merge_sces.R
```diff
@@ -0,0 +1,52 @@
...
+)
+
+# Setup ------------------------------------------------------------------------
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# check that file extension for output file is correct 
+
+# list of paths to salmon files
```
jashapiro, line 34:
```suggestion
# list of paths to sce files 
```

bin/merge_sces.R
```diff
@@ -0,0 +1,52 @@
...
+# Setup ------------------------------------------------------------------------
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# check that file extension for output file is correct 
+
+# list of paths to salmon files 
+input_sce_files <- unlist(stringr::str_split(opt$input_sce_files, ','))
+# pull library ids from list of
```
jashapiro, line 36:
```suggestion
# pull library ids from list 
```

jashapiro (commented):
This looks good to me, with only a few minor comments for now. I can't decide whether to approve it, but I think I will hold off until we are done with the next workflow release, just as a double check alongside the draft status.

bin/merge_sces.R
jashapiro: I added this check earlier, and the way I did it allows me to delete a step here, as the names are already set (and purrr preserves them). I wonder if we should add a check that the read objects are SCEs though?
```suggestion
sce_list <- purrr::map(input_sce_files, readr::read_rds)
```

integrate.nf
```diff
@@ -0,0 +1,90 @@
...
+
+// merge individual SCE objects into one SCE object
+process merge_sce {
+  container params.SCPCATOOLS_CONTAINER
+  label 'mem_16'
+  publishDir "${params.checkpoints_dir}/merged_sces"
+  input:
+    tuple val(integration_group), val(library_ids), path(scpca_nf_file)
+  output:
+    path merged_sce_file
```
jashapiro, line 33: One more suggestion: for future use, we may want to keep the integration group in the output as we do for most tasks in the main workflow.
```suggestion
    tuple val(integration_group), path(merged_sce_file)
```

allyhawkins:
Thanks for the helpful comments @jashapiro. I believe I addressed all of them and have tested the updates and everything works as it should. This should be ready for another look.

bin/merge_sces.R
```diff
@@ -0,0 +1,147 @@
...
+# Merge SCEs -------------------------------------------------------------------
+
+# get list of sces
+sce_list <- purrr::map(input_sce_files, readr::read_rds)
+
+
+# check that all input RDS files contain SCE objects
+sce_checks <- purrr::map(sce_list,
+                         \(sce)
+                         is(sce, "SingleCellExperiment"))
```
jashapiro, line 96: Without the brackets it seems strange to put this on a separate line. I'm also removing the assumption in the variable name?
```suggestion
                         \(x) is(x, "SingleCellExperiment"))
```

jashapiro (approved):
LGTM, with a couple of very small updates suggested.

PR #242, opened 2022-11-18
Change arguments to `read_alevin`

allyhawkins:
The changes we added to `scpcaTools::read_alevin` in https://github.com/AlexsLemonade/scpcaTools/pull/144 changed some argument names to remove the `intron_mode` and `which_counts` arguments and replace with one `include_unspliced`. Here I modified the script that generates the unfiltered SCE object and reads in the data to accommodate those changes. I know that `include_unspliced` is set to `TRUE` as the default in the function, but I just put it here to be explicit. I didn't add it as a param to the actual workflow, but we could if others think that's important? 

I also removed the `seq_unit` arg that was being passed into the script because we were using it before to figure out which counts to include spliced vs. unspliced. Now we are including both matrices for all of them so no longer need to know if the library is single-cell or nuclei.

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -93,16 +93,12 @@ mito_genes <- unique(scan(opt$mito_file, what = "character"))
 # read in gtf file (genes only for speed)
 gtf <- rtracklayer::import(opt$gtf_file, feature.type = "gene")
 
-# convert seq_unit to spliced or unspliced to determine which types of transcripts to include in final counts matrix
-which_counts <- dplyr::case_when(opt$seq_unit == "cell" ~ "spliced",
-                                 opt$seq_unit == "nucleus" ~ "unspliced")
-
 # parse sample id list
 sample_ids <- unlist(stringr::str_split(opt$sample_id, ",|;")) |> sort()
 
 # get unfiltered sce
 unfiltered_sce <- read_alevin(quant_dir = opt$alevin_dir,
-                              which_counts = which_counts,
+                              include_unspliced = TRUE,
                               usa_mode = TRUE,
                               tech_version = opt$technology,
                               library_id = opt$library_id,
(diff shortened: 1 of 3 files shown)


bin/generate_unfiltered_sce.R
```diff
@@ -112,7 +98,8 @@ unfiltered_sce <- read_alevin(quant_dir = opt$alevin_dir,
 # read and merge feature counts if present
 if (opt$feature_dir != ""){
   feature_sce <- read_alevin(quant_dir = opt$feature_dir,
-                             mtx_format = TRUE,
+                             include_unspliced = TRUE,
+                             usa_mode = TRUE,
                              library_id = opt$library_id,
                              sample_id = sample_ids)
```
jashapiro, line 104: Have we tested with feature counts? I just realized that this might actually be a bit of a problem... there is no such thing as spliced and unspliced for feature counts, so we should probably handle these without adding the same metadata (transcript type has no meaning for ADTs, for example) that we do otherwise. Alternatively, we could remove/replace it here, but I think handling it in the `read_alevin` function might be a better plan).

In either case, the options here should probably be `include_unspliced =FALSE, usa_mode = FALSE`, except that we probably need `usa_mode` for reading the mtx file?

Ugh... I think I just realized a problem. If there are any tags with a `-A` or `-U` suffix in the name, `fishpond` may cause trouble. We might need to add back `read_alevin_mtx()` for just these feature files where we do no combining.

jashapiro (commented):
I thought this was going to be simple, but I may have identified a problem which will require us to go back to the read_alevin function one more time. I wasn't thinking too much about feature files for CITE-seq and cellhash, but since they have an uncontrolled set of feature names, we probably don't want to use USA mode/fishpond to read them in. So I think we need to retain the `mtx_format = TRUE` option, or something similar to handle mappings with no spliced/intron/ambiguous reads.

allyhawkins:
Thanks for catching the feature counts potential issue. That should now be accounted for as of merging AlexsLemonade/scpcaTools#147. I incorporated those changes here and also add a `spliced_only` option. I don't think it's entirely necessary, but maybe we'll want it one day. This should be ready for another look.

jashapiro (approved):
LGTM!

PR #245, opened 2022-11-23
Small fixes to match updates in scpcaTools

allyhawkins:
In testing for the upcoming release of `scpcaTools`, I found a few small errors that needed to be addressed here. 

- Passing the processed SCE object through to the QC report 
- In passing the processed SCE object through, I also added some values to the `metadata.json` that gets produced to reflect filtering and normalization additions. 
- Adding a missing `opt`
- remove extra `--seq_unit` option

I also changed to use the tag for what is about to be the most recent release of `scpcaTools`. I did test it with the new `edge` tag that just built after merging https://github.com/AlexsLemonade/scpcaTools/pull/151 and things looked as expected, even in the QC report.

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -61,6 +61,7 @@ option_list <- list(
   make_option(
     opt_str = c("--spliced_only"),
     action = "store_true",
+    default = FALSE,
     help = "include only the spliced counts as the main counts assay in the returned SCE object"
   )
 )
@@ -93,7 +94,6 @@ sample_ids <- unlist(stringr::str_split(opt$sample_id, ",|;")) |> sort()
 
 # set include unspliced for non feature data
 include_unspliced <- !opt$spliced_only
-}
 
 # get unfiltered sce
 unfiltered_sce <- read_alevin(quant_dir = opt$alevin_dir,
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -92,7 +92,7 @@ gtf <- rtracklayer::import(opt$gtf_file, feature.type = "gene")
 sample_ids <- unlist(stringr::str_split(opt$sample_id, ",|;")) |> sort()
 
 # set include unspliced for non feature data
-include_unspliced <- !opts$spliced_only
+include_unspliced <- !opt$spliced_only
 }
 
 # get unfiltered sce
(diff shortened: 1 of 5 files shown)


bin/sce_qc_report.R
```diff
@@ -152,12 +159,16 @@ metadata_list <- list(
   has_cellhash = has_cellhash,
   filtered_cells = ncol(filtered_sce),
   unfiltered_cells = ncol(unfiltered_sce),
-  filtering_method = filtered_sce_meta$filtering_method,
+  empty_drops_filtering_method = filtered_sce_meta$filtering_method,
```
jashapiro, line 162: Do we use this downstream at all? If not yet, then a small change:
```suggestion
  droplet_filtering_method = filtered_sce_meta$filtering_method,
```

bin/generate_unfiltered_sce.R
```diff
@@ -92,7 +92,7 @@ gtf <- rtracklayer::import(opt$gtf_file, feature.type = "gene")
 sample_ids <- unlist(stringr::str_split(opt$sample_id, ",|;")) |> sort()
 
 # set include unspliced for non feature data
-if(!is.null(spliced_only)){
+if(!is.null(opt$spliced_only)){
   include_unspliced <- FALSE
 } else {
   include_unspliced <- TRUE
```
jashapiro, line 98: since we are using `action='store_true'`, we can simplify (and correct) this
```suggestion
include_unspliced <- !opts$spliced_only
```

jashapiro (commented):
LGTM with one fix and one question/suggestion. Not approving quite yet though, as we do want to check that it really works with 0.2.0

bin/sce_qc_report.R
allyhawkins: No we don't, so I'll change this.

allyhawkins:
Just noting that I tested this with scpcaTools v0.2.0 and everything works now other than the NA gene symbol which I accounted for in https://github.com/AlexsLemonade/scpcaTools/pull/155.

jashapiro (approved):
LGTM! Thanks for doing the testing!

One little question, just for my edification.

bin/generate_unfiltered_sce.R
```diff
@@ -61,6 +61,7 @@ option_list <- list(
   make_option(
     opt_str = c("--spliced_only"),
     action = "store_true",
+    default = FALSE,
```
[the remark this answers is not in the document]
jashapiro: Good to know! I guess it works differently from the equivalent in python!

PR #249, opened 2022-11-29
Add scpca-meta.json outputs to more publishDirs

jashapiro:
This PR adds scpca-meta.json outputs to a few more outputs, for future-proofing & potential tracing. The additions are to the feature rad & quant output directories and to the alevinfry quant directory. These outputs are only used if `--publish_fry_outs` is specified, so I wasn't thinking about them before, but I think it is useful to have the output files just in case!


allyhawkins (approved):
The changes look good, but the joke is definitely going over my head.

PR #251, opened 2022-11-30
Update output files section to include processed rds

### LINKED ISSUE #248: Update example output and example output description
We now have an additional `_processed.rds` file that is included in the output for each library. We should update the example output that is linked in the [examples section](https://github.com/AlexsLemonade/scpca-nf/tree/main/examples#example-output) so that it reflects the new structure. 

We also want to update the external instructions section on [output files](https://github.com/AlexsLemonade/scpca-nf/blob/development/external-data-instructions.md#output-files) to show the right output files and add in the `_processed.rds` file.

allyhawkins:
Closes #248 

Here I updated the section on the output files in the external users instructions to include mention of the `_processed.rds` file. There were some remaining mentions of the `publish` directory so I changed that too.

I also re-ran our example data through the current development version of the pipeline to update what the output structure will look like. I created a zip of that output and saved it to the same place we had it previously in `s3://scpca-references/example-data/scpca_out.zip` and made sure that the read access was set to world.

### DIFF
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -371,7 +371,7 @@ For more information on the contents of these files, see the [ScPCA portal docs 
 See below for the expected structure of the `results` folder:
 
 ```
-publish
+results
 └── sample_id
     ├── library_id_filtered.rds
     ├── library_id_metadata.json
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM!

PR #252, opened 2022-11-30
Change miQC default filtering

### LINKED ISSUE #250: Investigate post-process filtering
When looking at some preliminary results, there was a bit of an odd result with SCPCL000533 where the results from miQC don't seem to match the filtering that we actually want to apply. All of the filtering happens in the `post_process_sce.R`, so we should be able to fix this without going upstream to scpcaTools.

For reference, the miQC plot looks like this:
download

Which would seem to imply that a large number of cells should be kept. However, only 259 cells are kept and the keep/discard plot ends up looking like this:
download-1

We shoudl investigate what is happening here before cutting another release!

jashapiro:
In investigating #250, we found that the culprit, at least in part, is miQC's `enforce_left_cutoff` option. 

Here I am adding an option to our filtering script to make the default for our workflow to set that option as `FALSE`, and only set it as `TRUE` if the `--enforce_left_cutoff` flag is passed at the command line. 

Note that doing it this way requires no changes to the workflow itself!

⚠️ Requires changes to scpcaTools made in https://github.com/AlexsLemonade/scpcaTools/pull/156

closes #250

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -27,6 +27,12 @@ option_list <- list(
     help = "probability compromised cutoff used for filtering cells with miQC"
   ),
   make_option(
+    opt_str = c("--enforce_left_cutoff"),
+    action = "store_true",
+    default = FALSE,
+    help = "flag to set the miQC's enforce_left_cutoff option to TRUE"
+  ),
+  make_option(
    opt_str = c("-r", "--random_seed"),
    type = "integer",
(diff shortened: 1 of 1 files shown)


allyhawkins (commented):
These changes look good, but we should probably test it with the changes in AlexsLemonade/scpcaTools#156 before merging to be sure.

allyhawkins (approved):
🚀 🚀

PR #253, opened 2022-12-06
Update version to v0.4.0

allyhawkins:
This PR prepares for the next release of `scpca-nf` and ups the version tag to v0.4.0 throughout the docs and config file. We also did some rearranging of docs in this release, so I updated the release checklist to reflect those changes as well.


jashapiro (approved):
LGTM! I did a quick search to be sure you got all references to 0.3.x, and it seems you did. I would definitely have missed updating the release checklist, so thank you for catching that!

PR #255, opened 2022-12-06
Script for merging SCE objects 

### LINKED ISSUE #243: Actually merge and prepare SCEs for integration in merging script
In #240, we are initiating the integration workflow, the first step which includes merging all of the SCEs from an integration group into one merged SCE file. The focus of that PR is setting up the workflow to read in the metadata files and group the input SCE files by the integration group and pass those files into the merging script. Next we will need to actually input the steps in the merging script which will take the list of input files and produce a merged SCE object, using a function to combine a list of SCE objects. 

In addition to producing a merged SCE object we then need to identify the combined HVGs for that merged object and calculate the merged PCA matrix, storing both of those things in the object that is exported.

allyhawkins:
Closes #243 
⚠️ Stacked on #240 

Here I'm updating the script for merging SCE's that was initiated in #240 to actually perform merging, combined HVG selection, multi batch PCA, and calculate UMAP. The output is a merged SCE object containing HVGs and dim reduction results stored in "PCA" and "UMAP". 

A few things: 

- Right now this script has an option to input both the SCE file names and the library IDs. However the library IDs are optional and if they are not input, then they are grabbed directly from the SCE file name. I was debating removing the library ID option all together and from the channel that gets passed as input to this script as we should always be able to grab the library IDs from the input SCE files. 
- This currently doesn't work all the way through because `batchelor` isn't quite yet in the `scpcaTools` docker image. It will be once https://github.com/AlexsLemonade/scpcaTools/pull/157 is merged. Right now it fails because it can't find `batchelor`, which is the second to last step before writing out the object. Once that PR is merged then I'll test this again and make any necessary updates. I just wanted to get any initial thoughts on structure or decisions made.

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -32,20 +32,65 @@ opt <- parse_args(OptionParser(option_list = option_list))
 
 # list of paths to sce files 
 input_sce_files <- unlist(stringr::str_split(opt$input_sce_files, ','))
-# pull library ids from list 
-input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ','))
+
+# set up library ids
+if(is.null(opt$input_library_ids)){
+  # extract library ids from filename
+  input_library_ids <- stringr::word(basename(input_sce_files), 1, sep = "_")
+} else {
+  # pull library ids from list 
+  input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ',')) 
+}
 
 # check that input files exist 
+missing_sce_files <- input_sce_files[which(!file.exists(input_sce_files))]
+if(length(missing_sce_files) > 0){
+  stop(
+    glue::glue(
+      "\nMissing SCE object for {missing_sce_files}."
+    )
+  )
+}
+# Merge SCEs -------------------------------------------------------------------
 
 # get list of sces
 sce_list <- purrr::map(input_sce_files, readr::read_rds)
 names(sce_list) <- input_library_ids
 
-# create combined SCE object with function in scpcaTools 
+# create combined SCE object
+merged_sce <- scpcaTools::merge_sce_list(sce_list,
+                                         batch_column = "library_id",
+                                         preserve_rowdata_cols = "gene_symbol",
+                                         cell_id_column = "cell_id")
 
-# HVG calculation (from perform_hvg_selection in integration-helpers)
 
-# Dim Reduction PCA and UMAP 
+# HVG selection ----------------------------------------------------------------
+
+# extract the column with the block variable
+batch_column <- merged_sce$library_id
+
+# model gene variance
+gene_var_block <- scran::modelGeneVar(merged_sce,
+                                      block = batch_column)
+# identify subset of variable genes
+hvg_list <- scran::getTopHVGs(gene_var_block,
+                               n = 2000)
+
+metadata(merged_sce)$merged_highly_variable_genes <- hvg_list
+
+# Dim Reduction PCA and UMAP----------------------------------------------------
+
+# multi batch PCA on merged object
+multi_pca <- batchelor::multiBatchPCA(merged_sce,
+                                      subset.row = hvg_list,
+                                      batch = batch_column,
+                                      preserve.single = TRUE)
+
+# add PCA results to merged SCE object 
+reducedDim(merged_sce, "PCA") <- multi_pca@listData[[1]]
+
+# add UMAP 
+merged_sce <- scater::runUMAP(dimred = "PCA")
 
 # write out merged sce file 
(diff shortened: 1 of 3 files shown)


integrate.nf
```diff
@@ -25,6 +25,7 @@ if(param_error){
 // merge individual SCE objects into one SCE object
 process merge_sce {
   container params.SCPCATOOLS_CONTAINER
+  label 'mem_8'
```
sjspielman, line 28: Should this be bumped to 16 GB? It's ultimately the size of the SCEs for this process I would, so depends on our largest anticipated integration groups. I imagine this will be tested and maybe tweaked over time!

bin/merge_sces.R
```diff
@@ -32,20 +32,65 @@ opt <- parse_args(OptionParser(option_list = option_list))
...
+
+# extract the column with the block variable
+batch_column <- merged_sce$library_id
+
+# model gene variance
+gene_var_block <- scran::modelGeneVar(merged_sce,
+                                      block = batch_column)
+# identify subset of variable genes
+hvg_list <- scran::getTopHVGs(gene_var_block,
+                               n = 2000)
```
sjspielman, line 77: I might think we want to take advantage of this config param? 

https://github.com/AlexsLemonade/scpca-nf/blob/046ef7b014b23b0f82e93f15bf1f21cbc234d960/nextflow.config#L45

allyhawkins:
Thanks for taking a look at this @sjspielman! I incorporated your comments, adding in some checks for the input SCE files, using the param we have for the number of HVGs to use, and also bumping up the memory for now although I think we'll know more about memory when we test and we might find that we can bump it back down. Part of the benefit of nextflow is we have a retry strategy that upon failure due to memory reasons, it will retry and double the memory request the next time.

Also one of the checks I added was to throw an error if there's only one input file provided. I think for this workflow we want to do that because we don't want to be performing merging or integration on one object and want to stop. 

I'm going to wait to re-request review until I get a chance to fully test things with merging the integration functions into `scpcaTools` and adding `batchelor` to the docker image.

bin/merge_sces.R
```diff
@@ -29,23 +36,80 @@ option_list <- list(
...
+                                      batch = batch_column,
+                                      preserve.single = TRUE)
 
-# HVG calculation (from perform_hvg_selection in integration-helpers)
+# add PCA results to merged SCE object 
+reducedDim(merged_sce, "PCA") <- multi_pca@listData[[1]]
 
-# Dim Reduction PCA and UMAP 
+# add UMAP 
+merged_sce <- scater::runUMAP(dimred = "PCA")
```
jashapiro, line 112:
```suggestion
merged_sce <- scater::runUMAP(merged_sce, dimred = "PCA")
```

allyhawkins:
Just confirming that I was able to test this with the updates to `scpcaTools` now in and the workflow works as expected. Marking it ready for full review!

bin/merge_sces.R
```diff
@@ -29,23 +36,80 @@ option_list <- list(
...
+  # list of paths to sce files 
+  input_sce_files <- unlist(stringr::str_split(opt$input_sce_files, ',')) 
+}
 
-# list of paths to sce files 
-input_sce_files <- unlist(stringr::str_split(opt$input_sce_files, ','))
-# pull library ids from list 
-input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ','))
+if(length(input_sce_files) == 1){
+  stop("Only 1 input file provided, no merging or integration will be performed for this group")
```
sjspielman, line 52: Would this `stop()` kill the whole nf workflow? If not, we may not want to say anything about "integration" here since this script does no integrating. But if this will kill the whole thing, then it's more reasonable.

bin/merge_sces.R
```diff
@@ -29,23 +36,80 @@ option_list <- list(
...
+  # pull library ids from list 
+  input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ',')) 
+}
 
 # check that input files exist 
+missing_sce_files <- input_sce_files[which(!file.exists(input_sce_files))]
+if(length(missing_sce_files) > 0){
+  stop(
+    glue::glue(
+      "\nCannot find input file: {missing_sce_files}."
```
sjspielman, line 69:
```suggestion
      "\nCannot find input file(s): {missing_sce_files}."
```

sjspielman (approved):
This looks fine to me! I don't necessarily need to see it again.

bin/merge_sces.R
allyhawkins: It would kill the whole thing the way it's set up right now. We could alter how errors are handled for this particular script in the nextflow process, but for now it would stop it.

PR #258, opened 2022-12-12
Remove extra slash in vireo directory

allyhawkins:
Just a small change to remove the extra slash in setting the vireo directory.


jashapiro (approved):
I submitted the same thing, but should have checked my email first!

PR #268, opened 2023-02-07
Add stub to integration workflow

allyhawkins:
Stacked on #240 
Here I'm just adding a stub to the first process that's part of the integration workflow being added in #240. In the future we should be sure to include a stub in each of our processes for quick testing.


jashapiro (approved):
Well, that is pretty easy... Easier than all the suggestions that I made earlier!

Issue #237, opened 2022-11-15
Add process for integrating using fastMNN


allyhawkins:
After initiating the workflow for data integration and creating the merging step to merge the objects, we will want to perform integration. We should include a process that integrates using [`fastMNN()`](https://rdrr.io/github/LTLA/batchelor/man/fastMNN.html) from the `batchelor` package. The input to this process will be the merged `SingleCellExperiment` object containing all libraries being integrated and a column with the associated library ID for each cell. 

We will have to think about how we want to return the results from integration. We are interested specifically in the PCA results calculated from the corrected gene expression. We could grab just the PCA results and add as another assay in the `reducedDim` slot of the merged object with the name `fastmnn_PCA` rather than returning a separate integrated object for each method of integration.

allyhawkins:
Closed by #274

PR #274, opened 2023-02-10
Integrate SCE's using fastMNN

### LINKED ISSUE #247: Create a script for integrating SCE objects
Before we can add a process to the integration workflow started in #240 for integration with `harmony` and integration with `fastMNN`, we will probably want to have a script that can be used for integration that is run by each of those processes. 

One option is to have a script for each process, so one script that will integrate using `harmony` and then another separate script to integrate using `fastMNN`. However, I think a lot of steps might be repeated and we might be able to combine into one script with an integration method as the input. Provided with an integration method, the script would perform either `harmony` or `fastMNN` and then output the integrated object. 

Another alternative is to allow for multiple integration methods as the input and then perform integration for both methods with the same process in the workflow. But I think we want to keep these processes separate as it would be easier to remove one process or not perform one type of integration in the future. Either way, we could create one script that can perform both integration methods and takes as input a merged SCE object and an integration method to perform and use that same script in two processes.

allyhawkins:
Closes #247 and #237 

Here I added a script to be used for integrating a merged SCE object using either `fastMNN` or `harmony`. I tried to keep things pretty simple in terms of options and only have the input, output, and method as required options. I also included the harmony covariates as an option in case we ever do want to be able to incorporate that, we can then add a param to nextflow to do that. 

As far as the actual integration, I used the `scpcaTools::integrate_sces()` function that we have and incorporated the minor adjustment to add in a variable for the highly variable genes (https://github.com/AlexsLemonade/scpcaTools/pull/163). 

The other thing I did here was add a process to integrate using `fastMNN`, and the reason for that was so we could start seeing how this script would actually be used in the workflow. I wanted to get some feedback on some decisions I made in doing that: 

- The process that I added, and the script, take as input a merged SCE object and return a merged SCE object with corrected PCA and UMAP results added to the object. This means that each integration method will take the same input and then each produce an object that has the original merged counts and the new corrected PCA/UMAP. I think we want to keep these integration methods completely separate so that they each return their own unique objects and then in the last step (the creation of the report), we would read them both in and grab the results we need from each. What I'm leaning towards is not publishing the output from the `fastMNN` or `harmony`, but reading that output into a final step that will merge all of the results together to create one object that gets published. 
- I made the decision to never return the corrected expression results and therefore did not add an argument for it that could be used from the workflow. I don't really know why we would need them and don't think we want to make that an option. 

The last thing I did here was set the names of the `input_sce_files` list that was being created in the merging script. I noticed during testing that the output from merging contained numbers rather than library ids in the `library_id` column of the `colData`.

### DIFF
--- a/bin/integrate_sce.R
+++ b/bin/integrate_sce.R
@@ -29,7 +29,7 @@ option_list <- list(
     opt_str = c("--method"),
     type = "character",
     default = NULL,
-    help = "Integration method to use, either `fastMNN` or `harmony` (case-insensitive)."
+    help = "Integration method to use, either `fastMNN` or `harmony` (case-sensitive)."
   ),
   make_option(
     opt_str = c("--harmony_covariate_cols"),
@@ -74,6 +74,12 @@ if(is.null(opt$input_sce_file)) {
   }
 }
 
+# set up multiprocessing params
+if(opt$threads > 1){
+  bp_param = BiocParallel::MulticoreParam(opt$threads)
+} else {
+  bp_param = BiocParallel::SerialParam()
+}
 
 # Read in SCE file -------------------------------------------------------------
 merged_sce <- readr::read_rds(opt$input_sce_file)
@@ -88,14 +94,34 @@ merged_hvgs <- metadata(merged_sce)$merged_hvgs
 
 # Integration ------------------------------------------------------------------
 
+# create list of arguments for either integration method
+integration_args <- list(
+  sce = merged_sce,
+  integration_method = opt$method,
+  batch_column = "library_id",
+  seed = opt$seed,
+  return_corrected_expression = FALSE
+)
+
+# append fastMNN only args
+if(opt$method == "fastMNN"){
+  integration_args <- append(integration_args,
+                             list(
+                               subset.row = merged_hvgs,
+                               auto.merge = TRUE,
+                               BPPARAM = bp_param
+                             ))
+}
+
+# append harmony only args
+if(opt$method == "harmony"){
+  integration_arts <- append(integration_args,
+                             covariate_cols = opt$harmony_covariate_cols)
+}
+
 # perform integration
-integrated_sce <- scpcaTools::integrate_sces(merged_sce,
-                                             integration_method = opt$method,
-                                             batch_column = "library_id",
-                                             covariate_cols = opt$harmony_covariate_cols,
-                                             hv_genes = merged_hvgs,
-                                             return_corrected_expression = FALSE,
-                                             seed = opt$seed)
+integrated_sce <- do.call(scpcaTools::integrate_sces, integration_args)
+
 # calculate UMAP from corrected PCA
 integrated_sce <- integrated_sce |>
   scater::runUMAP(dimred = glue::glue("{opt$method}_PCA"),
--- a/bin/integrate_sce.R
+++ b/bin/integrate_sce.R
@@ -1,0 +1,118 @@
+# Script used to perform integration on a given merged SCE object using R-based methods
+#
+# A merged SCE file in RDS format is read in. Integration is performed with either
+# `fastMNN` or `harmony`. The integrated SCE object is saved as an RDS
+# file in the provided output file.
+#
+
+# import libraries
+library(optparse)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-i", "--input_sce_file"),
+    type = "character",
+    default = NULL,
+    help = "Path to RDS file that contains the merged SCE object to integrate"
+  ),
+  make_option(
+    opt_str = c("-o", "--output_sce_file"),
+    type = "character",
+    default = NULL,
+    help = "Path to RDS file where the integrated SCE object will be saved"
+  ),
+  make_option(
+    opt_str = c("--method"),
+    type = "character",
+    default = NULL,
+    help = "Integration method to use, either `fastMNN` or `harmony` (case-insensitive)."
+  ),
+  make_option(
+    opt_str = c("--seed"),
+    type = "integer",
+    default = NULL,
+    help = "random seed to set during integration"
+  ),
+  make_option(
+    opt_str = c("--harmony_covariate_cols"),
+    type = "character",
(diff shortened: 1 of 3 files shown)


bin/integrate_sce.R
```diff
@@ -0,0 +1,104 @@
...
+                                             integration_method = opt$method,
+                                             batch_column = "library_id",
+                                             covariate_cols = opt$harmony_covariate_cols,
+                                             hv_genes = merged_hvgs,
+                                             return_corrected_expression = FALSE,
+                                             seed = opt$seed)
+# calculate UMAP from corrected PCA
+integrated_sce <- integrated_sce |>
+  scater::runUMAP(dimred = glue::glue("{integration_method}_PCA"),
+                  name = glue::glue("{integration_method}_UMAP"))
```
jashapiro, line 101: I think
```suggestion
  scater::runUMAP(dimred = glue::glue("{opt$method}_PCA"),
                  name = glue::glue("{opt$method}_UMAP"))
```

bin/integrate_sce.R
```diff
@@ -0,0 +1,104 @@
...
+    opt_str = c("-o", "--output_sce_file"),
+    type = "character",
+    default = NULL,
+    help = "Path to RDS file where the integrated SCE object will be saved"
+  ),
+  make_option(
+    opt_str = c("--method"),
+    type = "character",
+    default = NULL,
+    help = "Integration method to use, either `fastMNN` or `harmony` (case-insensitive)."
```
jashapiro, line 32: Right now this is case-sensitive. Do we want to keep it that way? 

Checking either is easy, but we will want to standardize within the script after initial parsing, for the storage of `fastMNN_UMAP` etc. The easy way is to make this case sensitive and leave things as they are...

bin/merge_sces.R
```diff
@@ -63,11 +63,14 @@ if(is.null(opt$input_library_ids)){
   # extract library ids from filename
   input_library_ids <- stringr::word(basename(input_sce_files), 1, sep = "_")
 } else {
-  # pull library ids from list 
-  input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ',')) 
+  # pull library ids from list
+  input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ','))
```
jashapiro, line 67: We could save a step here, since we don't use the library_ids any other way (right?)
```suggestion
  names(input_sce_files) <- unlist(stringr::str_split(opt$input_library_ids, ','))
```

jashapiro (commented):
This looks pretty good, but I had a few fixes here and there.

My `do.call()` suggestion below is growing on me. I like the idea that we can leave `scpcaTools` alone if we do things that way. But I may be going down a bad route there.

bin/integrate_sce.R
allyhawkins: oh I meant to delete that. I copied some of this from the integration script we had in `sc-data-integration`, but the functions we have in `scpcaTools` are actually case sensitive so I think we want to just keep it that way throughout.

allyhawkins:
Thanks for the helpful suggestions! I actually don't mind the `do.call` option and like that we would be able to customize each of the options for fastMNN vs. harmony and change them in the future much easier without having to alter `scpcaTools`. Let me know what you think of the new changes and I can close https://github.com/AlexsLemonade/scpcaTools/pull/163 if we like this approach instead.

bin/merge_sces.R
```diff
@@ -50,24 +50,24 @@ if(!(stringr::str_ends(opt$output_sce_file, ".rds"))){
...
 if(is.null(opt$input_library_ids)){
   # extract library ids from filename
-  input_library_ids <- stringr::word(basename(input_sce_files), 1, sep = "_")
+  names(input_sce_files) <- stringr::word(basename(input_sce_files), 1, sep = "_")
 } else {
-  # pull library ids from list 
-  input_library_ids <- unlist(stringr::str_split(opt$input_library_ids, ',')) 
+  # pull library ids from list
+  names(input_sce_files) <- unlist(stringr::str_split(opt$input_library_ids, ','))
 }
```
jashapiro, line 69: Maybe sort here?

```suggestion

# sort inputs by library id
input_sce_files <- input_sce_files[order(names(input_sce_files)]

```

jashapiro (commented):
If you are happy with `do.call`, then I am too. 

My remaining thought is about integration order, where I do think we will want a couple of options. To my mind they are: input order, random, and auto. While I could imagine we might also want the ability to specify merge order manually, that seems like something that could wait. For the first two (and maybe all) to be stable to the vaguaries of nextflow though, I think we will want to sort as part of the merge script (maybe as an default option with a default true?).

bin/integrate_sce.R
```diff
@@ -0,0 +1,131 @@
...
+  seed = opt$seed,
+  return_corrected_expression = FALSE
+)
+
+# append fastMNN only args
+if(opt$method == "fastMNN"){
+  integration_args <- append(integration_args,
+                             list(
+                               subset.row = merged_hvgs,
+                               auto.merge = TRUE,
```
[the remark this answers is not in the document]
allyhawkins: Is this something you think I should do in this PR or file another PR to edit the merge script and work on that?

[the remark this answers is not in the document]
allyhawkins: Oh I see your comment below now, ignore this question.

allyhawkins:
> My remaining thought is about integration order, where I do think we will want a couple of options. To my mind they are: input order, random, and auto. While I could imagine we might also want the ability to specify merge order manually, that seems like something that could wait. For the first two (and maybe all) to be stable to the vaguaries of nextflow though, I think we will want to sort as part of the merge script (maybe as an default option with a default true?).

Okay I incorporated your remaining comments and included sorting in the merge script. What I did not do yet is make the `auto.merge` a part of the script arguments yet. I think what I would want to do is making a single `merge_order` argument for the script where we include options like input, random, and auto. And then based on that argument we set the `auto.merge`. I can do this in a separate PR though and file an issue if you are on board with this plan?

jashapiro (approved):
LGTM!

Issue #270, opened 2023-02-07
Add process for "training" cell type reference models


allyhawkins:
We would like to start working towards adding cell type annotations to our datasets, starting with using cell type annotations to evaluate integration. In order to pilot this, we should include two additional processes - one to "train" the model using a specified reference dataset and one to classify cell types using the trained model. 

In doing this we will need to add a column to the metadata file specifying which reference or group of references to use for each library. The "training" process should then run on all references present in the metadata file, only training once per reference. The output of the process should be the model that will be used as input to a separate process for classifying cell types.

To start I think we would just add a single column to the `scpca-integration-metadata` that contains the single reference to use, but we probably want to be able to use multiple references. To do that we could instead add a column for the group of references to use (e.g., all blood references) and then have a separate file that contains the mapping between the group name and the list of references.

allyhawkins:
After discussion in our single-cell planning meeting, we would like to include the cell type annotation as another module within `scpca-nf` rather than testing it out within data integration. This process should be added to `scpca-nf`, so we will need to think of how we want to specify which reference to be used for each library. One thought I had was to have a separate table that includes the project ID and the associated reference(s) to use.

allyhawkins:
Some more discussion on this today led to the decision of adding this process to the `build-index.nf` rather than the workflow itself. That should train all reference datasets for all celldex references that we have and do it once. Those trained models will then be used as input to the classification that will be added to `scpca-nf`.

allyhawkins:
Closed by #284

PR #280, opened 2023-02-16
Script for building SingleR model

allyhawkins:
Partially addresses #270 

Here I started to add in a step to the `scpca-nf` workflow for building and training the models for the reference datasets to use when classifying cell types with `SingleR`. I am just starting by adding in a script that will take as input the reference data, stored as a `SummarizedExperiment` in a rds file and output a rds file with a list of two models, one containing a model using the `label.fine` and one containing `label.main`. This script will then be executed within a process in the `scpca-nf` workflow for creating the models, where all references for the specified project(s) or samples from a project being processed will go through the training step. 

A few comments on decisions I made here: 

- Right now, and for the forseeable future we are only using the `celldex` references, so because of that I didn't make an option to specify which labels to build the model with. I am using both labels each time. This will not work if the reference is not `celldex`, because the labels will be different. One approach is to make the labels to use an argument. Rather than make an argument, I chose to make models with both types of labels for now for testing purposes, but if others think we should change it I'm open to trying it. 
- The other thing I had to incorporate here was reading in the tx2gene reference file we use during the mapping step. The reason for that is I think we want to provide the genes to restrict when creating the model, and all of the SCE objects will contain the genes in this file. Alternatively we could just not restrict the genes? 

Once we have this script in, we can then add a new process to the workflow to run this script on each reference file. We will also need to create a metadata file that has the project id's and references for each project before we do that. I started to play around with this, but it was getting kind of tricky so I wanted to get some feedback just on the building the model part first before getting to the nextflow part. The other thought I had was for development purposes to keep the cell type module separate right now so testing and development is much easier and faster. Then once we feel confident about that we can write the code to link `main.nf` to the cell type module.

### DIFF
--- a/bin/train_SingleR.R
+++ b/bin/train_SingleR.R
@@ -24,7 +24,7 @@ option_list <- list(
   make_option(
     opt_str = c("--output_file"),
     type = "character",
-    help = "path to output rds file to store trained model for reference dataset"
+    help = "path to output rds file to store trained model for reference dataset."
   ),
   make_option(
     opt_str = c("--fry_tx2gene"),
@@ -41,13 +41,11 @@ option_list <- list(
     opt_str = c("-t", "--threads"),
     type = "integer",
     default = 1,
-    help = "Number of multiprocessing threads to use"
+    help = "Number of multiprocessing threads to use."
   )
 )
 
 opt <- parse_args(OptionParser(option_list = option_list))
-
-opt$ref_file <- "/Users/allyhawkins/Documents/ALSF/git_repos/ScPCA-admin/sample_info/celltype_annotation/references/celldex-blueprint_encode.rds"
 
 # Set up -----------------------------------------------------------------------
 
@@ -84,6 +82,11 @@ tx2gene <- readr::read_tsv(opt$fry_tx2gene,
 
 # select genes to use for model restriction
 gene_ids <- unique(tx2gene$gene)
+
+# check that genes aren't empty
+if (length(gene_ids) == 0){
+  stop("Provided tx2gene tsv file does not contain any genes.")
+}
 
 # Train models -----------------------------------------------------------------
 
--- a/bin/train_SingleR.R
+++ b/bin/train_SingleR.R
@@ -1,0 +1,106 @@
+#!/usr/bin/env Rscript
+
+# This script is used to build and train a model to be used for SingleR. 
+# The expected input is a SummarizedExperiment containing a reference dataset and a
+# 3-column, transcript to gene tsv file (as used with Alevin-fry), to provide the expected
+# geneset to be used for training the reference dataset. 
+
+# import libraries
+suppressPackageStartupMessages({
+  library(optparse)
+  library(SingleCellExperiment)
+})
+
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = c("--ref_file"),
+    type = "character",
+    help = "path to rds file with input sce object to be processed"
+  ),
+  make_option(
+    opt_str = c("--output_file"),
+    type = "character",
+    help = "path to output rds file to store processed sce object. Must end in .rds"
+  ),
+  make_option(
+    opt_str = c("--fry_tx2gene"),
+    type = "character",
+    help = "path to "
+  ),
+  make_option(
+    opt_str = c("-r", "--random_seed"),
+    type = "integer",
+    help = "A random seed for reproducibility."
+  ),
+  make_option(
+    opt_str = c("-t", "--threads"),
+    type = "integer",
+    default = 1,
+    help = "Number of multiprocessing threads to use"
+  )
+)
+
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# Set up -----------------------------------------------------------------------
+
+# set seed
+set.seed(opt$random_seed)
+
+# check that input files exist
+if(!file.exists(opt$ref_file)){
+  stop("Missing input file with cell type reference.")
+}
+
+if(!file.exists(opt$fry_tx2gene)){
+  stop("Missing `fry_tx2gene` file.")
+}
+
+# set up multiprocessing params
+if(opt$threads > 1){
+  bp_param = BiocParallel::MulticoreParam(opt$threads)
+} else {
+  bp_param = BiocParallel::SerialParam()
+}
+
+# read in model 
+ref_data <- readr::read_rds(opt$ref_file)
+
+# read in tx2gene 
+tx2gene <- readr::read_tsv(opt$fry_tx2gene,
+                           col_names = c("transcript", "gene", "transcript_type"))
+
+# select genes to use for model restriction 
+gene_ids <- tx2gene |>
+  dplyr::select(gene) |>
+  unique()
+
+# Train models -----------------------------------------------------------------
+
+# train using fine labels
+fine_model <- SingleR::trainSingleR(
+  ref_data,
+  labels = ref_data$label.fine,
+  genes = "de", 
+  # only use genes found in index 
+  restrict = gene_ids,
+  BPPARAM = bp_param
+) 
+
+# train using main labels 
+main_model <- SingleR::trainSingleR(
+  ref_data,
+  labels = ref_data$label.main,
+  genes = "de", 
+  # only use genes found in index 
+  restrict = gene_ids,
+  BPPARAM = bp_param
+) 
+
+# combine into one object 
+all_models <- list(fine = fine_model,
+     main = main_model)
+
+# export models
+readr::write_rds(all_models, opt$output_file)
--- a/bin/train_SingleR.R
+++ b/bin/train_SingleR.R
@@ -19,7 +19,8 @@ option_list <- list(
     opt_str = c("--ref_file"),
     type = "character",
     help = "path to rds file with reference dataset to use for cell type annotation.
-      These reference datasets must contain annotations labeld with `label.fine` and `label.main`."
+      These reference datasets must contain annotations labeld with 
+      `label.main`, `label.fine`, and `label.ont`."
   ),
   make_option(
     opt_str = c("--output_file"),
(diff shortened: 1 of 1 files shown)


bin/train_SingleR.R
```diff
@@ -0,0 +1,106 @@
...
+  labels = ref_data$label.main,
+  genes = "de", 
+  # only use genes found in index 
+  restrict = gene_ids,
+  BPPARAM = bp_param
+) 
+
+# combine into one object 
+all_models <- list(fine = fine_model,
+     main = main_model)
```
sjspielman, line 103: I think spacing is funky here? I might expect `main` to have ended up right under `fine`.

bin/train_SingleR.R
```diff
@@ -0,0 +1,106 @@
...
+  library(optparse)
+  library(SingleCellExperiment)
+})
+
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = c("--ref_file"),
+    type = "character",
+    help = "path to rds file with input sce object to be processed"
```
sjspielman, line 19:
```suggestion
    help = "path to rds file with input sce object to build a SingleR model from"
```

bin/train_SingleR.R
```diff
@@ -0,0 +1,106 @@
...
+  ),
+  make_option(
+    opt_str = c("--output_file"),
+    type = "character",
+    help = "path to output rds file to store processed sce object. Must end in .rds"
```
sjspielman, line 24: Up to you

```suggestion
    help = "path to output rds file to store processed sce object. Must have `.rds` extension."
```

bin/train_SingleR.R
```diff
@@ -0,0 +1,106 @@
...
+  ),
+  make_option(
+    opt_str = c("--fry_tx2gene"),
+    type = "character",
+    help = "path to "
```
sjspielman, line 29: where?

bin/train_SingleR.R
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env Rscript
+
+# This script is used to build and train a model to be used for SingleR. 
+# The expected input is a SummarizedExperiment containing a reference dataset and a
+# 3-column, transcript to gene tsv file (as used with Alevin-fry), to provide the expected
```
sjspielman, line 5: Is this an internal `scpca-nf` format, or `Alevin-fry` format? My question really is whether it's worth explicitly stating what those 3 columns are.

bin/train_SingleR.R
```diff
@@ -0,0 +1,106 @@
...
+ref_data <- readr::read_rds(opt$ref_file)
+
+# read in tx2gene 
+tx2gene <- readr::read_tsv(opt$fry_tx2gene,
+                           col_names = c("transcript", "gene", "transcript_type"))
+
+# select genes to use for model restriction 
+gene_ids <- tx2gene |>
+  dplyr::select(gene) |>
+  unique()
```
sjspielman, line 77: This is a pretty simple "pipeline" so could be base R, but I don't feel strongly

```suggestion
gene_ids <- unique(tx2gene$gene)
```

bin/train_SingleR.R
```diff
@@ -0,0 +1,106 @@
...
+
+# Train models -----------------------------------------------------------------
+
+# train using fine labels
+fine_model <- SingleR::trainSingleR(
+  ref_data,
+  labels = ref_data$label.fine,
+  genes = "de",
```
sjspielman, line 85: Hm, this could be a function argument for other use cases?

bin/train_SingleR.R
sjspielman: We might expand this to state that it must have `label.fine` and `label.main`.
I'd also add some quick checks that those columns exist before training the model.

sjspielman (commented):
Looks about right to me, left some suggestions!

bin/train_SingleR.R
allyhawkins: My understanding based on the documentation is that this is the only option we can use with our specified input. Otherwise we would need to provide a list of genes as input, which we are not going to be doing here. 
https://rdrr.io/github/LTLA/SingleR/man/trainSingleR.html

allyhawkins:
Thanks for taking a look at this, I implemented your comments and added a check for the labels being in the `colData` of the ref data that we are reading in. Also apologies because I hadn't fully filled out the help messages the first time around, but I fixed that this time. This should be ready for another round of review.

bin/train_SingleR.R
```diff
@@ -0,0 +1,115 @@
...
+  make_option(
+    opt_str = c("--ref_file"),
+    type = "character",
+    help = "path to rds file with reference dataset to use for cell type annotation.
+      These reference datasets must contain annotations labeld with `label.fine` and `label.main`."
+  ),
+  make_option(
+    opt_str = c("--output_file"),
+    type = "character",
+    help = "path to output rds file to store trained model for reference dataset"
```
sjspielman, line 27:
```suggestion
    help = "path to output rds file to store trained model for reference dataset."
```

bin/train_SingleR.R
```diff
@@ -0,0 +1,115 @@
...
+  make_option(
+    opt_str = c("-r", "--random_seed"),
+    type = "integer",
+    help = "A random seed for reproducibility."
+  ),
+  make_option(
+    opt_str = c("-t", "--threads"),
+    type = "integer",
+    default = 1,
+    help = "Number of multiprocessing threads to use"
```
sjspielman, line 44:
```suggestion
    help = "Number of multiprocessing threads to use."
```

bin/train_SingleR.R
```diff
@@ -0,0 +1,115 @@
...
+  )
+)
+
+opt <- parse_args(OptionParser(option_list = option_list))
+
+opt$ref_file <- "/Users/allyhawkins/Documents/ALSF/git_repos/ScPCA-admin/sample_info/celltype_annotation/references/celldex-blueprint_encode.rds"
```
sjspielman, line 50: Ha, I'm no stranger to this kind of local development bug, glad to see I'm not alone 😉

bin/train_SingleR.R
```diff
@@ -0,0 +1,115 @@
...
+
+# Set up -----------------------------------------------------------------------
+
+# set seed
+set.seed(opt$random_seed)
+
+# check that input files exist
+if(!file.exists(opt$ref_file)){
+  stop("Missing input file with cell type reference.")
```
sjspielman, line 59:
```suggestion
  stop("Missing input file with cell type reference (`ref_file`).")
```

bin/train_SingleR.R
```diff
@@ -0,0 +1,115 @@
...
+if(!all(c("label.main", "label.fine") %in% colnames(colData(ref_data)))){
+  stop("Reference dataset must contain `label.main` and `label.fine` in `colData`.")
+}
+
+# read in tx2gene
+tx2gene <- readr::read_tsv(opt$fry_tx2gene,
+                           col_names = c("transcript", "gene", "transcript_type"))
+
+# select genes to use for model restriction
+gene_ids <- unique(tx2gene$gene)
```
sjspielman, line 86: Possibly overkill, but maybe we should check this !=0? Or warn if it's below a certain number?

sjspielman (approved):
Looks good to me!

jashapiro (commented):
This looks good to me, but as discussed in our meeting, I think we should add `label.ont` to the training here. Which makes three trainings, which passes my threshold for converting to `purrr` or the like...

So I would do something like:
```
label_cols <- c("label.main", "label.fine", "label.ont")
names(label_cols) <- label_cols

models <- label_cols |>
  purrr::map(\(col) dplyr::pull(ref_data, var = col) |>
  purrr::map(
  \(labels) {SingleR::trainSingleR(
    ref_data,
    labels = labels,
    genes = "de",
    # only use genes found in index
    restrict = gene_ids,
    BPPARAM = bp_param
  })
```
 

A nice side benefit is that if we do need to add support for other label columns from other data sources, we only have to change the input label column list. Well, for this section anyway.

allyhawkins:
I made a few updates to this to incorporate using the `label.ont` as we discussed yesterday and mentioned in https://github.com/AlexsLemonade/scpca-nf/pull/280#pullrequestreview-1308002408. It would be great if reviewers could take one more look since I made those changes.

bin/train_SingleR.R
```diff
@@ -0,0 +1,108 @@
...
+
+models <- label_cols |>
+ purrr::map(\(labels) {SingleR::trainSingleR(
+                   ref_data,
+                   labels = ref_data[[labels]],
+                   genes = "de",
+                   # only use genes found in index
+                   restrict = gene_ids,
+                   BPPARAM = bp_param
+                 )})
```
jashapiro, line 105: Mostly formatting, but also a variable renaming.
```suggestion
models <- label_cols |>
 purrr::map(\(label_col) {
   SingleR::trainSingleR(
     ref_data,
     labels = ref_data[[label_col]],
     genes = "de",
     # only use genes found in index
     restrict = gene_ids,
     BPPARAM = bp_param    
   )})
```

bin/train_SingleR.R
```diff
@@ -0,0 +1,108 @@
...
+
+# Train models -----------------------------------------------------------------
+
+label_cols <- c("label.main", "label.fine", "label.ont")
+names(label_cols) <- label_cols
+
+models <- label_cols |>
+ purrr::map(\(labels) {SingleR::trainSingleR(
+                   ref_data,
+                   labels = ref_data[[labels]],
```
jashapiro, line 100: I'm assuming this line is correct, though it is quite unintuitive! 

I might substitute in `colData(ref_data)[, label_col]` to be more explicit?

jashapiro (approved):
This looks good; I only had a couple of minor thoughts.

bin/train_SingleR.R
allyhawkins: Yes I tested it to make sure it works, but I can change it to be more explicit.

PR #282, opened 2023-02-21
Initiate integration report

### LINKED ISSUE #272: Initiate summary report for integration workflow
After completing integration for a given dataset, we will want to have some sort of summarized report or way to evaluate if integration was "successful". To start, we should create a template report that reads in the merged and integrated data and plots a before and after integration UMAP, coloring by original sample. 

We might also consider adding a table or stacked bar chart showing the total number of cells in each sample and printing out some basic information about the integrated dataset like number of samples.

allyhawkins:
Closes #272 

Here I started the initial steps for adding in the integration report to the integration workflow. I kept the actual report itself very simple and just am creating a UMAP colored by batches for now. I think we could also add in a table of number of cells per batch and then the UMAPs by cell type, but since we don't actually have cell types implemented yet I decided to hold off. The main thing I want to figure out here is how exactly we are going to read in all of the integrated objects and merged objects, combine everything, and create the report. 

What I chose to do here is directly read the merged sce, fastmnn sce, and harmony sce into the notebook. I also don't need to actually keep the integrated objects around, I just grab the dimension reductions from post integration and then add those to the merged SCE object. At the end of the report I'm saving the merged SCE object with the integrated results now being added in and that is getting published to the results directory. Technically we could have a separate script that does this and then the report could directly read in the merged object that already contains all the results. I could be convinced either way. 

The other thing I chose to do here was directly render the report in the Nextflow process, rather than writing a script to render it. I did think about creating a script that first does the data wrangling and exports the SCE object and then renders the template, but I had already spent a lot of time on this and wanted to get others opinions before I did that to see if it was even necessary. 

I did have a question on the best place to store the `.Rmd` files? I think we probably want to create a separate folder for templates? Any thoughts on naming that folder or do we want to move it to `bin` instead? 

Also you'll notice that I have two consecutive `.combine` steps to get a tuple containing the integration group, merged SCE file, fastmnn SCE file, and harmony SCE file. I tried to pass the merged SCE file as output from both integration processes and then just have one `.combine` step using `by: [0:1]`, but Nextflow kept skipping the process, making me think it didn't like passing the merged SCE file as input and output? 

I'm still doing some testing and debugging to get this to run to completion, but I wanted to get this up.

### DIFF
--- a/bin/integration-report.Rmd
+++ b/bin/integration-report.Rmd
@@ -4,6 +4,7 @@ params:
   merged_sce: NULL
   fastmnn_sce: NULL
   harmony_sce: NULL
+  output_file: "example_group.rds"
   batch_column: "batch"
 
 title: "`r glue::glue('Integration report for {params$integration_group}')`"
@@ -108,6 +109,10 @@ batch_umaps <- purrr::map(umap_names, \(name){
 
 patchwork::wrap_plots(batch_umaps, ncol = 2,
                       guides = "collect")
+```
+
+```{r}
+readr::write_rds(merged_sce, params$output_file)
 ```
 
 ```{r}
--- a/bin/integration-report.Rmd
+++ b/bin/integration-report.Rmd
@@ -1,0 +1,116 @@
+---
+params:
+  integration_group: "example_group"
+  merged_sce: NULL
+  fastmnn_sce: NULL
+  harmony_sce: NULL
+  batch_column: "batch"
+
+title: "`r glue::glue('Integration report for {params$integration_group}')`"
+author: "CCDL"
+date: "`r params$date`"
+output:
+  html_document:
+    toc: true
+    toc_depth: 3
+    toc_float: true
+---
+
+```{r setup, message = FALSE, echo = FALSE}
+# knitr options
+knitr::opts_chunk$set(
+  echo = FALSE
+)
+
+library(SingleCellExperiment)
+library(ggplot2)
+
+# Set default ggplot theme
+theme_set(theme_bw() + 
+          theme(plot.margin = margin(rep(20, 4))))
+```
+
+```{r}
+# check that input files are provided 
+if(is.null(params$merged_sce)){
+  stop("Must provide a path to a RDS file containing a merged SCE object.")
+}
+
+if(is.null(params$fastmnn_sce)){
+  stop("Must provide a path to a RDS file containing a SCE object post integration with fastMNN.")
+}
+
+if(is.null(params$harmony_sce)){
+  stop("Must provide a path to a RDS file containing a SCE object post integration with harmony.")
+}
+
+# define batch column 
+batch_column <- params$batch_column
+```
+
+
+```{r}
+integrated_sce_files <- list(fastmnn = params$fastmnn_sce,
+                             harmony = params$harmony_sce)
+
+# only load in the embeddings since we don't need to store the entire object 
+integrated_embeddings <- purrr::map(integrated_sce_files, readr::read_rds) |>
+  purrr::map(reducedDims) |> 
+  # select only those that have a method and then dimreduction name
+  purrr::map(~.x[grep("_PCA$|_UMAP$", names(.x))]) |>
+  # convert from SimpleList and then flatten
+  purrr::map(as.list) |> 
+  purrr::flatten()
+```
+
+
+```{r}
+# read in merged object
+merged_sce <- readr::read_rds(params$merged_sce)
+
+# get list of methods and dimreductions to be added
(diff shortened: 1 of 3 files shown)


jashapiro:
A few comments as you work on this:

I think I would separate out merging the integrated SCE objects into a separate process, then have the template take only a single integrated_sce.rds file which includes all of various reduced_dims. Right now we also have the two reductions happening in parallel; it may make sense to do them serially and just add each reduction to the same object: as long as there are only two steps I doubt the real time difference would be that great, and the integration functions don't need to know that there were other integrations applied previously, as long as they aren't removing anything, which they should not be. That saves us from having to do any `.combine` steps, which I think are a common source of error/confusion. For ease of use, we might want to add a metadata variable that details the integrations that were performed to make automatically accessing all of the integration results later easier & automated.

I like having the templates directory: you can see I did the same thing in #283; in case this becomes more complicated, it might make sense to put the report in its own subdirectory, but the current implementation is fine for now.

allyhawkins:
> Right now we also have the two reductions happening in parallel; it may make sense to do them serially and just add each reduction to the same object: as long as there are only two steps I doubt the real time difference would be that great, and the integration functions don't need to know that there were other integrations applied previously, as long as they aren't removing anything, which they should not be. That saves us from having to do any .combine steps, which I think are a common source of error/confusion. 

Do you mean have the output from fastmnn be the input of harmony? And then the result is one single object with merged counts and dim reductions from two different integrations. If we do that then we wouldn't need to have a separate script since everything would already be merged into one object.

jashapiro:
> Do you mean have the output from fastmnn be the input of harmony? And then the result is one single object with merged counts and dim reductions from two different integrations. If we do that then we wouldn't need to have a separate script since everything would already be merged into one object.

Yes, that is what I mean, and yes, it would save a script! (This is the result of me not looking at what I wrote first when I thought about something a few minutes later...)

allyhawkins:
Okay I like this plan! I'm going to make those changes.

allyhawkins:
I'm going to go ahead and request review on this because I've made the changes to pass the same object through to both integration processes now and removed the extra steps of putting all the results into one object and writing that out from the report. Everything seems to work now, and I've attached a copy of the rendered report. I think the actual publishing step is still a little wonky, but it is writing the output. 

The other thing I didn't do is incorporate a variable for the integration methods yet. I wasn't quite sure what was meant by doing that and if we wanted to pass those in as params in the notebook directly, or store that information in the metadata of the SCE objects? 

[rhabdomyosarcoma-dyer_chen_summary_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/10809503/rhabdomyosarcoma-dyer_chen_summary_report.html.zip)

jashapiro (commented):
This looks good to me overall. I added a couple of little comments, along with what I hope is a clarification of my comment about the integration methods in metadata. If you want to implement that, it seems like this is probably good time to do it. 

Obviously there are a lot of possible updates to the report itself, but this does _not_ seem like the place to do that!

allyhawkins:
Thanks for the suggestions. I incorporated your comments, including adding the integration method to the metadata of the SCE object. I also agree that this isn't where we should be making changes to the report, because I think it's definitely in infancy stages. This is just to get things started and then we can make the report "prettier" and incorporate more plots. 

That being said, I did stop printing out the color messages and added in the session info suggestion and it does look a little bit nicer for now. 
[acute_myeloid_leukemia-gawad_summary_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/10819159/acute_myeloid_leukemia-gawad_summary_report.html.zip)

bin/integrate_sce.R
```diff
@@ -131,5 +131,13 @@ integrated_sce <- integrated_sce |>
   scater::runUMAP(dimred = glue::glue("{opt$method}_PCA"),
                   name = glue::glue("{opt$method}_UMAP"))
 
+# add method to integrated sce
+existing_methods <- metadata(integrated_sce)$integration_methods
+if(!is.null(existing_methods)){
+  metadata(integrated_sce)$integration_methods <- c(existing_methods, opt$method)
+} else {
+  metadata(integrated_sce)$integration_methods <- opt$method
+}
```
jashapiro, line 140: I'm pretty sure that `c(NULL, something)` is equivalent to  `c(something)`, so I think you can reduce this to: 

```suggestion
existing_methods <- metadata(integrated_sce)$integration_methods
metadata(integrated_sce)$integration_methods <- c(existing_methods, opt$method)
```

jashapiro (approved):
LGTM, with only one little suggestion to save some logic.

PR #283, opened 2023-02-21
Add report template files and use them for generating reports

### LINKED ISSUE #267: Move QC report to repository (from scpcaTools)
Currently the reports we write are generated using a function in the [scpcaTools](https://github.com/AlexsLemonade/scpcaTools) package, but this makes updating the reports require a change to the package, which is a bit unwieldy, and makes for slower updates when changes to the report are required (requiring a release of  `scpcaTools`).

It _should_ be possible to move the report template from https://github.com/AlexsLemonade/scpcaTools/tree/main/inst/rmd to a directory within this project. To use this within the workflow, we would then change the report process to include as an input the _directory_ containing the rmd template file(s).  It will need to be the _directory_, as we use multi-part rmd documents, so all parts need to be included. The path to the template directory should be specified using the `${projectDir}` nextflow variable, which will point to the `scpca-nf` root directory.

We will also need to migrate the contents of `scpcaTools::generate_qc_report()` to the current [sce_qc_report.R](https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/sce_qc_report.R) script.

jashapiro:
Closes #267

Here I am adding the report template from scpcaTools to the scpca-nf repository and modifying existing code to use that report instead of the one built into the scpcaTools package.

The code here does work, but I do have some questions about where the files should go, as well as whether we should define the paths and such somewhere other than where I currently do, which is just in the call to the `sce_qc_report` process. 

Note that I pass in the full directory to the template files as the path input, which is required for all files to be added to the working directory. This does mean that the name of the main template file is hard-coded, but I could add that in as a separate argument, perhaps passing in a tuple of the directory (as a path object) and the file name as a `val`. If that sounds like a good idea, let me know!


jashapiro:
> I could add that in as a separate argument, perhaps passing in a tuple of the directory (as a path object) and the file name as a `val`. If that sounds like a good idea, let me know!

I decided to go ahead and do this in https://github.com/AlexsLemonade/scpca-nf/pull/283/commits/2c68a18291509c3083aae4d2390e8401fac64c00, because after writing it it sounded like a good idea to me.

main.nf
```diff
@@ -24,6 +24,11 @@ rna_techs = single_cell_techs.findAll{it.startsWith('10Xv')}
 citeseq_techs = single_cell_techs.findAll{it.startsWith('CITEseq')}
 cellhash_techs = single_cell_techs.findAll{it.startsWith('cellhash')}
 
+// report template path
+report_template_dir = file("${projectDir}/templates/qc_report", type: 'dir')
+report_template_file = "qc_report.rmd"
+report_template_tuple = tuple(report_template_dir, report_template_file)
```
allyhawkins, line 30: I haven't had time for a full review yet, but I did go to test this to see if this approach worked for you because defining it in the workflow did not work for me. That is, if you run using `nextflow run main.nf`. We probably do want it to work in both ways of running. When I tested it I got an error that the file doesn't exist because it's searching the project directory of where you submit it (your local directory if using `nextflow run main.nf`), and it doesn't pull the file at the beginning. If you set it as a parameter with the same path, then it should pull the file from the repo first, whether it be a local copy or from Github.

jashapiro: I think defining it as a `param` makes sense, but I want to check my understanding. Did you pull this locally to test it, or were your running it from github? I had been testing running locally, with `nextflow run main.nf` and that was working fine, but I had not yet tested running it with `nextflow run alexslemonade/scpca-nf -r jashapiro/template-report` (just started that test).

jashapiro: Okay, I just finished testing with both checking out the repo and running `nextflow run main.nf ` and separately running `nextflow run alexslemonade/scpca-nf -r jashapiro/template-report` and it seemed to work both ways. 

I know that in an earlier version I had tried to use `${projectDir}` within a process, and that failed, but here I am defining it outside the process itself, so I think it should work. The advantage of making it a param would be that we could potentially change it on the command line, but I don't tend to think that is going to be that useful to do, which I why I didn't make it a param at this point. 

But if it isn't working for you, we do need to track that down!

allyhawkins: I must have been running from a different branch because I tried to duplicate this error and it ran successfully now. I had seen that same missing file error before when I was using `${projectDir}` in the process, so I just assumed that was still the issue here, before checking the branch 🤦‍♀️

allyhawkins (approved):
After my failed test earlier that actually works as expected, everything else here looks good 👍

PR #284, opened 2023-02-27
Add process for training celldex references for SingleR

### LINKED ISSUE #270: Add process for "training" cell type reference models

allyhawkins:
Closes #270 

Here I added a process to the `build-index.nf` script to train the references for use with SingleR for cell type annotation. We probably could separate this out into another file completely... but they are all references that will be used for `scpca-nf` so if others want me to put it in its own file I can. 

The process takes as input a reference file and then runs the `train_SingleR.R` script to generate a rds file with a list of models, one for each of the labels present in the `celldex` reference. Overall it was pretty straight forward. The only question I really have for now is where to store everything on S3. Since we are still in development, I am storing everything in our private bucket that we have, rather than writing it to the public `scpca-references`, but if we are comfortable with storing the models there I can update that path. I also did test this and everything worked as expected.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -152,13 +152,13 @@ process train_singler_models {
 
 workflow {
   // generate splici and spliced cDNA reference fasta
-  //generate_reference(params.ref_fasta, params.ref_gtf, params.assembly)
+  generate_reference(params.ref_fasta, params.ref_gtf, params.assembly)
   // create index using reference fastas
-  //salmon_index(generate_reference.out.fasta_files, params.ref_fasta)
+  salmon_index(generate_reference.out.fasta_files, params.ref_fasta)
   // create cellranger index
-  //cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
+  cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
   // create star index
-  //star_index(params.ref_fasta, params.ref_gtf, params.assembly)
+  star_index(params.ref_fasta, params.ref_gtf, params.assembly)
 
 
   // create channel of cell type ref files and names
--- a/build-index.nf
+++ b/build-index.nf
@@ -125,13 +125,48 @@ process index_star{
     """
 }
 
+process train_singler_models {
+  container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.celltype_model_dir}"
+  label 'cpus_4'
+  label 'mem_16'
+  input:
+    tuple path(celltype_ref), val(ref_name)
+  output:
+    path celltype_model
+  script:
+    celltype_model = "${ref_name}_model.rds"
+    """
+    train_SingleR.R \
+      --ref_file ${celltype_ref} \
+      --output_file ${celltype_model} \
+      --seed ${params.seed} \
+      --threads ${task.cpus}
+
+    """
+
+
+}
+
 workflow {
   // generate splici and spliced cDNA reference fasta
-  generate_reference(params.ref_fasta, params.ref_gtf, params.assembly)
+  //generate_reference(params.ref_fasta, params.ref_gtf, params.assembly)
   // create index using reference fastas
-  salmon_index(generate_reference.out.fasta_files, params.ref_fasta)
+  //salmon_index(generate_reference.out.fasta_files, params.ref_fasta)
   // create cellranger index
-  cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
+  //cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
   // create star index
-  star_index(params.ref_fasta, params.ref_gtf, params.assembly)
+  //star_index(params.ref_fasta, params.ref_gtf, params.assembly)
+
+
+  // create channel of cell type ref files and names
+  celltype_refs_ch = Channel.fromPath(params.celltype_refs_metafile)
+    .splitCsv(header: true, sep: '\t')
+    .map{[
+      celltype_ref_file = "${params.celltype_ref_dir}/${it.filename}",
+      ref_name = it.reference
+    ]}
+
+  // train cell type references using SingleR
+  train_singler_models(celltype_refs_ch)
 }
--- a/build-index.nf
+++ b/build-index.nf
@@ -132,6 +132,7 @@ process train_singler_models {
   label 'mem_16'
   input:
     tuple path(celltype_ref), val(ref_name)
+    path tx2gene
   output:
     path celltype_model
   script:
@@ -140,6 +141,7 @@ process train_singler_models {
     train_SingleR.R \
       --ref_file ${celltype_ref} \
       --output_file ${celltype_model} \
+      --fry_tx2gene ${tx2gene} \
       --seed ${params.seed} \
       --threads ${task.cpus}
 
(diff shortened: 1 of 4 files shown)


config/reference_paths.config
```diff
@@ -19,3 +19,7 @@ t2g_bulk_path    = "${params.ref_dir}/annotation/Homo_sapiens.GRCh38.104.spliced
 
 // barcode files
 barcode_dir      = "${params.ref_rootdir}/barcodes/10X"
+
+// cell type references
+celltype_ref_dir = 's3://nextflow-ccdl-data/reference/celldex'
```
jashapiro, line 24: We might want to move these to something more general, as we may get reference files from more than just `celldex`.

In reference to your overall comment, I think putting these in the main reference S3 bucket is fine, and I might prefer that anything in `reference_paths.config` be public. If we want to keep these private for now, I would put the definition in the `profile_ccdl.config` directory as a way of indicating they are not publicly accessible.

jashapiro (commented):
This looks good, but I do think we should make the singleR "training" workflow a separate file, as we are likely to want to run it more often/separately from the main index building.

We could still have it run in the main index building workflow by using an `include` statement and referencing a named sub-workflow (see https://www.nextflow.io/docs/latest/dsl2.html#workflow-definition).

I also had an idea where the SingleR reference building would run on the fly as needed, kind of like how I made the cellhashing references work, but I pretty sure that is not worth the effort; better to update the reference more manually and intentionally for now.

build-index.nf
```diff
@@ -134,4 +159,16 @@ workflow {
   cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
   // create star index
   star_index(params.ref_fasta, params.ref_gtf, params.assembly)
```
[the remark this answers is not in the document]
jashapiro: IT does look like that is a mistake, yes!

allyhawkins:
Thanks for the comments @jashapiro. I went ahead and pulled out the process for building the celltype refs to it's own file where I defined the process and a workflow. I put this into a separate module and added it to the folder. Is this what you had in mind? 

My other thought was making it 100 percent independent from the `build-index.nf` and not including it there. 

The other thing I did was update the paths to the references to have them be stored in the public bucket. Let me know if you like the new paths and if so, we need to move the reference files from celldex initially which we can temporarily do as part of https://github.com/AlexsLemonade/ScPCA-admin/pull/491. We should probably move the download refs to `scpca-nf` sooner rather than later so that things are easier to keep in sync for this exact reason.

modules/build-celltype-ref.nf
```diff
@@ -0,0 +1,39 @@
...
+    // create channel of cell type ref files and names
+    celltype_refs_ch = celltype_refs_file
+        .map{[
+        celltype_ref_file = "${params.celltype_ref_dir}/${it.filename}",
+        ref_name = it.reference
+        ]}
+
+    // train cell type references using SingleR
+    train_singler_models(celltype_refs_ch, params.t2g_3col_path)
+}
```
jashapiro, line 39: This is close to what I was suggesting. I think we can actually make `build_celltype_ref` the whole workflow, input included, then make a separate default workflow that just calls that workflow within this document: 
```suggestion
workflow build_celltype_ref {
  // build celltype references
  celltype_refs_ch = Channel.fromPath(params.celltype_refs_metafile)
        .splitCsv(header: true, sep: '\t')
        .map{[
        celltype_ref_file = "${params.celltype_ref_dir}/${it.filename}",
        ref_name = it.reference
        ]}

    // train cell type references using SingleR
    train_singler_models(celltype_refs_ch, params.t2g_3col_path)
}

workflow{
    build_celltype_ref()
}
```

If the `build_celltype_ref` workflow does end up requiring a `take:` then I would make that just the refs metafile, but I don't _think_ it is required.


The idea behind doing it this way is that if we want to _only_ build the celltype references, we can do that with 
`nextflow run build-celltype-ref.nf` locally or [`nextflow run AlexsLemonade/scpca-nf -main-script build-celltype-ref.nf` or `nextflow run AlexsLemonade/scpca-nf -entry build-celltype-ref`] when pulling directly from github. (Either way, I would probably move this file out to the root directory, as it is a possible "entry point".)

If we want to run _all_ of the index building, then `nextflow run build-index.nf` (and variants) will still do the job. I suppose we could separate things out completely, but I

build-index.nf
```diff
@@ -133,5 +135,11 @@ workflow {
...
   cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
   // create star index
-  star_index(params.ref_fasta, params.ref_gtf, params.assembly)
+  index_star(params.ref_fasta, params.ref_gtf, params.assembly)
+
+  // build celltype references
+  celltype_refs_ch = Channel.fromPath(params.celltype_refs_metafile)
+        .splitCsv(header: true, sep: '\t')
+
+  build_celltype_ref(celltype_refs_ch)
```
jashapiro, line 144: If my previous comment works as I expect, then you should only need the reference to the other workflow here.
```suggestion
  build_celltype_ref()
```

config/reference_paths.config
```diff
@@ -19,3 +19,7 @@ t2g_bulk_path    = "${params.ref_dir}/annotation/Homo_sapiens.GRCh38.104.spliced
 
 // barcode files
 barcode_dir      = "${params.ref_rootdir}/barcodes/10X"
+
+// cell type references
+celltype_ref_dir = '${params.ref_rootdir}/celltype/references'
+celltype_model_dir = '${params.ref_rootdir}/celltype/singler_models'
```
jashapiro, line 25: These look good to me.

build-index.nf
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env nextflow
 nextflow.enable.dsl=2
 
+include { build_celltype_ref } from './modules/build-celltype-ref.nf'
```
jashapiro, line 4: Just noting I would might move the file out of the subdirectory and update this line to: 
```suggestion
include { build_celltype_ref } from './build-celltype-ref.nf'
```

jashapiro (commented):
Looks good overall, and close to what I was thinking... I made a couple suggestions to do what I was thinking, but I do wonder a bit if it is worth including the celltype references in `build-index.nf`. I think I lean toward yes, but I am not fully convinced. Either way, I don't think we want it in the `modules` directory, as we do want it to be runnable independently.

New file locations look good, so we should update the actual files to match.

I'm not approving mostly because I want to be sure my suggestions work (or know they don't) before merging.

allyhawkins:
Thanks for the suggestions @jashapiro. I moved everything over so it is entirely dependent and tested it out and it works. I also ran it from the `build-index.nf`. 

I do need to update how we build the filename based on changes to the metadata file, so once you are good with changes in https://github.com/AlexsLemonade/ScPCA-admin/pull/491, then I can update the line that creates the `celltype_ref_file` path.

jashapiro (approved):
LGTM

PR #286, opened 2023-03-09
Workflow for classifying cell types

### LINKED ISSUE #271: Process for classifying cell types
After the models have been trained using each input reference dataset (described in #270), we will need to perform cell type classification using the trained model. This should be done in a separate process that takes as input the trained model and the unannotated SCE object. The output would be the SCE object with a new `celltype` column added to the metadata. 

The main decision that we will need to make here within integration is if we want to classify once on the merged dataset or once on each individual SCE object prior to integration. The actual process itself should not change, the only thing that will be altered is where the process goes in the workflow with the Nextflow script. I'm inclined to think it will be faster if we do it once on the merged dataset because it will save time reading in and writing individual SCE objects.

allyhawkins:
Closes #271 
⚠️ Stacked on #284 

Here I'm adding a workflow to use for adding cell type annotations to the processed SCE files from `scpca-nf`. The way I approached this was first by creating an additional module stored in the `modules` folder that actually does the cell typing. The input to this is the metadata, processed SCE file, and list of references to use for annotation. I then created a separate workflow that we can use to independently run that module so that we can test cell type annotation separately from the rest of `scpca-nf`. 

I also did link it to the main `scpca-nf` though by pulling out the `meta` object and the processed SCE file from the `post_process_sce` output. This way it can be run by itself or as part of the whole workflow. 

As for the actual cell type annotation, I chose to perform annotation on all provided references and labels.  I used the naming convention we decided on in our meeting for the additions to the `colData` of `label.main-CelldexPackageName`. I did have to do a bit of weird wrangling to get the names in the correct order and format that I wanted. When I was converting back to the `colData` of the object, the `-` was getting converted back to `.` which was rather annoying, but I solved that with turning off the check for name syntax. 

I also felt like we still want to maintain a lot of the output from SingleR so we can use it in downstream analysis. So I decided to create two lists, one of all the score matrices and one of all the delta scores and store those in the metadata of the SCE object. The next step after getting this set up will be to start working on some exploratory analysis notebook, so having all of the relevant `SingleR` data I thought would be helpful.

I did test this out, but note that it uses the metadata file that is getting added as part of https://github.com/AlexsLemonade/ScPCA-admin/pull/491, so to be able to run this, I provided the path to the local copy of the metadata file after checking out the `allyhawkins/project-celltype-ref-metadata` branch in `ScPCA-admin` at the command line using the `--project_celltype_metafile` flag.

I'm going to request Josh for review, mainly because I would like his input on the Nextflow stuff, but @sjspielman if you would like to take a look for nextflow learning and also provide any input on the R script and SingleR stuff your 👀 are also welcome.


modules/classify-celltypes.nf
```diff
@@ -0,0 +1,54 @@
...
+          project_id = it.scpca_project_id,
+          singler_model = "${params.celltype_model_dir}/${it.celltype_ref_name}_model.rds"
+        ]}
+
+      // create channel with grouped meta, processed sce object, and all references to use
+      grouped_celltype_ch = processed_sce_channel
+        .map{[it[0]["project_id"]] + it}
+        .combine(celltype_ch, by: 0)
+        .map{it.drop(1)} // remove extra project ID
+        .groupTuple(by: 0) // group by meta
```
jashapiro, line 43: I have had trouble with grouping by `meta` (a complex data type); it sometimes doesn't do what I expect it to do. Most of the time it is fine, but it seems to be brittle. I think we might want to reverse the order here, and do the grouping within the celltype channel first, so each project has a list of models (where we can group by project_id), then do the `combine`.

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+# set seed
+set.seed(opt$random_seed)
+
+# check that input file file exists
+if(!file.exists(opt$input_sce_file)){
+  stop("Missing input SCE file")
+}
+
+# check that references all exist
+model_files <- unlist(stringr::str_split(opt$singler_models, ","))
```
jashapiro, line 54: We really need to update to stringr 1.5, because they finally added `str_split_1()` to save us from the tyranny of `unlist` (We are currently at 1.4.1, so nothing to do here)

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+if(!any(file.exists(model_files))){
+  missing_files <- model_files[!which(file.exists(model_files))]
+  glue::glue("
+             Missing model file: {missing_files}
```
jashapiro, line 58:
```suggestion
             Missing model file(s): {missing_files}
```

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+set.seed(opt$random_seed)
+
+# check that input file file exists
+if(!file.exists(opt$input_sce_file)){
+  stop("Missing input SCE file")
+}
+
+# check that references all exist
+model_files <- unlist(stringr::str_split(opt$singler_models, ","))
+if(!any(file.exists(model_files))){
```
jashapiro, line 55:
```suggestion
if(!all(file.exists(model_files))){
```

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+  missing_files <- model_files[!which(file.exists(model_files))]
```
jashapiro, line 56:
```suggestion
  missing_files <- model_files[which(!file.exists(model_files))]
```

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+# read in references as a list
+model_names <- stringr::str_remove(basename(model_files), "_model.rds")
+names(model_files) <- model_names
+model_list <- purrr::map(model_files, readr::read_rds) |>
+  # make sure that the labels have label type before reference dataset name
+  purrr::map2(model_names,
+              \(model, name){
+    names(model) <- glue::glue("{names(model)}-{name}")
+    model
+  }) |>
```
jashapiro, line 82: spacing
```suggestion
              \(model, name){
                names(model) <- glue::glue("{names(model)}-{name}")
                model
              }) |>
```

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+}
+
+# read in input rds file
+sce <- readr::read_rds(opt$input_sce_file)
+
+# read in references as a list
+model_names <- stringr::str_remove(basename(model_files), "_model.rds")
+names(model_files) <- model_names
+model_list <- purrr::map(model_files, readr::read_rds) |>
+  # make sure that the labels have label type before reference dataset name
```
jashapiro, line 77: This comment is a bit unclear to me. Part of that is that I'm not really sure the structure of the model objects, but part of it is just that I'm just confused by what a `label` and `label type` mean here, with the below code referring only to `name` and `model` (repeatedly!)

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+  purrr::map2(model_names,
+              \(model, name){
```
jashapiro, line 79: This suggestion going to conflict with my other, I'm sure, but... we can use `imap()` here!
```suggestion
  purrr::imap(\(model, name){
```

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+    names(model) <- glue::glue("{names(model)}-{name}")
+    model
+  }) |>
+  unname() |> # remove existing package names
+  unlist() # need to collapse into one list of all models
```
jashapiro, line 84: I'm confused again... If we start with one file per model, how are we ending up with a nested list here?

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+
+# run singleR for all provided models
+all_singler_results <- purrr::map(model_list,
+                                  \(model) SingleR::classifySingleR(
+                                    sce,
+                                    model,
+                                    fine.tune = TRUE,
+                                    BPPARAM = bp_param
+                                  )) |>
+  purrr::set_names(names(model_list))
```
jashapiro, line 96: I'm thinking about using more functional tricks here to maybe simplify a bit:
```suggestion
# create a partial function for mapping easily
classify_sce <- purrr::partial(SingleR::classifySingleR, 
                               test = sce, 
                               fine.tune=TRUE, 
                               BPPARAM = bp_param)
# run singleR for all provided models
all_singler_results <- model_list |>
    purrr::map(classify_sce)
```

I don't think the `set_names` should be needed at the end: they should be carried along by `purrr`

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+
+# Annotate sce -----------------------------------------------------------------
+
+# create a dataframe with a single column of annotations for each model used
+all_annotations_df <- purrr::map(all_singler_results,
+                                 \(result){result$pruned.labels}) |>
+  dplyr::bind_cols() |>
+  dplyr::mutate(barcode = colnames(sce))
```
jashapiro, line 104: Use `map_dfc` to get a data frame out faster
```suggestion
all_annotations_df <- all_singler_results |>
  purrr::map_dfc(\(result) result$pruned.labels ) |>
  dplyr::mutate(barcode = colnames(sce))
```

bin/classify_SingleR.R
```diff
@@ -0,0 +1,129 @@
...
+                          row.names = coldata_df$barcode,
+                          # prevent replacing "-" in annotation columns
+                          check.names = FALSE)
+
+# add in scores and delta to metadata of sce object
+metadata(sce)$singler_score <- purrr::map(all_singler_results,
+                                          \(result){result$scores})
+
+metadata(sce)$singler_delta <- purrr::map(all_singler_results,
+                                          \(result) {result$delta.next})
```
jashapiro, line 122: Is there a reason not to just add in the full `singleR` results objects? Size, I guess, but how much do you save by pulling these two out?

modules/classify-celltypes.nf
```diff
@@ -0,0 +1,54 @@
...
+          it[0], // meta
+          it[1][0], // processed rds
+          it[2] // tuple of reference models
+        ]}
+
+      classify_singleR(grouped_celltype_ch)
+
+    emit: classify_singleR.out
+
+}
```
jashapiro, line 54: github complains
```suggestion
}

```

jashapiro (commented):
The workflow elements look good; I think I had one little suggestion about order of operations there because I've been burned more than once and now am more than twice shy.

I also had some thoughts about the script, some of which are just because I was confused about a few steps, but I also made some fancier purrr suggestions that I hope will work and don't add extra confusion.

bin/classify_SingleR.R
allyhawkins: So it's not one file per model because each file that's being read in contains a list of three models, each model has been generated with the labels for `label.main`, `label.fine`, and `label.ont`. For reference the name that I am using is referring to the celldex package name, e.g. `HumanPrimaryCellAtlasData` and then the label or label type is `label.main` etc. I will update the comments to make that more clear.

bin/classify_SingleR.R
allyhawkins: It's not particularly large. I just thought it might be easier to access to grab down the line when we are interested in making plots, rather than storing the whole model, but I can change it.

allyhawkins:
Thanks for the additional `purrr` suggestions! I incorporated those in and also switched the order of the combine steps. 

I also tried to clarify some steps with additional comments in the script. The files that have the models being read in actually contain a list of models which is where I think things got confusing. Each reference package was trained using all of the label options (`label.main`, `label.ont`, and `label.fine`) so the resulting model objects are stored as a list in an rds file, which is what is getting read in. 

The one thing I did not update yet is whether or not we want to store the entire SingleR object or not. When testing I was using 2 reference datasets for a library so a total of 6 singleR result objects in a list which was ~ 30 MB. I was only storing them separately because I thought it might be easier to grab specific elements for making plots later on.

bin/classify_SingleR.R
jashapiro: The bulk of the object is going to be the scores matrix, so I think it is worth keeping the whole thing. It also will allow us to use some other `singleR` functions that expect the full results objects down the line if we want to. In particular, I can imagine us wanting to play around with `combineRecomputedResults` to see how that performs (particularly with the cell ontology results)

jashapiro:
```suggestion
# store results in metadata
metadata(sce)$singler_results <- all_singler_results
```

bin/classify_SingleR.R
```diff
@@ -0,0 +1,133 @@
...
+# extract coldata from sce object to join with annotations
+coldata_df <- as.data.frame(colData(sce)) |>
+  tibble::rownames_to_column("barcode") |>
+  dplyr::left_join(all_annotations_df)
+
+# add coldata with annotations back to sce
+colData(sce) <- DataFrame(coldata_df,
+                          row.names = coldata_df$barcode,
+                          # prevent replacing "-" in annotation columns
+                          check.names = FALSE)
```
jashapiro, line 119: I think we can safely assume that the orders are not changing here, since we are calculating the classifications from the same sce object in every case. So I think we can simplify this a lot:

```suggestion
# create a dataframe with a single column of annotations for each model used
all_annotations_df <- all_singler_results |>
  purrr::map_dfc(\(result) result$pruned.labels ) |>
  DataFrame(check.names = FALSE) # prevent replacing "-" in annotation columns

colData(sce) <- cbind(colData(sce), all_annotations_df)
```

Issue #281, opened 2023-02-17
Add diagram showing overall steps of scpca-nf


allyhawkins:
A few times in evals questions about what exactly the steps are in `scpca-nf` have been brought up. There's a lot of mystery on what actually happens to get FASTQ to SCE objects and right now we only include a few sentences describing that we use Alevin-fry and then a link to the portal docs. I think it would be a good idea to include an image with a very broad overview of the steps that are occurring when running `scpca-nf` so users have an idea of what's happening under the hood, other than quantifying single-cell gene expression data. A simple figure with one or two word descriptions of each of the steps happening could help provide a minimal understanding of what the workflow is doing.

allyhawkins:
I'm going to go ahead and remove my assignment and assign @dvenprasad, as she is going to update the diagram added in #288 to close this out.

dvenprasad:
@allyhawkins here is a draft of the figure. 
Let me know if this retains meaning from the current image. 
I noticed that the outputs were numbered but looking at the README, it doesn't look like we need them. Happy to add them in, if needed. 

Let me know if I missed anything else. 

Don't worry about the spacing and font sizes. I'll clean it up when  I iterate. 

![scpca-nf-outline](https://user-images.githubusercontent.com/15315514/234407893-2df908eb-a515-4682-8be0-427af0288ac9.png)

dvenprasad:
@allyhawkins 
Okay, I iterated based on your feedback

- Changed `tsv` to `mtx`
- Replaced `Convert` with `Read into R`
- Changed the shapes in the filtering steps to circles, and added some more of them to help illustrate the point about empty droplets
- I grouped the outputs together and put it in a labeled box. 
- As for the input and pipeline, I think color coding it or adding boxes will make it super busy. I added a title for the pipeline to make steps in the pipeline more obvious. The input is the starting point and is labelled, I think that will be sufficient. 

![scpca-nf-pipeline](https://user-images.githubusercontent.com/15315514/234642811-4f022c3c-f38b-41d0-81e1-3d7ef9ba1c61.png)

allyhawkins:
That's what I was trying to describe with the box around the output files! I love it! I think this looks great!

dvenprasad:
Okay perfect! Thank you. I'll PR to replace the image in the README.

jashapiro:
I think this looks great, but maybe we can tone down the red x on the trash can? It seems to put a lot of weight on that step!

[...]

PR #288, opened 2023-03-31
Add temporary overview diagram for pipeline

allyhawkins:
Starts to address #281 

Here I'm adding a temporary diagram for the overview of the pipeline that will act as a placeholder for now. The general details of the digram we plan to keep, but @dvenprasad has offered to help us out with some nice icons and expanding the diagram to be horizontal to take up less space. For now, I'm adding this in and she will modify and replace it when she is able to. 

I chose to add this to the main readme below the introduction and before instructions for processing, but let me know if there's a difference place we might think that users would find it helpful.


dvenprasad (approved):
LGTM!
🐠 🐠 🐠

PR #289, opened 2023-04-06
Initiate support for multiple organisms

allyhawkins:
Starts to address https://github.com/AlexsLemonade/ScPCA-admin/issues/447

Here I started making changes to `scpca-nf` to incorporate the use of references from different organisms. The overall goal is to use the second approach mentioned in https://github.com/AlexsLemonade/ScPCA-admin/issues/447#issuecomment-1341322507 where we will be able to align samples on a per sample basis, rather than having to use a single reference for all samples being processed. In doing this we also need to be able to build a reference for multiple organisms. Since we are moving towards running the pipeline with different references, in working on building the necessary indices, I thought it made sense to also alter the index building workflow to accommodate multiple organisms. 

To do this I first created a tsv file that has one row per organism that we will support, for now starting with just human and mouse. I used a tsv file because I needed to pull out organism, assembly, and version for naming various folders and files to keep naming conventions the same and not alter the human reference we have right now. This tsv file is then used to create a json file that has a dictionary where each organism contains a list of the corresponding reference paths needed for that organism and the key is the name used to refer to that reference. I chose to use a script to create this so that we can continue to add in organisms and then just create an updated json, adding in new entries whenever we have new organisms. 

Within the actual workflow, I have the `build-index.nf`, building references for each of the organisms supplied in the tsv file. It uses the reference name to grab all the relevant paths from the json file and create the indices for each organism. Here I used the existing `Utils.getMetaVal` to do this. 

Now all the params for references have been removed from the configs and have been transferred to the json file except for the root directory. As we want to still be able to change that. 

Following this, we would need to add a new column to the `scpca-library-metadata.tsv` to track the organism to use for that library. I started to adjust the main workflow to show how we would then capture this by using that column to parse the json file and grab the paths for that organism to the desired reference. 

I'm filing this as a draft right now for a few reasons: 

- I've tested the building index workflow, but not the main workflow as I want some input on the actual process before I file changes to the metadata and am able to fully test those. But those changes are less involved then the index building is. 
- I included the script to prepare the json file, the json file, and the tsv in this PR to go into this repo. I'm not totally sure that they belong here and maybe we should put them into `ScPCA-admin` instead, but I'm curious what others thoughts are. There's nothing private about them, they just are specific to using our references. 

Things that we will need to do but I think belong in subsequent PRs: 
- We will need to update documentation on how to specify which organism to run samples on. Additionally, if we keep the `references` folder that I've added, we should add a readme on how to maintain the files being added for data lab staff. 
- Right now we use the `t2g_3col_path` param in the cell typing workflow. Since this PR removes it, we will need to figure out how we want to handle that moving forward. For now, I just added it as a param within that workflow, which actually may be what we want to do temporarily anyways.


build-index.nf
```diff
@@ -95,47 +91,62 @@ process cellranger_index{
 
 process index_star{
```
sjspielman, line 92: let's go ahead and make this consistent; i'm making 2 suggestions to this end.

first suggestion:
```suggestion
process star_index{
```

build-index.nf
```diff
@@ -95,47 +91,62 @@ process cellranger_index{
...
+  generate_reference(ref_ch)
   // create index using reference fastas
-  salmon_index(generate_reference.out.fasta_files, params.ref_fasta)
+  salmon_index(generate_reference.out.fasta_files, ref_ch)
   // create cellranger index
-  cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
+  cellranger_index(ref_ch)
   // create star index
-  index_star(params.ref_fasta, params.ref_gtf, params.assembly)
+  index_star(ref_ch)
```
sjspielman, line 149: and second suggestion for making this consistent -

```suggestion
  star_index(ref_ch)
```

main.nf
```diff
@@ -95,15 +97,15 @@ workflow {
...
-      ref_fasta: params.ref_fasta,
-      ref_gtf: params.ref_gtf,
-      salmon_splici_index: params.splici_index,
-      t2g_3col_path: params.t2g_3col_path,
-      salmon_bulk_index: params.bulk_index,
-      t2g_bulk_path: params.t2g_bulk_path,
-      cellranger_index: params.cellranger_index,
-      star_index: params.star_index,
+      reference_name: ref_paths[it.sample_reference].reference_name,
+      ref_fasta: file("${params.ref_rootdir}/${ref_paths[it.sample_reference].ref_fasta}"),
```
sjspielman, line 101: I wondered if a small helper function would be nice to help create these paths, but I think it's probably unnecessary and also won't necessarily help the code look "cleaner," just differently busy (which is unavoidable given the task!).

references/scripts/create-reference-json.R
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env Rscript
+
+# Script for creating json file to hold path to reference files for organisms that can be
+# used with scpca-nf. The output of this script will create a json file where each key corresponds
+# to an organism and contains a dictionary of reference paths. The paths included here are specific
+# to the organization used for storing references in `s3://scpca-references`
+# To create the JSON file use a TSV file that contains three columns, `organism`, `assembly`, and `version`
```
sjspielman, line 7: or, make all the json's caps - 

```suggestion
# To create the json file, use a TSV file that contains three columns, `organism`, `assembly`, and `version`
```

references/scripts/create-reference-json.R
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env Rscript
+
+# Script for creating json file to hold path to reference files for organisms that can be
+# used with scpca-nf. The output of this script will create a json file where each key corresponds
+# to an organism and contains a dictionary of reference paths. The paths included here are specific
+# to the organization used for storing references in `s3://scpca-references`
```
sjspielman, line 6:
```suggestion
# to the organization used for storing references in `s3://scpca-references`.
```

sjspielman (commented):
> I included the script to prepare the json file, the json file, and the tsv in this PR to go into this repo. I'm not totally sure that they belong here and maybe we should put them into ScPCA-admin instead, but I'm curious what others thoughts are. 

I don't have many specific thoughts here, but it will probably be marginally more annoying to have it in `ScPCA-admin` when adding more organisms, because paths. As long as it's not a _problem_ to keep it here then, I think maybe we'll appreciate it later?

The index workflow is also running fine for me, no bugs that I can tell :)

config/reference_paths.config
```diff
@@ -1,21 +1,9 @@
-// reference params and files
-assembly         = 'Homo_sapiens.GRCh38.104'
+// path to reference files
+ref_json = "$projectDir/references/scpca-refs.json"
+ref_metadata = "$projectDir/references/ref-metadata.tsv"
```
jashapiro, line 3: Preserving spacing to make parallel structure easier to see. And brackets for consistency.
```suggestion
ref_json            = "${projectDir}/references/scpca-refs.json"
ref_metadata        = "${projectDir}/references/ref-metadata.tsv"
```

references/scripts/create-reference-json.R
```diff
@@ -0,0 +1,91 @@
...
+                            glue::glue("{reference_name}.star_idx")),
+    mito_file = file.path(ref_dir, "annotation",
+                           glue::glue("{reference_name}.mitogenes.txt")),
+    t2g_3col_path = file.path(ref_dir, "annotation",
+                               glue::glue("{reference_name}.spliced_intron.tx2gene_3col.tsv")),
+    t2g_bulk_path = file.path(ref_dir, "annotation",
+                               glue::glue("{reference_name}.spliced_cdna.tx2gene.tsv"))
+    ) |>
+    # unbox length one vectors
+    purrr::map(jsonlite::unbox)
```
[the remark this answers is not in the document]
allyhawkins: sadly that doesn't work. I had tried that first and had to move it to here to actually get it to print out a file without the boxes.

allyhawkins:
Thanks for the helpful suggestions! 

> The first one is to use the json paths as much as possible: If we already go to the trouble of defining the paths we will be using, we should refer to those as they were read in rather than recreating the same paths in the workflow code.

> The one place this breaks down is the generate_reference process, where we are writing files to multiple sub directories. The "best" solution I can think of at the moment is to not do that, and have separate processes for each directory of output, but that has its own problems with the way the scripts are designed, so I actually think the current method, with "annotations" and "fasta" as required directories is probably where we will have to end up.

Regarding this comment, I tried to use paths from the json file as much as possible. I was able to change most everything without issue except for the publish directories. I incorporated your suggestion to use `file(meta.salmon_index).parent` or similar for the publish directories, but then was consistently seeing an error that `meta` was `null` and had no attribute `file`. I tried multiple different ways including passing the publish directory as a param and nothing worked. So what I ended with was actually adding those directories to the json file as directories and not just the files and then using them to publish. I don't know if this is any better but maybe it will be easier to keep in sync if all the paths are in one central location? 

I also was working on changes related to https://github.com/AlexsLemonade/scpca-nf/pull/289#discussion_r1161795538, and was having a similar error where it could not get `ref_fasta` on a null object. I tried a few different ways, but thought I would pause and at least get the index sorted out before moving onto the workflow debugging. 

I did also return the old reference params so they are not removed yet. I also agree that we could stack all PRs related to this changes on this branch and then merge everything at once.

jashapiro (commented):
Looks good, with a few comments below, which include rebuilding the json file, and I think removing some changes that were meant for a separate branch/PR. 

I still hope to avoid the `meta.salmon_dir` etc., and I had one more idea to test there, which was to try constructs like `"${params.ref_rootdir}/${file(meta.splici_index).parent}"` It seemed to work with some limited testing I did; I think the difference may be that having all of the "code" enclosed in `${}` might work better in the directives like `publishDir`. But it also could be that my testing was insufficient!

references/scripts/create-reference-json.R
```diff
@@ -0,0 +1,93 @@
...
+                                  glue::glue("{reference_name}.spliced_cdna.txome")),
+    cellranger_dir = file.path(ref_dir, "cellranger_index"),
+    cellranger_index = file.path(ref_dir, "cellranger_index",
+                                 glue::glue("{reference_name}_cellranger_full")),
+    star_dir = file.path(ref_dir, "star_index"),
+    star_index = file.path(ref_dir, "star_index",
+                           glue::glue("{reference_name}.star_idx"))
+  ) |>
+    # unbox length one vectors
+    purrr::map(jsonlite::unbox)
```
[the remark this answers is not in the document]
allyhawkins: I swear when I did it this did not work, because it was the first thing I tried, but then when I tried again it worked.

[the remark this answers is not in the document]
jashapiro: I believe you! My guess is that there was subtle difference in the code at that point.

allyhawkins:
Okay I think I've got it now! The suggestion about using `${file(meta.salmon_index).parent}" works! Thank you. So I removed the need for the extra directory paths in the json file and also committed the updated json file (I had updated it on my end last time but missed actually committing it). I also fixed the unboxing issue. 

The last thing I did was remove anything that's being edited on `main.nf` or other modules for actually running the workflow to be in a separate branch so I can file a separate PR stacked on this one with those changes.

build-index.nf
```diff
@@ -93,49 +90,65 @@ process cellranger_index{
...
+    .splitCsv(header: true, sep: '\t')
+    .map{[
+      reference_name = "${it.organism}.${it.assembly}.${it.version}",
+      // grab file paths for each organism
+      ref_paths[reference_name]
+    ]}
+    .map{it +[
+      file("${params.ref_rootdir}/${it[1]["ref_fasta"]}"), // path to fasta
+      file("${params.ref_rootdir}/${it[1]["ref_gtf"]}") // path to gtf
+    ]}
```
jashapiro, line 141: We could potentially simplify just a bit more.
```suggestion
      ref_paths[reference_name],
      file("${params.ref_rootdir}/${ref_paths[reference_name]["ref_fasta"]}"), // path to fasta
      file("${params.ref_rootdir}/${ref_paths[reference_name]["ref_gtf"]}") // path to gtf
    ]}
```

references/ref-metadata.tsv
```diff
@@ -0,0 +1,3 @@
+organism	assembly	version
```
jashapiro, line 1: Realizing I suggested `species` in https://github.com/AlexsLemonade/ScPCA-admin/pull/512. I don't care which we use, but we should be consistent.

allyhawkins: I took these names directly from the Ensembl readmes when I was looking to see how they format their naming schemes. They use `organism.assembly.version` to name each release. So if you don't have a preference, then I think I'll change to `organism` in #512.

jashapiro: works for me!

PR #300, opened 2023-04-26
Replace pipeline image

### LINKED ISSUE #281: Add diagram showing overall steps of scpca-nf

dvenprasad:
closes #281 

Replaced the pipeline image with a new one. 

Screenshot:
Screenshot 2023-04-26 at 1 01 33 PM


allyhawkins (approved):
LGTM 🚀

PR #307, opened 2023-05-05
Cut feature barcode file for salmon indexing

sjspielman:
From this comment: https://github.com/AlexsLemonade/ScPCA-admin/issues/527#issuecomment-1534971056

> Since the changes in https://github.com/AlexsLemonade/scpca-nf/pull/305 are quite minimal and non-breaking, we should be able to create a PR that applies just those changes to `main` (along with version string updates) and create new point release. 

This PR makes the associated `cut` change to the feature barcode file on its way into `salmon index`, and bumps the minor release up accordingly. I think I got all the spots for the version string, but please let me know if I missed one!


jashapiro (approved):
I think you got all of them!

Issue #311, opened 2023-05-09
Multiple alternative experiments


sjspielman:
This issue is primarily intended for discussion as we move towards adding CITEseq processing (filtering/normalization) into the workflow. While working towards this goal, I noticed some aspects of the pipeline which may have been written with a longer term goal of accommodating multiple alternative experiments, although offline discussion has suggested this is a feature (⚠️ pun alert) that we don't implement.

I'm specifically looking at these places:

- In the `make_merged_unfiltered_sce` process, we assign a name for the alternative experiment based on the technology metadata. This choice on its own does not mean that we are setting up for multiple experiments, but it _could_ suggest that the altExp name should be sent into to later processes so that the right altExp can be pulled out to work with (for example, to be filtered on!).
https://github.com/AlexsLemonade/scpca-nf/blob/a6b87e474888cdfc4f11c137b5ab138b730c97ae/modules/sce-processing.nf#L30-L59
 
- In the post process Rscript itself `post_process_sce.R`, we have this for-loop that implies the potential existence of multiple named altExps. 
https://github.com/AlexsLemonade/scpca-nf/blob/a6b87e474888cdfc4f11c137b5ab138b730c97ae/bin/post_process_sce.R#L97-L105

To me, it seems the most immediate decision here is whether we need to be passing altExp _names_ through the pipeline to be able to isolate the CITEseq one for processing, in the event there are multiple altExps. If we always only expect 1 altExp, its name is essentially moot so that metadata isn't needed. Along the way, we could also get rid of that for-loop.

On the other hand, if we think we might in the longer term want to handle multiple altExps, then the metadata name does need to be passed through to save us from our future selves, and we'd keep the for-loop.

Tagging @jashapiro @allyhawkins for discussion!

jashapiro:
The name of the altExp are already passed along as `meta.feature_type` so there is no need for a change there at this time.

Updating the workflows to handle multiple AltExps is something we have not yet had to support, and will require substantial reengineering. The most likely scenario is that we will get data that has something like multiplexing and CITEseq or multiplexing and ATAC, in which case we would be adding a separate process and flow among processes. The merge script would have to change, as would how we designate those files in the `technology` column.

Which is basically to say, this a problem we can face when we actually have data with multiple altExps. I don't think there are any changes that need to be made at this time.

sjspielman:
[...]

PR #312, opened 2023-05-09
Add initial steps for ADT filtering

### LINKED ISSUE #293: Add ADT filtering to sce processing
The script [`post_process_sce.R`](https://github.com/AlexsLemonade/scpca-nf/blob/97c3035d946159d162b7248c83abf29ec4ae3169/bin/post_process_sce.R) should be updated to also perform ADT-level filtering on the `filtered_sce` object (which gets exported as the `<project>_processed_sce.RDS` file). This should include the following components:

- Throughout, keep track of filtering performed here (vs RNA-informed filtering) for use in QC
  - Maybe we add column(s) to `metadata` for this
- Estimate a baseline profile from the _raw unfiltered_ data using `DropletUtils::ambientProfileEmpty()`
- Get filtering QC metrics with `DropletUtils::cleanTagCounts()` and filter according to resulting `discard` column
- _If_ a negative control is present (we'll need to first get this information into `metadata` as well), filter out cells with non-0 counts as presumably contaminated.
  - This means we'll need to have information (in either a sample table or SCE's `metadata`?) indicating which ADT, if any, represents a negative control

sjspielman:
Closes #293 
(Originally I envisioned this being the first of several PRs to close out #293 - hence the branch name - but now I'm thinking for any next filtering steps I'll want to open new issues, so we can close out this issue now.)

This PR sets up the initial code for filtering on CITE-seq data. The code has been tested and is all running :tada:. I made a couple choices here, and left some for discussion during review. 

- I currently am using an Rscript argument `adt_name` that should point to the CITE-seq altExp. Based on conversation in #311, I do not actually think this argument is needed, but I've stubbed it out in case we want it. 
- For the filtering itself, I decided to make the feature barcode third column optional, which may help avoid parsing errors. If that column is not present, I assume that all ADTs are real targets and not controls. This choice is open for discussion!
- I set the filtering counts cutoff to a default of 25. Really this could use some testing (maybe it should depend on the specific library?), but benchmarking this excessively likely isn't worth our time. Any and all thoughts here are very welcome.
- Now, the filtering itself:
  - If no controls, filter based on `discard` qc stats and move on. (Question: Do we want to keep these QC stats in the SCE?)
  - If negative controls, then _also_ remove cells with high neg control counts. As it is possible to have >1 negative control ADT, I wrote this up as: Remove any cell where at least 1 negative control is >= threshold.
- How do we feel about retaining that for-loop over the altExps? Should we replace with an `if` to check for the _single_ altExp?

I see next steps after this PR as implementing some bookkeeping, e.g. how many cells get removed from RNA vs ADT filtering, which may be useful to include in the final QC report. Am I missing anything else re-filtering specifically (aside from docs)? 

I'll note I'm mostly expecting @jashapiro to review this but @allyhawkins may also have thoughts, feel free to chime in everyone!

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -147,7 +147,7 @@ if (!is.null(ambient_profile)) {
   if (!"target_type" %in% names(adt_barcode_df)) {
     adt_barcode_df$target_type <- "target"
   } 
-  if ( !all.equal( sort(adt_barcode_df$names), sort(rownames(altExp(filtered_sce, adt_exp))) )) {
+  if ( !all.equal( sort(adt_barcode_df$name), sort(rownames(altExp(filtered_sce, adt_exp))) )) {
     stop("Mismatch between provided ADT barcode file and ADTs in SCE.")
   }
   # Find any negative controls
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -147,9 +147,9 @@ if (!is.null(ambient_profile)) {
   if (!"target_type" %in% names(adt_barcode_df)) {
     adt_barcode_df$target_type <- "target"
   } 
- if( !all.equal(sort(adt_barcode_df$names), sort(rownames(altExp(filtered_sce, adt_exp)) ){
-   stop("Mismatch between provided ADT barcode file and ADTs in SCE.")
- }
+  if ( !all.equal( sort(adt_barcode_df$names), sort(rownames(altExp(filtered_sce, adt_exp))) )) {
+    stop("Mismatch between provided ADT barcode file and ADTs in SCE.")
+  }
   # Find any negative controls
   neg_controls <- adt_barcode_df |>
     dplyr::filter(target_type == "neg_control") |>
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -33,6 +33,19 @@ option_list <- list(
     help = "flag to set the miQC's enforce_left_cutoff option to TRUE"
   ),
   make_option(
+    opt_str = c("--adt_barcode_file"),
+    type = "character",
+    default = NULL,
+    help = "Optional path to an ADT barcode file, where the third column indicates
+      the ADT type (target, negative control, or positive control)."
+  ),
+  make_option(
+    opt_str = c("--adt_name"),
+    type = "character",
+    default = NULL,
+    help = "Name for the alternative experiment, if present, that contains ADT features"
+  ),
+  make_option(
    opt_str = c("-r", "--random_seed"),
    type = "integer",
    help = "A random seed for reproducibility."
@@ -62,13 +75,38 @@ if(!dplyr::between(opt$prob_compromised_cutoff, 0, 1)){
 # read in unfiltered rds file
 unfiltered_sce <- readr::read_rds(opt$unfiltered_file)
 
+# check if ADT data is present, and prep if so
+if (is.null(opt$adt_barcode_file)) {
+  ambient_profile <- NULL
+} else { 
+
+  # Create data frame of ADTs and their target types
+  # If `target_type` column is not present, assume all ADTs are targets
+  adt_barcode_df <- readr::read_tsv(
+    opt$adt_barcode_file, 
+    # if only 2 columns exist, only the first two col_names will be used
+    col_names = c("name", "barcode", "target_type")
+  ) 
+  if (!"target_type" %in% names(adt_barcode_df)) {
+    adt_barcode_df$target_type <- "target"
+  } 
+
+  # assign and check name for this alternative experiment
+  adt_exp <- opt$adt_name
+  if (!adt_exp %in% altExpNames(unfiltered_sce)) {
+    stop("Given named ADT alternative experiment not present in unfiltered SCE.")
+  }
+  # Calculate ambient profile from empty drops for later use
+  ambient_profile <- DropletUtils::ambientProfileEmpty( counts(altExp(unfiltered_sce, adt_exp)) )
+}
+
 # filter sce
 filtered_sce <- scpcaTools::filter_counts(unfiltered_sce)
 # remove unfiltered for memory saving
 rm(unfiltered_sce)
 
 # need to remove old gene-level rowData statistics first
-drop_cols = colnames(rowData(filtered_sce)) %in% c('mean', 'detected')
+drop_cols <- colnames(rowData(filtered_sce)) %in% c('mean', 'detected')
 rowData(filtered_sce) <- rowData(filtered_sce)[!drop_cols]
 
 # recalculate rowData and add to filtered sce
@@ -99,12 +137,41 @@ alt_names <- altExpNames(filtered_sce)
 
 for (alt in alt_names) {
   # remove old row data from unfiltered
-  drop_cols = colnames(rowData(altExp(filtered_sce, alt))) %in% c('mean', 'detected')
+  drop_cols <- colnames(rowData(altExp(filtered_sce, alt))) %in% c('mean', 'detected')
   rowData(altExp(filtered_sce, alt)) <- rowData(altExp(filtered_sce, alt))[!drop_cols]
 
   # add alt experiment features stats for filtered data
   altExp(filtered_sce, alt) <- scuttle::addPerFeatureQCMetrics(altExp(filtered_sce, alt))
 }
+
+
+# calculate filtering QC from ADTs, if present
+if (!is.null(ambient_profile)) {
+  
+  # Find any negative controls
+  neg_controls <- adt_barcode_df |>
+    dplyr::filter(target_type == "neg_control") |>
+    dplyr::pull(name)
+  
+  # Calculate QC stats, providing negative controls if present
+  # note: function fails if controls is length 0 or null, so keep the `if`
+  if (length(neg_controls) == 0) {
+    adt_qc_df <- DropletUtils::cleanTagCounts(
+      counts(altExp(filtered_sce, adt_exp)),
+      ambient = ambient_profile
+    )
+  } else {
+    adt_qc_df <- DropletUtils::cleanTagCounts(
+      counts(altExp(filtered_sce, adt_exp)),
+      ambient = ambient_profile, 
+      controls = neg_controls
+    )
+  }
+  
+  # Save QC stats to the altexp
+  colData(altExp(filtered_sce, adt_exp)) <- cbind(colData(altExp(filtered_sce, adt_exp)), adt_qc_df)
+}
+
 
 # write filtered sce to output
 readr::write_rds(filtered_sce, opt$filtered_file, compress = "gz")
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -42,7 +42,7 @@ option_list <- list(
   make_option(
     opt_str = c("--adt_name"),
     type = "character",
-    default = NULL,
+    default = "CITEseq",
     help = "Name for the alternative experiment, if present, that contains ADT features"
   ),
   make_option(
@@ -79,7 +79,12 @@ unfiltered_sce <- readr::read_rds(opt$unfiltered_file)
 if (is.null(opt$adt_barcode_file)) {
   ambient_profile <- NULL
 } else { 
-
+  # assign and check name for this alternative experiment
+  adt_exp <- opt$adt_name
+  if (!adt_exp %in% altExpNames(unfiltered_sce)) {
+    stop("Given named ADT alternative experiment not present in unfiltered SCE.")
+  }
+  
   # Create data frame of ADTs and their target types
   # If `target_type` column is not present, assume all ADTs are targets
   adt_barcode_df <- readr::read_tsv(
@@ -91,11 +96,6 @@ if (is.null(opt$adt_barcode_file)) {
     adt_barcode_df$target_type <- "target"
   } 
 
-  # assign and check name for this alternative experiment
-  adt_exp <- opt$adt_name
-  if (!adt_exp %in% altExpNames(unfiltered_sce)) {
-    stop("Given named ADT alternative experiment not present in unfiltered SCE.")
-  }
   # Calculate ambient profile from empty drops for later use
   ambient_profile <- DropletUtils::ambientProfileEmpty( counts(altExp(unfiltered_sce, adt_exp)) )
 }
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -84,17 +84,6 @@ if (is.null(opt$adt_barcode_file)) {
   if (!adt_exp %in% altExpNames(unfiltered_sce)) {
     stop("Given named ADT alternative experiment not present in unfiltered SCE.")
   }
-  
-  # Create data frame of ADTs and their target types
-  # If `target_type` column is not present, assume all ADTs are targets
-  adt_barcode_df <- readr::read_tsv(
-    opt$adt_barcode_file, 
-    # if only 2 columns exist, only the first two col_names will be used
-    col_names = c("name", "barcode", "target_type")
-  ) 
-  if (!"target_type" %in% names(adt_barcode_df)) {
-    adt_barcode_df$target_type <- "target"
-  } 
 
   # Calculate ambient profile from empty drops for later use
   ambient_profile <- DropletUtils::ambientProfileEmpty( counts(altExp(unfiltered_sce, adt_exp)) )
@@ -148,6 +137,17 @@ for (alt in alt_names) {
 # calculate filtering QC from ADTs, if present
 if (!is.null(ambient_profile)) {
   
+  # Create data frame of ADTs and their target types
+  # If `target_type` column is not present, assume all ADTs are targets
+  adt_barcode_df <- readr::read_tsv(
+    opt$adt_barcode_file, 
+    # if only 2 columns exist, only the first two col_names will be used
+    col_names = c("name", "barcode", "target_type")
+  ) 
+  if (!"target_type" %in% names(adt_barcode_df)) {
+    adt_barcode_df$target_type <- "target"
+  } 
+  
   # Find any negative controls
   neg_controls <- adt_barcode_df |>
     dplyr::filter(target_type == "neg_control") |>
@@ -169,7 +169,11 @@ if (!is.null(ambient_profile)) {
   }
   
   # Save QC stats to the altexp
-  colData(altExp(filtered_sce, adt_exp)) <- cbind(colData(altExp(filtered_sce, adt_exp)), adt_qc_df)
+  colData(altExp(filtered_sce, adt_exp)) <- cbind(colData(altExp(filtered_sce, adt_exp)), 
+                                                  adt_qc_df)
+  
+  # Add `target_type` to rowData
+  rowData(altExp(filtered_sce, adt_exp))$target_type <- adt_barcode_df$target_type
 }
 
 
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -147,7 +147,9 @@ if (!is.null(ambient_profile)) {
   if (!"target_type" %in% names(adt_barcode_df)) {
     adt_barcode_df$target_type <- "target"
   } 
-  
+ if( !all.equal(sort(adt_barcode_df$names), sort(rownames(altExp(filtered_sce, adt_exp)) ){
+   stop("Mismatch between provided ADT barcode file and ADTs in SCE.")
+ }
   # Find any negative controls
   neg_controls <- adt_barcode_df |>
     dplyr::filter(target_type == "neg_control") |>
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -122,12 +122,11 @@ if (!(is.null(opt$adt_barcode_file))) {
     stop("The ADT feature barcode file must have either 2 or 3 columns.")
   }
 
-  # TODO: Name for _this_ alternative experiment.
-  # One this is set, apply `altexp_name` as second argument when calling `altExp()`
-  #altexp_name <- opt$adt_name
+  # Name for this alternative experiment.
+  altexp_name <- opt$adt_name
 
   # Calculate ambient profile from unfiltered sce for later use
-  ambient_profile <- DropletUtils::ambientProfileEmpty( counts(altExp(sce)) )
+  ambient_profile <- DropletUtils::ambientProfileEmpty( counts(altExp(sce, altexp_name)) )
 
   # define indicator for if ADTs need to be processed
   process_adt <- TRUE
@@ -139,10 +138,6 @@ if (!(is.null(opt$adt_barcode_file))) {
 
 # filter sce using criteria in scpca_filter
 filtered_sce <- sce[, which(sce$scpca_filter == "Keep")]
-
-# replace existing stats with recalculated gene stats
-drop_cols <- colnames(rowData(filtered_sce, alt)) %in% c('mean', 'detected')
-rowData(filtered_sce) <- rowData(filtered_sce)[!drop_cols]
 
 filtered_sce <- filtered_sce |>
   scuttle::addPerFeatureQCMetrics()
@@ -161,7 +156,7 @@ for (alt in altExpNames(filtered_sce)) {
 if (process_adt) {
 
   adt_qc_df <- DropletUtils::cleanTagCounts(
-    counts(altExp(filtered_sce)),
+    counts(altExp(filtered_sce, altexp_name)),
     ambient = ambient_profile
   )
 
@@ -169,7 +164,7 @@ if (process_adt) {
   filtered_sce <- filtered_sce[, which(!(adt_qc_df$discard))]
 
   # TODO: Should we add `adt_qc_df` into the altExp? If we want it:
-  #colData(altExp(filtered_sce)) <- cbind(colData(altExp(filtered_sce)), adt_qc_df)
+  #colData(altExp(filtered_sce, altexp_name)) <- cbind(colData(altExp(filtered_sce, altexp_name)), adt_qc_df)
 
   # Filter on negative controls, if present
   neg_controls <- adt_barcode_df |>
@@ -178,7 +173,7 @@ if (process_adt) {
 
   if (length(neg_controls) > 0) {
     # counts for each negative control
-    neg_control_counts <- counts(altExp(filtered_sce))[neg_controls, ]
+    neg_control_counts <- counts(altExp(filtered_sce, altexp_name))[neg_controls, ]
 
     # remove any cell with >=threshold for any given negative control
     # first, determine which cells need to be removed
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -11,9 +11,15 @@ suppressPackageStartupMessages({
 # set up arguments
 option_list <- list(
   make_option(
-    opt_str = c("-i", "--input_sce_file"),
+    opt_str = c("-i", "--input_filtered_sce_file"),
     type = "character",
     help = "path to rds file with input sce object to be processed"
+  ),
+  make_option(
+    opt_str = c("-u", "--input_unfiltered_sce_file"),
+    type = "character",
+    help = "path to rds file with fully unfiltered (contains empty drops) input sce object. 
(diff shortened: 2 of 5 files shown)


main.nf
sjspielman, line 1: Noting that I made this change to match some changes in #289, since I didn't stack on #289. This will work itself out in the end I'm sure!

bin/post_process_sce.R
```diff
@@ -97,13 +149,53 @@ filtered_sce <- filtered_sce |>
...
+    neg_control_counts <- counts(altExp(filtered_sce, altexp_name))[neg_controls, ]
+
+    # remove any cell with >=threshold for any given negative control
+    # first, determine which cells need to be removed
+    cells_to_remove <- purrr::map(
+      neg_controls,
+      \(x) which(neg_control_counts[x,] >= opt$adt_negative_cutoff)
+    ) |>
+    unlist() |>
+    unique()
```
jashapiro, line 189: My general philosophy is that if we have a matrix, let's use the matrix:
    
```suggestion
    # remove any cell with >=threshold for any given negative control
    cells_to_remove <- colSums(neg_control_counts >= opt$adt_negative_cutoff) > 0
```
```

**Review by jashapiro — COMMENTED** [review]
My first major comment at this stage is that we don't want to be using file inputs that are not declared in the `input:` directive, as you do here for the `feature_barcode_file`.

So to fix that, you will want to update the inputs to:
```
    input:
        tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(feature_barcode_file)
```

Which will require modifying `main.nf` to add that file to the channel  that goes into `post_process_sce` with something like:
```
all_sce_ch = sce_ch.single.mix(genetic_demux_sce.out)
  .map{it.toList() + file(it[0].feature_barcode_file)]}
```

I _think_ this will work fine if the file does not actually exist, and you can check that as you already were.

I made a few other suggestions, but I did not have a close look at everything as yet, as I expect some iteration based on the changes suggested above.

File: modules/sce-processing.nf [hunk 2]
```diff
@@ -134,10 +134,20 @@ process post_process_sce{
         tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(processed_rds)
     script:
         processed_rds = "${meta.library_id}_processed.rds"
+
+        // Two checks for the feature barcode file, and provide argument if both true:
+        // - technology should be CITEseq
+        // - barcode file should exist
+        adt_barcode_option_bool = meta.technology ==~ '^CITEseq.+' & file(meta.feature_barcode_file).exists()
```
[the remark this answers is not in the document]
sjspielman: I'm pretty sure this needs `file()` - nextflow threw errors otherwise about how it did not like checking if a string exists. It only wants to know if a file exists. But otherwise can be simplified as you suggest!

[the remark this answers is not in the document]
jashapiro: It won't need file if you pass in the file as an input!

bin/post_process_sce.R
jashapiro: for complete parallelism, it should be:
```
cells_to_remove <- which(colSums(neg_control_counts >= opt$adt_negative_cutoff) > 0)
```

sjspielman: > No, my code is checking each cell, converting it to a boolean matrix, then adding the sums of TRUE and FALSE for each cell.

Ah, indeed! It would help if I more precisely noted where the parentheses are!

jashapiro: Minor followup: possibly better to do: 
```
cells_to_keep <- which(colSums(neg_control_counts >= opt$adt_negative_cutoff) == 0)
```
and save a negation

allyhawkins:
I took a brief look at this and just have some small replies to your questions. 

> I set the filtering counts cutoff to a default of 25. Really this could use some testing (maybe it should depend on the specific library?), but benchmarking this excessively likely isn't worth our time. Any and all thoughts here are very welcome.

Is there any guidance anywhere in OSCA that says to use 25? Or another number? Another thought could be to take the max count of the negative control and say anything that has a count in the top 75% of that value gets removed? I'm really shooting from the hip here, but trying to think of how we can account for differences in coverage of negative controls. 

> If no controls, filter based on discard qc stats and move on. (Question: Do we want to keep these QC stats in the SCE?)

I think we may want to keep some of them? Maybe `zero.ambient` would be helpful looking at the [docs](https://rdrr.io/github/MarioniLab/DropletUtils/man/cleanTagCounts.html)?

sjspielman:
> Is there any guidance anywhere in OSCA that says to use 25? Or another number?

The guidance is, not unexpectedly, that this is really dataset-specific..

http://bioconductor.org/books/3.17/OSCA.advanced/integrating-with-protein-abundance.html#applying-custom-qc-filters

> We identify affected cells as high outliers for the total control count (Figure [12.3](http://bioconductor.org/books/3.17/OSCA.advanced/integrating-with-protein-abundance.html#fig:cite-total-con-hist)); hard thresholds are more difficult to specify due to experiment-by-experiment variation in the expected coverage of ADTs.


Will look into `zero.ambient` too!

bin/post_process_sce.R
```diff
@@ -84,6 +105,37 @@ if(all(is.na(sce$miQC_pass))){
...
+    colnames(adt_barcode_df) <- c("name", "barcode", "target_type")
+  } else {
+    stop("The ADT feature barcode file must have either 2 or 3 columns.")
+  }
+
+  # Name for this alternative experiment.
+  altexp_name <- opt$adt_name
+
+  # Calculate ambient profile from unfiltered sce for later use
+  ambient_profile <- DropletUtils::ambientProfileEmpty( counts(altExp(sce, altexp_name)) )
```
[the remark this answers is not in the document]
sjspielman: Ah yes, thanks, need to add that input!

bin/post_process_sce.R
```diff
@@ -84,6 +105,37 @@ if(all(is.na(sce$miQC_pass))){
...
+  # Check if 3rd column is present and assign column names, values assuming
+  #   that all ADTs are targets if there is no 3rd column
+  if (ncol(adt_barcode_df) == 2) {
+    colnames(adt_barcode_df) <- c("name", "barcode")
+    adt_barcode_df$target_type <- "target"
+  } else if (ncol(adt_barcode_df) == 3) {
+    colnames(adt_barcode_df) <- c("name", "barcode", "target_type")
+  } else {
+    stop("The ADT feature barcode file must have either 2 or 3 columns.")
+  }
```
jashapiro, line 123: Do we really care if there are more than three columns? If not (and I don't think we should care for future robustness), we can simplify the reading here.

```suggestion
  # Create data frame of ADTs and their target types
  adt_barcode_df <- readr::read_tsv(
    opt$adt_barcode_file, 
    col_names = c("name", "barcode", "target_type")
    # if only 2 columns exist, only the first two col_names will be used
  ) 
    
  if (!"target_type" %in% names(adt_barcode_df)) {
    adt_barcode_df$target_type <- "target"
  } 
```

sjspielman:
Noting that the processing is working in https://github.com/AlexsLemonade/scpca-nf/pull/312/commits/57cfa4a28408eaa12cede1800c62633afb608a45, but celltyping now not working afterwards... I did not change the output from `post_process_sce` so I don't immediately see where I might have broken it. I confirmed the `celltype_ch` channel is still structured as `[[meta], processed rds]`, where there is actual content in the processed RDS file. 
I ended up with this error: 
```
fatal error: An error occurred (403) when calling the HeadObject operation: Forbidden
```
... feels permissions-y? If anyone wants to see the full log, [enjoy](https://github.com/AlexsLemonade/scpca-nf/files/11435688/nextflow.log).

bin/post_process_sce.R
sjspielman: I'll implement this but can't accept suggestion without stashing in-progress changes, and it's easier to copy/paste...

jashapiro:
> Noting that the processing is working in [57cfa4a](https://github.com/AlexsLemonade/scpca-nf/commit/57cfa4a28408eaa12cede1800c62633afb608a45), but celltyping now not working afterwards... I did not change the output from `post_process_sce` so I don't immediately see where I might have broken it. I confirmed the `celltype_ch` channel is still structured as `[[meta], processed rds]`, where there is actual content in the processed RDS file. I ended up with this error:

This does feel like a permissions issue, and I suspect it is related to Ally working on cell typing (the singleR model files changed today) We should probably comment out the cell typing process for now, as it is not fully implemented.

bin/post_process_sce.R
sjspielman: Noting this is done in https://github.com/AlexsLemonade/scpca-nf/pull/312/commits/0b38c3f48a6e015a3aeed628253305bdb395a7ac, along with a couple other updates to use unfiltered as well as filtered.

allyhawkins:
> This does feel like a permissions issue, and I suspect it is related to Ally working on cell typing (the singleR model files changed today) We should probably comment out the cell typing process for now, as it is not fully implemented.

Just confirming that this was me working on the `build-index.nf` workflow. I think because we have cell typing separate right now, we should remove running it from both the index and the main workflow until it's fully developed as we discussed in slack. Just commenting here for posterity.

sjspielman:
Back to filtering - the `discard` output from `cleanTagCounts` will be true if either `zero.ambient` is true, _or_ the overall count is too high (`extra` below), so `discard` more stringent.

https://github.com/MarioniLab/DropletUtils/blob/e44a19f907997bc88c682d353b4883708a7e86b2/R/cleanTagCounts.R#L111

bin/post_process_sce.R
```diff
@@ -97,13 +157,47 @@ filtered_sce <- filtered_sce |>
...
+  adt_qc_df <- DropletUtils::cleanTagCounts(
+    counts(altExp(filtered_sce, altexp_name)),
+    ambient = ambient_profile
+  )
+
+  # Filter according to `discard`
+  filtered_sce <- filtered_sce[, which(!(adt_qc_df$discard))]
+
+  # TODO: Should we add `adt_qc_df` into the altExp? If we want it:
+  #colData(altExp(filtered_sce, altexp_name)) <- cbind(colData(altExp(filtered_sce, altexp_name)), adt_qc_df)
```
jashapiro, line 179: I don't see any negative to including this, except that we need to include a note about it in the docs.

bin/post_process_sce.R
```diff
@@ -97,13 +157,47 @@ filtered_sce <- filtered_sce |>
...
 
   # add alt experiment features stats for filtered data
   altExp(filtered_sce, alt) <- scuttle::addPerFeatureQCMetrics(altExp(filtered_sce, alt))
 }
 
+# filter sce based on ADTs
+if (process_adt) {
+
+  adt_qc_df <- DropletUtils::cleanTagCounts(
+    counts(altExp(filtered_sce, altexp_name)),
```
jashapiro, line 171: We should use the `controls` argument in the case of negative controls. In this case we may be able to get rid of our own judgement about negative control counts, and the code for counting...

jashapiro (commented):
A few more mostly style comments, but the biggest comment is probably the last one, which makes part of my earlier review moot:

`cleanTagCounts` has a `controls` argument for handling negative control data, so we should use that if we can! SEems like a better plan than trying to create our own arbitrary cutoff: it will build the cutoff based on the data, which seems a good thing in case one of the negative controls has more nonspecific binding.

bin/post_process_sce.R
```diff
@@ -246,8 +246,8 @@ try({
 
 # Export --------------
 
-# write out original SCE with additional filtering column
-readr::write_rds(sce, opt$input_sce_file, compress = "gz")
+# write out _original_ filtered SCE with additional filtering column
+readr::write_rds(sce, opt$input_filtered_sce_file, compress = "gz")
```
jashapiro, line 250: Good argument to maybe rename this just `filtered_sce_file` 
```suggestion
readr::write_rds(sce, opt$input_filtered_sce_file, compress = "gz")
```

bin/post_process_sce.R
sjspielman: Still have to count how many negative controls, since `cleanTagCounts` would really _not_ like an empty `controls` argument - if it's `c()`, then it thinks everything is a negative control, and it errors out with `NULL`. Otherwise this is now implemented!

sjspielman:
@jashapiro I've updated this script substantially based on reviews, let me know what you think now!

jashapiro (commented):
Getting close! I think we want to change up order of operations just a bit though. 

For the RNA data we calculate all statistics including our filtering judgment in `scpca_filter`, add them to the input `sce`, and then in a later step perform the actual filtering (confusingly, we call this `filtered_sce`, but the input is already filtered in a way, so we might rename that for clarity). This allow us to write out all of the the data in an updated version of the original `sce`.

We can take the same approach with the ADT data: If we do the `cleanTagCounts()` and update the altExp colData before the RNA filtering, then we can combine both the RNA filtering and ADT filtering into the `scpca_filter` variable, and then filter once on that column .

In concrete terms, you would move most of the code added here (lines 168-197) to after the current line 125 (`sce <- readr::read_rds(opt$filtered_sce_file)`).

We would then update the code that creates `scpca_filter` to include the cleanTagCounts results, if present, and update the `scpca_filter_method` to include more information about whether ADT filtering was also used in the filters.

The key advantage here is that it allows users to go back to the `filtered_sce.rds` file and recreate the final filtering that resulted in `processed_sce.rds` without redoing any calculations, if they so desire.

One related thought: do we want to include the ambient profile result in the altExp metadata? We've done this with miQC, but it probably depends on how large the object is.

jashapiro:
Wait... are we doing all of this in the wrong function? Should we be doing the first step of this in  the `filter_sce` process?

jashapiro:
> > Wait... are we doing all of this in the wrong script? Should we be doing the first step of this (cleanTagCounts()) in the filter_sce process?
> 
> I think we're in the right place, but this also depends on one's philosophy for what is meant by "filtering." Maybe what you're getting at is - we could do calculations in `filter_sce` but then _apply_ the filtering in post process, which is analogous to how miQC, for example, is handled?

Yes, that is exactly what I am referring to. Most of what we are doing here with cleanTagCounts is analogous to the miQC step, so it would seem to make sense to do both of those in the same step of the workflow. It also saves us reading in the unfiltered data again, though that is a more minor benefit.

sjspielman:
Update - 
- I reorganized as discussed, and tested with an ADT data set. All working! Right now, I'm testing this out on libraries without CITEseq to see whether feeding an empty string into the filtering process works.
- I lost my mind at one point with our 2 different definitions of "filtered sce", so I changed the variable `filtered_sce` -> `processed_sce` in the `post_process_sce.R` script for personal sanity. I can revert this if I went to far here!

sjspielman:
Back to note that the empty string approach, with several tries of different syntax for specifying something empty (`""`, `file("")`, `path("")`, `"not_a_file.txt"`, `file("not_a_file.txt")`.........), does not seem to work. I've also tried playing around with how the input is specified, but no luck yet. Will be back in 1-1.5 hours to continue 🕵️

sjspielman:
Back to note that the empty string approach, with several tries of different syntax for specifying something empty (`""`, `file("")`, `path("")`, `"not_a_file.txt"`, `file("not_a_file.txt")`.........), does not seem to work. I've also tried playing around with how the input is specified, but no luck yet. Will be back in 1-1.5 hours to continue 🕵️

sjspielman:
Back to note that the empty string approach, with several tries of different syntax for specifying something empty (`""`, `file("")`, `path("")`, `"not_a_file.txt"`, `file("not_a_file.txt")`.........), does not seem to work. I've also tried playing around with how the input is specified, but no luck yet. Will be back in 1-1.5 hours to continue 🕵️

jashapiro:
Possibly helpful: https://nextflow-io.github.io/patterns/optional-input/

sjspielman:
This turns out to be a known problem with AWS: https://github.com/nextflow-io/nextflow/issues/1233 
We may _actually_ need a dummy file here for optional input, something like this - https://github.com/nextflow-io/nextflow/issues/1233#issuecomment-513121438. I'm not crazy enamored by this approach but I don't _hate_ it.

One low-hanging fruit other approach could be to "split up" `filter_sce` into two processes, `filter_sce_merged` and `filter_sce_unmerged` and avoid the whole optional argument altogether, but this is also kinda clunky! That said, I think I like it slightly more than dummy files?

jashapiro:
I actually _like_ the null file approach, so let's try that.

sjspielman:
> I actually _like_ the null file approach, so let's try that.

Ok, onwards! I've starting getting going on this, and a tricky aspect I'm encountering is how to tell nextflow about the file path in the first place. It doesn't really make sense to me to put a blank file literally as metadata param (weird UI for external users..). So, here's what I'm currently thinking - make use of the `files_directory` value in the metadata and actually just create a blank file there on the fly, which can be passed into the `filter_sce` process. I think that file would need to be created in its own `process` that can be called from the subworkflow `generate_sce()`.

How does something like this sound? Any ideas for simplifying?

jashapiro:
> > I actually _like_ the null file approach, so let's try that.
> 
> Ok, onwards! I've starting getting going on this, and a tricky aspect I'm encountering is how to tell nextflow about the file path in the first place. It doesn't really make sense to me to put a blank file literally as metadata param (weird UI for external users..). So, here's what I'm currently thinking - make use of the `files_directory` value in the metadata and actually just create a blank file there on the fly, which can be passed into the `filter_sce` process. I think that file would need to be created in its own `process` that can be called from the subworkflow `generate_sce()`.
> 
> How does something like this sound? Any ideas for simplifying?

To follow the `nf-core` convention (and similar to something we did before, see below), we can create an `assets` directory at the root of the workflow, and then refer to it using `${projectDir}/assets/NULL_FILE` (or whatever you want to name it, or if you want to stash it in a subdirectory).

This is the approach that we are using to distribute the Rmd templates, and we might consider moving those to a subdirectory of the `assets` directory, but that can come later.

sjspielman:
All tested and working with `NO_FILE.txt`, for both runs with and without CITEseq!

bin/filter_sce_rds.R
```diff
@@ -32,6 +32,19 @@ option_list <- list(
...
+    opt_str = c("--adt_barcode_file"),
+    type = "character",
+    default = NULL,
+    help = "Optional path to an ADT barcode file, where the third column indicates
+      the ADT type (target, negative control, or positive control)."
+  ),
+  make_option(
+    opt_str = c("--adt_name"),
+    type = "character",
+    default = NULL,
```
jashapiro, line 45: Adding the default. I'm a bit conflicted on this, but I think it is safe as we do have a `opt$adt_name %in% altExpNames(sce)` in there.
```suggestion
    default = "CITEseq",
```

bin/filter_sce_rds.R
```diff
@@ -62,13 +75,38 @@ if(!dplyr::between(opt$prob_compromised_cutoff, 0, 1)){
...
+  ) 
+  if (!"target_type" %in% names(adt_barcode_df)) {
+    adt_barcode_df$target_type <- "target"
+  } 
+
+  # assign and check name for this alternative experiment
+  adt_exp <- opt$adt_name
+  if (!adt_exp %in% altExpNames(unfiltered_sce)) {
+    stop("Given named ADT alternative experiment not present in unfiltered SCE.")
+  }
```
jashapiro, line 98: I'm a big proponent of failing fast in code (not in orgs) so I would move the check up to before reading any files. 
```suggestion
  # assign and check name for this alternative experiment
  adt_exp <- opt$adt_name
  if (!adt_exp %in% altExpNames(unfiltered_sce)) {
    stop("Given named ADT alternative experiment not present in unfiltered SCE.")
  }
  
  # Create data frame of ADTs and their target types
  # If `target_type` column is not present, assume all ADTs are targets
  adt_barcode_df <- readr::read_tsv(
    opt$adt_barcode_file, 
    # if only 2 columns exist, only the first two col_names will be used
    col_names = c("name", "barcode", "target_type")
  ) 
  if (!"target_type" %in% names(adt_barcode_df)) {
    adt_barcode_df$target_type <- "target"
  } 

```

jashapiro: Secondary thought: we don't actually use the `adt_barcode_df` until much later... do we want to save reading until that point? 

Alternatively, we could use it here to check that the names match the altExp rownames. That seems a check worth doing to make sure we have the right pairing.

Related: we should definitely add the `target_type` column to the altExp rowData. (I don't know if the barcode seq is already there, but it also doesn't seem important to add)

bin/filter_sce_rds.R
```diff
@@ -99,13 +137,42 @@ alt_names <- altExpNames(filtered_sce)
...
+  } else {
+    adt_qc_df <- DropletUtils::cleanTagCounts(
+      counts(altExp(filtered_sce, adt_exp)),
+      ambient = ambient_profile, 
+      controls = neg_controls
+    )
+  }
+  
+  # Save QC stats to the altexp
+  colData(altExp(filtered_sce, adt_exp)) <- cbind(colData(altExp(filtered_sce, adt_exp)), adt_qc_df)
```
jashapiro, line 172: Are there any attributes on `adt_qc_df` we might want to save to metadata?

bin/post_process_sce.R
```diff
@@ -47,28 +53,47 @@ option_list <- list(
...
+#  and string to include in `scpca_filter_method`
+alt_exp <- opt$adt_name
+if (!(alt_exp %in% altExpNames(sce))) {
+  adt_discard_rows <- c()
+  adt_filter_string <- ""
+} else {
+  adt_discard_rows <- which(colData(altExp(sce, alt_exp))$discard)
+  adt_filter_string <- ";cleanTagCounts"
+  
+}
```
jashapiro, line 87: Minor organization thought: move the ADT filter checking to closer to its use (current row 109)

bin/post_process_sce.R
```diff
@@ -81,75 +106,89 @@ if(all(is.na(sce$miQC_pass))){
   metadata(sce)$scpca_filter_method <- "miQC"
 }
 
+# include ADT filtering information
+colData(sce)$scpca_filter[adt_discard_rows] <- "Remove"
```
jashapiro, line 110: very minor, but I don't think we need the `colData()` here: we call it later as just `sce$scpca_filter`, so I think we should be able to be consistent? (Also modify existing usage above)

```suggestion
sce$scpca_filter[adt_discard_rows] <- "Remove"
```

jashapiro (commented):
Looking good! I just had a few more thoughts about adding a bit more metadata as we are performing the tag cleaning, and a tiny bit of organization.

bin/filter_sce_rds.R
sjspielman: I copied below an overview of the columns in `adt_qc_df`. Some ideas..
-  % of cells that passed the `discard` check. That said, this would be a percentage based on total cells in `unfiltered_sce`, even though this information would be in the `filtered_sce` object. We don't really have analogous information for RNA filtering though; seems like mostly parameters (not results) in the metadata. 
-  Something that I think could make good sense is an indicator for whether controls were used as part of the calculations

```
    zero.ambient, a logical field indicating whether each cell has zero ambient contamination.

    sum.controls, a numeric field containing the sum of counts for all control features. Only present if controls is supplied.

    high.controls, a logical field indicating whether each cell has unusually high control total. Only present if controls is supplied.

    ambient.scale, a numeric field specifying the relative amount of ambient contamination. Only present if controls is not supplied.

    high.ambient, a numeric field indicating whether each cell has unusually high ambient contamination. Only present if controls is not supplied.

    discard, a logical field indicating whether a column in x should be discarded.
```

jashapiro: Sorry, I was unclear. I don't think we need to add any fields of our own, and the question of whether there are controls should be possible to assay based on the `rowData` column I suggested, so I don't think we need to add that. I meant specifically anything that might be stored in the extended attributes of the `DataFrame` object, which may be accessible with `metadata(adt_qc_df)` (or possibly some other `@` slot)

sjspielman: Ah, gotcha. Nothing useful here though - 

```
> summary(adt_qc_df)
[1] "DataFrame object of length 4 with 0 metadata columns"
```

There are a couple slots, but their contents are either empty or are just reiterations of what's in the `DataFrame` (run both with and without controls). 

So, we're getting `target_type` into the `rowData` of the sce but nothing else to do that I can see!

sjspielman:
Noting that I need to go test the `development` merge, a lot of manual copy/paste there, but otherwise review changes are implemented!

bin/filter_sce_rds.R
```diff
@@ -99,13 +126,57 @@ alt_names <- altExpNames(filtered_sce)
...
+  # If `target_type` column is not present, assume all ADTs are targets
+  adt_barcode_df <- readr::read_tsv(
+    opt$adt_barcode_file, 
+    # if only 2 columns exist, only the first two col_names will be used
+    col_names = c("name", "barcode", "target_type")
+  ) 
+  if (!"target_type" %in% names(adt_barcode_df)) {
+    adt_barcode_df$target_type <- "target"
+  } 
+
```
jashapiro, line 150: Can we add a check like this? (possibly with sorting to make it more flexible)

```suggestion
 if( !all.equal(sort(adt_barcode_df$names), sort(rownames(altExp(filtered_sce, adt_exp)) ){
   stop("Mismatch between provided ADT barcode file and ADTs in SCE.")
 }
```

bin/filter_sce_rds.R
```diff
@@ -99,13 +126,59 @@ alt_names <- altExpNames(filtered_sce)
...
+    opt$adt_barcode_file, 
+    # if only 2 columns exist, only the first two col_names will be used
+    col_names = c("name", "barcode", "target_type")
+  ) 
+  if (!"target_type" %in% names(adt_barcode_df)) {
+    adt_barcode_df$target_type <- "target"
+  } 
+ if( !all.equal(sort(adt_barcode_df$names), sort(rownames(altExp(filtered_sce, adt_exp)) ){
+   stop("Mismatch between provided ADT barcode file and ADTs in SCE.")
+ }
```
jashapiro, line 152: My suggestion, my bad spacing
```suggestion
  # Check that barcodes match
  if( !all.equal(sort(adt_barcode_df$names), sort(rownames(altExp(filtered_sce, adt_exp)) ){
    stop("Mismatch between provided ADT barcode file and ADTs in SCE.")
  }

```

sjspielman: beat you to it.. https://github.com/AlexsLemonade/scpca-nf/pull/312/commits/40c5914367760d93863d2d8a3aaec017353792ff

sjspielman: honestly without the rainbow parentheses this seems ~impossible to get right!

jashapiro: And apparenlty parens (🤦🏼 )

modules/sce-processing.nf
```diff
@@ -162,12 +183,16 @@ workflow generate_merged_sce {
   // input is a channel with feature_meta, feature_quantdir, rna_meta, rna_quantdir
   take: feature_quant_channel
   main:
-
     feature_sce_ch = feature_quant_channel
       .map{it.toList() + [file(it[0].mito_file), file(it[0].ref_gtf)]}
```
jashapiro, line 187:
```suggestion
      .map{it.toList() + [file(it[2].mito_file), file(it[2].ref_gtf)]}
```

sjspielman:
🚀 for CITEseq data! Debugging for non-CITEseq data coming early next week!

modules/sce-processing.nf
```diff
@@ -150,9 +164,16 @@ workflow generate_sce {
     sce_ch = quant_channel
       .map{it.toList() + [file(it[0].mito_file), file(it[0].ref_gtf)]}
 
-    make_unfiltered_sce(sce_ch) \
-      | filter_sce
+    make_unfiltered_sce(quant_channel)
```
jashapiro, line 167: I think that this was the bug that snuck in with processing non-CITEseq samples 
```suggestion
    make_unfiltered_sce(sce_ch)
```
(@allyhawkins this is not in main)

sjspielman:
Yep that was definitely it, good Friday 👀 @jashapiro! Just ran through RNA-only, tested with `SCPCL000001`.
Screenshot 2023-05-15 at 8 02 25 AM

jashapiro (approved):
Looks good! And I think we can go ahead and merge this now that the 0.5.0 changes have been merged into `main`.

PR #314, opened 2023-05-10
Change mouse index to use ensembl 104

### LINKED ISSUE #308: Use ensembl version 104 for mouse reference
I didn't think about this much before, but for consistency with CellxGene, we should probably use `ensembl 104` for mouse too. 

See. https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#required-gene-annotations

This sparks a separate issue, but since we have not yet processed any mouse data yet, we should be fine as it just means making another reference!

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/306#discussion_r1186133667_

This should just entail changing the current reference files from version 109 to 104 for the mouse reference.

allyhawkins:
Closes #308 

Here I updated the reference that we use for mouse to use Ensembl 104 instead of 109. I changed the `version` in the tsv file and then re-ran the script to generate the `json` file. 

The final step was actually generating the index using `build-index.nf`. I did run into a few small issues that I had to fix along the way. The first was that the `generate_fasta` process was expecting the `fasta` outputs that were named with the index name and then appended with `.fa.gz` so I added that in. The second was that we were previously using `file("${meta.splici_index}").parent` to specify the publish directory, however whenever we used that it was grabbing the full path from my local directory and outputting the files in the S3 bucket with that organization rather than the correct parent directory from the json file. So I changed it to use the `meta.ref_dir/salmon_index` or the associated index folder name. If reviewers have other ideas I'm definitely open to them. 

I also had to remove outputting the `ref_fasta` and `ref_gtf` from the `generate_reference` process because otherwise it outputs those files to the main `ref_dir` which we don't need them to do. We also don't pass that output to anything else. 

Also, @jashapiro, I did look at the files in the bucket when I tested running this morning to see if they were set to public after being created, and they unfortunately were not. So the updated settings did not seem to work. They do make new folders within the existing subfolders so probably that's why.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -17,8 +17,8 @@ process generate_reference{
     tuple path(splici_fasta), path(spliced_cdna_fasta), emit: fasta_files
     tuple path("annotation/*.gtf.gz"), path("annotation/*.tsv"), path("annotation/*.txt"),  emit: annotations
   script:
-    splici_fasta = "fasta/" + file("${meta.splici_fasta}").name
-    spliced_cdna_fasta =  "fasta/" + file("${meta.splici_cdna_fasta}").name
+    splici_fasta = "fasta/" + file("${meta.splici_index}").name + ".fa.gz"
+    spliced_cdna_fasta =  "fasta/" + file("${meta.salmon_bulk_index}").name + ".fa.gz"
     """
     make_reference_fasta.R \
       --gtf ${gtf} \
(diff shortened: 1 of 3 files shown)


jashapiro:
> Also, @jashapiro, I did look at the files in the bucket when I tested running this morning to see if they were set to public after being created, and they unfortunately were not. So the updated settings did not seem to work. They do make new folders within the existing subfolders so probably that's why.

The files may not appear to be public, but I wonder if they can be accessed publicly, which is all we really need.

jashapiro (approved):
LGTM!

I think the permissions thing is actually solved by the change I made; while when I look at the console the files are still listed as read/write owner/owner, I am able to list and download the files from an account with no general access.

PR #316, opened 2023-05-10
Remove cell typing from main workflow

### LINKED ISSUE #313: Remove cell typing from main workflow
> This does feel like a permissions issue, and I suspect it is related to Ally working on cell typing (the singleR model files changed today) We should probably comment out the cell typing process for now, as it is not fully implemented.

Just confirming that this was me working on the `build-index.nf` workflow. I think because we have cell typing separate right now, we should remove running it from both the index and the main workflow until it's fully developed as we discussed in slack. Just commenting here for posterity.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpca-nf/issues/312#issuecomment-1540852615_

We will want to make sure that this is removed until we release the new changes to support multiple organisms. We should create a branch off `dev` to remove the call to cell typing and then after making the release/ when we are ready to add cell typing back in we can revert those changes.

allyhawkins:
Closes #313 

This just removes the call to build celltype refs and to add cell types within the main workflow. These can both still be run independently and we will add them back in after we are confident in adding annotations to the workflow.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -136,7 +136,4 @@ workflow {
   cellranger_index(params.ref_fasta, params.ref_gtf, params.assembly)
   // create star index
   index_star(params.ref_fasta, params.ref_gtf, params.assembly)
-
-  // build celltype references
-  build_celltype_ref()
 }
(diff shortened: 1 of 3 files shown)


jashapiro (approved):
LGTM!

PR #317, opened 2023-05-10
Skip feature mapping

### LINKED ISSUE #315: Add RAD checkpoint for barcode features
Because feature mapping is fast, we hadn't bothered with keeping a checkpoint file for those steps. However, for transmitting results without underlying sequence data from external users, it would be nice if we also had checkpoints for the feature data RADs as well.

We probably could keep the `alevin-fry` output matrices instead of the RAD files, as I do not expect much change in the quantification steps for feature barcodes, but either way, we should probably have _some_ checkpoint for these files, modeled after our other checkpoint steps.

allyhawkins:
Closes #315 

Here I updated the `af-feature` module to publish the rad directories for feature libraries. I also added the ability to skip mapping for feature libraries similar to how we do it for other mapping steps. I did test this with both skipping mapping and running mapping and it worked as expected. 

Also noting that this is stacked on #316 because of [github actions not working right now](https://www.githubstatus.com/).

### DIFF
--- a/modules/af-features.nf
+++ b/modules/af-features.nf
@@ -124,8 +124,7 @@ workflow map_quant_feature{
       .unique()
     index_feature(feature_barcodes_ch)
 
-    // create tuple of [metadata, [Read1 files], [Read2 files]]
-    // We start by including the feature_barcode file so we can combine with the indices, but that will be removed
+    // add the publish directories to the channel and branch based on existing rad files
     feature_ch = feature_channel
       .map{it.feature_rad_publish_dir = "${params.checkpoints_dir}/rad/${it.library_id}";
            it.feature_rad_dir = "${it.feature_rad_publish_dir}/${it.run_id}-features";
@@ -137,6 +136,8 @@ workflow map_quant_feature{
 
     // pull out files that need to be repeated
     feature_reads_ch = feature_ch.make_rad
+      // create tuple of [metadata, [Read1 files], [Read2 files]]
+      // We start by including the feature_barcode file so we can combine with the indices, but that will be removed
       .map{meta -> tuple(meta.feature_barcode_file,
                          meta,
                          file("${meta.files_directory}/*_R1_*.fastq.gz"),
@@ -145,6 +146,8 @@ workflow map_quant_feature{
       .combine(index_feature.out, by: 0) // combine by the feature_barcode_file (reused indices, so combine is needed)
       .map{it.drop(1)} // remove the first element (feature_barcode_file)
 
+    // // if the rad directory has been created and repeat_mapping is set to false
+    // create tuple of metdata map (read from output) and rad_directory to be used directly as input to alevin-fry quantification
     feature_rad_ch = feature_ch.has_rad
       .map{meta -> tuple(Utils.readMeta(file("${meta.feature_rad_dir}/scpca-meta.json")),
                          file(meta.feature_rad_dir)
--- a/modules/af-features.nf
+++ b/modules/af-features.nf
@@ -128,6 +128,7 @@ workflow map_quant_feature{
     feature_ch = feature_channel
       .map{it.feature_rad_publish_dir = "${params.checkpoints_dir}/rad/${it.library_id}";
            it.feature_rad_dir = "${it.feature_rad_publish_dir}/${it.run_id}-features";
+           it.barcode_file = "${params.barcode_dir}/${params.cell_barcodes[it.technology]}";
            it}
       .branch{
           has_rad: !params.repeat_mapping && file(it.feature_rad_dir).exists()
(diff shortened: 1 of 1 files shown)


modules/af-features.nf
```diff
@@ -124,24 +124,46 @@ workflow map_quant_feature{
...
     index_feature(feature_barcodes_ch)
 
-    // create tuple of [metadata, [Read1 files], [Read2 files]]
-    // We start by including the feature_barcode file so we can combine with the indices, but that will be removed
-    feature_reads_ch = feature_channel
+    // add the publish directories to the channel and branch based on existing rad files
+    feature_ch = feature_channel
+      .map{it.feature_rad_publish_dir = "${params.checkpoints_dir}/rad/${it.library_id}";
+           it.feature_rad_dir = "${it.feature_rad_publish_dir}/${it.run_id}-features";
+           it}
```
jashapiro, line 131: For parallelism with the af-rna, we might define the barcode file here.
```suggestion
           it.feature_rad_dir = "${it.feature_rad_publish_dir}/${it.run_id}-features";
           it.barcode_file = "${params.barcode_dir}/${params.cell_barcodes[it.technology]}";
           it}
```

modules/af-features.nf
```diff
@@ -124,24 +124,46 @@ workflow map_quant_feature{
...
+    feature_rad_ch = feature_ch.has_rad
+      .map{meta -> tuple(Utils.readMeta(file("${meta.feature_rad_dir}/scpca-meta.json")),
+                         file(meta.feature_rad_dir)
+                         )}
 
     // run Alevin on feature reads
     alevin_feature(feature_reads_ch)
+
+    // combine output from running alevin step with channel containing libraries that skipped creating RAD file
+    all_feature_rad_ch = alevin_feature.out.mix(feature_rad_ch)
```
jashapiro, line 160: If we do add the barcode file to meta, we can add to the channel here. 

(My recent reading suggests we can add it as a bare string, and the fact that it would be taken in as a `path` means we can skip the  `file()` statement)

```suggestion
    all_feature_rad_ch = alevin_feature.out.mix(feature_rad_ch)
      .map{it.toList() + it[0].barcode_file}
```

jashapiro, line 160: If we do add the barcode file to meta, we can add to the channel here. 

(My recent reading suggests we can add it as a bare string, and the fact that it would be taken in as a `path` means we can skip the  `file()` statement)

```suggestion
    all_feature_rad_ch = alevin_feature.out.mix(feature_rad_ch)
      .map{it.toList() + it[0].barcode_file}
```

modules/af-features.nf
```diff
@@ -124,24 +124,46 @@ workflow map_quant_feature{
...
+
+    cellbarcode_ch = feature_channel
+      .map{file("${params.barcode_dir}/${params.cell_barcodes[it.technology]}")}
+
     // quantify feature reads
-    fry_quant_feature(alevin_feature.out, cellbarcode_ch)
+    fry_quant_feature(all_feature_rad_ch, cellbarcode_ch)
```
jashapiro, line 166: If we have added the barcode file to the channel, then we can make this a single input (`fry_quant_feature` needs updating to do this, but I can't comment there)
```suggestion
    fry_quant_feature(all_feature_rad_ch)
```

modules/af-features.nf
```diff
@@ -124,24 +124,46 @@ workflow map_quant_feature{
...
+                         )}
 
     // run Alevin on feature reads
     alevin_feature(feature_reads_ch)
+
+    // combine output from running alevin step with channel containing libraries that skipped creating RAD file
+    all_feature_rad_ch = alevin_feature.out.mix(feature_rad_ch)
+
+    cellbarcode_ch = feature_channel
+      .map{file("${params.barcode_dir}/${params.cell_barcodes[it.technology]}")}
```
jashapiro, line 163: and if we have the barcode file in the path, we don't need this!

```suggestion
```

jashapiro (commented):
This looks good, but I was comparing it to what we do in `af-rna.nf` and noted some differences. I think it would be worth making them as parallel as possible, which means writing the barcode file into `meta` at the start and then working with the updates that flow from that...

allyhawkins:
I made the edits to incorporate the barcode file to `meta` earlier, like we do in `af-rna` like you suggested. However, I still needed to use the `file()` option to pass the barcode file to the fry quant process. I kept getting `Execution aborted due to unexpected error` but then when I used `file()` it worked.

jashapiro (approved):
LGTM!

> However, I still needed to use the `file()` option to pass the barcode file to the fry quant process. I kept getting Execution aborted due to unexpected error but then when I used `file()` it worked.

I guess I still don't fully understand the rules.

PR #321, opened 2023-05-12
Update to v0.5.0 and tag scpcaTools

allyhawkins:
Just updating the version numbers throughout and then also making sure `scpcaTools` is using the most recent tagged version.


jashapiro (approved):
LGTM!

PR #326, opened 2023-05-15
Add CPU & memory checks and stubs

### LINKED ISSUE #151: Idea: Add resource limits to facilitate local running
We claim in the docs for external users that local running is possible, but having moved the cpu and memory requirements to the process base this may not be true. If we try to run locally on a machine with insufficient resources (i.e. if we request more cores than available on the local machine), the steps will simply fail.

An example error message:

```
Error executing process > 'map_quant_rna:alevin_rad (SCPCR999991-rna)'

Caused by:
  Process requirement exceed available CPUs -- req: 12; avail: 4
```

We could also solve this with a parameter like `max_cpu` and some comparisons in the resource tags to enforce these limits (which is what [nf-core does](https://github.com/nf-core/tools/blob/e0d00d3e09a996f59eb5ce6ebc4795452a5b1d39/nf_core/pipeline-template/nextflow.config#L178-L180)). This allows a single place to change a parameter in their config file that we could point people to. 

We could also create a `local` profile which sets limits much lower (this could be combined with the previous option). The trouble there is that we can't really run all the steps with less than 32GB RAM (and some datasets may require more on failure) and they will be slow with small numbers of CPUs, so it isn't really practical except for test datasets. 

We could also just change docs to describe minimum computing requirements, which is probably the right option in my estimation, but I wanted to put other options on the table for discussion.

### LINKED ISSUE #290: Add stubs to workflow
When we get a chance we should consider going back to add stubs to process's in our workflow. 

              Just a note that we should probably add these to scpca-nf as we go. Sometime when we have a chance to make sure everything has one...

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/CART_Monje/pull/131#discussion_r1157230370_

jashapiro:
The goal of the PR is to enable local running of the workflow, including basic checks of workflow logic using stubs.

Closes #290 
Closes #151 (though docs are needed here)

To do this, I added `stub:` blocks to all current processes, with the exception of the cell type classification process, which is not currently in the main workflow.

Unfortunately, adding stubs is insufficient for allowing local running, as the memory and cpu requirements are still used for the stub. So to fix that, I added some very simple functions inspired by nf-core to modify our memory and cpu settings, checking new `params` for the max memory and cpu and returning the lesser of the two values. 

So to run locally, you can now do something like:

```
nextflow main.nf -profile ccdl -stub --max_cpus 2 --max_memory 2.GB
```

I put this parameter in the `process_base.config` file, as that is where they are used, but it might make sense to move it to the main `nextflow.config` file to make it easier to users to find and set for their own configuration. On the other hand, we encourage people to make their own `config` file, and any value they put there will override anyway, so it may be more a documentation question. Opinions welcome.

If this is working for others, I will update the docs to include these parameters (and that you can run with less RAM, though performance may be affected and no guarantees for something like Cell Ranger).

I expect to leave the `-stub` functionality undocumented, except perhaps in development instructions, as I don't think that is something we should expect end users to use.

One other thing I noticed while doing this (since I was hitting almost every process) was that we have some inconsistency in the number of spaces for indentation; I might want to fix that in this PR, but I didn't want too many modified lines for the review!

### DIFF
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -25,7 +25,8 @@
 
  ## Overview
 
-Using `scpca-nf` to process your own single-cell and single-nuclei RNA-seq data requires access to a high performance computing (HPC) environment that can ideally accommodate up to 24 GB of RAM and 12 CPUs.
+Using `scpca-nf` to process your own single-cell and single-nuclei RNA-seq data requires access to a high performance computing (HPC) environment that can accommodate up to 24 GB of RAM and 12 CPU cores.
+Some datasets and processes (genetic demultiplexing and spatial transcriptomics) may require additional resources, and our default configuration allows up to 96 GB of RAM and 24 CPU cores.
 While the workflow does support scaling down requirements in lower-resource environments, we have not tested extensively in those conditions, and some components may fail.
 After identifying the system that you will use to execute the nextflow workflow, you will need to follow the steps outlined in this document to complete the set up process.
 Here we provide an overview of the steps you will need to complete:
@@ -150,7 +151,7 @@ Note that *workflow* parameters such as `--run_metafile` and `--outdir` are deno
 
 There are also a number of optional parameters that can be set, either at the command line or in a configuration file, including:
 
-- `max_cpus`: the maximum number of CPUs to use for a single process (default: 24)
+- `max_cpus`: the maximum number of CPU cores to use for a single process (default: 24)
 - `max_memory`: the maximum amount of memory to use for a single process (default: `96.GB`)
 
 Other customizable parameters can be found in the `nextflow.config` file in the repository.
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -25,7 +25,8 @@
 
  ## Overview
 
-Using `scpca-nf` to process your own single-cell and single-nuclei RNA-seq data requires access to a high performance computing (HPC) environment that can accomodate up to 24 GB of RAM and 12 CPUs.
+Using `scpca-nf` to process your own single-cell and single-nuclei RNA-seq data requires access to a high performance computing (HPC) environment that can ideally accommodate up to 24 GB of RAM and 12 CPUs.
+While the workflow does support scaling down requirements in lower-resource environments, we have not tested extensively in those conditions, and some components may fail.
 After identifying the system that you will use to execute the nextflow workflow, you will need to follow the steps outlined in this document to complete the set up process.
 Here we provide an overview of the steps you will need to complete:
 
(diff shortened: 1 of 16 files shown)


allyhawkins (commented):
This overall looks good to me. And worked fine.

In regards to your question about where to put the parameters, I think you could leave them where they are now, but I would add them to the template config file that we provide to users. Then I would include instructions in the docs on if you are running locally, that you want to change them. I also would comment around them in the template config file that they only need to be changed in the case of running locally and maybe even put some warnings regarding running spatial data. 

I'm fine with leaving running with the `-stub` option undocumented as that's specific to nextflow development, but we should definitely make sure the two new parameters are documented.

jashapiro:
I added some docs about the new parameters as well as a few places that I ran into little mistakes. This should be ready for another review!

external-instructions.md
```diff
@@ -147,6 +148,14 @@ These parameters can be set at the command line using `--run_metafile <path to r
 
 Note that *workflow* parameters such as `--run_metafile` and `--outdir` are denoted at the command line with double hyphen prefix, while options that affect Nextflow itself have only a single hyphen.
 
+There are also a number of optional parameters that can be set, either at the command line or in a configuration file, including:
+
+- `max_cpus`: the maximum number of CPUs to use for a single process (default: 24)
+- `max_memory`: the maximum amount of memory to use for a single process (default: `96.GB`)
```
allyhawkins, line 154: Just noting the inconsistencies between the default in the template and the default in the config file. I think they should be the same?

external-instructions.md
```diff
@@ -25,7 +25,8 @@
 
  ## Overview
 
-Using `scpca-nf` to process your own single-cell and single-nuclei RNA-seq data requires access to a high performance computing (HPC) environment that can accomodate up to 24 GB of RAM and 12 CPUs.
+Using `scpca-nf` to process your own single-cell and single-nuclei RNA-seq data requires access to a high performance computing (HPC) environment that can ideally accommodate up to 24 GB of RAM and 12 CPUs.
```
allyhawkins, line 28: We say that these are the requirements but then have defaults that are above this. We should probably explicitly state that this is a minimum and you can go over.

external-instructions.md
```diff
@@ -157,8 +166,12 @@ We could first create a file `my_config.config` (or a filename of your choice) w
 // my_config.config
 params.run_metafile = '<path to run_metafile>'
 params.outdir = '<path to output>'
+params.max_cpus = 12
+params.max_memory = 24.GB
```
allyhawkins, line 170: Different suggestion here than the default we mention above. Is that on purpose?

allyhawkins (commented):
Mostly noted some inconsistencies in what we suggest to use for the memory and cpus. Maybe that's on purpose, but if we are suggesting different things than what we set as defaults we should probably explain it.

external-instructions.md
jashapiro: I wasn't sure; If everything goes to plan we should be fine with lower suggestions for users, but it probably does make sense to make these the defaults.

external-instructions.md
jashapiro: Never mind, I'll make it match

allyhawkins (approved):
Docs look good and ran this locally after the filtering change was added in and everything worked as expected 👍

PR #327, opened 2023-05-16
Update external docs to explain `feature_barcode_file` third column

### LINKED ISSUE #299: Update external docs with ADT information
Related to #298 
We'll need to update external docs regarding ADT processing. These sections will need updating, at least:
- [x] https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md#prepare-the-metadata-file
  - Specifically, third column in `feature_barcode_file` for target type  
- [x] https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md#special-considerations-for-specific-data-types

sjspielman:
Closes #299 

This PR adds documentation to explain that a third column can optionally be included in the `feature_barcode_file` to specify the ADT "target type," as I've been calling it.

Questions for reviewers..
- Did I miss anywhere else in these docs that need to be updated?
- Is it clear enough that the potential values are `target`, `neg_control`, and `pos_control` or would you suggest rephrasing there?
- Should we also say something like "at this time, positive controls are not considered in the pipeline"?

### DIFF
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -127,7 +127,7 @@ The following columns may be necessary for running other data modalities (CITE-s
 
 | column_id       | contents                                                       |
 |-----------------|----------------------------------------------------------------|
-| `feature_barcode_file` | path/uri to file containing the feature barcode sequences (only required for CITE-seq and cellhash samples)  |
+| `feature_barcode_file` | path/uri to file containing the feature barcode sequences (only required for CITE-seq and cellhash samples); for CITE-seq samples, this file can optionally indicate whether antibodies are targets or controls.  |
 | `feature_barcode_geom` | A salmon `--read-geometry` layout string. <br> See https://github.com/COMBINE-lab/salmon/releases/tag/v1.4.0 for details (only required for CITE-seq and cellhash samples) |
 | `slide_section`   | The slide section for spatial transcriptomics samples (only required for spatial transcriptomics) |
 | `slide_serial_number`| The slide serial number for spatial transcriptomics samples (only required for spatial transcriptomics)   |
@@ -313,6 +313,18 @@ For example:
 TAG01	CATGTGAGCT
 TAG02	TGTGAGGGTG
 ```
+
+For CITE-seq data, you can optionally include a third column in the `feature_barcode_file` to indicate whether the given antibody is a true target (`target`) or a control, either a spike-in positive control (`pos_control`) or negative control (`neg_control`).
+
+For example, the following shows that two antibodies are targets and one is a negative control:
+
+```
+TAG01	CATGTGAGCT	target
+TAG02	TGTGAGGGTG	neg_control
+TAG03	GTAGCTCCAA	target
+```
+
+If negative control antibodies are indicated, this information will be considered during post-processing filtering and normalization.
 
 
 ### Multiplexed (cellhash) libraries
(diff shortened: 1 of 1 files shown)


external-instructions.md
```diff
@@ -314,6 +314,18 @@ TAG01	CATGTGAGCT
 TAG02	TGTGAGGGTG
 ```
 
+For CITE-seq data, you can optionally include a third column in the `feature_barcode_file` to indicate whether the given antibody is a true target (`target`) or a control, either a spike-in positive control (`pos_control`) or negative control (`neg_control`).
```
jashapiro, line 317: I think this would be more clear that these are the only three valid values
```suggestion
For CITE-seq data, you can optionally include a third column in the `feature_barcode_file` to indicate the purpose of each antibody, which can take one of the following three values:

- `target`:  antibody is a true target 
- `neg_control`: a negative control antibody
- `pos_control`: a spike-in positive control
```

What happens if there are invalid values? Do we give a warning/failure? (I've forgotten)

We should also probably say what happens if this column is missing (everything is a target).

external-instructions.md
```diff
@@ -314,6 +314,18 @@ TAG01	CATGTGAGCT
...
+
+For example, the following shows that two antibodies are targets and one is a negative control:
+
+```
+TAG01	CATGTGAGCT	target
+TAG02	TGTGAGGGTG	neg_control
+TAG03	GTAGCTCCAA	target
+```
+
+If negative control antibodies are indicated, this information will be considered during post-processing filtering and normalization.
```
jashapiro, line 327:
```suggestion
If there are negative control antibodies, these will be taken into account during post-processing filtering and normalization.
Positive controls are currently unused, but this label will be included in final output files.
```

jashapiro (commented):
I think we should be explicit about the allowed values, so I suggested a change there. 

I also wanted just a bit more information about what "optional" really means and I did think we should mention that positive controls are not used (but the information is retained)

external-instructions.md
sjspielman: > What happens if there are invalid values? Do we give a warning/failure?

This is a great point, since we actually "didn't" handle this at all, but we should!
https://github.com/AlexsLemonade/scpca-nf/blob/1bece995d363813ccfbf4720851f2f8739e13b5f/bin/filter_sce_rds.R#L142-L157

The only action that is taken on these values is to find the ones annotated as `neg_control`. But, we should give a warning if invalid names were found and tell users we've therefore assumed it's a target (aka the default). I'll open up an issue for this.

sjspielman:
Given parallel updates in #331, I think this is now about where we want to end up.

external-instructions.md
```diff
@@ -328,6 +328,26 @@ TAG01	CATGTGAGCT
 TAG02	TGTGAGGGTG
 ```
 
+For CITE-seq data, you can optionally include a third column in the `feature_barcode_file` to indicate the purpose of each antibody, which can take one of the following three allowed values:
```
jashapiro, line 331:
```suggestion
For CITE-seq data, you can optionally include a third column in the `feature_barcode_file` to indicate the purpose of each antibody, which can take one of the following three values:
```

jashapiro (approved):
LGTM, but I removed an `allowed` as we are being a bit more permissive, and "one of three values" seemed definitive enough.

external-instructions.md
sjspielman: > I removed an allowed as we are being a bit more permissive

agreed!

Issue #328, opened 2023-05-18
Make genetic demultiplexing optional


jaclyn-taroni:
_For now, this is a placeholder issue to be filled in by someone closer to the implementation details._

jashapiro:
Noting that this should be created as a branch off `main` and filed as PR thereto. There are a number of other changes in `development` that we will not want to incorporate yet.

jashapiro:
Closed now that development has the same code

PR #329, opened 2023-05-23
Make genetic demultiplexing optional

jashapiro:
Addresses #328, but I do not want to close that until these changes are merged both into `main` and `development`.

Note that this incorporates changes in #326 that were used for testing... I could strip those out if we want to.

The main changes are to make the workflow more robust to missing bulk samples, and to skip genetic demultiplexing in the case when those bulk samples are not present. 

I do this by adding a new list of library ids `genetic_multiplex_libs` that is filtered to only those multiplexed libraries for which every sample has matching bulk RNA data. That is then used to filter the input to the genetic demultiplexing part of the workflow. The cellhash demultiplexing has already been performed (this is done for any library with cellhash features already), so we can just skip that extra step and then proceed as before. This also fixes some potential errors that could have occurred when _some_ bulk samples were present: I actually don't know if the workflow would have failed in that case or just performed genetic demultiplexing ignoring the missing data. Either is not ideal!

One thing that is not yet present is a manual `params`/cli option to skip genetic demultiplexing even if bulk data is present. I don't think it would be difficult to add, but I also wasn't sure it was necessary. Let me know what you think.

I checked that the QC report already (mostly) accommodates missing genetic demux. I think that only thing we might need to change is some of the wording, but as the `main` branch does not yet have the changes that incorporate the in-workflow templates, it is probably best to make those changes in `development`.

Testing is ongoing (hence draft status), but I thought this was probably ready for an initial look, as the tests seem to be going well.

### DIFF
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -1,25 +1,25 @@
 // Resource maximum settings
 params.max_cpus = 24
-params.max_memory = 128.GB
+params.max_memory = 96.GB
 
 process{
-  memory = { check_memory(4.GB * task.attempt) }
+  memory = {check_memory(4.GB * task.attempt)}
 
   errorStrategy = { task.attempt < 3 ? 'retry' : 'finish' }
   maxRetries = 2
   maxErrors = '-1'
 
   withLabel: mem_8 {
-    memory = { check_memory(8.GB * task.attempt) }
+    memory = {check_memory(8.GB * task.attempt)}
   }
   withLabel: mem_16 {
-    memory = { check_memory(16.GB * task.attempt) }
+    memory = {check_memory(16.GB * task.attempt)}
   }
   withLabel: mem_24 {
-    memory = { check_memory(24.GB * task.attempt) }
+    memory = {check_memory(24.GB * task.attempt)}
   }
   withLabel: mem_32 {
-    memory = { check_memory(32.GB * task.attempt) }
+    memory = {check_memory(32.GB * task.attempt)}
   }
   withLabel: cpus_2  {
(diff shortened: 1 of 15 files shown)


allyhawkins:
This looks good to me. I think if you want to keep the changes that add in the `stubs` and memory limits that's fine, but I think we also need the documentation associated with #326 to be added in as well. Otherwise I would remove it. 

Also I looked at the QC report and we would probably be okay except we may want to note if the `vireo` results are missing then that library didn't have bulk RNA-seq data to perform demultiplexing. I'm good with those changes happening in `development`, but then are we planning on adding in the templates to the next release? If so, that means a release of `scpcaTools` too.

jashapiro:
> 



> This looks good to me. I think if you want to keep the changes that add in the `stubs` and memory limits that's fine, but I think we also need the documentation associated with #326 to be added in as well. Otherwise I would remove it.

I took them out... there is some risk of strange conflicts now, but I'll hopefully be able to handle that with careful merges

> Also I looked at the QC report and we would probably be okay except we may want to note if the `vireo` results are missing then that library didn't have bulk RNA-seq data to perform demultiplexing. I'm good with those changes happening in `development`, but then are we planning on adding in the templates to the next release? If so, that means a release of `scpcaTools` too.

jashapiro:
Okay, I have done a bunch of tests with various data types, and I'm pretty confident that things are working as expected, so I am marking this ready for review/merge!

allyhawkins (approved):
LGTM! 

Just noting that we will also want to update the docs to note that bulk sequencing is no longer required too.

PR #331, opened 2023-05-24
Check ADT type values

### LINKED ISSUE #330: Check ADT target type for invalid values
In #327, we realized that we have not checked that valid values are supplied for the ADT type. The valid values are `target`, `neg_control`, and `pos_control`.

We should check the values and print a _warning_ if invalid values are found, and state that they will be considered targets; I think it seems a little harsh to _fail_ on this..

sjspielman:
Closes #330 

This PR adds a quick check for ADT target type values, printing a nice warning if anything is invalid that it will default to `target`. I added a few comments that seemed helpful along the way.

### DIFF
--- a/bin/filter_sce_rds.R
+++ b/bin/filter_sce_rds.R
@@ -152,9 +152,16 @@ if (!is.null(ambient_profile)) {
   }
   # Check for valid target types and warn if invalid found
   allowed_adt_values <- c("target", "neg_control", "pos_control")
-  if (!all(adt_barcode_df$target_type) %in% allowed_adt_values) {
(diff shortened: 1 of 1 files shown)


jashapiro (commented):
I don't like changing user inputs, even if they are "wrong", so I suggested removing that chunk of code and softening some of the warnings.

sjspielman:
Thanks for the feedback, agreed all around! Confirmed the code runs through fine ✅

jashapiro (approved):
LGTM!

sjspielman:
For some reason #330 was not automatically closed ¯\_(ツ)_/¯

PR #332, opened 2023-05-25
Add stub config and automatic testing

jashapiro:
Inspired by a conversation with @sjspielman, this PR adds a full "stub" profile and test files to the repository, and, as a treat, a new GHA to test the workflow logic with every PR. 

This is partly a proof-of concept, as it does not yet contain tests of all paths through the workflow, but adding more paths should be pretty straightforward. 

The basic concept is that you should now be able to run a test run with the following command:

```
nextflow run main.nf -stub -profile stub
```

This will start a local run using a set of dummy files, including runs, metadata, references, etc. that are found in the `test` directory, saving us from having to stage large files just to test that the workflow logic is sound. 

Right now the test files are one rna-only and one cite-seq data set. Adding more files should just mean adding rows to the stub-metadata.tsv file, and appropriately named dummy files.

As part of testing this, I found a place where we were missing a `file()` wrapper, so that is added in as well. (I also think I may have figured out the logic there: if we have an `s3://` input, we don't _need_ the `file()`wrapper, but we do need it for local files?)

A separate PR will document this mode of testing in the internal instructions, and we can add to the tests over time.


.github/workflows/nextflow-stub-check.yaml
```diff
@@ -0,0 +1,28 @@
...
+        uses: docker://nextflow/nextflow:21.10.6
+        with:
+          args: nextflow -log stub-run.log run main.nf -stub -profile stub -ansi-log false
+      
+      - name: Upload nextflow log
+        if: ${{ always() }}
+        uses:  actions/upload-artifact@v3
+        with:
+          name: nextflow-log
+          path: stub-run.log
```
sjspielman, line 28: github requests a new line

test/stub-metadata.tsv
```diff
@@ -0,0 +1,4 @@
+scpca_run_id	scpca_library_id	scpca_sample_id	scpca_project_id	technology	seq_unit	sample_reference	files_directory	feature_barcode_file	feature_barcode_geom	slide_section	slide_serial_number
+STUBR01	STUBL01	STUBS01	STUBP01	10Xv2	cell	stub	test/runs/STUBR01	NA	NA	NA	NA
```
sjspielman, line 2: Upon reading these ids I made the important conclusion that this will be referred to as the stubble workflow.

sjspielman (approved):
Wow, who know I was so inspirational 🌟 ! This is a really neat addition! I've locally run `nextflow -stub` and also did a run with a citeseq dataset to make extra sure everything is still behaving nicely with the `file()` fix ✅ 

All looks good to move along, just needs one new line in the workflow to appease github.

test/stub-metadata.tsv
jashapiro: 💯

Issue #334, opened 2023-05-25
More stub tests


jashapiro:
In https://github.com/AlexsLemonade/scpca-nf/pull/332, I added a stub workflow that can automatically test that workflow logic is sound without requiring a full workflow run.

Currently, the test only tests the paths used for an RNA-seq run and a CITE-seq run. We should add more test cases by adding new entries to `tests/stub-metadata.tsv` along with the input files to support those runs (these can generally just be blank files created with `touch`)

Some of the data types we should add:

- bulk RNA-seq
- spatial (this may require setting a docker image for Space Ranger in `config/profile_stub.config`, but that image does _not_ need to contain Space Ranger)
- cellhashed with genetic demultiplexing
- cellhashed without genetic demultiplexing

We also may want to consider whether we need to test any other particular parameter flags that we set. At the moment, the only one I can think of would be `--skip-genetic-demux`, but we would really only be testing that it ran successfully unless we added some extra logic.

We could also add some logic to test that the expected output files/directories are generated after each run, but that gets more complicated as we add more inputs, so I'm not sold on it.

jashapiro:
I think this was closed by #376

PR #335, opened 2023-05-25
Add cellhash-only to development branch

### LINKED ISSUE #328: Make genetic demultiplexing optional

jashapiro:
This was originally just intended to merge changes from #329 to the `development` branch, and as such this

Closes #328

While I was at it, I also decided to partially address #334, so here I am adding lines to the `stub-metadata.tsv` file that include paired- and single-end bulk data, as well as cellhash multiplexed data with and without accompanying bulk data.

This test uncovered another missing `file()` wrapper, so that is now added in as well (do we want to cherry pick that fix over to `main` before we cut that release)?


jashapiro:
Thought I had requested review for this a while ago, but it was just GitHub suggesting a reviewer. 🤦

modules/bulk-salmon.nf
```diff
@@ -160,7 +160,7 @@ workflow bulk_quant_rna {
                 file(it[1][0].t2g_bulk_path)]}
 
         // create tsv file and combined metadata for each project containing all libraries
-        merge_bulk_quants(grouped_salmon_ch, params.run_metafile)
+        merge_bulk_quants(grouped_salmon_ch, file(params.run_metafile))
```
allyhawkins, line 163: Since I'm just reviewing this now, I didn't see this before I did the release yesterday, do we want to get this over to main and into that release...?

test/stub-metadata.tsv
```diff
@@ -1,4 +1,11 @@
 scpca_run_id	scpca_library_id	scpca_sample_id	scpca_project_id	technology	seq_unit	sample_reference	files_directory	feature_barcode_file	feature_barcode_geom	slide_section	slide_serial_number
 STUBR01	STUBL01	STUBS01	STUBP01	10Xv2	cell	stub	test/runs/STUBR01	NA	NA	NA	NA
-STUBR02	STUBL02	STUBS02	STUBP02	10Xv3.1	cell	stub	test/runs/STUBR01	NA	NA	NA	NA
-STUBR03	STUBL02	STUBS02	STUBP02	CITEseq_10Xv3	cell	stub	test/runs/STUBR02	test/references/barcodes/adt.stub.tsv	2[1-15]	NA	NA
+STUBR02	STUBL02	STUBS02	STUBP02	10Xv3.1	cell	stub	test/runs/STUBR02	NA	NA	NA	NA
+STUBR03	STUBL02	STUBS02	STUBP02	CITEseq_10Xv3	cell	stub	test/runs/STUBR03	test/references/barcodes/adt.stub.tsv	2[1-15]	NA	NA
+STUBR04	STUBL04	STUBS04	STUBP02	paired_end	bulk	stub	test/runs/STUBR04	NA	NA	NA	NA
+STUBR08	STUBL08	STUBS10	STUBP03	single_end	bulk	stub	test/runs/STUBR08	NA	NA	NA	NA
```
allyhawkins, line 6: This is really minor, but is there a reason you went from 4 to 10 for the samples?

allyhawkins (approved):
This looks good to me. I just had a question about what we want to do regarding the `file()` you added here and then just curious about the jump in sample/library numbers, but ultimately the numbers used don't really matter.

modules/bulk-salmon.nf
jashapiro: We probably don't need to, but it might be worth it. While this comes up when using a local file rather than a remote address, and I think we can expect external users to do that, we don't expect external users at this point to be including bulk samples, which is the only path this affects. So I'm torn.

jashapiro (commented):
Thanks for the review! Sorry the request was delayed. See below about my thoughts on porting the file change...

modules/bulk-salmon.nf
jashapiro: It won't fail for our process, as we use S3 for the `run_metafile`. Just if the metafile is local, I think.

allyhawkins: Sorry I clearly did not fully read your previous comment. I think if its just for local files then we are okay.

PR #341, opened 2023-05-31
Add some documentation on skipping genetic demultiplexing

allyhawkins:
I realized we hadn't removed the text around requiring RNA-seq for all samples and always running genetic demultiplexing. I added a little bit of language here including instructions on how to skip genetic demultiplexing. I wasn't quite sure where in the multiplexed section this should go, so definitely open to other ideas.

### DIFF
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -317,15 +317,18 @@ TAG02	TGTGAGGGTG
 
 ### Multiplexed (cellhash) libraries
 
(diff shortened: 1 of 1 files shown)


jashapiro (commented):
I made a suggestion to split out the genetic demux a bit more... Couldn't make it as a direct suggestion though, so you will have to implement it yourself, I guess!

jashapiro (approved):
LGTM!

PR #343, opened 2023-05-31
Add CITEseq postprocessing to QC report

### LINKED ISSUE #336: Update QC report based on CITE-seq post-processing
Following changes made to incorporate CITE-seq post-processing into the pipeline, we'll need to update QC report.

#### [`qc_report.rmd`](https://github.com/AlexsLemonade/scpca-nf/blob/development/templates/qc_report/qc_report.rmd) 
  - Update relevant phrasing throughout
    - It would be nice to be able to have logic for whether certain text appears, for example here: https://github.com/AlexsLemonade/scpca-nf/blob/6953ae72d0b8600b09bc8d50205260bb611a4e7e/templates/qc_report/qc_report.rmd#L473 
  - The string for `scpca_filter_method` metadata has now changed, so think about that...
  -  Some changes needs to happen to reflect that filtering in the `_processed.rds` object is filtered for _both_ RNA and ADTs.
  - Update the `basic_statistics` table appropriately. Needs an `if` to check for CITE-seq and rows for more stats https://github.com/AlexsLemonade/scpca-nf/blob/6953ae72d0b8600b09bc8d50205260bb611a4e7e/templates/qc_report/qc_report.rmd#L233-L238
  - TBD: Should this plot only show RNA-based filtering? As written, it will include ADT-based filtering as well, which seems reasonable to me if there are ADT counts! Just need to update text (and maybe make the text conditional) https://github.com/AlexsLemonade/scpca-nf/blob/6953ae72d0b8600b09bc8d50205260bb611a4e7e/templates/qc_report/qc_report.rmd#L442-L450

#### [`cite_qc.rmd`](https://github.com/AlexsLemonade/scpca-nf/blob/development/templates/qc_report/cite_qc.rmd) 
  - Also select target type https://github.com/AlexsLemonade/scpca-nf/blob/6953ae72d0b8600b09bc8d50205260bb611a4e7e/templates/qc_report/cite_qc.rmd#L38
    - TBD: We end up creating this column if users didn't provide. Do we want to tweak the implementation to add a metadata indicator if this information was user-provided?

CC @allyhawkins @jashapiro , tagging (_this is a pun_) if you have thoughts on some of the TBD items above, or anything else! This might not be an exhaustive list of changes, but it's enough to get started and more will gel once I start getting going.

sjspielman:
Closes #336 

This PR updates the QC report to have a bit more information about CITEseq post-processing. There wasn't much that really needed to be done it seemed in the end? I've done where what I think is the minimum we need (note one `TODO` comment in the `qc_report.rmd` file which is a question for reviewers!), but we could add more details, for example, maybe the number of cells removed based on RNA info vs. ADT info. Any recommendations for other stats to include would be helpful! Tagging a couple people for review to crowd source any other ideas for info to include.

While I was here, I also fixed a couple things..
- The `scpca_filter` column is present in both `_filtered` and `_processed` RDS files, so we remove a bit of wrangling
- I cleaned up a couple spacing around parentheses for styling 
- I tweaked, in line 444 of `qc_report.rmd`, a phrase that I think should not be there - I deleted the bit "calculated by `miQC`", since we then proceed to explain that maybe it wasn't calculated by miQC! Was this a good move on my part?

When testing this, I found that running `nextflow` with `-resume` did not want to use the updated rmds (is the `template` directory not copied over fresh, I wonder? any insights as to why `nextflow` didn't want to use? i wasn't pulling anything fresh from git, just running `main.nf`...). So, I just locally knitted with whatever package versions I have, which is maybe why we have the warning on top, unless we do need to update the `mutate()` code there. Local knitting is definitely responsible for the title though! Here's the report now:  
[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/11618036/qc_report.html.zip)

### DIFF
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -203,6 +203,12 @@ if (alt_exp %in% altExpNames(processed_sce)) {
 
     # Add result_matrix back into correct SCE as logcounts assay
     logcounts(altExp(processed_sce, alt_exp)) <- result_matrix
+    
+    # Also, calculate & save most variable ADTs from `adt_sce` 
+    top_adts <- adt_sce |>
+      scran::modelGeneVar() |>
+      scran::getTopHVGs()
+    metadata(processed_sce)$highly_variable_adts <- top_adts
   }
 }
 
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -177,12 +177,13 @@ if (alt_exp %in% altExpNames(processed_sce)) {
   adt_sce <- altExp(processed_sce, alt_exp)
   adt_sce <- adt_sce[,adt_sce$discard == FALSE]
 
-  # Only perform normalization if size factors are all positive
-  # Use the `adt_sce` variable here, since that's what we'll be normalizing
+  # If any size factors are not positive, simply use log1p
   if ( any( adt_sce$sizeFactor <= 0 ) ) {
-    warning("Failed to normalize ADT counts.")
+    metadata(processed_sce)$adt_normalization <- "log-normalization"
+    logcounts(adt_sce) <- log1p(counts(adt_sce))
   } else {
-    # Apply normalization
+    # Apply normalization using size factors
+    metadata(processed_sce)$adt_normalization <- "median-based"
     adt_sce <- scuttle::logNormCounts(adt_sce)    
 
     # Add this logcounts matrix with NA values added for cells not included in normalization
@@ -203,12 +204,6 @@ if (alt_exp %in% altExpNames(processed_sce)) {
 
     # Add result_matrix back into correct SCE as logcounts assay
     logcounts(altExp(processed_sce, alt_exp)) <- result_matrix
-    
-    # Also, calculate & save most variable ADTs from `adt_sce` 
-    top_adts <- adt_sce |>
-      scran::modelGeneVar() |>
-      scran::getTopHVGs()
-    metadata(processed_sce)$highly_variable_adts <- top_adts
   }
 }
 
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -40,7 +40,8 @@ antibody_tags <- as.data.frame(rowData(cite_exp)) |>
   arrange(desc(mean)) |>
   select("Antibody",
          "Mean UMI count per cell" = mean,
-         "Percent of cells detected" = detected)
+         "Percent of cells detected" = detected, 
+         "ADT target type" =  target_type)
 
 knitr::kable(antibody_tags, digits = 2) |>
   kableExtra::kable_styling(bootstrap_options = c("striped", "condensed"),
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -56,12 +56,12 @@ knitr::kable(antibody_tags, digits = 2) |>
 
 ```{r}
 
-filtered_cell_count <- sum(filtered_sce$adt_scpca_filter == "Remove")
+filtered_cell_count <- sum(filtered_sce$adt_scpca_filter == "Keep")
 
 basic_statistics <- tibble::tibble(
     "Method used to identify cells targeted for filtering" = format(processed_meta$adt_scpca_filter_method),
-    "Number of cells targeted for filtering"  = format(filtered_cell_count),
-    "Percent of cells targeted for filtering" =
+    "Number of cells that pass filtering thresholds"  = format(filtered_cell_count),
+    "Percent of cells that pass filtering thresholds" =
       paste0(round(filtered_cell_count/nrow(processed_sce) * 100, digits = 2), "%")
   ) |>
   mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) |> # reformat nulls
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -2,7 +2,7 @@
 
 This section details quality control statistics from the ADT (antibody-derived tag) component of CITE-seq experiments.
 
-## CITE-seq Statistics
+## CITE-seq Experiment Statistics
 
 ```{r}
 # add rowData if missing
@@ -13,17 +13,17 @@ if (is.null(rowData(adt_exp)$detected)){
 cell_adt_counts <- Matrix::colSums(counts(adt_exp))
 
 adt_information <- tibble::tibble(
-  "Number of ADTs assayed" = 
+  "Number of ADTs assayed" =
     format(nrow(adt_exp), big.mark = ',', scientific = FALSE),
-  "Number of reads sequenced" = 
+  "Number of reads sequenced" =
     format(adt_meta$total_reads, big.mark = ',', scientific = FALSE),
-  "Percent reads mapped to ADTs" = 
+  "Percent reads mapped to ADTs" =
     paste0(round(adt_meta$mapped_reads/adt_meta$total_reads * 100, digits = 2), "%"),
-  "Percent of ADTs in cells" = 
+  "Percent of ADTs in cells" =
     paste0(round(sum(cell_adt_counts)/adt_meta$mapped_reads * 100, digits = 2), "%"),
-  "Percent of cells with ADTs" = 
+  "Percent of cells with ADTs" =
     paste0(round(sum(cell_adt_counts > 0)/length(cell_adt_counts) * 100, digits = 2), "%"),
-  "Median ADT UMIs per cell" = 
+  "Median ADT UMIs per cell" =
     format(median(cell_adt_counts), big.mark = ',', scientific = FALSE)
   )|>
   t()
@@ -35,14 +35,14 @@ knitr::kable(adt_information, align = 'r') |>
   kableExtra::column_spec(2, monospace = TRUE)
 ```
 
-## Antibody-Derived Tag Statistics 
+## ADT Statistics
 ```{r}
 antibody_tags <- as.data.frame(rowData(adt_exp)) |>
   tibble::rownames_to_column("Antibody") |>
   arrange(desc(mean)) |>
   select("Antibody",
          "Mean UMI count per cell" = mean,
-         "Percent of cells detected" = detected, 
+         "Percent of cells detected" = detected,
          "ADT target type" =  target_type)
 
 knitr::kable(antibody_tags, digits = 2) |>
@@ -50,5 +50,27 @@ knitr::kable(antibody_tags, digits = 2) |>
                             full_width = FALSE,
                             position = "left",
                             ) |>
-  kableExtra::column_spec(2:3, monospace = TRUE) 
+  kableExtra::column_spec(2:3, monospace = TRUE)
+```
+
+## ADT Post-processing Statistics
+
+```{r}
+
+filtered_cell_count <- sum(processed_sce$adt_scpca_filter_method == "Remove")
+
+basic_statistics <- basic_statistics |>
+  mutate(
+    "Method used to identify cells targeted for filtering" = format(processed_meta$adt_scpca_filter_method)
+    "Number of cells targeted for filtering"  = format(filtered_cell_count),
+    "Percent of cells targeted for filtering" =
+      paste0(round(filtered_cell_count/nrow(processed_sce) * 100, digits = 2), "%")
+  )
+
+  # TODO: check rendering here & tweak kable settings
+  knitr::kable(basic_statistics, align = 'r') |>
+  kableExtra::kable_styling(bootstrap_options = "striped",
+                            full_width = FALSE,
+                            position = "left") |>
+  kableExtra::column_spec(2, monospace = TRUE)
 ```
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -193,8 +193,8 @@ if (has_processed) {
 
   # grab expression for top ADTs from counts
   var_adt_exp_df <- logcounts(altExp(processed_sce))[top_adts,] |>
-    as.matrix() |>
     t() |>
+    as.matrix() |>
     as.data.frame() |>
     tibble::rownames_to_column("barcode") |>
     # combine all ADTs into a single column for  faceting 
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -49,28 +49,100 @@ knitr::kable(antibody_tags, digits = 2) |>
   kableExtra::kable_styling(bootstrap_options = c("striped", "condensed"),
                             full_width = FALSE,
                             position = "left",
-                            ) |>
-  kableExtra::column_spec(2:3, monospace = TRUE)
+                            ) 
 ```
 
 ## ADT Post-processing Statistics
 
 ```{r}
 
-filtered_cell_count <- sum(processed_sce$adt_scpca_filter_method == "Remove")
+filtered_cell_count <- sum(filtered_sce$adt_scpca_filter_method == "Remove")
 
-basic_statistics <- basic_statistics |>
-  mutate(
-    "Method used to identify cells targeted for filtering" = format(processed_meta$adt_scpca_filter_method)
+basic_statistics <- tibble::tibble(
+    "Method used to identify cells targeted for filtering" = format(processed_meta$adt_scpca_filter_method),
     "Number of cells targeted for filtering"  = format(filtered_cell_count),
     "Percent of cells targeted for filtering" =
       paste0(round(filtered_cell_count/nrow(processed_sce) * 100, digits = 2), "%")
-  )
+  ) |>
+  mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) |> # reformat nulls
+  t()
 
-  # TODO: check rendering here & tweak kable settings
-  knitr::kable(basic_statistics, align = 'r') |>
+knitr::kable(basic_statistics, align = 'r') |>
   kableExtra::kable_styling(bootstrap_options = "striped",
                             full_width = FALSE,
                             position = "left") |>
   kableExtra::column_spec(2, monospace = TRUE)
 ```
+
+
+## Expression of highly variable ADTs
+
+The plots below show UMAP embeddings calculated from RNA expression where each cell is colored by the expression level of the labeled 
+ADT. 
+The ADTs chosen for plotting are the top 4 most variable ADTs in the library.
+
+
+```{r message=FALSE, results='asis'}
+# only create plot if variable ADTs are present
+if (!is.null(processed_meta$highly_variable_adts)) {
+  
+  # select top ADTs to plot 
+  top_adts <- processed_meta$highly_variable_adts |>
+    head(n = 4)
+  
+  # grab expression for top ADTs from counts
+  var_adt_exp <- logcounts(adt_exp[top_adts,]) |>
+    as.matrix() |>
+    t() |>
+    as.data.frame() |>
+    tibble::rownames_to_column("barcode")
+  
+  # extract umap embeddings as a dataframe
+  umap_df <- reducedDim(processed_sce, "UMAP") |>
+    as.data.frame() |>
+    tibble::rownames_to_column("barcode") |>
+    rename("UMAP1" = "V1",
+           "UMAP2" = "V2") 
+  
+  # combine ADT expression with coldata, umap embeddings, and create data frame to use for plotting
+  coldata_df <- colData(processed_sce) |>
+    as.data.frame() |>
+    tibble::rownames_to_column("barcode") |>
+    # TODO: FILTER TO ONLY ADTS WHOSE KEEP==TRUE???
+    # combine with gene expression
+    left_join(var_adt_exp, by = "barcode") |>
+    # combine with umap embeddings 
+    left_join(umap_df, by = "barcode") |>
+    # combine all ADTs into a single column for easy faceting 
+    tidyr::pivot_longer(cols = starts_with("CD"),
+                        names_to = "ADT",
+                        values_to = "adt_expression")
+  
+  
+  ggplot(coldata_df, aes(x = UMAP1, y = UMAP2, color = adt_expression)) + 
+    geom_point(alpha = 0.1, size = 0.001) + 
+    facet_wrap(vars(ADT)) +
+    scale_color_viridis_c() +
+    labs(
+      color = "Log-normalized ADT expression"
+    ) +
+    theme(legend.position = "bottom", 
+          axis.text = element_text(size = 8, color = "black"),
+          axis.title = element_text(size = 8, color = "black"),
+          strip.text = element_text(size = 8),
+          strip.background = element_rect(fill = 'transparent'),
+          legend.title = element_text(size = 8),
+          legend.text = element_text(size = 8)) + 
+    guides(colour = guide_colorbar(title.position = "bottom", title.hjust = 0.5))
+  
+} else {
+  glue::glue("
+    <div class=\"alert alert-warning\">
+    
+    This library does not contain a set of highly variable ADTs. 
+    
+    </div>
+  ")
+}
+```
+
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -165,7 +165,6 @@ if (has_processed) {
   # Calculate variance for each ADT
   adt_var <- altExp(processed_sce) |>
     logcounts() |>
-    as.matrix() |> 
     apply(1, var, na.rm=TRUE) 
   
   # Get the top 4
@@ -188,7 +187,7 @@ if (has_processed) {
 The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT. 
 
 
-```{r fig.alt="UMAP colored by expression of highly variable ADTs", warning = FALSE}
+```{r fig.alt="Density plot showing normalized expression of highly variable ADTs", warning = FALSE}
 if (has_processed) {
 
   # grab expression for top ADTs from counts
@@ -212,7 +211,7 @@ if (has_processed) {
 ```
 
 
-```{r fig.alt="UMAP colored by expression of highly variable ADTs"}
+```{r fig.alt="UMAP calculated from RNA expression but colored by normalized expression of highly variable ADTs", fig.height = 6}
 if (has_processed) {
   # create data frame of UMAPs and expression 
    umap_df <- scuttle::makePerCellDF(processed_sce) |>
@@ -233,8 +232,13 @@ if (has_processed) {
     # remove axis numbers and background grid
     scale_x_continuous(labels = NULL, breaks = NULL) + 
     scale_y_continuous(labels = NULL, breaks = NULL) + 
-    theme(legend.position = "bottom")  + 
-    guides(colour = guide_colorbar(title.position = "bottom", title.hjust = 0.5)) 
+    coord_fixed() +
+    theme(legend.position = "bottom",
+          axis.title = element_text(size = 8, color = "black"),
+          strip.text = element_text(size = 8),
+          legend.title = element_text(size = 9),
+          legend.text = element_text(size = 8)) +
+    guides(colour = guide_colorbar(title.position = "bottom", title.hjust = 0.5))
 }
 ```
 
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -62,7 +62,7 @@ if (has_processed) {
   filtered_cell_count <- sum(processed_sce$adt_scpca_filter == "Keep")
   
   basic_statistics <- tibble::tibble(
-    # Note that the `adt_scpca_filter_method` column is only present in the `processed_sce` object
+    # Note that the adt_scpca_filter_method column is only present in the processed_sce object
     "Method used to identify cells to filter" = format(processed_meta$adt_scpca_filter_method),
     "Number of cells that pass filtering threshold"  = format(filtered_cell_count),
     "Percent of cells that pass filtering threshold" =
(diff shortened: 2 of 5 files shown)


templates/qc_report/qc_report.rmd
```diff
@@ -426,26 +441,31 @@ In such situations, the calculated probability of compromise may not be valid (s
...
-Cells that should be removed are those that are identified to be low quality cells, such as cells with high probability of being compromised, calculated by `miQC`.
+Cells that should be removed based on RNA counts are those that are identified to be low quality cells, such as cells with high probability of being compromised.
 The method of filtering is indicated above the plot as either `miQC` or `Minimum gene cutoff`. 
 If `miQC`, cells below the specified probability compromised cutoff and above the minimum number of unique genes identified are kept for downstream analyses. 
 If only a `Minimum gene cutoff` is used, then `miQC` is not used and only those cells that pass the minimum number of unique genes identified threshold are retained. 
 The dotted vertical line indicates the minimum gene cutoff used for filtering. 
 
+```{r eval = has_cite, results='asis'}
+glue::glue(
+"Next, additional low quality cells were removed based on ADT counts (<code>cleanTagCounts</code>), such that cells with high levels of ambient contamination and/or isotype control tags (if present) were removed."
```
jashapiro, line 452: I think you can still use markdown here. Eitehr way, in keeping with previous usage, we should probably make this a full reference with a link to the package.

I'm also splitting into two sentences here. 

Update: I just realized that you were probably referring to the text label above with your code! That is fine, but now that I think about it, we should probably make that more descriptive, so that the `scpca_filter_method` encodes whether negative controls were used or not. 

```suggestion
"Next, additional low quality cells were removed based on ADT counts ([`DropletUtils::cleanTagCounts()`](https://bioconductor.org/packages/release/bioc/html/DropletUtils.html)).
Cells with high levels of ambient contamination and/or isotype control tags (if present) were removed."
```

jashapiro (commented):
Looks good overall, though I realized mid-review that we may want to be more specific in the filtering label we apply in the workflow. I think it is worth surfacing up front if the filtering used negative controls or not.

allyhawkins:
So I took a look at this and I have some thoughts which may be kind of crazy, so you can tell me if they are not feasible at this time. You all may disagree with me but... Based on my understanding of the changes that have taken place, for any library that contains CITE-seq data we have now filtered on both the RNA and the ADT counts so that the final `processed.rds` file contains only cells that pass both of those thresholds? So that means only those cells are included in normalization and PCA/UMAP? 

If the above is true and we don't want to change that then the changes I'm about to propose will not be as applicable... 
However, I can very much argue for having `processed.rds` files that contain the cells that have been filtered just on RNA. The CITE-seq library is handled separately from the gene expression library, so there could be various reasons why you throw out a cell that does poorly in CITE-seq but want to keep that cell to look at it's gene expression. So if users want to look at these samples regardless of CITE-seq, then they are going to lose all of those cells that have good quality RNA data. I noticed this because of the miQC plot where there are a lot of cells getting labeled as remove that I would want to keep. 

Because of that, I think it's important to separate the RNA and the ADT filtering here and actually move all of the mention of ADT filtering and downstream plots that would incorporate ADT filtering to the CITE-seq specific section. Then you show the data as a result of just analyzing the RNA data vs. RNA plus ADT analysis. Again, this may be an unpopular opinion and I know it's later in the game, but I am worried about losing information for RNA that we want to keep.

templates/qc_report/qc_report.rmd
```diff
@@ -227,18 +227,33 @@ basic_statistics <- tibble::tibble(
   )
 
 # if processed sce exists add filtering and normalization table
-if(has_processed){
+if (has_processed) {
   processed_meta <- metadata(processed_sce)
   
+  # TODO: Do we want to separate RNA and ADT rows? ADT is always the same, so may not be needed, but
```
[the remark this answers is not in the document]
sjspielman: The reason I split it is really only because I worried it would look confusing, especially since the header for this section is RNA-sequencing. Of course, RNA is also being filtered, just that ADT information is being used to determine the filtering... @allyhawkins, any stylistic opinions here? 

I definitely take your point about future breakage, though! I had thought at one point about whether we might actually want two different metadata fields for this (such as, `scpca_filter_method_rna` and `scpca_filter_method_adt`), but that will already break enough code to not be worth it! Plus, while we need to keep track of which RNA filtering method was used, we only have 1 ADT filtering method so there's less ambiguity.

sjspielman:
@allyhawkins I don't think this is "crazy" at all! You made a crux point here - 

> So if users want to look at these samples regardless of CITE-seq, then they are going to lose all of those cells that have good quality RNA data. I noticed this because of the miQC plot where there are a lot of cells getting labeled as remove that I would want to keep.

My immediate question now is whether this implies we need a 4th SCE export. This may not be necessary since there's a `discard` column in the CITE-seq colData that can be used to filter more cells out, but users would have to dig in themselves.

jashapiro:
> @allyhawkins I don't think this is "crazy" at all! You made a crux point here -
> 
> > So if users want to look at these samples regardless of CITE-seq, then they are going to lose all of those cells that have good quality RNA data. I noticed this because of the miQC plot where there are a lot of cells getting labeled as remove that I would want to keep.
> 
> My immediate question now is whether this implies we need a 4th SCE export. This may not be necessary since there's a `discard` column in the CITE-seq colData that can be used to filter more cells out, but users would have to dig in themselves.

Memories... I feel like this was a discussion a while ago when we initially decided not to worry about ADT filtering at that time! But I agree 100% that we would not want to make this a third export.

I think one option is that we modify the `scpca_filter` to be more fine grained than `keep` and `remove` to reflect which components are considered "okay". The idea would be that we could add a `remove_adt` or something like that which could be easily filtered on without diving into the `altExp`, but would be kept in the processed data. We could then add a second plot (or update this one) to show which cells would be filtered by ADT data specifically.

allyhawkins:
> My immediate question now is whether this implies we need a 4th SCE export. This may not be necessary since there's a discard column in the CITE-seq colData that can be used to filter more cells out, but users would have to dig in themselves.

I think my ideal scenario would be having a separate file that contains the ADT processed data that has been filtered and normalized using ADT counts rather than RNA. However, that means a lot of changes including on the actual portal end and I don't think we want to take that route, at least not right now. 

In the mean time, I think your suggestion where we only filter on RNA counts and then include a separate column indicating if cells would be filtered using ADT counts would be sufficient. We would then need to add that to user facing documentation. We did a similar thing before the processed data even existed where we had a `ccdl_filter` column present in the `filtered.rds` files to suggest which cells to remove. So here we would do the same with the CITE-seq. This would allow us to make a separate plot showing filtering of both RNA + ADT in the CITE-seq section without actually removing any cells that are informative when it comes to RNA.

jashapiro:
One other thought: we probably want UMAPs with the ADT expression! Maybe pick the 4 with the greatest variation and plot those?

sjspielman:
#moreUMAPSmorebetter

templates/qc_report/qc_report.rmd
```diff
@@ -125,52 +125,53 @@ if(has_multiplex){
...
     format(unfiltered_meta$total_reads, big.mark = ',', scientific = FALSE),
-  "Percent of RNA-seq reads mapped to transcripts" = 
+  "Percent of RNA-seq reads mapped to transcripts" =
     paste0(round((unfiltered_meta$mapped_reads/unfiltered_meta$total_reads) *100, 2), "%")
 )
-if (has_adt){
-  adt_exp <- altExp(filtered_sce, "adt")
+
+if (has_adt) {
+  adt_exp <- altExp(processed_sce, "adt")
```
sjspielman, line 146: noting this is needed for UMAPs, and it has no other effect on output.

sjspielman:
And we're back for more QC report updates! I've made the following changes:

- In `post_process_sce.R`, I now also calculate the most variable ADTs. I'm not sure about a modality-specific approach here, so I just did the same approach as for genes. I also stashed this in metadata as `highly_variable_adts` for later use in the QC report. 
  - ⚠️ If we keep this strategy, we'll need to add this metadata variable to documentation as well.
- In the QC report itself, I fixed a couple minor spacing spots and changed `~` -> `vars()` for faceting, since I was there.
- In the QC report, I also added a section in the `cite_qc.rmd` for a UMAP of top 4 variable ADTs, using the vector that is now in the metadata, and I tweaked some plot size aspects for visibility.
  -  I left one `TODO` here - currently all cells are being plotted, including those with `NA` logcounts. We could filter those cells out before plotting, but the points are so small that you don't really perceive the `NA` points anyways? 

HTML - [qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/11708350/qc_report.html.zip)

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,114 @@ knitr::kable(adt_information, align = 'r') |>
...
+
+## ADT Post-processing Statistics
+
+```{r}
+
+filtered_cell_count <- sum(filtered_sce$adt_scpca_filter == "Remove")
+
+basic_statistics <- tibble::tibble(
+    "Method used to identify cells targeted for filtering" = format(processed_meta$adt_scpca_filter_method),
+    "Number of cells targeted for filtering"  = format(filtered_cell_count),
```
allyhawkins, line 63:
```suggestion
    "Number of cells that pass filtering thresholds"  = format(filtered_cell_count),
```
I think we probably want to be consistent with how we report in the RNA section, which is the cells that are present after filtering. Also the targeted wording was confusing to me.

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,114 @@ knitr::kable(adt_information, align = 'r') |>
...
-                            ) |>
-  kableExtra::column_spec(2:3, monospace = TRUE) 
+                            ) 
 ```
+
+## ADT Post-processing Statistics
+
+```{r}
+
+filtered_cell_count <- sum(filtered_sce$adt_scpca_filter == "Remove")
```
allyhawkins, line 59:
```suggestion
filtered_cell_count <- sum(filtered_sce$adt_scpca_filter == "Keep")
```
I think I would capture the keep cells, see my comment below.

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,114 @@ knitr::kable(adt_information, align = 'r') |>
...
+
+basic_statistics <- tibble::tibble(
+    "Method used to identify cells targeted for filtering" = format(processed_meta$adt_scpca_filter_method),
+    "Number of cells targeted for filtering"  = format(filtered_cell_count),
+    "Percent of cells targeted for filtering" =
```
allyhawkins, line 64:
```suggestion
    "Percent of cells that pass filtering thresholds" =
```

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,114 @@ knitr::kable(adt_information, align = 'r') |>
...
+
+knitr::kable(basic_statistics, align = 'r') |>
+  kableExtra::kable_styling(bootstrap_options = "striped",
+                            full_width = FALSE,
+                            position = "left") |>
+  kableExtra::column_spec(2, monospace = TRUE)
+```
+
+
+## Expression of highly variable ADTs
```
allyhawkins, line 78: Before you get to this I think we need a plot showing the filtering that is suggested. We could include another plot like the keep/remove colored plot in RNA but color cells by RNA filter, ADT filter, all filter, or keep.

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,114 @@ knitr::kable(adt_information, align = 'r') |>
...
+
+The plots below show UMAP embeddings calculated from RNA expression where each cell is colored by the expression level of the labeled
```
allyhawkins, line 80: Alternatively (or additionally?) we could show a faceted ridge plot with the top variable ADT tags on the X axis and the Y axis is the normalized expression.

allyhawkins (commented):
Thanks for making these changes! I left a few comments on where I think we could clarify things and then also how we could add in some more plots that I think would be helpful either in addition to or instead of the UMAPs.

templates/qc_report/cite_qc.rmd
sjspielman: I ended up using that wording all around because we don't actually _perform_ filtering, just identify cells to filter. On one hand I agree about consistency, but on the other hand we aren't actually doing the same thing here! Either way I've accepted suggestions and will fiddle with other wording to be more clear.

sjspielman:
This should be ready for a new look! I've made the following changes:
- Include a faceted plot of all the ADT/RNA combinations of filtering, again showing the miQC plot
  - I had some trouble here with `glue::glue()`! Note the printed statement that will end up being deleted but is here now for debugging... why doesn't `glue::glue()` print inside of the `if`? I played around with this a bunch, and found more generally that none of the glues inside of `if` blocks are printing.. Does anyone have ideas?!
- I fixed a bug with adt output - much of the `has_adt` logic assumed `has_processed` is true, which is not a safe assumption! I updated in some spots to avoid this trap.
- I added a ridgeplot into the cite report, currently living before the UMAPs. Here, I found that the chunk only wants to print out one plot (another confusing rmarkdown thing 🤷‍♀️), so I ended up just using `patchwork` to ensure they are both printed.   
Here's what the report currently looks like: 
[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/11871915/qc_report.html.zip)

templates/qc_report/qc_report.rmd
sjspielman: Noting this was not, indeed, true since `has_processed` could be false! It was fixed in recent commits.

jashapiro (commented):
I left a bunch of little comments, some of which are minor formatting, some of which may address your questions, and some of which are just me being picky. 

I think the biggest one is that we probably don't want to use `getHVGs` for ADTs, as there is not enough data to fit the model well. Do we can just calculate variance from the matrix on the fly. Similarly, if there is no `logcounts` matrix because normalization failed, we can still make one for plotting here, which I think is better than skipping the plots. We just need to add a warning that normalization has not been performed.

bin/post_process_sce.R
```diff
@@ -203,6 +203,12 @@ if (alt_exp %in% altExpNames(processed_sce)) {
 
     # Add result_matrix back into correct SCE as logcounts assay
     logcounts(altExp(processed_sce, alt_exp)) <- result_matrix
+    
+    # Also, calculate & save most variable ADTs from `adt_sce` 
+    top_adts <- adt_sce |>
+      scran::modelGeneVar() |>
+      scran::getTopHVGs()
+    metadata(processed_sce)$highly_variable_adts <- top_adts
```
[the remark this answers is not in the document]
sjspielman: > I don't think we want to do modeling here: I don't think there is enough data

:100: makes more sense to me.

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+    
+    
+    adt_umap <- ggplot(coldata_df, aes(x = UMAP1, y = UMAP2, color = adt_expression)) + 
+      geom_point(alpha = 0.1, size = 0.1) + 
+      facet_wrap(vars(ADT)) +
+      scale_color_viridis_c() +
+      labs(
+        color = "Log-normalized ADT expression"
+      ) +
+      theme(legend.position = "bottom",
```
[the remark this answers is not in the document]
sjspielman: > Also, let's maybe remove numbers from the UMAP?

I can do that all around!

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+  ggplot(filtered_coldata_df, aes(x = detected, y = subsets_mito_percent, color = filter_summary)) +
+    geom_point(alpha = 0.5, size = 1) +
+    geom_vline(xintercept = min_gene_cutoff, linetype = "dashed") +
+    labs(x = "Number of genes detected",
+         y = "Mitochondrial percentage",
+         title = "Combined RNA and ADT filters") +
+    scale_color_viridis_d(direction = -1) +
+    facet_wrap(vars(filter_summary)) +
+    theme(plot.title = element_text(hjust = 0.5),
+          legend.position = "none")
```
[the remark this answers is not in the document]
sjspielman: Yep, got this all into a `theme_set()` on my end already when I was making plotting changes during your review!

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+            axis.text = element_text(size = 8, color = "black"),
+            axis.title = element_text(size = 8, color = "black"),
+            strip.text = element_text(size = 8),
+            strip.background = element_rect(fill = 'transparent'),
+            legend.title = element_text(size = 8),
+            legend.text = element_text(size = 8)) + 
+      guides(colour = guide_colorbar(title.position = "bottom", title.hjust = 0.5)) 
+    
+    # Combine with patchwork so that both plots are reliably printed
+    adt_ridgeplot / adt_umap
```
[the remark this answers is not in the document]
sjspielman: No need, can easily separate and that will make for clearer alt text to add in as well.

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+      # combine all ADTs into a single column for easy faceting 
+      tidyr::pivot_longer(cols = starts_with("CD"),
+                            names_to = "ADT",
+                            values_to = "adt_expression")
+    
+    # First, expression ridgeplot ridgeplots
+    adt_ridgeplot <- ggplot(var_adt_exp_df, 
+           aes(y = ADT, x = adt_expression, fill = ADT)) + 
+      ggridges::geom_density_ridges(alpha = 0.5) + 
+      scale_fill_viridis_d() +
```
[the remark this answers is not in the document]
sjspielman: Yes, I also thought of this while you were reviewing and I was making plot changes! I ended up changing to use a 2 color scheme for the faceted keep/remove plot to _match_ the keep/remove colors in the earlier cell filtering plot that is in the main QC report. For this plot, I ended up just using default ggplot fills (since empty density plots look weird to me).

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+      head(n = 4)
+    
+    # grab expression for top ADTs from counts
+    var_adt_exp_df <- logcounts(adt_exp_processed[top_adts,]) |>
+      as.matrix() |>
+      t() |>
+      as.data.frame() |>
+      tibble::rownames_to_column("barcode") |>
+      # combine all ADTs into a single column for easy faceting 
+      tidyr::pivot_longer(cols = starts_with("CD"),
```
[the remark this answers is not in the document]
sjspielman: > not make an assumption about ADT names?

:+1:

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+             "UMAP2" = "V2") 
+    
+    # combine ADT expression with coldata, umap embeddings, and create data frame to use for plotting
+    coldata_df <- colData(processed_sce) |>
+      as.data.frame() |>
+      tibble::rownames_to_column("barcode") |>
+      # combine with gene expression
+      left_join(var_adt_exp_df, by = "barcode") |>
+      # combine with umap embeddings 
+      left_join(umap_df, by = "barcode")
```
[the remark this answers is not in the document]
sjspielman: 👀

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+
+First, ridgeplots showing normalized expression of these 4 ADTs are shown.
+Second, UMAP embeddings calculated from RNA expression are shown, where each cell is colored by the expression level of the given ADT. 
+
+
+```{r fig.height=10, fig.width=8, message=FALSE, warning=FALSE, results='asis'}
+if (has_processed) {
+  
+  # only create plot if variable ADTs are present
+  if (!is.null(processed_meta$highly_variable_adts)) {
```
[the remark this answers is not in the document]
allyhawkins: For non-CITE-seq samples, the backup is to log transform the counts if something fails with the default normalization method. On the one hand, I like the idea of mirroring that and including log-transformed counts for CITE-seq. On the other hand, I am wary of providing what may appear like normalized counts to users and them using those without re-doing normalization. I think ultimately I'm in favor of adding it to the object rather than failing normalization, but we definitely need to make sure documentation is up to date.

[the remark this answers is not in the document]
sjspielman: >  I think ultimately I'm in favor of adding it to the object rather than failing normalization, but we definitely need to make sure documentation is up to date.

👍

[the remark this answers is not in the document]
sjspielman: Hm, I think we might also want another metadata variable here to match RNA.... `adt_normalization = <median-based/log-normalization>`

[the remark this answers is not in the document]
allyhawkins: > Hm, I think we might also want another metadata variable here to match RNA.... `adt_normalization = <median-based/log-normalization>`

I agree with this.

templates/qc_report/cite_qc.rmd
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/f51e0466dd4355fc942b2e80272b70c9c5503348

templates/qc_report/cite_qc.rmd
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/c566fdd62711087930028ca0e45c24fe01e33182
I directly used `log1p` here, and added the metadata string.

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+```
+
+## Removing low quality cells based on ADT counts
+
+```{r results='asis'}
+
+glue::glue("
+  This glue is OUTSIDE the if and it prints just fine. 
+  Where are my glues INSIDE the ifs? :( 
+  ")
```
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/f51e0466dd4355fc942b2e80272b70c9c5503348

templates/qc_report/cite_qc.rmd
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/fac803cda9c5519a70f3cb60449cdd623948ba6f & https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/f51e0466dd4355fc942b2e80272b70c9c5503348

templates/qc_report/qc_report.rmd
```diff
@@ -28,9 +28,10 @@ knitr::opts_chunk$set(
 library(SingleCellExperiment)
 library(dplyr)
 library(ggplot2)
+library(patchwork)
 
 # Set default ggplot theme
-theme_set(theme_bw() + 
+theme_set(theme_bw() +
           theme(plot.margin = margin(rep(20, 4))))
```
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/ab677de1bb8b5e26894d1e3ce49bb0ddb63f5374

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,210 @@ knitr::kable(adt_information, align = 'r') |>
...
+      t() |>
+      as.data.frame() |>
+      tibble::rownames_to_column("barcode") |>
+      # combine all ADTs into a single column for easy faceting 
+      tidyr::pivot_longer(cols = starts_with("CD"),
+                            names_to = "ADT",
+                            values_to = "adt_expression")
+    
+    # First, expression ridgeplot ridgeplots
+    adt_ridgeplot <- ggplot(var_adt_exp_df,
```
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/f51e0466dd4355fc942b2e80272b70c9c5503348

templates/qc_report/cite_qc.rmd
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/343/commits/f51e0466dd4355fc942b2e80272b70c9c5503348

sjspielman:
Thanks @jashapiro and @allyhawkins a lot for all your comments! I've responded to everything here, and we're moving right along. I've tested workflow changes as well.

At this point, this PR is getting ~too big. I suggest, unless there is a _major_ problem, let's merge this in soon and if there are more smaller changes needed we can tackle them in a subsequent PR?

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,206 @@ knitr::kable(adt_information, align = 'r') |>
...
+```{r}
+# Calculate ADT variance
+if (has_processed) {
+  
+  top_n <- 4 # we want the top 4 ADTs
+
+  # Calculate variance for each ADT
+  adt_var <- altExp(processed_sce) |>
+    logcounts() |>
+    as.matrix() |>
```
jashapiro, line 168: Not sure this is needed: `apply` functions should work fine on `Matrix` objects too. But I could be wrong,

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,206 @@ knitr::kable(adt_information, align = 'r') |>
...
+The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT. 
+
+
+```{r fig.alt="UMAP colored by expression of highly variable ADTs", warning = FALSE}
+if (has_processed) {
+
+  # grab expression for top ADTs from counts
+  var_adt_exp_df <- logcounts(altExp(processed_sce))[top_adts,] |>
+    as.matrix() |>
+    t() |>
```
jashapiro, line 197: Likely a bit more efficient to swap the order here and let the `Matrix` package work its magic.
```suggestion
    t() |>
    as.matrix() |>
```

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,206 @@ knitr::kable(adt_information, align = 'r') |>
...
+
+  # Calculate variance for each ADT
+  adt_var <- altExp(processed_sce) |>
+    logcounts() |>
+    as.matrix() |> 
+    apply(1, var, na.rm=TRUE) 
+  
+  # Get the top 4
+  top_adts <- adt_var[ order(adt_var, decreasing=TRUE)[1:top_n] ] |>
+    names()
```
jashapiro, line 173: Annoying that `order` doesn't preserve the names!

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,206 @@ knitr::kable(adt_information, align = 'r') |>
...
+  ggplot(umap_df, aes(x = UMAP1, y = UMAP2, color = adt_expression)) + 
+    geom_point(alpha = 0.1, size = 0.1) + 
+    facet_wrap(vars(ADT)) +
+    scale_color_viridis_c() +
+    labs(
+      color = "Log-normalized ADT expression"
+    ) +
+    # remove axis numbers and background grid
+    scale_x_continuous(labels = NULL, breaks = NULL) + 
+    scale_y_continuous(labels = NULL, breaks = NULL) +
```
jashapiro, line 235: Did you try adding in `coord_fixed()`? I'm thinking we want to add that to all UMAP plots to make them plot with the same aspect ratio.

templates/qc_report/qc_report.rmd
```diff
@@ -28,10 +28,13 @@ knitr::opts_chunk$set(
...
 library(dplyr)
 library(ggplot2)
+library(patchwork)
 
 # Set default ggplot theme
-theme_set(theme_bw() + 
-          theme(plot.margin = margin(rep(20, 4))))
+theme_set(
+  theme_bw() +theme(plot.margin = margin(rep(20, 4)), 
+                    strip.background = element_rect(fill = 'transparent'))
```
jashapiro, line 36: Just formatting
```suggestion
  theme_bw() +
  theme(plot.margin = margin(rep(20, 4)), 
        strip.background = element_rect(fill = 'transparent'))
```

jashapiro (approved):
This looks good to me, with some very minor suggestions, except for the `coord_fixed()` one, which I think we should do as long as it doesn't look awful.

templates/qc_report/cite_qc.rmd
sjspielman: rude indeed!!! will re-look at this..

templates/qc_report/cite_qc.rmd
sjspielman: thanks for the reminder here!

templates/qc_report/cite_qc.rmd
sjspielman: Seems ok to me actually, see example - 

```
> vec <- c("A" = 1.5, "B" = 2.5, "C" = 0.5, D = "10")
> vec[ order(vec, decreasing=TRUE)[1:2] ]
    B     D 
"2.5"  "10" 
> vec[ order(vec, decreasing=TRUE)[1:2] ] |> names()
[1] "B" "D"
```

jashapiro: No, it is fine! I was thinking of doing things in a different way with `vec |> order() |> head() |> names()` to save some nesting, but you can't do that because `order()` drops the names.

sjspielman: gotcha :)

templates/qc_report/cite_qc.rmd
```diff
@@ -35,19 +35,206 @@ knitr::kable(adt_information, align = 'r') |>
...
+
+    No ADT post-processing was performed on this library.
+
+    </div>
+  ")
+}
+```
+
+
+The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT.
```
allyhawkins, line 187: Should this be below the next chunk? I think you want it right before the UMAP is actually plotted?

allyhawkins (approved):
I had one minor comment about the placement of the UMAP statement. Other than that I'm going to approve, since I'm trusting you're fixing all of Josh's suggestions. I don't need to see this again.

templates/qc_report/cite_qc.rmd
sjspielman: yes good catch, thanks!

sjspielman:
Here's the "final" version of the report for posterity - 
[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/11898907/qc_report.html.zip)

Main notable change is that using `coord_fixed()` means the plot display area needs to be more explicitly sized, so facet labels aren't cut off. The cutting off could still happen, but you'd need a gene symbol that's _longer_ than an Ensembl ID, which is not going to be very likely, and the benefit from `coord_fixed()` seemed greater.

PR #345, opened 2023-06-01
Remove ADT filtering with new filter indicator

### LINKED ISSUE #344: Update ADT filtering to not perform filtering
Inspired by conversations including these comments:
- https://github.com/AlexsLemonade/scpca-nf/pull/343#issuecomment-1572385042
- https://github.com/AlexsLemonade/scpca-nf/pull/343#issuecomment-1572389263

ADT filtering should be updated such that all filtering calculations remain, but filtering is not actually _applied_. Instead, we'll want to add slightly more detail to keep/remove indicators about whether given cells might be good to remove based on ADT stats. Happily this should not affect ADT normalization.

sjspielman:
Closes #344 

Here, I've updated our approach to ADT filtering:

- `scpca_filter` now contains "Remove" and "Keep" based only on RNA
- A new column `adt_filter` contains "Remove" and "Keep" based only on ADT. I chose this variable name and values for rough consistency with `scpca_filter`.
- Filtering is now back to just RNA-level filtering, and as a consequence that extra `;cleanTagCounts` string is gone from `scpca_filter_method`.
- Some workarounds were needed for normalization, since cells that get filtered out are going to be (under most circumstances) ones with size factors = 0. So, there is "temporary" filtering just for the purpose of performing normalization, and then the logcounts matrix is rebuilt with `NA` values for cells that would have been filtered out due to ADTs. But, size factors are still calculated for all cells.
  - Includes some dimension and value checks (do we want more checks? eg I thought of checking row/col names but those are pretty locked down.)
  - When adding this matrix back into the SCE, I ended up coercing it to a sparse `Matrix` (capital M). I would have assumed that `SingleCellExperiment()` would coerce on my behalf, but the printing suggested it did not.
- Fixed a bug! `warn()`, which is in fact not an R function, has been changed to `warning()`, which indeed is an R function. 

This has been tested both as stand-alone Rscript and from the pipeline ✅  
Let me know what you think of this approach! After this goes in, we'll move to update `scpca-docs` accordingly and get back to the QC report.

### DIFF
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -73,7 +73,6 @@ if(!(stringr::str_ends(opt$output_sce_file, ".rds"))){
 sce <- readr::read_rds(opt$filtered_sce_file)
 
 # create scpca_filter column -----------
-
 if(all(is.na(sce$miQC_pass))){
   sce$scpca_filter <- ifelse(
     sce$detected >= opt$gene_cutoff,
@@ -90,27 +89,25 @@ if(all(is.na(sce$miQC_pass))){
   )
   metadata(sce)$scpca_filter_method <- "miQC"
 }
+metadata(sce)$scpca_filter_method <- paste0(metadata(sce)$scpca_filter_method)
 
-# if present, add ADT filtering to `scpca_filter` and
-# setup associated metadata string
+# add min gene cutoff to metadata
+metadata(sce)$min_gene_cutoff <- opt$gene_cutoff
+
+
+# create adt_filter column, if CITEseq ----------
 alt_exp <- opt$adt_name
 if (alt_exp %in% altExpNames(sce)) {
-  # Add column for whether cells are "recommended" to be removed
-  #  based on ADT stats
   sce$adt_filter <-  ifelse(
     altExp(sce, alt_exp)$discard,
     "Remove",
     "Keep"
   )
 }
-metadata(sce)$scpca_filter_method <- paste0(metadata(sce)$scpca_filter_method)
-
-# add min gene cutoff to metadata
-metadata(sce)$min_gene_cutoff <- opt$gene_cutoff
 
 # Perform filtering -----------------
 
-# filter sce using criteria in scpca_filter
+# filter sce using criteria in scpca_filter (not adt_filter)
 processed_sce <- sce[, which(sce$scpca_filter == "Keep")]
 
 # replace existing stats with recalculated gene stats
@@ -156,6 +153,7 @@ processed_sce <- scuttle::logNormCounts(processed_sce)
 
 # Try to normalize ADT counts, if present
 if (alt_exp %in% altExpNames(processed_sce)) {
+  
   # need `all()` since,if present, this is an array
   if( !all(is.null(metadata(altExp(processed_sce, alt_exp))$ambient_profile))){
     # Calculate median size factors from the ambient profile
@@ -167,13 +165,49 @@ if (alt_exp %in% altExpNames(processed_sce)) {
     # if ambient profile is not present, set sizeFactor to 0 for later warning.
     altExp(processed_sce, alt_exp)$sizeFactor <- 0
   }
-
+  
+  # Perform filtering specifically to allow for normalization
+  adt_sce <- altExp(processed_sce, alt_exp)
+  adt_sce <- adt_sce[,adt_sce$discard == FALSE]
+  
   # Only perform normalization if size factors are all positive
-  if ( any( altExp(processed_sce, alt_exp)$sizeFactor <= 0 ) ) {
+  # Use the `adt_sce` variable here, since that's what we'll be normalizing
+  if ( any( adt_sce$sizeFactor <= 0 ) ) {
     warning("Failed to normalize ADT counts.")
   } else {
     # Apply normalization
-    altExp(processed_sce, alt_exp) <- scuttle::logNormCounts(altExp(processed_sce, alt_exp))
+    adt_sce <- scuttle::logNormCounts(adt_sce)    
+
+    # Add this logcounts matrix with NA values added for cells not included in normalization
+    na_barcodes <- altExp(processed_sce, alt_exp)[,altExp(processed_sce, alt_exp)$discard]
+    # First, create matrix of NA values with barcode column names
+    na_matrix <- matrix(NA, 
+           nrow = nrow(altExp(processed_sce, alt_exp)), 
+           ncol = ncol(na_barcodes)
+    )
+    colnames(na_matrix) <- colnames(na_barcodes)
+    
+    # Combine matrices and reorder columns to match SCE object (row order already matches)
+    combined_matrix <- logcounts(adt_sce) |>
+      as.matrix() |>
+      cbind(na_matrix) 
+    combined_matrix <- combined_matrix[, colnames(altExp(processed_sce, alt_exp))]
+    
+    # Check dimensions
+    if (nrow(combined_matrix) != nrow(altExp(processed_sce, alt_exp)) | 
+        ncol(combined_matrix) != ncol(altExp(processed_sce, alt_exp))) {
+      stop("ADT normalization error.")
+    }
+    
+    # Check correct number of NAs:
+    observed_na_count <- sum(is.na(combined_matrix))
+    expected_na_count <- nrow(adt_sce) * (ncol(altExp(processed_sce, alt_exp)) - ncol(adt_sce))
+    if (observed_na_count != expected_na_count) {
+      stop("Incorrect number of normalized NAs recovered during ADT normalization.")
+    }
+
+    # Add combined_matrix back into correct SCE as logcounts assay
+    logcounts(altExp(processed_sce, alt_exp)) <- Matrix::Matrix(combined_matrix, sparse = TRUE)
   }
 }
 
--- a/bin/post_process_sce.R
+++ b/bin/post_process_sce.R
@@ -94,15 +94,16 @@ if(all(is.na(sce$miQC_pass))){
 # if present, add ADT filtering to `scpca_filter` and
 # setup associated metadata string
 alt_exp <- opt$adt_name
-if (!(alt_exp %in% altExpNames(sce))) {
-  adt_filter_string <- ""
-} else {
-  adt_discard_rows <- which(altExp(sce, alt_exp)$discard)
-  sce$scpca_filter[adt_discard_rows] <- "Remove"
-  adt_filter_string <- ";cleanTagCounts"
+if (alt_exp %in% altExpNames(sce)) {
+  # Add column for whether cells are "recommended" to be removed
+  #  based on ADT stats
+  sce$adt_filter <-  ifelse(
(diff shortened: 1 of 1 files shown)


bin/post_process_sce.R
```diff
@@ -90,26 +89,25 @@ if(all(is.na(sce$miQC_pass))){
...
-}
-metadata(sce)$scpca_filter_method <- paste0(metadata(sce)$scpca_filter_method,
-                                            adt_filter_string)
+metadata(sce)$scpca_filter_method <- paste0(metadata(sce)$scpca_filter_method)
 
 # add min gene cutoff to metadata
 metadata(sce)$min_gene_cutoff <- opt$gene_cutoff
 
+
+# create adt_filter column, if CITEseq ----------
```
jashapiro, line 97: I think I might prefer to keep a single `scpca_filter` column, but with more than two values. If we do want to split this up, I think we would want them to be something like `scpca_rna_filter` and `scpca_adt_filter` for clarity and consistency.

For values, I might do something like `keep_all`, `keep_rna`, `keep_adt`, `remove`? On reflection, I don't necessarily love that, so I'm open to other ideas. (Maybe switching the order? `keep`, `remove_rna`, `remove_adt`, `remove_all`)

jashapiro: @allyhawkins should weigh in here!

bin/post_process_sce.R
```diff
@@ -166,13 +165,49 @@ if (alt_exp %in% altExpNames(processed_sce)) {
...
   # Only perform normalization if size factors are all positive
-  if ( any( altExp(processed_sce, alt_exp)$sizeFactor <= 0 ) ) {
-    warn("Failed to normalize ADT counts.")
+  # Use the `adt_sce` variable here, since that's what we'll be normalizing
+  if ( any( adt_sce$sizeFactor <= 0 ) ) {
+    warning("Failed to normalize ADT counts.")
   } else {
     # Apply normalization
-    altExp(processed_sce, alt_exp) <- scuttle::logNormCounts(altExp(processed_sce, alt_exp))
+    adt_sce <- scuttle::logNormCounts(adt_sce)
```
jashapiro, line 178: An alternative way to do this, which may be more efficient, would be to create a full matrix from the dimensions of the `processed_sce` and fill only the columns:

```
# get the counts matrix and make it NA
result_matrix <- counts(altExp(processed_sce, alt_exp))
result_matrix[TRUE] <- NA

# get the computed values & fill them in
result_matrix[, rownames(adt_sce)] <- logcounts(adt_sce) 

logcounts(altExp(processed_sce, alt_exp)) <- result_matrix
```

sjspielman: ^ Should be`colnames()` in the index-y line, but yes this should work, and fewer checks!

bin/post_process_sce.R
allyhawkins: Hmmm. I personally favor two columns. The reason is because you are filtering on `scpca_filter` to get the processed SCE object, but you won't do any filtering on an `adt_filter` column. It feels like two separate tasks. 

I'm also thinking about explaining it in the docs and I think it's easier to keep the explanation of the `scpca_filter` column as is, only discussing the filtering cutoffs related to RNA. Then you add a row to the table we have for explaining the [contents of the `colData` of alt experiments specifically](https://scpca.readthedocs.io/en/latest/sce_file_contents.html#additional-singlecellexperiment-components-for-cite-seq-libraries). Also, in the case where there is no `adt` data then we want to keep the values of `scpca_filter` consistent, so I feel like having `keep` and `remove` when it's just rna data there makes more sense.

jashapiro: That makes sense to me, but I'll again advocate for `scpca_rna_filter` and `scpca_adt_filter`

allyhawkins: I think that makes sense, just unfortunate that it means more changes. 

Does this mean we need to re-process every project though?? So they all have the correct column names?

jashapiro: I suppose it does, but that should be quick? I suppose another option is to name the new column `CITEseq_scpca_filter` for consistency with the `altExp` name. (`altexps_CITEseq_scpca_filter` seemed a bridge too far, though it is most parallel)

jashapiro: I _am_ thinking again we might want to rename the modality to `adt` though... The whole method of RNA + ADT is technically what CITE-seq is, so I think we can do that without worry, but it does require more docs & code updates.

allyhawkins: > Update: I think I like that as a good compromise: leave scpca_filter alone, prefix for other modality filters.

I think this sounds good and I like this approach. 

> I am thinking again we might want to rename the modality to adt though... The whole method of RNA + ADT is technically what CITE-seq is, so I think we can do that without worry, but it does require more docs & code updates.

I do have to agree that it might make more sense to name to `adt`.

sjspielman: Sorry I missed all this chat, but I agree with its conclusions so that's good!!

bin/post_process_sce.R
sjspielman: Done in https://github.com/AlexsLemonade/scpca-nf/pull/345/commits/0b7a88d3a29197caa5a4562e5fb3e5235e0a784c.

bin/post_process_sce.R
sjspielman: > I do have to agree that it might make more sense to name to `adt`.

To clarify do we mean the alternate experiment name here? I don't see any `modality` in metadata.

jashapiro: Yes, I mean altExp

sjspielman:
I think this is about there for this PR, and I'll follow up separately with CITEseq renaming (#347).

jashapiro (commented):
Mostly there... I think we want to update the metadata we add for the adt filtering, and some questions about potentially unnecessary matrix type conversions.

bin/post_process_sce.R
```diff
@@ -166,13 +164,37 @@ if (alt_exp %in% altExpNames(processed_sce)) {
...
+    
+    # now get the computed logcounts & fill them in
+    result_matrix[, colnames(adt_sce)] <- as.matrix(logcounts(adt_sce))
+    
+    # Check correct number of NAs:
+    observed_na_count <- sum(is.na(result_matrix))
+    expected_na_count <- nrow(adt_sce) * (ncol(altExp(processed_sce, alt_exp)) - ncol(adt_sce))
+    if (observed_na_count != expected_na_count) {
+      stop("Incorrect number of normalized NAs recovered during ADT normalization.")
+    }
```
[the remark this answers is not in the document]
sjspielman: I think I probably meant to remove this check altogether, but I guess we can also set this to instead fail if `observed_na_count <= expected_na_count`?

bin/post_process_sce.R
```diff
@@ -90,26 +89,24 @@ if(all(is.na(sce$miQC_pass))){
...
-if (!(alt_exp %in% altExpNames(sce))) {
-  adt_filter_string <- ""
-} else {
-  adt_discard_rows <- which(altExp(sce, alt_exp)$discard)
-  sce$scpca_filter[adt_discard_rows] <- "Remove"
-  adt_filter_string <- ";cleanTagCounts"
-}
-metadata(sce)$scpca_filter_method <- paste0(metadata(sce)$scpca_filter_method,
-                                            adt_filter_string)
+metadata(sce)$scpca_filter_method <- metadata(sce)$scpca_filter_method
```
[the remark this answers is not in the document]
sjspielman: What is happening here is I had to remove `paste` code but then apparently did not look at the resulting line to see...we don't need this line at all!
But yes, let's add `adt_scpca_filter_method` as a separate metadata "cleanTagCounts".

bin/post_process_sce.R
```diff
@@ -166,13 +164,37 @@ if (alt_exp %in% altExpNames(processed_sce)) {
...
   } else {
     # Apply normalization
-    altExp(processed_sce, alt_exp) <- scuttle::logNormCounts(altExp(processed_sce, alt_exp))
+    adt_sce <- scuttle::logNormCounts(adt_sce)    
+
+    # Add this logcounts matrix with NA values added for cells not included in normalization
+    
+    # first, get the counts matrix and make it NA
+    result_matrix <- counts(altExp(processed_sce, alt_exp))
+    result_matrix[TRUE] <- NA
```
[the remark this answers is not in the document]
sjspielman: I think the comma actually, more expilcitly "all rows, all columns"

bin/post_process_sce.R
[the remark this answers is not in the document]
sjspielman: I went ahead and changed to `<` in https://github.com/AlexsLemonade/scpca-nf/pull/345/commits/18783e93564f7cc6c0723649353cbee5c2d85144

bin/post_process_sce.R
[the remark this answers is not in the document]
sjspielman: - Silly line removed in https://github.com/AlexsLemonade/scpca-nf/pull/345/commits/39d3a43887b452482eed527fc6f1a5c9ff0d7cd0
- Added adt metadata in https://github.com/AlexsLemonade/scpca-nf/pull/345/commits/18783e93564f7cc6c0723649353cbee5c2d85144

bin/post_process_sce.R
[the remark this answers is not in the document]
jashapiro: works for me.

bin/post_process_sce.R
```diff
@@ -91,25 +90,23 @@ if(all(is.na(sce$miQC_pass))){
...
 
+# create adt_scpca_filter column, if CITEseq ----------
+alt_exp <- opt$adt_name
+if (alt_exp %in% altExpNames(sce)) {
+  sce$adt_scpca_filter <-  ifelse(
+    altExp(sce, alt_exp)$discard,
+    "Remove",
+    "Keep"
+  )
+  metadata(sce)$adt_scpca_filter_method <- "cleanTagCounts"
```
jashapiro, line 104: Again, I'd like this value to depend on whether or not negative controls were used. Can we take advantage of the fact that the output columns vary to do so?

sjspielman: Ah, sorry yes forgot that part!! Existing columns can definitely work for checking this.

How about....
- `cleanTagCounts; isotype controls`
- `cleanTagCounts; no isotype controls` 

?

jashapiro: Maybe no punctuation? "cleanTagCounts with isotype controls" & "cleanTagCounts without isotype controls"? A bit wordy, but once you get past a single word it probably doesn't matter, unless you want to try to come up with a one-word encoding.

sjspielman: all set in https://github.com/AlexsLemonade/scpca-nf/pull/345/commits/9de1a0e51b37e5d3d20a5ec454c64cfc5f0a33ba

jashapiro (approved):
LGTM, but I'd probably want @allyhawkins to weigh in on the `adt_scpca_filter_method` names.

allyhawkins (approved):
This LGTM and I like the options used for the adt filter method.

PR #346, opened 2023-06-01
Check index paths match when skipping mapping for all mapping steps

### LINKED ISSUE #302: Check that references are the same when skipping mapping for all modalities
It seems like we only check that the reference is the same for the bulk mapping... Not sure why we don't do that check with the alevin data too? It seems like we should (and maybe check the actual index file rather than the assembly name?) I would do this in a separate PR...

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/301#pullrequestreview-1402666552_

allyhawkins:
Closes #302 

This adds a check to all places where we skip mapping for the path of the reference to be the same as what's present in the `scpca-meta.json` file. I added this to the af-rna, af-feature, genetic demux workflow, and spaceranger. I also modified what was in the bulk workflow to look specifically for the index path rather than the assembly name and tx2gene file path. 

The only one that still checks for the reference name is the `af-feature`, since we create the index uniquely for each of those and the path to the index is not written out as part of `scpca-meta.json`.

### DIFF
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -112,8 +112,7 @@ workflow map_quant_rna {
       .branch{
           has_rad: (!params.repeat_mapping
                     && file(it.rad_dir).exists()
-                    && Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "ref_assembly") == "${it.ref_assembly}"
-                    && Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "t2g_3col_path") == "${it.t2g_3col_path}"
+                    && Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "salmon_splici_index") == "${it.salmon_splici_index}"
                     )
           make_rad: true
        }
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -111,7 +111,7 @@ workflow map_quant_rna {
        // split based in whether repeat_mapping is false and a previous dir exists
       .branch{
           has_rad: (!params.repeat_mapping
-                    && file(it.rad_dir).exists()
+                    && file("${it.rad_dir}/map.rad").exists()
                     && Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "ref_assembly") == "${it.ref_assembly}"
                     )
           make_rad: true
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -111,7 +111,7 @@ workflow map_quant_rna {
        // split based in whether repeat_mapping is false and a previous dir exists
       .branch{
           has_rad: (!params.repeat_mapping
-                    && file("${it.rad_dir}/map.rad").exists()
+                    && file(it.rad_dir).exists()
                     && Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "ref_assembly") == "${it.ref_assembly}"
                     )
           make_rad: true
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -110,7 +110,11 @@ workflow map_quant_rna {
            it}
        // split based in whether repeat_mapping is false and a previous dir exists
       .branch{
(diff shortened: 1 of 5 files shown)


modules/af-rna.nf
```diff
@@ -110,7 +110,10 @@ workflow map_quant_rna {
            it}
        // split based in whether repeat_mapping is false and a previous dir exists
       .branch{
-          has_rad: !params.repeat_mapping && file(it.rad_dir).exists()
+          has_rad: (!params.repeat_mapping
+                    && file(it.rad_dir).exists()
+                    && Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "salmon_splici_index") == "${it.salmon_splici_index}"
```
jashapiro, line 115: I'm trying to decide if we really need the paths to match or if the assembly name is sufficient. While this is a better match for local repeatability, it may be too much detail for later runs, especially if we are reprocessing data that was initially externally processed.  In that case we would not expect the index path to match, but the assembly would still match.

allyhawkins: I think you make a really good point here. I was going off this part of the comment in the issue: 
> Not sure why we don't do that check with the alevin data too? It seems like we should (and maybe check the actual index file rather than the assembly name?) I would do this in a separate PR...

But we filed that before we really got started with the call for contributions so I can revert to checking the reference name instead.

allyhawkins: Updated in https://github.com/AlexsLemonade/scpca-nf/pull/346/commits/3472b60fd01429f5b09355e955b1056713e98c2e

jashapiro (approved):
This looks good, and I'm approving, but I had a thought to look beyond the top directories only because I bit myself with this in the last couple weeks. The new stub profile with local running should mean this shouldn't come up in this exact way, but it could come up if there are other failure modes that sneak through. I don't think we want to check every file within the directories, but picking one might sometimes be useful?

allyhawkins:
> This looks good, and I'm approving, but I had a thought to look beyond the top directories only because I bit myself with this in the last couple weeks. The new stub profile with local running should mean this shouldn't come up in this exact way, but it could come up if there are other failure modes that sneak through. I don't think we want to check every file within the directories, but picking one might sometimes be useful?

I added a check for one other internal file or directory inside the output directory based on this suggestion in https://github.com/AlexsLemonade/scpca-nf/pull/346/commits/37b9bc634df90aa8abf3102e5f4e084487f231df. Because of that, I'm going to re-request review.

jashapiro:
> > This looks good, and I'm approving, but I had a thought to look beyond the top directories only because I bit myself with this in the last couple weeks. The new stub profile with local running should mean this shouldn't come up in this exact way, but it could come up if there are other failure modes that sneak through. I don't think we want to check every file within the directories, but picking one might sometimes be useful?
> 
> I added a check for one other internal file or directory inside the output directory based on this suggestion in [37b9bc6](https://github.com/AlexsLemonade/scpca-nf/commit/37b9bc634df90aa8abf3102e5f4e084487f231df). Because of that, I'm going to re-request review.

I actually just realized we _don't_ want to check for files, because the test I am running checks twice, to make sure the step-skipping path works, and this will break that check (not break it, but not allow us to check the way we were). So I think we have to check for the directories only as before and just deal with the potential for errant stub runs to break things in a way that requires a rerun. With the stub profile, that should not happen often, but I can't think of a way around it at the moment, without adding lots of conditional checks.

allyhawkins:
> I actually just realized we don't want to check for files, because the test I am running checks twice, to make sure the step-skipping path works, and this will break that check (not break it, but not allow us to check the way we were). So I think we have to check for the directories only as before and just deal with the potential for errant stub runs to break things in a way that requires a rerun. With the stub profile, that should not happen often, but I can't think of a way around it at the moment, without adding lots of conditional checks.

Okay I went back to just the directories. Going to go ahead and merge this.

jashapiro:
> Okay I went back to just the directories. Going to go ahead and merge this.

Sorry for the detour!

PR #348, opened 2023-06-02
Rename CITEseq to adt

sjspielman:
**Stacked on #345**

This PR updates the workflow to use `adt` as the CITEseq name, not `CITEseq`, which involved a bit of R and a bit of groovy (....gRoovy? sorry.). I decided to change the `feature_type` variable we have floating around in the meta list/dictionary/whatever it technically is called, and this information will just flow forward with updated R option defaults. 

All tested and runs through with final `_filtered.rds` & `_processed.rds` looking as expected!

### DIFF
--- a/modules/sce-processing.nf
+++ b/modules/sce-processing.nf
@@ -46,7 +46,9 @@ process make_merged_unfiltered_sce{
         // add feature metadata as an element of the main meta object
         meta['feature_type'] = feature_meta.technology.split('_')[0]
         meta['feature_meta'] = feature_meta
-
+        
+        // If feature_type is "CITEseq", make it "adt"
+        meta['feature_type'] = meta['feature_type'] == "CITEseq" ? "adt":meta['feature_type']
 
         """
         generate_unfiltered_sce.R \
@@ -82,10 +84,10 @@ process filter_sce{
         filtered_rds = "${meta.library_id}_filtered.rds"
 
         // Three checks for whether we have ADT data:
-        // - technology should be CITEseq
+        // - technology should be adt
         // - barcode file should exist
         // - barcode file should _not_ be the empty file NO_FILE.txt
-        adt_present = meta.feature_type == 'CITEseq' &
+        adt_present = meta.feature_type == 'adt' &
           feature_barcode_file.exists() &
           feature_barcode_file.name != "NO_FILE.txt"
 
@@ -217,6 +219,7 @@ workflow generate_merged_sce {
     feature_sce_ch = feature_quant_channel
       // RNA meta is in the third slot here
       .map{it.toList() + [file(it[2].mito_file), file(it[2].ref_gtf)]}
+
 
     make_merged_unfiltered_sce(feature_sce_ch)
 
--- a/modules/sce-processing.nf
+++ b/modules/sce-processing.nf
@@ -220,7 +220,6 @@ workflow generate_merged_sce {
       // RNA meta is in the third slot here
       .map{it.toList() + [file(it[2].mito_file), file(it[2].ref_gtf)]}
 
-
     make_merged_unfiltered_sce(feature_sce_ch)
 
     // append the feature barcode file
--- a/modules/sce-processing.nf
+++ b/modules/sce-processing.nf
@@ -48,7 +48,9 @@ process make_merged_unfiltered_sce{
         meta['feature_meta'] = feature_meta
         
         // If feature_type is "CITEseq", make it "adt"
-        meta['feature_type'] = meta['feature_type'] == "CITEseq" ? "adt":meta['feature_type']
+        if (meta['feature_type'] == "CITEseq") {
+          meta['feature_type'] = "adt"
+        }
 
         """
         generate_unfiltered_sce.R \
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -1,5 +1,7 @@
 # CITE-seq Experiment Summary
 
+This section details quality control statistics from the ADT (antibody-derived tag) component of CITE-seq experiments.
+
 ## CITE-seq Statistics
 
 ```{r}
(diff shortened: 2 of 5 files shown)


modules/sce-processing.nf
```diff
@@ -46,7 +46,9 @@ process make_merged_unfiltered_sce{
         // add feature metadata as an element of the main meta object
         meta['feature_type'] = feature_meta.technology.split('_')[0]
         meta['feature_meta'] = feature_meta
-
+        
+        // If feature_type is "CITEseq", make it "adt"
+        meta['feature_type'] = meta['feature_type'] == "CITEseq" ? "adt":meta['feature_type']
```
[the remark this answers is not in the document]
jashapiro: We could do it all in one line, with something like:
```
meta['feature_type'] = feature_meta.technology.split('_')[0].replaceFirst(/^CITEseq$/, "adt")
```

But if we want to keep separate steps, I'd use a simple `if` rather than ternary operator. I think I prefer this version, for clarity.
```suggestion
        if (meta['feature_type'] == "CITEseq") {
          meta['feature_type'] = "adt"
        }
```

jashapiro (commented):
I made a small suggestion, but in addition that that I believe there are some changes in the QC report that would require changes from an assumption of `"CITEseq"` both in text and code.

sjspielman:
> but in addition that that I believe there are some changes in the QC report

Oh, absolutely there will be _several_. I was planning to tackle this when I hopped back to #343 after this, but can do them in this PR if preferred!

modules/sce-processing.nf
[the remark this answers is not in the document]
sjspielman: I also prefer the `if` for clarity.

jashapiro:
> > but in addition that that I believe there are some changes in the QC report
> 
> Oh, absolutely there will be _several_. I was planning to tackle this when I hopped back to #343 after this, but can do them in this PR if preferred!

I think it makes sense to do all the changes here, otherwise we risk some breakage in intermediate runs.

sjspielman:
I've updated some of the naming in the QC reports:
- Main report: https://github.com/AlexsLemonade/scpca-nf/pull/348/commits/da7217d81b5ff2153cda70d893436e6cffbf9833
- CITE-seq report: https://github.com/AlexsLemonade/scpca-nf/pull/348/commits/448b56420e3ca5b8ede903b30cfeb5e90160d480

Not entirely sure I stand by all these changes, or whether more are needed. But, the main goal is consistency, so I tried to get us closer to that goal! More 👀 definitely appreciated here.

templates/qc_report/cite_qc.rmd
```diff
@@ -13,9 +15,9 @@ cell_adt_counts <- Matrix::colSums(counts(cite_exp))
 cite_information <- tibble::tibble(
   "Number of ADTs assayed" = 
     format(nrow(cite_exp), big.mark = ',', scientific = FALSE),
-  "CITE-seq reads sequenced" = 
+  "Number of reads sequenced" =
```
allyhawkins, line 18: I think we still want to specify here that these are reads only coming from the ADTs. Even though this is the adt only section, it still might be a good idea to decrease potential confusion. 
```suggestion
  "Number of ADT reads sequenced" = 
```

templates/qc_report/cite_qc.rmd
```diff
@@ -13,9 +15,9 @@ cell_adt_counts <- Matrix::colSums(counts(cite_exp))
 cite_information <- tibble::tibble(
   "Number of ADTs assayed" = 
     format(nrow(cite_exp), big.mark = ',', scientific = FALSE),
-  "CITE-seq reads sequenced" = 
+  "Number of reads sequenced" = 
     format(cite_meta$total_reads, big.mark = ',', scientific = FALSE),
```
allyhawkins, line 19:
```suggestion
    format(adt_meta$total_reads, big.mark = ',', scientific = FALSE),
```
Should we alos be changing these variable names to refer to as `adt_meta` and `adt_exp`?

allyhawkins (commented):
I took a quick look and this mostly makes sense to me, I just had a few small comments I wanted to add.

templates/qc_report/cite_qc.rmd
sjspielman: yeah, good call!

templates/qc_report/cite_qc.rmd
sjspielman: This I ended up deleting because it sounded weird to me, in the next line, to say "ADT reads mapped to ADTs," but on the other hand I guess it is what it is!

allyhawkins: I would leave the next one as is. This is the total reads where as the next one is the percent of the total reads that map to ADT. So maybe this should really be `Total number of reads sequenced (ADT library)`. Or something to reflect that it is specifically just the ADT library.

sjspielman: > Total number of reads sequenced (ADT library)

I like this!

sjspielman: Anyways while implementing this and reading some of the other tables again, I've decided "Number of ADT reads sequenced" is now my preference, so you were right! ;)

templates/qc_report/qc_report.rmd
```diff
@@ -163,7 +163,7 @@ if (has_cellhash){
     mutate(
       "Number of HTOs assayed" = 
         format(nrow(multiplex_exp), big.mark = ',', scientific = FALSE),
-      "Multiplex reads sequenced" = 
+      "Number of multiplex reads sequenced" = 
         format(multiplex_meta$total_reads, big.mark = ',', scientific = FALSE),
       "Percent of multiplex reads mapped to HTOs" =
```
jashapiro, line 168: I think "multiplex reads" is probably too generic... bad earlier choice of words.
```suggestion
      "Number of cellhash reads sequenced" = 
        format(multiplex_meta$total_reads, big.mark = ',', scientific = FALSE),
      "Percent of cellhash reads mapped to HTOs" = 
```

templates/qc_report/cite_qc.rmd
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/348/commits/e55cd07f98bb5236a79ff9c3263cd863e1edb1fd

allyhawkins (approved):
LGTM

sjspielman:
Not merging yet just in case @jashapiro wants to have another look, fyi!

jashapiro (approved):
LGTM

PR #349, opened 2023-06-09
Add `.gitkeep` to scpca-references but ignore the rest

sjspielman:
After much very slight irritation when running `git status`, I decided to implement this - keep the `scpca-references/` folder, but ignore its contents. 

Not necessary but if others agree that it's helpful, we can merge in...and if it's not helpful, we can not merge it in!


.gitignore
```diff
@@ -3,3 +3,7 @@
 
 # ignore work directory
 work/
+
+# ignore contents of scpca-references
+scpca-references/
```
allyhawkins, line 8: I think you could ignore this whole folder? This should only be in the folder if you've run the `get_refs.py` script. I think nextflow will also pull the references if you're running locally, but either way we don't need this folder in the repo at all.

sjspielman: even better!

allyhawkins (approved):
🚀

PR #350, opened 2023-06-09
Process to export SCE as AnnData

### LINKED ISSUE #226: Add process to convert SCE to AnnData

We want to be able to output the gene by cell counts matrices as AnnData objects stored as H5 files as well as output the SCE objects we already produce. Currently there are multiple SCE objects that get output as RDS files, the unfiltered, filtered, and processed RDS files. We probably want to have all of them also available as AnnData objects so that they are directly compatible with python. 

We can create a process that converts RDS files with SCE objects to H5 files with AnnData objects and then run that process 3 times, once for each type of SCE file. Alternatively we could have a process that converts all three in one go using the same input that the QC report uses and then runs a script that converts all 3. I think I might favor the second option, but I'm curious if others have other ideas. 

Either way we will definitely need a script that takes RDS files as input and outputs H5 files using the `scpcaTools::sce_to_anndata` function. 

**Additional context**

We may need to break this up into additional issues as we start to work on this but I wanted to file this to get us started on adding AnnData conversion. We want to do this before we have our next big release and re-process projects.

allyhawkins:
Closes #226 

This PR adds in both the script and process to the workflow to convert an RDS file containing an SCE object to an HDF5 file containing an AnnData object. 

- The script is pretty straightforward and takes in one SCE file at a time. I'm using the function we have in `scpcaTools` to do the conversion and export the AnnData object. 
- The process is also pretty simple and takes in a tuple of the `meta` object, the sce file to be converted, and then the type of file, e.g., if the file represents an `unfiltered`, `filtered`, or `processed` SCE object. The script for converting to AnnData is then run within that process. 
- The workflow that I added here is where most of the work is being done. The output of the `post_process_sce` which contains the `meta` object and all three files, is passed as the input. I then mapped the channel so that each entry of the channel was a tuple with the `meta` and then a single SCE file which gets passed to the process for converting. The output from the process is then grouped back together by library ID to create a single tuple with the `meta` object and all three AnnData files. 

Food for thought/ next steps: 
- For right now I have this publishing to the same results directory as the SCE objects and organized in the same way. This means all RDS and HDF5 files for a single sample will live in the same folder. We may want to re-arrange this depending on how downloads will be organized. I think if the goal is to allow for people to download things as either RDS or HDF5 then we probably want to have them separate? 
- Once we have a tuple with all of the AnnData files, we can then pass them onto additional downstream processes. One of our goals with this is to make our processed data CZI compliant. Part of that is making sure that things are in the right places and labeled correctly in the AnnData object. I am envisioning a separate process that runs a python script that works directly with the AnnData to make any modifications that we need.


modules/export-anndata.nf
```diff
@@ -0,0 +1,46 @@
...
+      // combine all anndata files by library id
+      // creates anndata channel with [meta, unfiltered, filtered, processed]
+      anndata_ch = export_anndata.out
+        .map{ [it[0]["library_id"], it[0], it[1]] }
+        .groupTuple(by: 0)
+        .map{it.drop(1)}
+
+    emit: anndata_ch
+
+}
```
sjspielman, line 46: github gods demand a ~sacrifice~ new line.

bin/sce_to_anndata.R
```diff
@@ -0,0 +1,46 @@
...
+}
+
+# Convert to AnnData -----------------------------------------------------------
+
+# read in sce
+sce <- readr::read_rds(opt$input_sce_file)
+
+# export sce as anndata object
+scpcaTools::sce_to_anndata(sce,
+                           anndata_file = opt$output_h5_file)
```
jashapiro, line 46: This is incredibly minor, but follows from my feeling that we may want to standardize on tidyverse code formatting. 
```suggestion
scpcaTools::sce_to_anndata(
  sce,
  anndata_file = opt$output_h5_file
)
```

I'm also just wondering how this function does handling ADT data? Do we need to convert that separately if it is present? (I have not looked, but I'm kind of assuming you did when writing this)

modules/export-anndata.nf
```diff
@@ -0,0 +1,51 @@
...
+                 ]}
+
+      // export each anndata file
+      export_anndata(sce_ch)
+
+      // combine all anndata files by library id
+      // creates anndata channel with [meta, unfiltered, filtered, processed]
+      anndata_ch = export_anndata.out
+        .map{ [it[0]["library_id"], it[0], it[1]] }
+        .groupTuple(by: 0)
```
jashapiro, line 46: Minor addition here, but since we know the size, we can make this a bit more efficient (It won't wait for the end to emit tuples that are done).
```suggestion
        .groupTuple(by: 0, size: 3, remainder: true)
```

The `remainder` argument means that we would emit size 2 items if there were any left. I don't expect this, so maybe we don't want it, but otherwise they just get lost?

modules/export-anndata.nf
```diff
@@ -0,0 +1,51 @@
...
+        .map{it.drop(1)}
```
jashapiro, line 47: Do we want to modify this to pull out just the info we need? I think the output of `groupTuple` will look like this:

`[library_id, [meta_unfiltered, meta_filtered, meta_processed], [anndata_unfiltered, anndata_filtered, anndata_processed]]`

I think we can trust the order to be maintained, which is good, but I also don't trust that, if you know what I mean.

So the trusting version would be:
```suggestion
        .map{ [it[1][0]] +  it[2] }
```
Where we use a single instance from the `meta` list and then separate out the other three items.

I am willing to go with just this for now, with the thought that if it turns out we can't trust the order, we could come back and make it more complicated.

modules/export-anndata.nf
```diff
@@ -0,0 +1,51 @@
+
+// process for converting rds files containing sce to h5 containing anndata
+process export_anndata{
+    container params.SCPCATOOLS_CONTAINER
+    label 'mem_8'
```
jashapiro, line 5: I have a feeling we might run into memory issues here with some frequency, so I would bump this.
```suggestion
    label 'mem_16'
```

jashapiro (commented):
Thanks for doing this! I think it looks good, with a modification to the final mapping that might be needed.

> For right now I have this publishing to the same results directory as the SCE objects and organized in the same way. This means all RDS and HDF5 files for a single sample will live in the same folder. We may want to re-arrange this depending on how downloads will be organized. I think if the goal is to allow for people to download things as either RDS or HDF5 then we probably want to have them separate?

I am not worried about this for this stage. I think at ingestion to the website they can be moved around as needed, so I don't think we should worry about subfolders.

bin/sce_to_anndata.R
jashapiro: > Yes, so this does not handle alternative experiments. Our function that we are using will only convert the main experiment. Looking back on [AlexsLemonade/scpcaTools#115](https://github.com/AlexsLemonade/scpcaTools/issues/115), we had some initial thoughts on how we wanted to handle it, one of which was outputting a separate file for each `altExp`. I think we may want to address this in a separate issue/PR here because we will have to think about what we want the output to look like there.
> 
> In looking briefly at the Scanpy documentation, it looks like they store everything in one matrix with the adt data as additional rows in the gene by cell counts matrix. Maybe we could do something similar prior to converting to anndata to keep everything in one file? https://scanpy-tutorials.readthedocs.io/en/multiomics/cite-seq/pbmc5k.html

Doing that seems kind of hacky, and I don't love it. I think maybe the right approach going forward is to to export `mudata` objects? https://mudata.readthedocs.io/en/latest/. This allows wrapping multiple anndata objects in a way much more similar to SCE. The accessing the underlying AnnData objects is done with calls like `mudata['rna']`. 

There are some remaining questions though: For example: do we want all files to be `mudata` for output, even if there is only RNA data?

All of this I think falls into future discussion, but we should probably resolve it pretty soon to prevent too much rewriting later.

allyhawkins: I'm going to file a new issue about this, but think we should tackle it after this goes in/ this sprint. Another question I have is about making sure our output is compliant with CZI cellxgene, since that is part of the goal of creating the `AnnData` output. I think because of that we may have to keep everything as `AnnData` rather than use `mudata`? 
https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#general-requirements

That being said, it sounds like they are starting to figure out CITE-seq, so maybe we could discuss with them, how best to store the CITE-seq data.

jashapiro: Agree on a new issue. But I think that we can do both pretty easily, as we should be able to make the AnnData RNA object within the `mudata` compatible with cellxgene.

allyhawkins:
Just noting that I tested this with a library with CITE-seq and things do still work, we just only get the RNA experiment in the output right now. This is ready for another review.

bin/sce_to_anndata.R
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env Rscript
+
+# This script takes a SingleCellExperiment stored in a .rds file and converts it
+# to an AnnData object saved as an hdf5 file
```
jashapiro, line 4: Expanding the comment a touch...
```suggestion
# This script takes a SingleCellExperiment stored in a .rds file and converts the main experiment
# (usually RNA) to an AnnData object saved as an hdf5 file
```

main.nf
```diff
@@ -212,6 +213,9 @@ workflow {
   // generate QC reports
   sce_qc_report(post_process_sce.out, report_template_tuple)
 
+  // convert to anndata
```
jashapiro, line 216:
```suggestion
  // convert RNA component of SCE object to anndata
```

modules/export-anndata.nf
```diff
@@ -0,0 +1,51 @@
+
+// process for converting rds files containing sce to h5 containing anndata
```
jashapiro, line 2:
```suggestion
// process for converting rds files containing an SCE to h5 containing anndata containing the RNA data
```

jashapiro (approved):
I added a few comments to clarify that (at least for now) the output file only includes RNA data. 

Other than that this looks good.

PR #355, opened 2023-06-26
Account for `altExp` in converting SCE to AnnData

### LINKED ISSUE #351: Handle altExps with AnnData output
> Yes, so this does not handle alternative experiments. Our function that we are using will only convert the main experiment. Looking back on [AlexsLemonade/scpcaTools#115](https://github.com/AlexsLemonade/scpcaTools/issues/115), we had some initial thoughts on how we wanted to handle it, one of which was outputting a separate file for each `altExp`. I think we may want to address this in a separate issue/PR here because we will have to think about what we want the output to look like there.
> 
> In looking briefly at the Scanpy documentation, it looks like they store everything in one matrix with the adt data as additional rows in the gene by cell counts matrix. Maybe we could do something similar prior to converting to anndata to keep everything in one file? https://scanpy-tutorials.readthedocs.io/en/multiomics/cite-seq/pbmc5k.html

Doing that seems kind of hacky, and I don't love it. I think maybe the right approach going forward is to to export `mudata` objects? https://mudata.readthedocs.io/en/latest/. This allows wrapping multiple anndata objects in a way much more similar to SCE. The accessing the underlying AnnData objects is done with calls like `mudata['rna']`. 

There are some remaining questions though: For example: do we want all files to be `mudata` for output, even if there is only RNA data?

All of this I think falls into future discussion, but we should probably resolve it pretty soon to prevent too much rewriting later.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/350#discussion_r1235438921_

One other thing we will want to consider is how to make sure that the output is compliant with CZI cellxgene, which requires `hdf5` files. We may be able to still use `muData` and make the individual objects CZI compliant.

allyhawkins:
Closes #351 

Here I am accounting for the presence of an `altExp` when converting SCE objects to `AnnData` objects. 

- I updated the existing Nextflow process to convert SCE to AnnData to optionally output a second `hdf5` file corresponding to the `altExp` object. 
 - In the Rscript responsible for converting SCE to AnnData, I included additional arguments where users can provide an optional `feature_name` and output file for feature data. When `feature_name` is specified, it extracts the corresponding `altExp` from the object and writes it to a separate AnnData object.
- All outputs from the export process are mixed together into a single channel. I updated the `groupTuple` option to account for up to 6 files keeping `remainder:true` so that the smaller tuples with only three files won't be discarded. 
- I tested this with both a CITE-seq library and an RNA-only library.

### DIFF
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -16,9 +16,15 @@ option_list <- list(
     help = "path to rds file with input sce object to be converted"
   ),
   make_option(
-    opt_str = c("-o", "--output_h5_file"),
+    opt_str = c("--output_rna_h5"),
     type = "character",
-    help = "path to output hdf5 file to store AnnData object. Must end in .hdf5 or .h5"
+    help = "path to output hdf5 file to store RNA counts as AnnData object. Must end in .hdf5 or .h5"
+  ),
+  make_option(
+    opt_str = c("--output_feat_h5"),
+    type = "character",
+    help = "path to output hdf5 file to store feature counts as AnnData object. 
+    Only used if the input SCE contains an altExp. Must end in .hdf5 or .h5"
   )
 )
 
@@ -32,8 +38,8 @@ if(!file.exists(opt$input_sce_file)){
 }
 
 # check that output file is h5
-if(!(stringr::str_ends(opt$output_h5_file, ".hdf5|.h5"))){
-  stop("output file name must end in .hdf5 or .h5")
+if(!(stringr::str_ends(opt$output_rna_h5, ".hdf5|.h5"))){
+  stop("output rna file name must end in .hdf5 or .h5")
 }
 
 # Convert to AnnData -----------------------------------------------------------
@@ -44,5 +50,32 @@ sce <- readr::read_rds(opt$input_sce_file)
 # export sce as anndata object
 scpcaTools::sce_to_anndata(
   sce,
-  anndata_file = opt$output_h5_file
+  anndata_file = opt$output_rna_h5
 )
+
+# check for altExp
+alt_name <- altExpNames(sce)
+
+# if altExp exists then export as hdf5
+if(!is.null(alt_name) & length(alt_name) == 1){
+  
+  # check for output file 
+  if(!(stringr::str_ends(opt$output_feat_h5, ".hdf5|.h5"))){
+    stop("output feature file name must end in .hdf5 or .h5")
+  }
+  
+  # extract altExp 
+  alt_sce <- altExp(sce, alt_name)
+  
+  # export altExp sce as anndata object
+  scpcaTools::sce_to_anndata(
+    alt_sce,
+    anndata_file = opt$output_feat_h5
+  )
+  
+}
+
+# warn that only one altExp will be exported 
+if (!is.null(alt_name) & length(alt_name) > 1){
+  warning("Only 1 altExp named with {alt_name[1]} will be exported to HDF5")
+}
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -21,6 +21,11 @@ option_list <- list(
     help = "path to output hdf5 file to store RNA counts as AnnData object. Must end in .hdf5 or .h5"
   ),
   make_option(
+    opt_str = c("--feature_name"),
+    type = "character",
+    help = "Feature type. Must match the altExp name, if present."
+  ),
+  make_option(
     opt_str = c("--output_feat_h5"),
     type = "character",
     help = "path to output hdf5 file to store feature counts as AnnData object. 
@@ -53,11 +58,13 @@ scpcaTools::sce_to_anndata(
   anndata_file = opt$output_rna_h5
 )
 
-# check for altExp
-alt_name <- altExpNames(sce)
-
-# if altExp exists then export as hdf5
-if(!is.null(alt_name) & length(alt_name) == 1){
+# if feature data exists, grab it and export to AnnData 
+if(!is.null(opt$feature_name)){
+  
+  # make sure the feature data is present 
+  if(!(opt$feature_name %in% altExpNames(sce))){
+    stop("feature_name must match name of altExp in provided SCE object.")
+  }
   
   # check for output file 
   if(!(stringr::str_ends(opt$output_feat_h5, ".hdf5|.h5"))){
@@ -65,7 +72,7 @@ if(!is.null(alt_name) & length(alt_name) == 1){
   }
   
   # extract altExp 
-  alt_sce <- altExp(sce, alt_name)
+  alt_sce <- altExp(sce, opt$feature_name)
   
   # export altExp sce as anndata object
   scpcaTools::sce_to_anndata(
@@ -73,9 +80,4 @@ if(!is.null(alt_name) & length(alt_name) == 1){
     anndata_file = opt$output_feat_h5
   )
   
-}
-
-# warn that only one altExp will be exported 
-if (!is.null(alt_name) & length(alt_name) > 1){
-  warning("Only 1 altExp named with {alt_name[1]} will be exported to HDF5")
 }
--- a/modules/export-anndata.nf
+++ b/modules/export-anndata.nf
@@ -13,12 +13,13 @@ process export_anndata{
     script:
       rna_hdf5_file = "${meta.library_id}_${file_type}.hdf5"
       feature_hdf5_file = "${meta.library_id}_${file_type}_feature.hdf5"
-
+      feature_present = meta.feature_type in ["adt", "cellhash"]
       """
       sce_to_anndata.R \
         --input_sce_file ${sce_file} \
         --output_rna_h5 ${rna_hdf5_file} \
-        --output_feat_h5 ${feature_hdf5_file}
+        --output_feat_h5 ${feature_hdf5_file} \
+        ${feature_present ? "--feature_name ${meta.feature_type}" : ''}
       """
     stub:
       hdf5_file = "${meta.library_id}_${file_type}.hdf5"
--- a/modules/export-anndata.nf
+++ b/modules/export-anndata.nf
@@ -8,19 +8,24 @@ process export_anndata{
     input:
       tuple val(meta), path(sce_file), val(file_type)
     output:
-      tuple val(meta), path(hdf5_file)
+      tuple val(meta), path(rna_hdf5_file), emit: rna_anndata
+      tuple val(meta), path(feature_hdf5_file), emit: feature_anndata, optional:true
     script:
-      hdf5_file = "${meta.library_id}_${file_type}.hdf5"
+      rna_hdf5_file = "${meta.library_id}_${file_type}.hdf5"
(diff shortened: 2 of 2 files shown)


modules/export-anndata.nf
```diff
@@ -41,9 +47,9 @@ workflow sce_to_anndata{
 
       // combine all anndata files by library id
       // creates anndata channel with [library_id, unfiltered, filtered, processed]
-      anndata_ch = export_anndata.out
+      anndata_ch = export_anndata.out.rna_anndata.mix(export_anndata.out.feature_anndata)
         .map{ [it[0]["library_id"], it[0], it[1]] }
```
jashapiro, line 51: I haven't tested this, but nextflow does have a feature to use variable group sizes with `groupKey`. I believe the syntax should look something like this, followed by removing the `size:` parameter in groupTuple. I'm also using the lambda syntax here to keep the mapping process a bit more a bit more clear, I hope, by automatically unpacking the input tuple to a series of variables (`meta` and `hdf5file`). (I'm realizing this syntax might be useful elsewhere where keeping track of what is where in a channel might get confusing, rather than it[0] all the time).

```suggestion
        .map{ meta, hdf5_file -> tuple(
          groupKey(meta.library_id, meta.feature_type in ["adt", "cellhash"]? 6 : 3),
          meta, 
          hdf5_file)
        }
```

I just had this thought while waiting to be picked up for dinner, so I will do more review later!

modules/export-anndata.nf
```diff
@@ -41,9 +47,12 @@ workflow sce_to_anndata{
...
       // creates anndata channel with [library_id, unfiltered, filtered, processed]
-      anndata_ch = export_anndata.out
-        .map{ [it[0]["library_id"], it[0], it[1]] }
-        .groupTuple(by: 0, size: 3, remainder: true)
+      anndata_ch = export_anndata.out.rna_anndata.mix(export_anndata.out.feature_anndata)
+        .map{ meta, hdf5_file -> tuple(
+          groupKey(meta.library_id, meta.feature_type in ["adt", "cellhash"]? 6 : 3),
+          meta,
+          hdf5_file)
+        }
```
jashapiro, line 55: I think you still need a `groupTuple` here?
```suggestion
        }
        .groupTuple(by: 0, remainder: true)
```

bin/sce_to_anndata.R
```diff
@@ -16,9 +16,20 @@ option_list <- list(
...
-    help = "path to output hdf5 file to store AnnData object. Must end in .hdf5 or .h5"
+    help = "path to output hdf5 file to store RNA counts as AnnData object. Must end in .hdf5 or .h5"
+  ),
+  make_option(
+    opt_str = c("--feature_name"),
+    type = "character",
+    help = "Feature type. Must match the altExp name, if present."
+  ),
+  make_option(
+    opt_str = c("--output_feat_h5"),
```
jashapiro, line 29: Three extra characters seem worth it here
```suggestion
    opt_str = c("--output_feature_h5"),
```

bin/sce_to_anndata.R
```diff
@@ -44,5 +55,29 @@ sce <- readr::read_rds(opt$input_sce_file)
...
+# if feature data exists, grab it and export to AnnData 
+if(!is.null(opt$feature_name)){
+  
+  # make sure the feature data is present 
+  if(!(opt$feature_name %in% altExpNames(sce))){
+    stop("feature_name must match name of altExp in provided SCE object.")
+  }
+  
+  # check for output file 
+  if(!(stringr::str_ends(opt$output_feat_h5, ".hdf5|.h5"))){
```
jashapiro, line 70: following earlier suggestion
```suggestion
  if(!(stringr::str_ends(opt$output_feature_h5, ".hdf5|.h5"))){
```

bin/sce_to_anndata.R
```diff
@@ -44,5 +55,29 @@ sce <- readr::read_rds(opt$input_sce_file)
...
+    stop("output feature file name must end in .hdf5 or .h5")
+  }
+  
+  # extract altExp 
+  alt_sce <- altExp(sce, opt$feature_name)
+  
+  # export altExp sce as anndata object
+  scpcaTools::sce_to_anndata(
+    alt_sce,
+    anndata_file = opt$output_feat_h5
```
jashapiro, line 80: following earlier suggestion
```suggestion
    anndata_file = opt$output_feature_h5
```

modules/export-anndata.nf
```diff
@@ -8,19 +8,25 @@ process export_anndata{
...
       tuple val(meta), path(sce_file), val(file_type)
     output:
-      tuple val(meta), path(hdf5_file)
+      tuple val(meta), path(rna_hdf5_file), emit: rna_anndata
+      tuple val(meta), path(feature_hdf5_file), emit: feature_anndata, optional:true
     script:
-      hdf5_file = "${meta.library_id}_${file_type}.hdf5"
-
+      rna_hdf5_file = "${meta.library_id}_${file_type}.hdf5"
+      feature_hdf5_file = "${meta.library_id}_${file_type}_feature.hdf5"
```
jashapiro, line 15: Should we make the file name more informative?
```suggestion
      feature_hdf5_file = "${meta.library_id}_${file_type}_{meta.feature_type}.hdf5"
```

I guess we should also decide if we want to use `_rna.hdf5` for the RNA file name?

modules/export-anndata.nf
```diff
@@ -8,19 +8,25 @@ process export_anndata{
...
+      feature_present = meta.feature_type in ["adt", "cellhash"]
```
jashapiro, line 16: If the above suggestion fails when `meta.feature_type` is not defined, we could do:
```suggestion
      if (meta.feature_type in ["adt", "cellhash"]){
        feature_present = true
        feature_hdf5_file = "${meta.library_id}_${file_type}_{meta.feature_type}.hdf5"
      } else {
        feature_present = false
      }
```

modules/export-anndata.nf
```diff
@@ -8,19 +8,25 @@ process export_anndata{
...
         --input_sce_file ${sce_file} \
-        --output_h5_file ${hdf5_file}
+        --output_rna_h5 ${rna_hdf5_file} \
+        --output_feat_h5 ${feature_hdf5_file} \
+        ${feature_present ? "--feature_name ${meta.feature_type}" : ''}
       """
     stub:
-      hdf5_file = "${meta.library_id}_${file_type}.hdf5"
+      rna_hdf5_file = "${meta.library_id}_${file_type}.hdf5"
+      feature_hdf5_file = "${meta.library_id}_${file_type}_feature.hdf5"
```
jashapiro, line 26: This should probably match the logic and filenames above, whatever we end up deciding there.

modules/export-anndata.nf
```diff
@@ -8,19 +8,25 @@ process export_anndata{
     input:
       tuple val(meta), path(sce_file), val(file_type)
     output:
-      tuple val(meta), path(hdf5_file)
+      tuple val(meta), path(rna_hdf5_file), emit: rna_anndata
+      tuple val(meta), path(feature_hdf5_file), emit: feature_anndata, optional:true
```
jashapiro, line 12: I'm wondering if we might want to handle this a bit differently so that we have all of the hdf5 files in a single object in the output.

If we use the fact that `path` allows for globs, we could make the output something like:
```
tuple val(meta), path("{${meta.library_id}_${file_type}*.hdf5")
```
I think that would catch both files. We then would need to handle the possibility of multiple files in downstream steps, but I don't think that would be too bad, and it would save us from having to deal with any of the weird `groupKey` and `groupTuple` stuff for this part of the workflow.

jashapiro (commented):
My review is a bit backward here, because I started with the `groupKey` stuff, but I think we can get rid of it all by using an output `path` glob rather than optional outputs, which seems more efficient and in keeping with what we might ultimately want. (In some ways it might obviate my desire for  using `muData` objects as well).

I'm hoping this can work for us, but let me know what problems you run into, as I am sure there will be some, as any time we try to use a new Nextflow feature!

modules/export-anndata.nf
allyhawkins: I thought about using RNA, but then that means every file is going to have `_rna` even in the absence of any additional ADT or cellhash data. But maybe that's not a bad thing. 
I was also trying to be consistent with the SCE filenames.

allyhawkins:
Thanks for the helpful suggestions. I adjusted the outputs to use the path glob rather than optional outputs. I did not add in the `if` statement around checking for `meta.feature_type`. I tested it with a library with and without an ADT or hash library and it still works regardless of the `meta.feature_type` being present or not. 

Regarding the grouping of the output, I kept it in with some slight modifications. I still want to output a tuple with the `meta` and all of the `hdf5` files corresponding to that library. I imagine as we start making modifications with the `hdf5` files and modifying the `AnnData` objects in the subsequent processes that we may go back and change this.

I also added the `_rna.hdf5` to the RNA only files. I think because each of these files only holds a single modality, we should specify that. This should be ready for another look.

modules/export-anndata.nf
```diff
@@ -42,8 +47,13 @@ workflow sce_to_anndata{
...
-        .map{ [it[0]["library_id"], it[0], it[1]] }
-        .groupTuple(by: 0, size: 3, remainder: true)
+        .map{ meta, hdf5_files -> tuple(
+          groupKey(meta.library_id, meta.feature_type in ["adt", "cellhash"]? 6 : 3),
+          meta,
+          hdf5_files.collect()
+        )}
+        .groupTuple(by: 0, remainder: true)
+        //.groupTuple(by: 0, size: meta.feature_type in ["adt", "cellhash"]? 6 : 3, remainder: true)
         .map{ [it[1][0]] +  it[2] }
```
jashapiro, line 57: I'm pretty sure all of this is redundant at this point? 

The output from the workflow should be `tuple(meta, path(hdf5_files))` so the later processing should not be required, as that is the format we already want, isn't it?
```suggestion

```

allyhawkins: I don't think so. Because we will have: 
```
[[meta, unfiltered_rna, unfiltered_feature], 
 [ meta, filtered_rna, filtered_feature],
 [ meta, processed_rna, processed_feature]]
```
But what I want is: 
```
[meta, [unfiltered_rna, unfiltered_feat, filtered_rna, filtered_feat, processed_rna, processed_feat]]
```
So having the grouping should produce this output?

jashapiro: Ah, you are correct! But I think we may want it to be a bit different from how you show it: 

```
[meta, [unfiltered_hdf5s, filtered_hdf5s, processed_hdf5s]]
```

Which is to say that the hdf5s will be in subgroups, which I think is how we probably want it. (I don't think we need the `.collect()`, but let me know if I am missing something else!

But with that, we can  still simplify a bit to: 
```suggestion
        .map{ meta, hdf5_files -> tuple(
          meta.library_id, 
          meta,
          hdf5_files
        )}
        .groupTuple(by: 0, size: 3, remainder: true)
        .map{ [it[1][0]] +  it[2] }
```

allyhawkins:
I tested the most recent additions, so this is ready for another round of review.

jashapiro (approved):
LGTM!

PR #357, opened 2023-06-27
Fix metadata `has_citeseq` bug

sjspielman:
Earlier today, I noticed that some samples were outputting `has_cite` to be False in the metadata TSV even though I knew cite-seq was there! I mentioned it to Ally, who pointed out that indeed it said `has_cite` was True, and she was right! Alas, I had confused myself - the portal is fine (metadata from which I sent her), but the pipeline is not. I have now gotten back on track to fix this.

In #348 we changed `"CITEseq"` --> `"adt"` for the altExp name, but we missed a spot, which was leading to wrong metadata outputs! Here, I've updated `bin/sce_qc_report.R` to take in the altexp name from an `optparse` argument, with `"adt"` as the default, as we have done previously, e.g. - 
https://github.com/AlexsLemonade/scpca-nf/blob/19722f43c8b5dd8ebacd0b3202ebee0b0a18ce90/bin/post_process_sce.R#L23-L28
https://github.com/AlexsLemonade/scpca-nf/blob/19722f43c8b5dd8ebacd0b3202ebee0b0a18ce90/modules/sce-processing.nf#L182

Note that the most recent commit says "untested" but it has now been tested and works as expected.

### DIFF
--- a/modules/qc-report.nf
+++ b/modules/qc-report.nf
@@ -31,7 +31,8 @@ process sce_qc_report{
           --genome_assembly "${meta.ref_assembly}" \
           --workflow_url "${workflow_url}" \
           --workflow_version "${workflow_version}" \
-          --workflow_commit "${workflow.commitId}"
+          --workflow_commit "${workflow.commitId}" \
+          --adt_name ${meta.feature_type}
         """
     stub:
         qc_report = "${meta.library_id}_qc.html"
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -156,7 +156,7 @@ if (!is.null(sce_meta$library_id)){
 
 # check for alt experiments (CITE-seq, etc)
 alt_expts <- altExpNames(unfiltered_sce)
-has_citeseq <- "CITEseq" %in% alt_expts
+has_citeseq <- "adt" %in% alt_expts
 has_cellhash <- "cellhash" %in% alt_expts
 
 
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -88,6 +88,12 @@ option_list <- list(
     help = "workflow commit hash"
   ),
   make_option(
+    opt_str = c("--adt_name"),
+    type = "character",
+    default = "adt",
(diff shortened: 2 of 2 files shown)


modules/qc-report.nf
```diff
@@ -31,7 +31,8 @@ process sce_qc_report{
           --genome_assembly "${meta.ref_assembly}" \
           --workflow_url "${workflow_url}" \
           --workflow_version "${workflow_version}" \
-          --workflow_commit "${workflow.commitId}"
+          --workflow_commit "${workflow.commitId}" \
+          --adt_name ${meta.feature_type}
```
jashapiro, line 35: Invoking it this way will make cellhash samples show up as having citeseq. I think we want to not have this as an option, but rather to just change the one line in the R script.

sjspielman: > Invoking it this way will make cellhash samples show up as having citeseq

Ahhh, I see. Will simplify!

sjspielman:
In https://github.com/AlexsLemonade/scpca-nf/pull/357/commits/eb7a32d73eb16da6be4bac8a6026ec70c263782b, I simplified this to just changing the string directly, and I got excitable with some anonymous function syntax...

sjspielman:
Hm actually, thinking about @jashapiro's comment where `cellhash` data would have incorrectly gotten labeled as ADT, I think this is possibly also a problem in the `sce-processing` code I linked when I opened this PR. If `meta.feature_type` is "cellhash", then this code will _still_ get run (and probably error out entirely), which is not ideal..

https://github.com/AlexsLemonade/scpca-nf/blob/19722f43c8b5dd8ebacd0b3202ebee0b0a18ce90/bin/post_process_sce.R#L96-L103

Perhaps this should also be `if ("adt" %in% altExpNames(sce))` and kill the `adt_name` option? I think that is something to take care of here?

jashapiro:
> Perhaps this should also be `if ("adt" %in% altExpNames(sce))` and kill the `adt_name` option? I think that is something to take care of here?

You can probably leave this code as is and just remove the `--adt_name` argument when calling the script.

PR #362, opened 2023-06-29
Really use the QC report template

jashapiro:
Apparently the last time I said I was adding the QC template, I added a line to the wrong place. This moves it to what should be the right place (and changes `scpcaTools` to `edge` for testing).


sjspielman (approved):
🚀 thanks for tracking this down!!

PR #364, opened 2023-06-30
Some QC report tweaks

sjspielman:
This PR cleans up some aspects of the QC report that I noticed (some choices I made are open for discussion!) - 

- I dealt with deprecation warnings. (did I deprecate the deprecation warnings? you decide!)
  - There are two warnings we get from `tidyverse` - use of `dplyr::across` now needs a `.cols` argument, and in `ggplot2` the `size` aesthetic is being deprecated in favor of `linewidth`. 
  - Rather than turning warnings off, I just went ahead and updated the code. 
  - As part of this, I wrote a little helper function `reformat_nulls()`, which uses `across()`, to make code a bit easier to maintain and work with. I did some testing and it works fine as expected.
- I added some backticks where they were missing around function names written in the text in the multiplex rmd
- I rephrased `emptyDrops()` -> `emptyDropsCellRanger()` as that is what we use
- One plot size tweak for fun.

Here's the current QC report for a multiplex library. Have a look also at the plot rendered for UMAPs, where we can see an example of UMAPs where the overall plot ends up very vertical as consequence of `coord_fixed()`, which doesn't seem like a problem to me since the coordinates are what the coordinates are! Just wanted to make sure folks have a sense of how `coord_fixed()` is ending up for a different dataset.
[qc_report.html.txt](https://github.com/AlexsLemonade/scpca-nf/files/11917679/qc_report.html.txt)

### DIFF
--- a/templates/qc_report/cite_qc.rmd
+++ b/templates/qc_report/cite_qc.rmd
@@ -68,9 +68,9 @@ if (has_processed) {
     "Number of cells that pass filtering threshold"  = format(filtered_cell_count),
     "Percent of cells that pass filtering threshold" =
         paste0(round(filtered_cell_count/nrow(processed_sce) * 100, digits = 2), "%"), 
-    "Normalization method" = format(processed_meta$adt_normalization)
+    "Normalization method" = format(processed_meta$adt_normalization) 
     ) |>
-    mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) |> # reformat nulls
+    reformat_nulls() |>
     t()
   
   knitr::kable(basic_statistics, align = 'r') |>
--- a/templates/qc_report/qc_report.rmd
+++ b/templates/qc_report/qc_report.rmd
@@ -36,6 +36,13 @@ theme_set(
   theme(plot.margin = margin(rep(20, 4)), 
         strip.background = element_rect(fill = 'transparent'))
 )
+
+# Helper function to change NULL -> "N/A" in a data frame
+reformat_nulls <- function(df) {
+  df |>
+    mutate(across(everything(), 
+                  .fns = ~ifelse(.x == "NULL", "N/A", .x))) 
+}
 ```
 
 ```{r sce_setup}
@@ -177,7 +184,7 @@ if (has_cellhash) {
 }
 
 sample_information <- sample_information |>
-  mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) |> # reformat nulls
+  reformat_nulls() |>
   t()
 
 # make table with sample information
@@ -205,7 +212,7 @@ processing_info <- tibble::tibble(
       transcript_type == "spliced" ~ "Spliced only",
       TRUE ~ transcript_type)
   ) |>
-  mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) |>
+  reformat_nulls() |>
   t()
 
 
@@ -251,7 +258,7 @@ if (has_processed) {
 }
 
 basic_statistics <- basic_statistics |>
-  mutate(across(.fns = ~ifelse(.x == "NULL", "N/A", .x))) |> # reformat nulls
+  reformat_nulls() |> # reformat nulls
   t()
 
 # make table with basic statistics
@@ -338,7 +345,7 @@ ggplot(grouped_celldata, aes(x = med_rank, y = med_sum, color = pct_passed)) +
   geom_point(mapping = aes(x = rank, y = sum, color = filter_pct),
              data = top_celldata,
              alpha = 0.5) +
-  geom_line(size = 2, lineend = "round", linejoin= "round") +
+  geom_line(linewidth = 2, lineend = "round", linejoin= "round") +
   scale_x_log10(labels = scales::label_number(accuracy = 1)) +
   scale_y_log10(labels = scales::label_number(accuracy = 1)) +
   scale_color_gradient2(low = "grey70",
@@ -352,7 +359,7 @@ ggplot(grouped_celldata, aes(x = med_rank, y = med_sum, color = pct_passed)) +
   ) +
   theme(legend.position = c(0, 0),
         legend.justification = c(0,0),
-        legend.background = element_rect(color = "grey20", size = 0.25),
+        legend.background = element_rect(color = "grey20", linewidth = 0.25),
         legend.box.margin = margin(rep(5, 4)))
 ```
 
@@ -379,11 +386,11 @@ ggplot(filtered_celldata,
        color = "Percent reads\nmitochondrial") +
   theme(legend.position = c(0, 1),
         legend.justification = c(0,1),
-        legend.background = element_rect(color = "grey20", size = 0.25),
+        legend.background = element_rect(color = "grey20", linewidth = 0.25),
         legend.box.margin = margin(rep(5, 4)))
 ```
 
-The above plot of cell metrics includes only droplets which have passed the `emptyDrops()` filter.
+The above plot of cell metrics includes only droplets which have passed the `emptyDropsCellRanger()` filter.
 The plot will usually display a strong (but curved) relationship between the total UMI count and the number of genes detected.
 Cells with low UMI counts and high mitochondrial percentages may require further filtering.
 
@@ -411,7 +418,7 @@ if (skip_miQC) {
          y = "Percent reads mitochondrial") +
     theme(legend.position = c(1, 1),
           legend.justification = c(1,1),
-          legend.background = element_rect(color = "grey20", size = 0.25),
+          legend.background = element_rect(color = "grey20", linewidth = 0.25),
           legend.box.margin = margin(rep(5, 4)))
 }
 ```
@@ -459,7 +466,7 @@ if (has_filtered & has_processed) {
     theme(plot.title = element_text(hjust = 0.5),
           legend.position = c(1, 1),
           legend.justification = c(1,1),
-          legend.background = element_rect(color = "grey20", size= 0.25),
+          legend.background = element_rect(color = "grey20", linewidth = 0.25),
           legend.title = element_text(hjust = 0.5))
 } else {
   glue::glue("
--- a/templates/qc_report/multiplex_qc.rmd
+++ b/templates/qc_report/multiplex_qc.rmd
@@ -53,7 +53,7 @@ knitr::kable(hto_tags, digits = 2) |>
 
 ## Demultiplexing Sample Calls 
 
-Demultiplexing was performed using both [DropletUtils::hashedDrops](https://rdrr.io/github/MarioniLab/DropletUtils/man/hashedDrops.html) and [Seurat::HTOdemux](https://rdrr.io/github/satijalab/seurat/man/HTODemux.html) only on the _filtered_ cells, using default parameters for each.
+Demultiplexing was performed using both [`DropletUtils::hashedDrops()`](https://rdrr.io/github/MarioniLab/DropletUtils/man/hashedDrops.html) and [`Seurat::HTOdemux()`](https://rdrr.io/github/satijalab/seurat/man/HTODemux.html) only on the _filtered_ cells, using default parameters for each.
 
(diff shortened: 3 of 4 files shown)


sjspielman:
In the spirit of our new planning approach, I'm going to wait to request a reviewer here so that we can assign during planning next sprint. Of course, feel free to poke around before then if you are so moved!

sjspielman:
In https://github.com/AlexsLemonade/scpca-nf/pull/364/commits/2640b652b297f264d9757b9278291566135cf2b4 I fixed a bug left that snuck in in #343 wherein CITE-seq runs fail to knit because `t()` doesn't want to work on a `Matrix`, only `matrix`. I simply swapped `t() |> as.matrix()`  --> `as.matrix() |> t()` and now it's all working. In this commit, I also fiddled with a couple print statements introducing "the plot below..." - they should only print if there _is_ a plot below (which would occur if CITEseq data is present but not a processed SCE, a niche condition....).

In https://github.com/AlexsLemonade/scpca-nf/pull/364/commits/6a531261b51677b38e07ee9d11bc6da220f97439 I finally got irritated enough with `github` asking me to commit my changes to `qc_report.html` 😑 I could make this `.gitignore` more general (`*html`?) if that's preferred!

sjspielman:
Found another bug, and fixed one styling aspect in https://github.com/AlexsLemonade/scpca-nf/pull/364/commits/fc6e054ce1861b4f587fb49b81bcd21ef87397c4 - 
- Bug: When calculating the percent of cells retained, it seems better to divide by total cells and not total genes...! The code fixed is `nrow()` -> `ncol()`.
- Styling: The table that prints out ADT stats was currently ordered just by mean UMI count. I thought it might be clearer to arrange _both_ on mean and target type. The code ensures that `target` ADTs come first, followed by `neg_control` and `pos_control`, and within each grouping the rows are arranged by mean. 

At this point my own testing is complete, so this is now ready for review!

templates/qc_report/cite_qc.rmd
```diff
@@ -97,16 +102,16 @@ if (has_processed) {
...
 The plot below displays an overall view of cell filtering based on both ADT and RNA counts. Cells are labeled as one of the following:\n\n
 
-- 'Keep': This cell is retained based on both RNA and ADT counts.\n
-- 'Filter (RNA only) ': This cell is filtered based on only RNA counts.\n
-- 'Filter (ADT only)': This cell is filtered based on only ADT counts.\n\n
-- 'Filter (RNA & ADT)': This cell is filtered based on both RNA and ADT counts.\n
+- "Keep": This cell is retained based on both RNA and ADT counts.\n
+- "Filter (RNA only)": This cell is filtered based on only RNA counts.\n
+- "Filter (ADT only)" This cell is filtered based on only ADT counts.\n\n
+- "Filter (RNA & ADT)": This cell is filtered based on both RNA and ADT counts.\n
```
jashapiro, line 111: Just noticing all the `\n` here, which I would not have expected to be needed.

templates/qc_report/cite_qc.rmd
```diff
@@ -97,16 +102,16 @@ if (has_processed) {
 if (has_processed) { 
   
   # DO NOT TAB TEXT IN, for formatting reasons
```
jashapiro, line 104: I'm guessing you didn't add all of this arbitrarily, but it is not in keeping with my understanding of the functions involved, so I want to push it a bit in case we can get rid of some extra formatting.

My main thought is that`glue` ignores leading spaces, unless you tell it to keep them. So I would expect the following to work just fine, based on my experience with `glue` and some quick testing in a separate notebook:

```
  glue::glue('
    The plot below displays an overall view of cell filtering based on both ADT and RNA counts. Cells are labeled as one of the following:
  
    - "Keep": This cell is retained based on both RNA and ADT counts.
    - "Filter (RNA only)": This cell is filtered based on only RNA counts.
    - "Filter (ADT only)" This cell is filtered based on only ADT counts.
    - "Filter (RNA & ADT)": This cell is filtered based on both RNA and ADT counts.
  
  ')  |> print()
```

This is following the rules described in https://glue.tidyverse.org/reference/trim.html. 
I also removed all of the explicit `\n`, because I have never before needed those with `glue`. 
I might imagine you could need a final `\n`, but making that literal (as above) should be fine too.

templates/qc_report/cite_qc.rmd
```diff
@@ -208,11 +218,16 @@ if (has_processed) {
...
 
-The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT. 
 
 
-```{r fig.alt="UMAP calculated from RNA expression but colored by normalized expression of highly variable ADTs", fig.height = 6}
+```{r fig.alt="UMAP calculated from RNA expression but colored by normalized expression of highly variable ADTs", fig.height = 6, results='asis'}
 if (has_processed) {
+    
+  glue::glue(
+"The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT.\n\n"
```
jashapiro, line 227:
```suggestion
    "The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT.\n\n"
```

templates/qc_report/qc_report.rmd
```diff
@@ -36,6 +36,13 @@ theme_set(
   theme(plot.margin = margin(rep(20, 4)), 
         strip.background = element_rect(fill = 'transparent'))
 )
+
+# Helper function to change NULL -> "N/A" in a data frame
+reformat_nulls <- function(df) {
+  df |>
+    mutate(across(everything(), 
+                  .fns = ~ifelse(.x == "NULL", "N/A", .x)))
```
jashapiro, line 44: Slight restyling.
```suggestion
  df |> mutate(
    across(everything(),
      .fns = \(x) ifelse(x == "NULL", "N/A", x)
    )
  )
```

templates/qc_report/cite_qc.rmd
```diff
@@ -182,16 +187,21 @@ if (has_processed) {
...
 
-The plot below displays normalized expression of these four ADTs, with one ADT shown per panel.
 
 
-```{r fig.alt="Density plot showing normalized expression of highly variable ADTs", warning = FALSE}
+```{r fig.alt="Density plot showing normalized expression of highly variable ADTs", warning = FALSE, results='asis'}
 if (has_processed) {
+  
+  glue::glue(
+"The plot below displays normalized expression of these four ADTs, with one ADT shown per panel.\n\n"
```
jashapiro, line 196: I feel like one trailing newline ought to be enough?
```suggestion
    "The plot below displays normalized expression of these four ADTs, with one ADT shown per panel.\n"
```

jashapiro (commented):
This overall looks fine, but I was confused by some formatting decisions that I would like to undo if possible. See comments below.

I also had some ugly thoughts about forcing square UMAP plots, but I don't want to add that level of complexity.

templates/qc_report/.gitignore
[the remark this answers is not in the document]
sjspielman: not really, can do

templates/qc_report/cite_qc.rmd
```diff
@@ -39,7 +39,12 @@ knitr::kable(adt_information, align = 'r') |>
 ```{r}
 antibody_tags <- as.data.frame(rowData(adt_exp)) |>
   tibble::rownames_to_column("Antibody") |>
-  arrange(desc(mean)) |>
+  # ensure `target` is the first level of target_type
+  mutate(target_type = forcats::fct_relevel(target_type, "target")) |>
+  # group by target_type before arranging
+  group_by(target_type) |>
+  # use .by_group = TRUE to arrange by target type first, then by mean
+  arrange(desc(mean), .by_group = TRUE) |>
```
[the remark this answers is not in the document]
sjspielman: Am I crazy or is this a new-ish feature then? I feel like every time I've tried to make `arrange` work with two variables it has failed me...

templates/qc_report/cite_qc.rmd
sjspielman: :100: thank you, i got so turned around in my formatting journey it seems...

templates/qc_report/qc_report.rmd
sjspielman: noting I implemented this but not via suggestion

templates/qc_report/cite_qc.rmd
sjspielman: done separately.

templates/qc_report/cite_qc.rmd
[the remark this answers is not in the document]
sjspielman: honestly i'm going with crazy...

sjspielman:
I've responded to reviews, thanks for helping clean this up! Here's a current version - 
[qc_report.html.txt](https://github.com/AlexsLemonade/scpca-nf/files/11969596/qc_report.html.txt)

.gitignore
```diff
@@ -6,3 +6,7 @@ work/
 
 # ignore scpca-references
 scpca-references/
+
+
+# ignore template htmls
+templates/qc_report/*html
```
jashapiro, line 12: I don't know if we might expect these files to be littered about other places, so I made this both more specific and more general. Also an extra space. Not sure what the _best_ option is here, but this was more what I was thinking for moving to root `.gitignore`.
```suggestion
qc_report.html
*_qc.html

```

templates/qc_report/cite_qc.rmd
```diff
@@ -208,11 +213,17 @@ if (has_processed) {
...
 
 
-```{r fig.alt="UMAP calculated from RNA expression but colored by normalized expression of highly variable ADTs", fig.height = 6}
+```{r fig.alt="UMAP calculated from RNA expression but colored by normalized expression of highly variable ADTs", fig.height = 6, results='asis'}
 if (has_processed) {
+    
+  # 2 \n\n here avoids text wrapping the plot in wide screens
+  glue::glue(
+  "The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT.\n\n"
+  ) |> print()
```
jashapiro, line 224: An alternative version here. Mostly me being opposed to `\n`, but not important.
```suggestion
  # Extra blank line here keeps the plot in a separate paragraph
  glue::glue(
    "The plot below displays UMAP embeddings calculated from RNA expression, where each cell is colored by the expression level of the given ADT.
    
    "
  ) |> print() 
```

templates/qc_report/cite_qc.rmd
```diff
@@ -96,17 +98,15 @@ if (has_processed) {
...
-- 'Filter (RNA & ADT)': This cell is filtered based on both RNA and ADT counts.\n
-
-Each panel is labeled with the given category of cells as well as the total number of cells in that category.
-  ")  |> print()
+  glue::glue('
+    The plot below displays an overall view of cell filtering based on both ADT and RNA counts. Cells are labeled as one of the following:
+  
+    - "Keep": This cell is retained based on both RNA and ADT counts.
+    - "Filter (RNA only)": This cell is filtered based on only RNA counts.
+    - "Filter (ADT only)" This cell is filtered based on only ADT counts.
```
jashapiro, line 106: Missing one colon
```suggestion
    - "Filter (ADT only)": This cell is filtered based on only ADT counts.
```

jashapiro (approved):
I made a couple of minor suggestions, but this looks good overall and I don't think I should need to see it again unless you have questions.

templates/qc_report/cite_qc.rmd
sjspielman: Fine with me, I'll test it!

sjspielman: Noting this does prevent the ugly wrapping!

templates/qc_report/cite_qc.rmd
```diff
@@ -39,7 +39,9 @@ knitr::kable(adt_information, align = 'r') |>
 ```{r}
 antibody_tags <- as.data.frame(rowData(adt_exp)) |>
   tibble::rownames_to_column("Antibody") |>
-  arrange(desc(mean)) |>
+  # ensure `target` is the first level of target_type
+  mutate(target_type = forcats::fct_relevel(target_type, "target")) |>
+  arrange(target_type, desc(mean)) |>
```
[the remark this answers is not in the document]
sjspielman: > When we have a lot of ADTs, this table gets kind of unreasonably long. 

_yes, yes it does._

I'll open an issue for circling back to this.

PR #374, opened 2023-07-07
Merging in `development` for `v0.5.2` release

sjspielman:
Towards #354 
As discussed in #353, we're sending all of `development` into the next release, which will happen right after this PR goes in.

PR #375, opened 2023-07-07
Fix dev -> main merge conflict

sjspielman:
I filed #374, but github is sad about conflicts. So, here, I merged `main` into development, and accepted the `development` changes causing the conflict (`v0.5.1` -> `v0.5.2`). Of course this ends up with no file diffs in the end, but syncs up the histories so that we can proceed to #374. Seems like the bump to `v0.5.1` happened before this `development` history was established back in May/June, hence the need to sync up I suppose.


jashapiro (approved):
> Seems like the bump to `v0.5.1` happened before this `development` history was established back in May/June, hence the need to sync up I suppose.

0.5.1 was a "hotfix"-type release, with cherry-picked commits from dev, so this isn't too surprising. Part of the reason I didn't want to continue that pattern with this release.

PR #376, opened 2023-07-07
Add more stubs and instructions

### LINKED ISSUE #333: Document stub workflow
In https://github.com/AlexsLemonade/scpca-nf/pull/332, I added some test files (all blank) to use when testing with the `-stub` workflow processes, which can save lots of time downloading big files & also adds a way to test workflow logic automatically.

This needs some documentation in the internal instructions for development, along with adding the requirement of a `stub:` for each new process we add.

### LINKED ISSUE #334: More stub tests

jashapiro:
Closes #334
Closes #333

Here I am adding a few more stub tests, with the main addition being the visium testing. I also added some instructions to the contribution guide describing the general usage of the stub workflows.

I expect that this document could use more love in the future, but I wanted to knock out this part while I was thinking about it, and to test whether we need to have a (fake) image for spaceranger or not.


jashapiro:
Looks like it all worked without having to set a docker image. Hooray!

CONTRIBUTING.md
```diff
@@ -7,16 +7,24 @@ We welcome contributions to the `scpca-nf` workflow, including usage reports and
...
-To allow for efficient review, please include in any pull request a concise and clear explanation of the changes you have made and the issues addressed. 
+All pull requests must be reviewed before merging to `development`.
+To allow for efficient review, please include in any pull request a concise and clear explanation of the changes you have made and the issues addressed.
 
 When the changes in `development` merit a new release, a pull request will be filed to merge the current version of the `development` branch into `main`, followed by tagging a release on the `main` branch.
 
+## Stub workflows
+
+All Nextflow processes should include a [`stub` block](https://www.nextflow.io/docs/latest/process.html#stub) with a minimal script that can be run quickly to produce files in the expected output locations.
+At this stage this is purely used to allow for testing of the main workflow logic, and does not need to test internal logic of the process.
```
sjspielman, line 26:
```suggestion
At this stage this is purely used to allow for testing of the main workflow logic rather than the internal logic of different processes.
```

CONTRIBUTING.md
```diff
@@ -7,16 +7,24 @@ We welcome contributions to the `scpca-nf` workflow, including usage reports and
...
+
+The `stub-metadata.tsv` file is used to define input libaries that will be used for testing, so any additions to the overall workflow that will allow processing of a new library type should be added there, along with the appropriate input files (usually empty files with the expected names) for that library type in the `test/runs` directory.
```
sjspielman, line 28: splitting this sentence up a bit so it's easier to digest, typo fix, and add a rel link.
```suggestion
The [`stub-metadata.tsv`](test/stub-metadata.tsv) file is used to define input libraries that will be used for testing.
Any additions to the overall workflow that will allow processing of a new library type should be added into `stub-metadata.tsv, along with the appropriate input files (usually empty files with the expected names) for that library type in the `test/runs` directory.
```

CONTRIBUTING.md
```diff
@@ -7,16 +7,24 @@ We welcome contributions to the `scpca-nf` workflow, including usage reports and
...
+If a new reference type is needed, that should be defined in the `stub-refs.json` file.
```
sjspielman, line 29:
```suggestion
If a new reference type is needed, that should be defined in the [`stub-refs.json`](test/stub-refs.json) file.
```

test/stub-metadata.tsv
sjspielman, line 1: noting that everything here looks good 👍

sjspielman (approved):
LGTM! Just made a few wording suggestions. I wonder if we'll want a more involved `test/README.md` eventually. It's not _crucial_ at this stage I think, but if this directory continues to grow, it seems useful. If you want to pop something in now go for it, or we can circle back

Btw do we want this in `v0.5.2`?

jashapiro:
Used your suggestions, with slight modifications. I was doing this all in vscode though, which interestingly doesn't give you the credit you would get if I had clicked the accept buttons on the website. Oh well.

I think it makes sense to lump this into 0.5.2, but probably not a big deal.

PR #379, opened 2023-07-14
Fix ADT filtering

### LINKED ISSUE #378: Bug: ADT post-processing fails when cleanTagCounts assumptions are not met
After the `v0.5.2` release 😞 , we have observed that a small amount of datasets fail during `post_process_sce` specifically because of ADT quality control statistics.
In certain (infrequent!) circumstances, the assumptions of `cleanTagCounts()`, which uses `ambientContibSparse()` in the _absence_ of isotype controls, are not met ([source](https://github.com/MarioniLab/DropletUtils/blob/e44a19f907997bc88c682d353b4883708a7e86b2/R/ambientContribSparse.R)).
From the source:
>  The assumption here is that of sparsity, i.e., no more than \code{prop * nrow(y)} features should be actually present in each cell with a non-zero number of molecules.
> This is reasonable for most tag-based applications where we would expect only 1-2 tags (for cell hashing) or a minority of tags (for general CITE-seq) to be present per cell.
> Thus, counts for all other features must be driven by ambient contamination, allowing us to estimate a scaling factor for each cell based on the ratio to the ambient profile.

So, we'll need to get a patch in for this. We have discussed taking the following strategy - 
- If `discard` has `NA` values, instead use `zero.ambient` for filtering: cells get "Keep" if  `zero.ambient` is `TRUE` and "Remove" if it is `FALSE`.
- If _both_ `discard` and `zero.ambient` have `NA` values, we might just fail on the post-processing altogether? Or, open for discussion here!

Worth noting that `zero.ambient` is always returned by `cleanTagCounts()` regardless of whether any isotype controls are present, so this strategy can be used universally.

CC @jashapiro @allyhawkins

sjspielman:
Closes #378 

This PR fixes the ADT filtering bug, with some aspects up for discussion during review. I took the approach outlined in the corresponding issue:
- If both of the `cleanTagCounts` output columns `discard` and `zero.ambient` have `NA`s, totally bail on filtering (rather that just "skipping" filtering on cells with `NA` - I thought a uniform approach would be cleaner).
  - In this case, `metadata(processed_sce)$adt_scpca_filter_method` is given a value of `No filter`, as the kids say these days. 
  - The `colData` column `processed_sce$adt_scpca_filter` is given a value of `NULL`.
  - `log1p` normalization will be performed here on all cells
- Otherwise, first try to filter on `discard`. If this column has `NA`'s, fall back to filtering on `zero.ambient`
  - `sizeFactor`s are calculated here for normalization under this circumstance, as before

I also found another tab-related bug (!) wherein `log1p` normalization was not actually put back into the SCE object after calculation, so that's good! I moved this code back out of the `if`.
 - One note here: for the case of `No filter`, I could have just calculated `log1p` directly on the SCE object, rather than the "code detour" to ensure filtered out cells have `NA` normalized values. But I could add an `if` before entering the normalization to check if no filter was performed, and then just straight up do a `logcounts(altExp(sce)) <- log1p(counts(altExp(sce))`.

I then updated the QC report accordingly. 
- I added more logic for some of the output in `cite.rmd`; rather than only checking for `has_processed`, we need to also check that generally check that `!is.null(processed_sce$adt_scpca_filter)`. 
- I also rearranged some text to better match the final context.

I'm attaching two versions of a rendered QC report on a library that first caught this error:
- [`SCPCL000706_qc.html.txt`](https://github.com/AlexsLemonade/scpca-nf/files/12051089/SCPCL000706_qc.html.txt) is the QC report as usual run through the full workflow
- [`qc_report-no-filtering.html.txt`](https://github.com/AlexsLemonade/scpca-nf/files/12051090/qc_report-no-filtering.html.txt) is a modified version of the above. To create this notebook, I manually modified the processed SCE so that `processed_sce$adt_scpca_filter <- NULL` and `metadata(processed_sce)$adt_scpca_filter_method <- "No filter"` and directly rendered the report (not via the workflow).

I know this may be a lot to digest here, so please let me know where I can clarify!! Tagging everyone for review since it would be good to have a couple eyes on this, I think.


bin/post_process_sce.R
```diff
@@ -96,20 +96,43 @@ metadata(sce)$min_gene_cutoff <- opt$gene_cutoff
...
-  )
   
-  # assign `adt_scpca_filter_method` metadata based on colData contents
-  if ("sum.controls" %in% names(colData(altExp(sce, alt_exp)))) {
-    metadata(sce)$adt_scpca_filter_method <- "cleanTagCounts with isotype controls"
-  } else if ("ambient.scale" %in% names(colData(altExp(sce, alt_exp)))) {
-    metadata(sce)$adt_scpca_filter_method <- "cleanTagCounts without isotype controls"
+  # check for NAs in filtering columns
+  use_discard <- sum(is.na(altExp(sce, alt_exp)$discard)) == 0
+  use_zero.ambient <- sum(is.na(altExp(sce, alt_exp)$zero.ambient)) == 0
```
jashapiro, line 102: So this is what I was asking about earlier, but the way I asked might have been confusing. In the case where we have `discard` for some cells but for other cells it is NA, should we not use it for those cells?

Which is to say should we be doing this fallback on a cell-by-cell basis rather than as a whole?

I am somewhat surprised that you can have `NA` for some but not all high ambient calls, but if that is really happening, then I think we probably want to use the information where we have it.

bin/post_process_sce.R
```diff
@@ -96,20 +96,43 @@ metadata(sce)$min_gene_cutoff <- opt$gene_cutoff
...
+  
+  # Fail right away if both are FALSE
+  if (!use_discard & !use_zero.ambient) {
+    warning("Failed to filter ADT data.")
+    sce$adt_scpca_filter <- NULL
```
jashapiro, line 107: I think it will cause fewer problems downstream to leave this column present, with the warning and `"No filter"`  for the method sufficient.
```suggestion
    sce$adt_scpca_filter <- "Keep"
```

bin/post_process_sce.R
sjspielman: Honestly, I'm of two minds about this. I understand why we might want to use what filtering information we have, but I am specifically wary about using _different_ kinds of filtering for the same dataset. I can imagine a situation where we have..
- some cells filtered based on `discard`
- some cells filtered based on `zero.ambient`
- some cells unfiltered only because they have `NA` for both of these values. For these cells, I would imagine something funky could be going on and maybe they should be filtered, but they would get retained in a way that it might _appear_ as though they are fine to retain. This is also sort of related to a comment you left about `NULL` -> `"Keep"`.. this also makes me nervous in similar vein because it may give the impression to users that we have a good reason to keep those cells, when actually we have no reliable information either way. 

All of this could be documented of course, but I do feel uncomfortable about using different filtering approaches for different cells in the same library. I'm happy to be convinced otherwise, but I need a little more convincing..

sjspielman: > If you can look at one of these datasets and give a breakdown, that would be helpful for evaluation. 

good call, will do!


> that is for the case where we do no filtering, so the column is describing what we _did_ not what we are endorsing. What I want to avoid is a large number of places where we and downstream users have to start checking `is.null(sce$adt_scpca_filter)` before using it

👍 convinced!

sjspielman: > This is just a column with suggestions and not actual filtering, so I think it might be helpful to inform users that a specific cell received an `NA` and no call was made by `cleanTagCounts` regarding whether the cell should be discarded or not.

I like this idea! Full inventory of `NA` cells coming soon...

bin/post_process_sce.R
sjspielman: Done in a93152c (not via suggestion).

bin/post_process_sce.R
```diff
@@ -97,33 +97,25 @@ metadata(sce)$min_gene_cutoff <- opt$gene_cutoff
...
-  # check for NAs in filtering columns
-  use_discard <- sum(is.na(altExp(sce, alt_exp)$discard)) == 0
-  use_zero.ambient <- sum(is.na(altExp(sce, alt_exp)$zero.ambient)) == 0
+  # set up filter with all Keep, and then override to Remove where necessary
+  sce$adt_scpca_filter <- "Keep"
+  sce$adt_scpca_filter[which(altExp(sce, alt_exp)$discard)] <- "Remove"
+
+  # Warnings for different types of failure:
+  fail_all_removed <- sum(sce$adt_scpca_filter == "Remove") == length(sce$adt_scpca_filter)
+  fail_all_na <- sum(is.na(altExp(sce, alt_exp)$discard)) == length(sce$adt_scpca_filter)
```
jashapiro, line 106: Simplify and speed up.
```suggestion
  fail_all_removed <- all(sce$adt_scpca_filter == "Remove")
  fail_all_na <- all(is.na(altExp(sce, alt_exp)$discard)) 
```

bin/post_process_sce.R
```diff
@@ -133,6 +125,10 @@ if (alt_exp %in% altExpNames(sce)) {
     }
   }
 
+  # make extra sure there are no NAs in `adt_scpca_filter`
+  if (sum(is.na(sce$adt_scpca_filter)) != 0) {
```
jashapiro, line 129: another little simplification
```suggestion
  if (any(is.na(sce$adt_scpca_filter))) {
```

jashapiro (approved):
I think this looks fine, but I made a couple of simplification suggestions (`all()` and `any()`).

sjspielman:
I hadn't tested yet actually, just coming back to do that now! Will accept suggestions and then give it a full go to be extra sure!

PR #382, opened 2023-07-17
Only use `label.ont` in SingleR workflow

### LINKED ISSUE #380: Use only `label.ont` reference label for SingleR celltyping
Broken out of #365 

The cell type annotation workflow should be updated to only deal with _one_ reference label. In our case, we'll default to `label.ont`. 

- The `build-celltype-ref.nf` workflow will need to be updated to only return the label defined by `params.label_type` (`params.label_name`?) which will be assigned `"label.ont"`.
- Similarly, the Rscripts `bin/train_SingleR.R` and `bin/classify_SingleR.R` should have a new option `label_<type/name>` with default `"label.ont"`, and associated code updates to only do 1 label (RIP `purrr::map()`) (and associated changes when calling scripts from workflow).

sjspielman:
Closes #380

This PR updates the `build-celltype-ref.nf` workflow to only build a model from the given `params.label_name` (default `label.ont`).  Note at one point I kept getting confused b/w two directory params, so I added some comments into `config/reference_paths.config` for my 🧠 .

I have not yet run this through to update files on `S3`, but I've tested with local output directory and it's working for me fine there!

### DIFF
--- a/bin/train_SingleR.R
+++ b/bin/train_SingleR.R
@@ -19,7 +19,7 @@ option_list <- list(
     opt_str = c("--ref_file"),
     type = "character",
     help = "path to rds file with reference dataset to use for cell type annotation.
-      These reference datasets must contain annotations labeled with the given 
+      These reference datasets must contain annotations labeled with the given
       `label_name` (default 'label.ont')."
   ),
   make_option(
@@ -80,7 +80,6 @@ ref_data <- readr::read_rds(opt$ref_file)
 
 # check that ref data contains correct labels, and set up label column for later
 label_col <- opt$label_name
-names(label_col) <- label_col
 if(!label_col %in% colnames(colData(ref_data))){
   stop(
     glue::glue("Reference dataset must contain `{label_col}` in `colData`.")
@@ -99,7 +98,7 @@ if (length(gene_ids) == 0){
   stop("Provided tx2gene tsv file does not contain any genes.")
 }
 
-# Train models -----------------------------------------------------------------
+# Train model -----------------------------------------------------------------
 
 singler_model <- SingleR::trainSingleR(
   ref_data,
@@ -109,5 +108,27 @@ singler_model <- SingleR::trainSingleR(
   restrict = gene_ids,
   BPPARAM = bp_param
 )
+
+# If the label is `label.ont` (default), save corresponding cell label names
+if (label_col == "label.ont") {
+  cl_ont <- ontoProc::getCellOnto()
+
+  # grab ontology ids from singler_model which are stored as
+  # `original.exprs` names (`original.exprs` itself is an empty list)
+  ontology_ids <- names(singler_model$original.exprs)
+
+  # grab corresponding cell names
+  cell_names <- purrr::map(
+    ontology_ids,
+    \(cl_id) cl_ont$name[cl_id]
+  ) |> unlist()
+
+  # save data frame with _matching_ ontology ids and cell names to model object
+  # use tibble to avoid rownames
+  singler_model$cell_ontology_df <- tibble::tibble(
+    ontology_id = names(cell_names),
+    ontology_cell_names = cell_names
+  )
+}
 # export models
 readr::write_rds(singler_model, opt$output_file)
--- a/bin/train_SingleR.R
+++ b/bin/train_SingleR.R
@@ -19,13 +19,19 @@ option_list <- list(
     opt_str = c("--ref_file"),
     type = "character",
     help = "path to rds file with reference dataset to use for cell type annotation.
-      These reference datasets must contain annotations labeled with
-      `label.main`, `label.fine`, and `label.ont`."
+      These reference datasets must contain annotations labeled with the given 
+      `label_name` (default 'label.ont')."
   ),
   make_option(
     opt_str = c("--output_file"),
     type = "character",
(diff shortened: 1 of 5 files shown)


bin/train_SingleR.R
```diff
@@ -72,9 +78,13 @@ if(opt$threads > 1){
 # read in model
 ref_data <- readr::read_rds(opt$ref_file)
 
-# check that ref data contains correct labels
-if(!all(c("label.main", "label.fine", "label.ont") %in% colnames(colData(ref_data)))){
-  stop("Reference dataset must contain `label.main`, `label.fine`, and `label.ont` in `colData`.")
+# check that ref data contains correct labels, and set up label column for later
+label_col <- opt$label_name
+names(label_col) <- label_col
```
allyhawkins, line 83: Do you need to do this? It doesn't look like you use the names anywhere else?

bin/train_SingleR.R
```diff
@@ -91,19 +101,13 @@ if (length(gene_ids) == 0){
...
-   )})
-
+singler_model <- SingleR::trainSingleR(
+  ref_data,
+  labels = colData(ref_data)[, label_col],
+  genes = "de",
+  # only use genes found in index
+  restrict = gene_ids,
+  BPPARAM = bp_param
+)
```
allyhawkins, line 111: Related to my comment in your other PR https://github.com/AlexsLemonade/scpca-nf/pull/383#discussion_r1265811230, I think you could add an additional step here where we use `ontoProc` to grab the cell ontology names and store them inside the model object. Then in the script we can grab the cell ontology names from the model that match up with the `pruned.labels`. I feel like adding it here makes more sense so that we can ensure the `ontoProc` mapping is only done once for each reference and the same labels are used across all datasets mapped with a given reference?

allyhawkins (commented):
My only real comment here is about where to put the `ontoProc` mapping. I'm definitely open to other opinions if you think it makes more sense to put in the other PR. Thanks again for helping with this!

bin/train_SingleR.R
sjspielman: Thanks for reminder, on it!!

bin/train_SingleR.R
sjspielman: Probably not I think.. this was just left over from "transitioning" code to a single label but could have been deleted!

sjspielman:
Womp womp, noting I had accidentally set the target of this PR to `main`..and then I merged in `main` when github told me my branch was out of date 🥴 , and at this point I thought "wait, huh?" So, I promptly updated target to `development`, and happily there are no meaningful consequences to this silliness.

sjspielman:
I've updated this to grab and save cell names as well. To fully ensure correct mapping between ontology IDs and cell names, I ended up adding a data frame called `cell_ontology_df` to the model object that stores this information. I could also have made this a named list (ontology ids as names and cell names as values), but I figure later we'll end up `inner_join`'ing this information with the cell annotations, so a data frame seemed better for that next step.

allyhawkins (commented):
I don't have any additional comments. Once `ontoProc` is added to `scpcaTools` and then Josh's other comments are addressed this should be good to go.

bin/train_SingleR.R
```diff
@@ -89,21 +98,37 @@ if (length(gene_ids) == 0){
...
+
+  # grab ontology ids from singler_model which are stored as
+  # `original.exprs` names (`original.exprs` itself is an empty list)
+  ontology_ids <- names(singler_model$original.exprs)
+
+  # grab corresponding cell names
+  cell_names <- purrr::map(
+    ontology_ids,
+    \(cl_id) cl_ont$name[cl_id]
+  ) |> unlist()
```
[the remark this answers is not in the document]
sjspielman: I'm back to looking at this, and a bunch of this code needs to change for "mystery reason." Somehow when I previously ran `SingleR::trainSingleR()` the output contents were wrong (why? who knows, it's right now!), so this whole little bit of code needs to be updated. This comment will become outdated but the new code will be simplified nonetheless.

bin/train_SingleR.R
```diff
@@ -89,21 +98,37 @@ if (length(gene_ids) == 0){
...
+  labels = colData(ref_data)[, label_col],
+  genes = "de",
+  # only use genes found in index
+  restrict = gene_ids,
+  BPPARAM = bp_param
+)
 
+# If the label is `label.ont` (default), save corresponding cell label names
+if (label_col == "label.ont") {
+  cl_ont <- ontoProc::getCellOnto()
```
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/382/commits/30a218645d200ac75196a6e7853152260c4fa034

bin/train_SingleR.R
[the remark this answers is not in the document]
sjspielman: anyways, all cleaned up in https://github.com/AlexsLemonade/scpca-nf/pull/382/commits/30a218645d200ac75196a6e7853152260c4fa034

sjspielman:
Noting that `ontoProc` was added to `scpcaTools` - https://github.com/AlexsLemonade/scpcaTools/pull/186

allyhawkins (approved):
👍

PR #383, opened 2023-07-17
Use single model for celltype annotation

### LINKED ISSUE #381: Use only one reference model per project
Broken out of #365 

The cell type annotation workflow should be updated to only take one model. 

Thoughts to this end:

- `groupTuple()` no longer needed https://github.com/AlexsLemonade/scpca-nf/blob/4b4de3c8b10415288982aabfb0c31378270ec834/modules/classify-celltypes.nf#L31-L37
  - associated comment change here https://github.com/AlexsLemonade/scpca-nf/blob/4b4de3c8b10415288982aabfb0c31378270ec834/modules/classify-celltypes.nf#L47 
- This will now have only one `singler_model` (not plural!) https://github.com/AlexsLemonade/scpca-nf/blob/4b4de3c8b10415288982aabfb0c31378270ec834/modules/classify-celltypes.nf#L8-L13
- Only 1 model! https://github.com/AlexsLemonade/scpca-nf/blob/4b4de3c8b10415288982aabfb0c31378270ec834/bin/classify_SingleR.R#L23-L28 
  - R code updates accordingly, mostly a lot less `purrr::map()`.

sjspielman:
**⚠️ Stacked on #380**
Closes #381 

This PR updates the celltype annotation workflow to only expect a single model per flow. Mostly this involved some code simplification and variable renaming for consistency with the current approach.

To run this...
- I made a dummy metadata file `test-celltype-metadata.tsv`, which I've committed here just for testing - it will need to be removed before merge!
- I copied a reference, named as described in ☝️,  that only contains `label.ont` (aka, a single set of labels generated by #380 changes) to `s3://scpca-reference/celltype/singler_models/test-BlueprintEncodeData_model.rds` for use by workflow

I ran as:
```
op run --nextflow run add-celltypes.nf -profile ccdl,batch \
  --celltype_refs_metafile /full/path/to/test-celltype-metadata.tsv \  # update your path to test!
  --run_ids SCPCL000341  # an SCPCP000007 project
```

You can see the resulting annotated RDS in `s3://scpca/processed/results/SCPCP000007/SCPCS000244/SCPCL000341_annotated.rds`, and as expected it contains the `pruned.labels` column in `colData` and the full `SingleR` output in `metadata(sce)$singler_results`.

### DIFF
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -23,15 +23,7 @@ option_list <- list(
   make_option(
     opt_str = c("--singler_model_file"),
     type = "character",
-    help = "path to file containing model generated for use with SingleR containing labels
-      for the given `label_name`."
-  ),
-  # TODO: do we even need this? We could check that the label is correct, but maybe unnecessary since only 1 is expected anyways
-  make_option(
-    opt_str = c("--label_name"),
-    type = "character",
-    default = "label.ont",
-    help = "name of cell type label to use for annotation in the reference dataset."
+    help = "path to file containing a single model generated for SingleR annotation"
   ),
   make_option(
     opt_str = c("--seed"),
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -80,14 +80,14 @@ singler_model <- readr::read_rds(singler_model_file)
 # SingleR classify and annotate SCE-------------------------------------------------------------
 
 singler_results <- SingleR::classifySingleR(
-  ref = singler_model,
+  trained = singler_model,
   test = sce,
   fine.tune = TRUE,
   BPPARAM = bp_param
 )
 
 # add annotations to SCE colData, simply
-sce$singler_results$pruned.labels
+sce$pruned.labels <- singler_results$pruned.labels
 
 # store full SingleR results in metadata
 metadata(sce)$singler_results <- singler_results
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -21,11 +21,17 @@ option_list <- list(
     help = "path to output rds file to store processed sce object. Must end in .rds"
   ),
   make_option(
-    opt_str = c("--singler_models"),
+    opt_str = c("--singler_model_file"),
     type = "character",
-    help = "list of models generated for use with SingleR. Each input file contains 
-      a list of models generated from a single reference, one each for each label type:
-      `label.main`, `label.fine`, and `label.ont`."
+    help = "path to file containing model generated for use with SingleR containing labels
+      for the given `label_name`."
+  ),
+  # TODO: do we even need this? We could check that the label is correct, but maybe unnecessary since only 1 is expected anyways
+  make_option(
+    opt_str = c("--label_name"),
+    type = "character",
+    default = "label.ont",
+    help = "name of cell type label to use for annotation in the reference dataset."
   ),
   make_option(
     opt_str = c("--seed"),
@@ -53,13 +59,9 @@ if(!file.exists(opt$input_sce_file)){
 }
 
 # check that references all exist
-model_files <- unlist(stringr::str_split(opt$singler_models, ","))
-if(!all(file.exists(model_files))){
-  missing_files <- model_files[which(!file.exists(model_files))]
-  glue::glue("
-             Missing model file(s): {missing_files}
-             ")
-  stop("Please make sure that all provided SingleR models exist.")
+singler_model_file <- opt$singler_model_file
+if(!file.exists(singler_model_file)) {
+  stop(glue::glue("Provided model file {singler_model_file} is missing."))
 }
 
 # set up multiprocessing params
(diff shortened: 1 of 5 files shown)


modules/classify-celltypes.nf
```diff
@@ -44,7 +39,7 @@ workflow annotate_celltypes {
         .map{[
           it[0], // meta
           it[1], // processed rds
-          it[2] // tuple of reference models
+          it[2] // reference model file
```
jashapiro, line 42: I'm not sure what is being done here? It seem like you are just taking the first three elements, which could be done with just `it[0..2]`, but I'm not sure if that is really what you will want, as the format of `processed_sce_channel` is not specified here.

jashapiro (commented):
Just a couple little nextflow comments...

modules/classify-celltypes.nf
allyhawkins: You should be able to use this approach that Josh suggested. I would just add a comment here that the `processed_sce_channel` is `[meta, unfiltered, filtered, processed]`. I would also add a comment before the `.map{ it[0..2] }` showing what the final structure of the channel is.

bin/classify_SingleR.R
```diff
@@ -72,44 +66,23 @@ if(opt$threads > 1){
...
+
+singler_results <- SingleR::classifySingleR(
+  trained = singler_model,
+  test = sce,
+  fine.tune = TRUE,
+  BPPARAM = bp_param
+)
+
+# add annotations to SCE colData, simply
+sce$pruned.labels <- singler_results$pruned.labels
```
allyhawkins, line 82: hmm I'm torn on this. I don't think I want to just add `pruned.labels` as the column. I liked how we had it named with `label.ont_BlueprintEncodeData`. I think what I would do is add an argument to read in the `celltype_reference_name` from the metadata file and include that in the input channel. I also want to add both the ontology ID's and the full cell ontology names. I wonder if that should be something we do here or something we add when we build the reference?

allyhawkins (commented):
This looks good to me, just some minor comments. I love how much simpler this is now!

sjspielman:
Noting the force push above was from committing to wrong branch, but back in line now!

bin/classify_SingleR.R
sjspielman: I agree we want to be able to have information about the underlying reference! Here's a different idea that might work also - since we're going to use one model for `SingleR`, would it actually makes sense to store some of this in the sce's metadata? Something like, 
```
metadata(sce)$singler_reference <- "BlueprintEncodeData"
metadata(sce)$singler_reference_label <- "label.ont"
```

I like this because the variable name can be a little shorter and more manageable for working with, fewer periods/underscores etc. Either way, that variable name should be more information, and we're going to include cell names also, so maybe something like `"celltype_annotation_ontology"` and `"celltype_annotation"` for ontology and cell name columns respectively.

In terms of when to capture that information, I wonder if makes sense to store this information in the singler model itself when building the reference rather than passing it in from nextflow. Then in the Rscript, you just pull it out of the reference object - the reference object will then also always have clear information about where its contents came from besides just its file name.

Lots of thoughts, let me know what you think!

jashapiro: The column name should have `singler` in it somewhere in anticipation of other annotations being added. I might do something like `singler_pruned.label_ont` and `singler_pruned.label_name`? 

I would agree about keeping the `singler_reference` as a metadata variable, but we might want just a bit more data there (including the package, namely), and I agree we should try to get that from the model object rather than requiring another variable.

allyhawkins: I like this! Just one edit. Since we are going to be using both `SingleR` and `CellAssign` annotations, how about we do `singler_celltype_ontology` and `singler_celltype_annotation`? Or something along those lines to indicate the annotations for `SingleR`.

test-celltype-metadata.tsv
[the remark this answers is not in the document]
sjspielman: removed in https://github.com/AlexsLemonade/scpca-nf/pull/383/commits/c12f3ad8e70d46ad4aeb43e3160d89d53031ec6d

bin/classify_SingleR.R
```diff
@@ -72,44 +66,23 @@ if(opt$threads > 1){
 # read in input rds file
 sce <- readr::read_rds(opt$input_sce_file)
```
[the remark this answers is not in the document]
sjspielman: This was actually already in the script, but that probably was not apparent during review since it was in unmodified lines (49-50) that were hidden from view here.

sjspielman:
Thanks everyone for feedback and discussion! I've updated and successfully tested 🎉 some updates!

- The R script `train_SingleR.R` now also saves the reference label and reference name to the model object as `model$reference_name` (e.g., `BlueprintEncodeData`) and `model$reference_label` (e.g., `label.ont`).
  -  For saving the `reference_name`, I just did some quick regex to extract the name from the output model file, which mirrors how this file name is constructed in nextflow from the reference name in the first place.
- The R script `classify_Single.R` now saves this information into metadata fields, and saves cell type annotations, as follows:
  - `metadata(sce)$singler_reference` holds info stored in `model$reference_name` 
  - `metadata(sce)$singler_reference_label` holds info stored in `model$reference_label`
  -  `colData(sce)$singler_celltype_annotation` holds the cell name annotation
  - `colData(sce)$singler_celltype_ontology` holds the cell name ontology, but _only_ if `label.ont` was used as the label during training. Therefore, in nextflow, `param.label_name` is passed into this R script to handle the associated logic. 
- The file `s3://scpca-references/celltype/singler_models/BlueprintEncodeData_model.rds` holds the model, as produced by `build-celltype-ref.nf`. 
- I tested the `add-celltypes.R` workflow with `SCPCL000290` (an `SCPCP000007` project), and you can see the annotated file in `s3://nextflow-ccdl-results/scpca/processed/results/SCPCP000007/SCPCL000290_annotated.rds` looking as expected!

The one thing I didn't do was in response to this comment
> but we might want just a bit more data there (including the package, namely),

Since everything saved in the SCE is being prefaced with `singler`, I wasn't sure what precisely this means in contrast to those variables, and further what it should be named... `metadata(sce)$singler_method <- "SingleR"` felt a bit odd?

allyhawkins (approved):
This looks good to me. I just have one question mostly for my understanding.

bin/classify_SingleR.R
```diff
@@ -72,44 +72,44 @@ if(opt$threads > 1){
...
+
+  colData(sce) <- colData(sce) |>
+    as.data.frame() |>
+    dplyr::left_join(
+      singler_model$cell_ontology_df,
+      by = c("singler_celltype_ontology" = "ontology_id")
+    ) |>
+    # this is the new column that was joined in with the cell names
+    dplyr::rename(singler_celltype_annotation = ontology_cell_names) |>
+    DataFrame()
```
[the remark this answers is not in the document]
sjspielman: Yup, they're `NULL` now... fix incoming!

sjspielman:
https://github.com/AlexsLemonade/scpca-nf/pull/383/commits/38849f30bef42599e9ce60da21b994f47f9dd892 restores the `NULL` colnames to be barcodes, which you can now see in the tested run - `s3://nextflow-ccdl-results/scpca/processed/results/SCPCP000007/SCPCS000216/SCPCL000290_annotated.rds`

Issue #384, opened 2023-07-17
Process for creating marker gene refs to use with CellAssign


allyhawkins:
Prior to running CellAssign we will need to create the binary marker gene by cell matrices. We will need to create a reference matrix for each tissue type that is present on the portal. This process should take as input the [PanglaoDB cell type tsv](https://panglaodb.se/markers.html?cell_type=%27choose%27) and an organ to create a reference for. For each organ, the cell types present in that organ, along with cells that contain the `Immune cell` tag in the organ column, will be selected to produce a binary gene by cell matrix. 

We should store these in the `s3://scpca-references/celltype` folder where the SingleR refs live.

allyhawkins:
Closed by #391

PR #388, opened 2023-07-20
Add SingleR and CellAssign metadata

### LINKED ISSUE #279: Create independent workflow for grabbing and saving refs for SingleR
Related to #269, I started a script that is used to download, save, and sync references that we can use as input to cell type assignment with SingleR in https://github.com/AlexsLemonade/ScPCA-admin/pull/487. However, there is not really a good reason that it needs to permanently live in that repo vs. here in a separate workflow. We could then also build on that workflow to grab additional references for more sources other than just `celldex` as we expand cell typing. 

The first step in doing that would be to port over the scripts that are living in `ScPCA-admin` to this repo.

allyhawkins:
Closes #279 and starts to address #384

As I started to work on #384, the first thing that I had to figure out was how we are going to store the metadata for the references used with `CellAssign`, and how will that relate to the way we currently store the references for `SingleR`. Previously, we had been keeping a list of references to use with `SingleR` in `ScPCA-admin`, but there is no reason to me why the list of available references for use with the workflow should be there and not in this repo. This is similar to the metadata we now have for which references are available for single-cell (`ref-metadata.tsv`). That data is stored in the `references` folder as part of the repo, so here I am porting over the `celltype-reference-metadata.tsv` and making follow-up changes to account for that. 

- In preparation for having references for both `SingleR` and `CellAssign`, I added in new columns to account for the `celltype_method`. Now we are storing both what package/ database the references come from and the cell type method. I use that column to split the channel in the workflow.
- Related to the above point, I added an `organs` column that will be used to indicate which `organs` from `PanglaoDB` should be grouped for each `ref_name`/ tissue type. I included `muscle`, `brain`, and `blood` right now as those are the most straight forward to me and probably going to be the most highly used refs anyways. 
- I added a process to grab the references from `celldex` and save them (as was the original purpose of #279). The output is the `rds` file containing the `celldex` reference and becomes the input to the training process we already had. 
- Again to prep for `CellAssign` and addressing #384, I created the channel that will become input to a process for generating marker gene reference matrices. I plan on creating the process and the script that gets run to build the marker gene matrices as a separate PR.
- I also added the PanglaoDB tsv file to the `references` folder. 
- I haven't been able to test this fully, because I need `celldex` to be part of `scpcaTools`. Once that is in, I can test this and request a formal review.


allyhawkins:
Just noting that this has now been tested and works successfully. The saved celldex refs can be found in the `s3://scpca-references/celltype/singler_references` folder and then models are in `singler_models`.

build-celltype-ref.nf
```diff
@@ -3,42 +3,88 @@ nextflow.enable.dsl=2
...
+process save_singler_refs {
+  container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.celltype_ref_dir}/singler_references"
+  label 'mem_8'
+  input:
+    tuple val(ref_name), val(ref_database)
+  output:
+    tuple val(ref_name), path(ref_file)
+  script:
+    ref_file = "${ref_database}-${ref_name}.rds"
```
sjspielman, line 15:
```suggestion
    ref_file = "${ref_source}-${ref_name}.rds"
```

build-celltype-ref.nf
```diff
@@ -3,42 +3,88 @@ nextflow.enable.dsl=2
...
+    """
+    save_singler_refs.R \
+     --ref_name ${ref_name} \
+     --ref_file ${ref_file}
+    """
+  stub:
+    ref_file = "${ref_database}-${ref_name}.rds"
```
sjspielman, line 22:
```suggestion
    ref_file = "${ref_source}-${ref_name}.rds"
```

build-celltype-ref.nf
```diff
@@ -3,42 +3,88 @@ nextflow.enable.dsl=2
...
   // train cell type references using SingleR
-  train_singler_models(celltype_refs_ch, params.t2g_3col_path)
+  train_singler_models(save_singler_refs.out, params.t2g_3col_path)
+
+  // cellassign refs
+  cellassign_refs_ch = celltype_refs_ch.cellassign
+    // create a channel with ref_name, database, organs
+    .map{[
+      ref_name: it.celltype_ref_name,
+      ref_database: it.celltype_ref_database,
```
sjspielman, line 85:
```suggestion
      ref_source: it.celltype_ref_source,
```

build-celltype-ref.nf
```diff
@@ -3,42 +3,88 @@ nextflow.enable.dsl=2
...
+  // download and save reference files
+  save_singler_refs(singler_refs_ch)
 
   // train cell type references using SingleR
-  train_singler_models(celltype_refs_ch, params.t2g_3col_path)
+  train_singler_models(save_singler_refs.out, params.t2g_3col_path)
+
+  // cellassign refs
+  cellassign_refs_ch = celltype_refs_ch.cellassign
+    // create a channel with ref_name, database, organs
```
sjspielman, line 82:
```suggestion
    // create a channel with ref_name, source, organs
```

config/reference_paths.config
```diff
@@ -7,5 +7,6 @@ ref_rootdir  = 's3://scpca-references'
 barcode_dir  = "${params.ref_rootdir}/barcodes/10X"
 
 // cell type references
-celltype_ref_dir = "${params.ref_rootdir}/celltype/references" // original (generally celldex) files themselves
-celltype_model_dir = "${params.ref_rootdir}/celltype/singler_models" // models build from original (celldex) files
+celltype_ref_dir = "${params.ref_rootdir}/celltype"
+celltype_ref_metadata = "${projectDir}/references/celltype-reference-metadata.tsv"
+panglao_marker_genes = "${projectDir}/references/PanglaoDB_markers_27_Mar_2020.tsv"
```
sjspielman, line 12:
```suggestion
panglao_marker_genes_file = "${projectDir}/references/PanglaoDB_markers_27_Mar_2020.tsv"
```

to confirm, this is just getting set up for next steps and isn't used yet, seems fine! Suggested name change just to be clear it's a _file_ and not, for example, straight-up a list of genes.

sjspielman (commented):
Looks good overall! I most have organizational and naming comments. Note that I did not test this myself at this stage, but plan to do so on 2nd round of review (also in part to test that all the suggestions, should you decide to accept them, are ok!).

- Since `references` now contains references of two different flavors for for two different workflows (e.g. transcriptome references vs cell type annotation references), time to add a quick `references/README.md`. (Later, we'll want more docs about, for example, adding more references to this workflow; can we open an issue for that?) For this `README.md`, here's some particular aspects I think we should make sure are there, beyond the baseline "what are these files" content - 
	- Indicate that the `organ` column in `celltype-reference-metadata.tsv` has relevant values from the `organ` column in the PanglaoDB
	- References (no pun intended, probably, ok maybe a little) for `celldex` and `PanglaoDB`
- I'm not sure the variable name (and its "downstream" variables `celltype_ref_database`) is the right name here, since `celldex` isn't exactly a database (though PanglaoDB is). I wondered if "source" might be a better overall term here? If you agree, I've made suggestions throughout to use the term "source" instead of "database", e.g. `celltype_ref_source` in the TSV and `ref_source` in the workflow. Or, if you strongly disagree, I'm not offended if you resolve/ignore all these suggestions! 
- I wonder if we want real params for `singler_models/` and `singler_references/` paths, e.g. 
  ```
  # path in save_singler_refs() process
  params.singler_references_dir = "${params.celltype_ref_dir}/singler_references" 
  # path in train_singler_models() process
  params.singler_models_dir = "${params.celltype_ref_dir}/singler_models"
  ```
  This will make our lives easier should these values ever have to change, since it would all be defined in the references config and not directly within the workflow

allyhawkins:
> Since references now contains references of two different flavors for for two different workflows (e.g. transcriptome references vs cell type annotation references), time to add a quick references/README.md. (Later, we'll want more docs about, for example, adding more references to this workflow; can we open an issue for that?) For this README.md, here's some particular aspects I think we should make sure are there, beyond the baseline "what are these files" content -

I definitely agree we need docs, but I think that it makes more sense to do it in a separate issue, so I filed #389. 

> I'm not sure the variable name (and its "downstream" variables celltype_ref_database) is the right name here, since celldex isn't exactly a database (though PanglaoDB is). I wondered if "source" might be a better overall term here? If you agree, I've made suggestions throughout to use the term "source" instead of "database", e.g. celltype_ref_source in the TSV and ref_source in the workflow. Or, if you strongly disagree, I'm not offended if you resolve/ignore all these suggestions!

I liked this idea! So I went ahead and took your suggestions. I knew `database` wasn't right either, but couldn't find the right word, so thank you. 

> I wonder if we want real params for singler_models/ and singler_references/ paths, e.g

I had thought about this and was on the fence a lot. But I think you're right that changing them later will be easier to manage by making them params. Additionally, I would like to ultimately make a `json` file to hold the refs/paths that we have for cell typing just like we do for the regular workflow. 

@sjspielman This is ready for another look.

sjspielman (approved):
LGTM! I just ran `build-celltype-ref.nf`, and all looks as expected in `s3://scpca-references/celltype`. My only comment is...

> Additionally, I would like to ultimately make a json file to hold the refs/paths that we have for cell typing just like we do for the regular workflow.

Makes sense to me! Should this be an issue too (and we do should it come before or after #389? - your call!)?

PR #391, opened 2023-07-24
Generate marker gene references for CellAssign

### LINKED ISSUE #384: Process for creating marker gene refs to use with CellAssign

allyhawkins:
Closes #384 

This PR builds on the additions from #388 and adds in a process for creating the marker gene x cell type matrices need for `CellAssign`. 

- I created a script that takes as input a marker gene list, a list of organs, and a gtf file and outputs a binary gene x cell type matrix with ensembl ID's. I used the same method as our data integration process to create the matrix. First, I filtered the larger marker gene table by organs and then converted the resulting gene list into a binary matrix.
- I did add options to the script for the columns that contain the cell type and the gene id. In case we ever want to use a different marker table as the reference we could still use this script but change those arguments. The only thing that I don't make optional is the `organs`, because we are specifying that in the metadata table. 
- I wanted the output to contain the genes as ensembl IDs and not as human gene symbols, which is how PanglaoDB shows them. Previously, we were grabbing the ensembl IDs from the rownames of the SCE object we are annotating. Here, we want to create a single reference that can be used on all SCE objects, so I imported the `gtf` file that we use to build our reference and grab the ids from there. 
- I tested this and the new marker gene matrices are in `s3://scpca-references/celltype/cellassign_references`

### DIFF
--- a/bin/generate_cellassign_refs.R
+++ b/bin/generate_cellassign_refs.R
@@ -15,7 +15,7 @@ option_list <- list(
     opt_str = c("--marker_gene_file"),
     type = "character",
     help = "File containing a list of marker genes for all cell types and all organs. 
-      Must contain the `organ`, `cell_type_column` and `gene_id_column` columns. "
+      Must contain the `organ` column, `--cell_type_column` and `--gene_id_column`. "
   ),
   make_option(
     opt_str = c("--cell_type_column"),
@@ -28,17 +28,20 @@ option_list <- list(
     default = "official gene symbol"
   ),
   make_option(
-    opt_str = c("--gtf"),
+    opt_str = c("--gtf_file"),
     type = "character",
     help = "reference gtf to use"
   ),
   make_option(
-    opt_str = c("--ref_file"),
+    opt_str = c("--ref_mtx_file"),
     type = "character",
     help = "path to save reference file"
   )
 )
 
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# Set up -----------------------------------------------------------------------
 
 if(!file.exists(opt$marker_gene_file)){
   stop("Provided `marker_gene_file` does not exist.")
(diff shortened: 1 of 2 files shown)


bin/generate_cellassign_refs.R
```diff
@@ -0,0 +1,120 @@
...
+if(!(gene_id_column %in% colnames(marker_gene_df))){
+  stop("Specified `gene_id_column` does not exist as a column in `marker_gene_file`")
+}
+
+# Create marker gene matrix ----------------------------------------------------
+
+# grab marker genes for specified organs from marker gene reference
+organ_gene_df <- marker_gene_df |>
+  dplyr::filter(organ %in% organs) |>
+  dplyr::select({{cell_type_column}}, {{gene_id_column}})
```
sjspielman, line 76: Oooh I've never used `{{}}` like this before but I like it! I would have done something like - 
```
dplyr::select(all_of(c(cell_type_column, gene_id_column)))
```

....but what you have done is so much cleaner :)

bin/generate_cellassign_refs.R
```diff
@@ -0,0 +1,120 @@
...
+  as.data.frame() |>
+  dplyr::select(
+    "ensembl_id" = "gene_id",
+    "gene_symbol" = "gene_name"
+  ) |>
+  dplyr::filter(.data$gene_symbol != "NA") |>
+  dplyr::distinct() |>
+  ## in case there are any duplicate gene_ids (there shouldn't be!)
+  dplyr::group_by(.data$ensembl_id) |>
+  dplyr::summarise(gene_symbol = paste(.data$gene_symbol, collapse = ";"))
```
sjspielman, line 110: we don't need `.data$` right? (also i snuck a `z` in there 🇺🇸 )


```suggestion
  dplyr::filter(gene_symbol != "NA") |>
  dplyr::distinct() |>
  ## in case there are any duplicate gene_ids (there shouldn't be!)
  dplyr::group_by(ensembl_id) |>
  dplyr::summarize(gene_symbol = paste(gene_symbol, collapse = ";"))
```

bin/generate_cellassign_refs.R
```diff
@@ -0,0 +1,120 @@
...
+  make_option(
+    opt_str = c("--cell_type_column"),
+    type = "character",
+    default = "cell type"
+  ),
+  make_option(
+    opt_str = c("--gene_id_column"),
+    type = "character",
+    default = "official gene symbol"
+  ),
```
sjspielman, line 29: Can we add `help` here?

build-celltype-ref.nf
```diff
@@ -54,6 +55,32 @@ process train_singler_models {
     """
 }
 
+process generate_cellassign_refs {
+  container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.celltype_ref_dir}/cellassign_references"
+  label 'mem_8'
+  input:
+    tuple val(ref_name), val(ref_database), val(organs)
```
sjspielman, line 63:
```suggestion
    tuple val(ref_name), val(ref_source), val(organs)
```

build-celltype-ref.nf
```diff
@@ -54,6 +55,32 @@ process train_singler_models {
...
+    path(marker_gene_file)
+    path(ref_gtf)
+  output:
+    path ref_file
+  script:
+    ref_file="${ref_database}-${ref_name}.tsv"
```
sjspielman, line 69:
```suggestion
    ref_file="${ref_source}-${ref_name}.tsv"
```

build-celltype-ref.nf
```diff
@@ -54,6 +55,32 @@ process train_singler_models {
...
+    """
+    generate_cellassign_refs.R \
+      --organs "${organs}" \
+      --marker_gene_file ${marker_gene_file} \
+      --gtf_file ${ref_gtf} \
+      --ref_mtx_file ${ref_file}
+    """
+  stub:
+    ref_file="${ref_database}-${ref_name}.tsv"
```
sjspielman, line 78:
```suggestion
    ref_file="${ref_source}-${ref_name}.tsv"
```

build-celltype-ref.nf
sjspielman: I guess this reference _is_ a database, but suggesting for overall consistency

sjspielman:
Overall looks good to me, just a couple comments to solidify it a bit!

bin/generate_cellassign_refs.R
allyhawkins: Whoops... I took a lot of this directly from the function that adds gene symbols in `scpcaTools`.

allyhawkins:
Thanks for the review @sjspielman! I incorporated your suggestions and then also added in the help messages that were missing. This is ready for another look.

sjspielman (approved):
🚀

Issue #394, opened 2023-07-28
Incorporate CellAssign predictions into annotated SCE object


allyhawkins:
CellAssign produces a predictions matrix as the output. We will need to take the predictions, get the cell type assignments, and add those to the processed SCE object that also has SingleR annotations. 

We should use the [`get_celltype_assignments` from `sc-data-integration`](https://github.com/AlexsLemonade/sc-data-integration/blob/140ce34ecf15e2af2b3c7d7efdc1a32435859eae/celltype_annotation/utils/cellassign-helper-functions.R#L36) as a reference when putting this together.

allyhawkins:
Closed by #402

PR #396, opened 2023-07-31
Add CellAssign predictions to cell type annotation workflow

### LINKED ISSUE #303: Add CellAssign to cell typing workflow
In the cell typing workflow we would like to be able to cell type using [`CellAssign`](https://docs.scvi-tools.org/en/stable/user_guide/models/cellassign.html). This means we will need to add in an additional process to the cell typing workflow that runs `CellAssign` and adds cell type annotations to the SCE object. 

The ultimate goal of adding this in is that we would have one input SCE object and then two cell typing methods are performed - `SingleR` and `CellAssign`, with the output annotation from each being added to the SCE object. 

This will also require a different reference file than the `SingleR` process, so we will need to prep and create the marker genes matrix as input for this process. We may want to break this out into multiple issues, but I'm filing this together for now and we can break things up as we start working on it.

allyhawkins:
Closes #303 

This PR adds a process for obtaining `CellAssign` predictions to the main cell type annotation workflow. 

- I added this process into the current `classify_celltypes.nf` process where we now get cell type annotations using both `SingleR` and `CellAssign`. Right now for `CellAssign` I only output the predictions and plan to add one more process to merge in the predictions from `CellAssign` to the annotated rds file with `SingleR` annotations. 
- This does rely on the assumption that every project will have both a `SingleR` reference and a `CellAssign` reference. I think that's okay, because I don't see any scenarios as of yet that we wouldn't have them. 
- I used the highest memory and cpu label that we have. During testing, this was immediately failing due to memory errors, but is running on the third retry. I'm using the max to try and eliminate the failing and re-submitting as much as possible.
- Also see [AlexsLemonade/ScPCA-admin#](https://github.com/AlexsLemonade/ScPCA-admin/pull/604) for the changes to the metadata file that I made. There I added columns for the `SingleR` ref file and `CellAssign` ref file and am directly using them in the workflow rather than creating the paths. 

~~**Note:** I am still running a test to make sure that this completes fully, but it's been running for some time so I think most of the errors have been accounted for. I'll request a review once I ensure this is able to run to completion.~~

Edit: The test completed successfully, but it did take ~16 hours as a warning for reviewers that would like to test it. 😬

### DIFF
--- a/add-celltypes.nf
+++ b/add-celltypes.nf
@@ -52,10 +52,12 @@ workflow {
              || (it.submitter == params.project)
              || (it.project_id == params.project)
             }
-    // tuple of meta, processed rds file to use as input to cell type annotation
-    .map{meta -> tuple(meta,
-                       file("${params.results_dir}/${meta.project_id}/${meta.sample_id}/${meta.library_id}_processed.rds")
-                       )}
+      // tuple of meta, processed RDS file to use as input to singleR
+      .map{meta -> tuple(meta,
+                        file("${params.results_dir}/${meta.project_id}/${meta.sample_id}/${meta.library_id}_processed.rds"),
+                        file("${params.results_dir}/${meta.project_id}/${meta.sample_id}/${meta.library_id}_processed_rna.hdf5")
+                        )}
 
     annotate_celltypes(processed_sce_ch)
+
 }
--- a/add-celltypes.nf
+++ b/add-celltypes.nf
@@ -52,7 +52,7 @@ workflow {
              || (it.submitter == params.project)
              || (it.project_id == params.project)
             }
-      // tuple of meta, processed RDS file to use as input to singleR
+      // tuple of meta, processed RDS file, processed hdf5 file to use as input to classifying celltypes
       .map{meta -> tuple(meta,
                         file("${params.results_dir}/${meta.project_id}/${meta.sample_id}/${meta.library_id}_processed.rds"),
                         file("${params.results_dir}/${meta.project_id}/${meta.sample_id}/${meta.library_id}_processed_rna.hdf5")
--- a/bin/classify_cellassign.py
+++ b/bin/classify_cellassign.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3
+
+# This script takes an AnnData object and a reference marker gene matrix
+# where genes are rows and columns are cell types
+# CellAssign is run and the predictions matrix is returned
+
+
+import os
+import anndata as adata
+import scvi
+from scvi.external import CellAssign
+from scvi import settings
+import pandas as pd
+import numpy as np
+import argparse
+import re
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-i', '--input_anndata',
+                    dest = 'input_anndata',
+                    required = True,
+                    help = 'Path to HDF5 file with processed AnnData object to annotate')
+parser.add_argument('-o', '--output_predictions',
+                    dest = 'output_predictions',
+                    required = True,
+                    help = 'Path to directory to save the predictions, should end in tsv')
+parser.add_argument('-r', '--reference',
+                    dest = 'reference',
+                    required = True,
+                    help = 'Path to marker by cell type reference file')
+parser.add_argument('-s', '--seed',
+                    dest = 'seed',
+                    type=int,
+                    default = None,
+                    help = 'Random seed to set for scvi.')
+parser.add_argument('-t', '--threads',
+                    dest = 'threads',
+                    default = 1,
+                    type = int,
+                    help = 'Number of threads to use when training the CellAssign model')
+
+args = parser.parse_args()
+
+# set seed
+scvi.settings.seed = args.seed
+
+# define the number of threads to use
+settings.num_threads = args.threads
+
+# compile extension regex
+file_ext = re.compile(r"\.hdf5$|.h5$", re.IGNORECASE)
+
+# check that input file exists, if it does exist, make sure it's an h5 file
+if not os.path.exists(args.input_anndata):
+    raise FileExistsError("--input_anndata file not found.")
+elif not file_ext.search(args.input_anndata):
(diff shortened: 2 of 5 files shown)


add-celltypes.nf
```diff
@@ -52,10 +52,12 @@ workflow {
              || (it.submitter == params.project)
              || (it.project_id == params.project)
             }
-    // tuple of meta, processed rds file to use as input to cell type annotation
-    .map{meta -> tuple(meta,
-                       file("${params.results_dir}/${meta.project_id}/${meta.sample_id}/${meta.library_id}_processed.rds")
-                       )}
+      // tuple of meta, processed RDS file to use as input to singleR
```
sjspielman, line 55: comment needs to be updated for 2 methods now

bin/classify_cellassign.py
```diff
@@ -0,0 +1,92 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-i', '--input_hdf5_file',
+                    dest = 'input_hdf5_file',
+                    required = True,
+                    help = 'Path to HDF5 file with processed AnnData object to annotate')
+parser.add_argument('-o', '--output_predictions',
+                    dest = 'output_predictions',
+                    required = True,
+                    help = 'Path to directory to save the predictions, should end in tsv')
```
sjspielman, line 26: directories dont end in tsv?

bin/classify_cellassign.py
```diff
@@ -0,0 +1,92 @@
...
+parser.add_argument('-r', '--reference',
+                    dest = 'reference',
+                    required = True,
+                    help = 'Path to marker by cell type reference file')
```
sjspielman, line 30: can you indicate in help too if this is csv or tsv? (see comments below also)

bin/classify_cellassign.py
```diff
@@ -0,0 +1,92 @@
...
+elif not file_ext.search(args.input_hdf5_file):
+    raise ValueError("--input_hdf5_file must end in either .hdf5 or .h5 and contain a processed AnnData object.")
+
+# check that marker file exists and make sure its a csv
+if not os.path.exists(args.reference):
+    raise FileExistsError("--reference file not found.")
+
+# make sure output file path is tsv file
+if not ".tsv" in args.output_predictions:
+    raise ValueError("--output_predictions must provide a file path ending in tsv")
```
sjspielman, line 65:
```suggestion
if not args.output_predictions.endswith(".tsv"):
    raise ValueError("--output_predictions must provide a file path ending in tsv")
```

bin/classify_cellassign.py
```diff
@@ -0,0 +1,92 @@
...
+# compile extension regex
+file_ext = re.compile(r"\.hdf5$|.h5$", re.IGNORECASE)
+
+# check that input file exists, if it does exist, make sure it's an h5 file
+if not os.path.exists(args.input_hdf5_file):
+    raise FileExistsError("--input_hdf5_file file not found.")
+elif not file_ext.search(args.input_hdf5_file):
+    raise ValueError("--input_hdf5_file must end in either .hdf5 or .h5 and contain a processed AnnData object.")
+
+# check that marker file exists and make sure its a csv
```
sjspielman, line 59: wait, csv or tsv? later you parse it as tsv?

bin/classify_cellassign.py
```diff
@@ -0,0 +1,92 @@
...
+if not os.path.exists(args.reference):
+    raise FileExistsError("--reference file not found.")
+
+# make sure output file path is tsv file
+if not ".tsv" in args.output_predictions:
+    raise ValueError("--output_predictions must provide a file path ending in tsv")
+
+# read in references as marker gene tables
+ref_matrix = pd.read_csv(args.reference, sep = "\t", index_col='ensembl_id')
```
sjspielman, line 68: noting the reference is parsed here as tsv

bin/classify_cellassign.py
```diff
@@ -0,0 +1,92 @@
...
+# This script takes an AnnData object and a reference marker gene matrix
+# where genes are rows and columns are cell types
+# CellAssign is run and the predictions matrix is returned
+
+
+import os
+import anndata as adata
+import scvi
+from scvi.external import CellAssign
+from scvi import settings
```
sjspielman, line 12: Is this needed given line 10? On line 45 when you set the seed, you're doing `scvi.settings.seed` anyways, so I think this can go.

modules/classify-celltypes.nf
```diff
@@ -43,10 +69,24 @@ workflow annotate_celltypes {
...
+                                                                                                  processed_hdf5,
+                                                                                                  cellassign_ref_file
+                                                                                                  )}
+
+
+      classify_singleR(singler_input_ch)
+
+      predict_cellassign(cellassign_input_ch)
 
     emit: classify_singleR.out
```
sjspielman, line 91: Would we want to emit a tuple of both outputs here?

sjspielman (commented):
Overall this looks good! Returning these initial comments before I get to testing it out since I understand it's gonna take a while...Has `CellAssign` always taken _this_ long for us? Curious which library you used here, so I can maybe use a smaller one 😂

modules/classify-celltypes.nf
allyhawkins: I'm going to temporarily update the output to have both, but note that this will probably change as part of #394.

allyhawkins:
Thanks for the review @sjspielman. I updated the comments and addressed all of your feedback. I did temporarily update the output, but again I do think that will ultimately change. 

For testing I used, SCPCL000295. I will say that I tested before I upped the cpu requirement, so it only used 4 cpus. Increasing that to 12 should hopefully decrease the amount of time we need to run.

sjspielman:
I'll try the same library then, let's see how the change in cpu/ram affected runtime!

sjspielman (approved):
Successfully ran through the same sample at a runtime of ~7 hours (not including submission time), which is better than 16! A couple stats from this process..
 
Screenshot 2023-08-02 at 1 02 34 PM

Given next steps coming in #394, this seems good to go for this stage!

PR #398, opened 2023-08-01
Use reference json file to define paths in building celltype refs

### LINKED ISSUE #390: Create cell type reference json
> I wonder if we want real params for singler_models/ and singler_references/ paths, e.g

I had thought about this and was on the fence a lot. But I think you're right that changing them later will be easier to manage by making them params. Additionally, I would like to ultimately make a `json` file to hold the refs/paths that we have for cell typing just like we do for the regular workflow. 

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpca-nf/issues/388#issuecomment-1646277746_

allyhawkins:
Closes #390 
Stacked on #396 

This PR updates the `build-celltype-refs.nf` workflow to use the paths defined in `scpca-refs.json` rather than re-defining them at the top of the script. Originally in #390 I was thinking I would also add a cell type specific json file, but as I worked on it that seemed less helpful. We only have a few parameters that are mostly used to define publish directories and then the other ones are paths to reference files we don't want to store in S3. So in conclusion, I didn't fully make a new cell type specific json file, but instead updated this to remove hard coding paths.

### DIFF
--- a/build-celltype-ref.nf
+++ b/build-celltype-ref.nf
@@ -53,7 +53,7 @@ process train_singler_models {
 
 process generate_cellassign_refs {
   container params.SCPCATOOLS_CONTAINER
-  publishDir "${cellassign_ref_dir}"
+  publishDir "${params.cellassign_ref_dir}"
   label 'mem_8'
   input:
     tuple val(ref_name), val(ref_source), val(organs), path(ref_gtf)
--- a/build-celltype-ref.nf
+++ b/build-celltype-ref.nf
@@ -79,7 +79,7 @@ process generate_cellassign_refs {
 workflow build_celltype_ref {
 
   // read in json file with all reference paths
-  ref_paths = Utils.getMetaVal(file(params.ref_json), "Homo_sapiens.GRCh38.104")
+  ref_paths = Utils.getMetaVal(file(params.ref_json), params.celltype_organism)
 
   // create channel of cell type ref files and names
   celltype_refs_ch = Channel.fromPath(params.celltype_ref_metadata)
--- a/build-celltype-ref.nf
+++ b/build-celltype-ref.nf
@@ -1,17 +1,14 @@
 #!/usr/bin/env nextflow
 nextflow.enable.dsl=2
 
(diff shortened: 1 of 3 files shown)


build-celltype-ref.nf
```diff
@@ -83,6 +78,9 @@ process generate_cellassign_refs {
 
 workflow build_celltype_ref {
 
+  // read in json file with all reference paths
+  ref_paths = Utils.getMetaVal(file(params.ref_json), "Homo_sapiens.GRCh38.104")
```
sjspielman, line 82: Do you mean...
```suggestion
  ref_paths = Utils.getMetaVal(file(params.ref_json), params.celltype_organism)
```

sjspielman (commented):
I think this looks good! I didn't directly suggest this, but I wonder if we should swap `tx2gene` --> `t2g_3col_path` for a bit clearer readability here...

- `save_singler_refs` returns `t2g_3col_path` in 3rd spot in tuple (L11)
- then gets immediately consumed by `train_singler_models` which refers to that as `tx2gene`(L33 from L104)

does this make sense to you? also asking because i want to make sure i followed all the moving parts correctly, lots of scrolling..!

allyhawkins:
Thanks for the review @sjspielman. I addressed your comments including the use of `t2g_3col_path` throughout rather than switching variable names. This should be ready for another look.

build-celltype-ref.nf
```diff
@@ -60,9 +56,8 @@ process generate_cellassign_refs {
   publishDir "${cellassign_ref_dir}"
```
sjspielman, line 56:
```suggestion
  publishDir "${params.cellassign_ref_dir}"
```

sjspielman (commented):
When testing the workflow, it ran without error... but while all of the SingleR references were created as expected, the CellAssign ones weren't actually saved. I'm 99% sure my suggestion will fix this; running it through again now, we'll find out shortly!

sjspielman (approved):
Yup, that fixed it! Happy to approve now, just made sure to accept the suggestion 🚀

Issue #366, opened 2023-06-30
Add Cell level metadata to AnnData objects


allyhawkins:
To make our `AnnData` output as compliant as possible with CZI, we need to update the existing cell metadata present in the `AnnData` objects to include the necessary entries. 
The full documentation can be found here: https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#obs-cell-metadata

In brief we will need to add: 
- [x] Assay (e.g. 10X 3' v2)
- [x] Developmental stage
- [x] Disease ontology 
- [x] Donor ID
- [x] Organism ontology 
- [x] Ethnicity 
- [x] Sex
- [x] Suspension type (e.g 10X profiling) 
- [x] Tissue ontology 

We also need to include cell type ontology, but that is a separate addition we are tracking. 

One thing to think about as we are doing this, is if we want to add in these items to the SCE objects before conversion or if we want to add them into the AnnData after conversion. I do think when we generate merged SCE objects some of these things we may want (like disease, tissue, sex) to be in our SCE objects. So maybe instead of including a process to do that in AnnData we just apply it to SCE objects.

allyhawkins:
Now that we are incorporating the sample level metadata as part of the unfiltered SCE objects, this data will already be present in the `AnnData.uns` slot. We will need to take the list of metadata present and use it to populate the cell metadata columns that are required by CZI.

allyhawkins:
Before I start to implement these changes, I wanted to outline my ideas for implementation. Right now, the SCE objects each have `sample_metadata` stored in the SCE object as a single-row data frame. The contents of the `sample_metadata` need to be stored as columns in the cell-level metadata in the AnnData object. Additionally, we need to pull out the 10X kit and suspension type from the `library_metadata` to add to the cell-level metadata. 

I think that means we need to add two steps to the script that converts SCE objects to AnnData. So the new `sce_to_anndata.R` would have the following outline: 

1. Grab library metadata for 10X kit and suspension type and add to colData (new function in `scpcaTools` or we could just keep this within the script run in the workflow)
	1. I think we only want the CZI required columns to be added here, so `assay` and `suspension_type`
2. Add sample metadata to colData (new function in `scpcaTools`)
	1. This would take all columns of the `sample_metadata` and add them as new columns in `colData`
3. Convert SCE to AnnData (already exists) 

[...]

PR #400, opened 2023-08-10
Add sample metadata to unfiltered objects

### LINKED ISSUE #395: Add sample metadata to unfiltered SCE objects
Our goal is to include the sample metadata for every dataset in the SCE objects produced by the workflow. We then will add the sample metadata as cell level metadata to the AnnData objects to be compliant with CZI. 

The first step in doing this is reading in the `scpca-sample-metadata.tsv` to the Rscript that generates the unfiltered SCE objects. The sample metadata will then be stored as a list in the `metadata` of the SCE object.

One thing to note is that we have both human readable values and ontology ids for our metadata now, do we want to include both in the objects? I think we should do that unless anyone objects.

allyhawkins:
Closes #395 

This PR adds sample level metadata to the unfiltered sce objects which in turn adds the metadata to all objects produced by the workflow. A sample metadata file is now an added param, similar to the run level metadata. This is then directly read into the script that generates the unfiltered SCE object. 

After the sample metadata is read in and filtered to the specified sample, a list of all relevant sample-level metadata is created and added to the metadata of the SCE object. Generally, this is pretty straightforward, but I was concerned about this working for all types of libraries, including multiplexed libraries. I added a single `sample_metadata` list to the overall SCE metadata to deal with this. Within that list is a list of all samples, where each entry contains the metadata for a single sample. This means that to access the sample-level metadata for a sample, you would use: `metadata(sce)$sample_metadata[[sample_id]]`. 

An alternative to this would be to collapse all the values for all samples in the library to a comma-separated list for each piece of metadata, e.g., `age: 5,6,7,8`. The problem with that is it relies on the assumption that the order of the sample ids is the same as the order of the contents of the metadata list. 

I think I favor the first option slightly more because there will always be a named list with the metadata contents for every sample, regardless of library type. But if we are not concerned about the sample metadata for multiplexed libraries, then I would be more inclined to go with the second option. Very curious what others think here!


allyhawkins:
Checks keep failing because of the following error, any ideas @jashapiro? 
```
[d7/ad9d01] Submitted process > genetic_demux_vireo:starsolo_map:index_bam (1)
WARN: Killing pending tasks (2)
Join mismatch for the following entries: key=STUBL10 values=
```

bin/generate_unfiltered_sce.R
```diff
@@ -125,6 +139,35 @@ unfiltered_sce <- unfiltered_sce |>
...
+                                       organism = single_sample_df$organism,
+                                       development_stage_ontology_term_id = single_sample_df$development_stage_ontology_term_id,
+                                       sex_ontology_term_id = single_sample_df$sex_ontology_term_id,
+                                       organism_ontology_id = single_sample_df$organism_ontology_id,
+                                       self_reported_ethnicity_ontology_term_id = single_sample_df$self_reported_ethnicity_ontology_term_id,
+                                       disease_ontology_term_id = single_sample_df$disease_ontology_term_id,
+                                       tissue_ontology_term_id = single_sample_df$tissue_ontology_term_id
+                                     )
+                                   }) |>
+  purrr::set_names(sample_ids)
```
jashapiro, line 164: Any reason not to just keep this as a data frame? Then we could just do:
```
sample_metadata <- sample_metadata_df  |>
  dplyr::filter(scpca_sample_id %in% sample_ids)
```

If you want to, you could add row names to make selection later a bit easier, or go all the way and turn it into a `DataFrame` for consistency with other elements in SCE.

modules/sce-processing.nf
```diff
@@ -198,7 +200,7 @@ workflow generate_sce {
     sce_ch = quant_channel
       .map{it.toList() + [file(it[0].mito_file), file(it[0].ref_gtf)]}
 
-    make_unfiltered_sce(sce_ch)
+    make_unfiltered_sce(sce_ch, params.sample_metafile)
```
jashapiro, line 203: I think I found the cause of the error you are seeing when I looked locally I  got an earlier error:
```
DEBUG nextflow.Session - Session aborted -- Cause: Not a valid path value: 'test/stub-sample-metadata.tsv'
```
which pointed me here:
```suggestion
    make_unfiltered_sce(sce_ch, file(params.sample_metafile))
```

Which is to say you need to pass in the file object, not just the string. Tested locally, and it works.

allyhawkins: Thank you!

bin/generate_unfiltered_sce.R
jashapiro: I think the level of complexity for access either way is about the same, so we may as well make the creation easier! Yes, 1 row data frames are strange... but this makes combining them for merging SCE objects pretty easy and logical. (Should that addition be part of this PR? I know we are not using the merging yet, but we might as well make it work with the new data)

allyhawkins: > Yes, 1 row data frames are strange... but this makes combining them for merging SCE objects pretty easy and logical. (Should that addition be part of this PR? I know we are not using the merging yet, but we might as well make it work with the new data)

What addition are you referring to? Merging SCE objects in the integration workflow?

jashapiro: "addition" is the wrong word, but yes, accounting for the new metadata in the integration workflow. But I see that is mostly a function in `scpcaTools`, so we will need to account for it there. Which makes me wonder if the code for adding sample data table to metadata should be there as well? Probably add both in a single Issue/PR.

allyhawkins: Okay right now I changed this to add in the data frame in this script. I'll file an issue regarding making sure the merging step accounts for the sample metadata. I don't know how I feel about adding in the sample metadata as part of a function in scpcaTools. Do you mean adding another argument to `read_alevin` that is the `sample_metadata`?

jashapiro: I think I would just add a `add_sample_metadata()` function that does much of the same work as here. The idea would really be just to make the location & format of this data canonical within the scpcaTools SCE objects.

bin/generate_unfiltered_sce.R
```diff
@@ -125,6 +135,15 @@ unfiltered_sce <- unfiltered_sce |>
   add_gene_symbols(gene_info = gtf) |>
   scuttle::addPerFeatureQCMetrics()
 
+# read in sample metadata and filter to sample ids
+sample_metadata_df <- readr::read_tsv(opt$sample_metadata_file) |> 
+  # rename sample id column
+  dplyr::rename(sample_id = "scpca_sample_id")
```
jashapiro, line 141: I think the renaming here is fine. I prefer to leave the scpcaTools function a bit more prescriptive & less specific to this project.

The quotes here aren't really needed, but I think it makes the funciton a bit more clear/consistent? I go back and forth, partly because I think this is one of the less well-designed functions in `dplyr`. Convenient, but confusing.
```suggestion
  dplyr::rename("sample_id" = "scpca_sample_id")
```

main.nf
```diff
@@ -36,7 +36,7 @@ include { map_quant_feature } from './modules/af-features.nf' addParams(cell_bar
 include { bulk_quant_rna } from './modules/bulk-salmon.nf'
 include { genetic_demux_vireo } from './modules/genetic-demux.nf' addParams(cell_barcodes: cell_barcodes, bulk_techs: bulk_techs)
 include { spaceranger_quant } from './modules/spaceranger.nf'
-include { generate_sce; generate_merged_sce; cellhash_demux_sce; genetic_demux_sce; post_process_sce} from './modules/sce-processing.nf'
+include { generate_sce; generate_merged_sce; cellhash_demux_sce; genetic_demux_sce; post_process_sce} from './modules/sce-processing.nf' addParams(sample_metafile: params.sample_metafile)
```
jashapiro, line 39: While this works, I don't love it. I think I would rather have the `generate_sce` workflow take the `sample_metafile` as a second input and pass things through that way.

jashapiro (commented):
This looks good, but I think there are a few places I think we may want/need to make some adjustments. 

The first is in a comment below about the way we pass the sample info to the sub-workflow. I think we can pass the file in directly as an argument to the workflow, rather than passing the `param` down. I think you could do something like update the workflow  in `sce-processing.nf` as:

```
workflow generate_sce {
  // generate rds files for RNA-only samples
  take: 
    quant_channel, sample_metafile
  main:
    sce_ch = quant_channel
      .map{it.toList() + [file(it[0].mito_file), file(it[0].ref_gtf)]}

    make_unfiltered_sce(sce_ch, sample_metafile)
...
```

And then change `main.nf` to include something more like:
```
rna_sce_ch = generate_sce(rna_quant_ch, file(params.sample_metafile))
```

Whether we do that or not, we do need to also update the `generate_merged_sce` workflow to also incorporate the sample metadata.

allyhawkins:
> While this works, I don't love it. I think I would rather have the generate_sce workflow take the sample_metafile as a second input and pass things through that way.

I updated the workflow accordingly to pass the `sample_metafile` as a second input to the `generate_sce` workflow. 

> Whether we do that or not, we do need to also update the generate_merged_sce workflow to also incorporate the sample metadata.

I also incorporated the sample metadata into the `generate_merged_sce` workflow.

main.nf
```diff
@@ -159,7 +164,7 @@ workflow {
   rna_quant_ch = map_quant_rna.out
     .filter{it[0]["library_id"] in rna_only_libs.getVal()}
   // make rds for rna only
-  rna_sce_ch = generate_sce(rna_quant_ch)
+  rna_sce_ch = generate_sce(rna_quant_ch, file(params.sample_metafile))
```
jashapiro, line 167:
```suggestion
  rna_sce_ch = generate_sce(rna_quant_ch, sample_metafile)
```

main.nf
```diff
@@ -172,7 +177,7 @@ workflow {
     .join(map_quant_rna.out.map{[it[0]["library_id"]] + it }, by: 0, failOnDuplicate: true, failOnMismatch: false)
     .map{it.drop(1)} // remove library_id index
   // make rds for merged RNA and feature quants
-  feature_sce_ch = generate_merged_sce(feature_rna_quant_ch)
+  feature_sce_ch = generate_merged_sce(feature_rna_quant_ch, file(params.sample_metafile))
```
jashapiro, line 180:
```suggestion
  feature_sce_ch = generate_merged_sce(feature_rna_quant_ch, sample_metafile)
```

jashapiro (approved):
LGTM, but I couldn't resist one more little tweak. I don't think it needs to be incorporated... up to you.

allyhawkins:
Just noting that the changes in here work, but cause an issue once we get to the conversion to AnnData. I'm planning on addressing #366 shortly so will file a PR stacked on this branch with those changes. Once everything is put back together and working we can merge it all at once.

PR #402, opened 2023-08-10
Add CellAssign assignments to SCE object

### LINKED ISSUE #394: Incorporate CellAssign predictions into annotated SCE object

allyhawkins:
Closes #394 

This PR adds in a process to incorporate the predictions output from running `CellAssign` to the annotated SCE object. The input here is a SCE file, the predictions file, and the name of the reference used with `CellAssign`. This is set up so that the SCE object produced by the `SingleR` process is the input of this process. The final annotated object then contains the cell type assignments from both `SingleR` and `CellAssign`. I mirrored how we named both the assignments and the metadata from `SingleR`. 

The only tricky thing here was deciding how to pass the reference name through to this process. In `SingleR` we can add that to the model and store it in the object and then grab it that way. However here, the reference file is just a marker genes tsv file so we can't really store the name there. Instead, I grab it directly from the cell type metadata that we read in. This also means I had to pass it through the initial `classify_cellassign` process even though it isn't actively used there. 

I also updated this only to publish the final SCE object with both annotations. Do we want to continue publishing both? 

Note: This is a draft PR because I'm currently running a test to evaluate this which takes some time to complete. I will update this once that test is complete. I temporarily added a publish step to the predictions step in case the new step I added failed. I want to be able to troubleshoot it easily. Maybe this is an argument for adding in the predictions output to our `checkpoints` folder for any future issues.


allyhawkins:
This now should be ready for a formal review. I did end up changing the predictions publishing step to write to the checkpoints directory. I also added a check that the cell type assignments were getting assigned to the correct original barcodes.

sjspielman:
Major caveat that I haven't looked at the code yet, but I did have an immediate thought about this PR comment - 

> The only tricky thing here was deciding how to pass the reference name through to this process. In SingleR we can add that to the model and store it in the object and then grab it that way. However here, the reference file is just a marker genes tsv file so we can't really store the name there. 

We _could_ actually do the same thing as in `SingleR` with a bit of a modification to the [`generate_cellassign_refs.R`](https://github.com/AlexsLemonade/scpca-nf/blob/2a78a01e105745c625885335e9b76db0bd57bd79/bin/generate_cellassign_refs.R) script. Rather than exporting a TSV, we could export an RDS file which contains a list with two named items:
- the reference name
- the data frame (formerly the whole TSV)

This might help to simplify some of the nextflow code, but it would also involve some backtracking and I don't want that to be too tricky/time-consuming!

allyhawkins:
> We could actually do the same thing as in SingleR with a bit of a modification to the [generate_cellassign_refs.R](https://github.com/AlexsLemonade/scpca-nf/blob/2a78a01e105745c625885335e9b76db0bd57bd79/bin/generate_cellassign_refs.R) script. Rather than exporting a TSV, we could export an RDS file which contains a list with two named items:
> 
> the reference name
> the data frame (formerly the whole TSV)
> This might help to simplify some of the nextflow code, but it would also involve some backtracking and I don't want that to be too tricky/time-consuming!

We actually can't do that because the predictions file is output by python, so I don't think we can create an rds file.

sjspielman:
> We actually can't do that because the predictions file is output by python, so I don't think we can create an rds file.

argh, yes, this is a python library...oh well!

bin/classify_cellassign.R
```diff
@@ -0,0 +1,67 @@
...
+# read in input files
+sce <- readr::read_rds(opt$input_sce_file)
+predictions <- readr::read_tsv(opt$cellassign_predictions)
+
+celltype_assignments <- predictions |>
+  tidyr::pivot_longer(!barcode,
+                      names_to = "celltype",
+                      values_to = "prediction") |>
+  dplyr::group_by(barcode) |>
+  dplyr::slice_max(prediction, n = 1) |>
```
sjspielman, line 55: oh this is neat! I was only aware of its less exciting relative `dplyr::slice()`

modules/classify-celltypes.nf
```diff
@@ -60,7 +81,8 @@ workflow annotate_celltypes {
         .map{[
           project_id = it.scpca_project_id,
           singler_model_file = "${params.singler_models_dir}/${it.singler_ref_file}",
-          cellassign_ref_file = "${params.cellassign_ref_dir}/${it.cellassign_ref_file}"
+          cellassign_ref_file = "${params.cellassign_ref_dir}/${it.cellassign_ref_file}",
+          cellassign_ref_name = it.cellassign_ref_name
```
sjspielman, line 85: I'd add a really quick comment here for future us, that it is indeed correct that the `cellassign_ref_name` needs to be passed in but not the `singler_ref_name`.

modules/classify-celltypes.nf
```diff
@@ -72,24 +94,33 @@ workflow annotate_celltypes {
 
       // creates [meta, processed, SingleR reference model]
```
sjspielman, line 95: This comment needs to be updated with the cellassign parts

modules/classify-celltypes.nf
```diff
@@ -72,24 +94,33 @@ workflow annotate_celltypes {
...
-        .map{meta, processed_rds, processed_hdf5, singler_model_file, cellassign_ref_file -> tuple(meta,
-                                                                                                  processed_rds,
-                                                                                                  singler_model_file
-                                                                                                  )}
+        .map{meta, processed_rds, processed_hdf5, singler_model_file, cellassign_ref_file, cellassign_ref_name -> tuple(meta,
+                                                                                                                        processed_rds,
+                                                                                                                        singler_model_file
+                                                                                                                        )}
 
+      // creates [meta, processed hdf5, cellassign ref file, cell assign ref name]
```
sjspielman, line 102:
```suggestion
      // creates [meta, processed rds, processed hdf5, cellassign ref file, cell assign ref name]
```

modules/classify-celltypes.nf
```diff
@@ -72,24 +94,33 @@ workflow annotate_celltypes {
...
-    emit:
-      classify_singleR.out
-      predict_cellassign.out
+      // add cellassign annotations to the object with singleR results
+      all_celltype_assignments_ch = classify_singleR.out
+        // combines using meta from both singleR and cellassign, they should be the same
+        // resulting tuple should be [meta, singleR annotated rds, cellassign predictions]
+        .combine(predict_cellassign.out, by: 0)
+
+      classify_cellassign(all_celltype_assignments_ch)
```
sjspielman, line 122: I'd add a comment here, something like this to mirror `classify_singler` comment:

```suggestion
      // get CellAssign cell type predictions and add them to SCE
      classify_cellassign(all_celltype_assignments_ch)
```

modules/classify-celltypes.nf
```diff
@@ -3,7 +3,6 @@ process classify_singleR {
     container params.SCPCATOOLS_CONTAINER
     label 'mem_8'
     label 'cpus_4'
-    publishDir "${params.results_dir}/${meta.project_id}/${meta.sample_id}", mode: 'copy'
```
sjspielman, line 6: For my understanding... This deletion is indeed intentional, since at this point `classify_cellassign` will do the final writing to publishDir since that process is taking as input the SCE that SingleR annotations are already saved in. So, only `classify_cellassign` needs the publishDir, right?

bin/classify_cellassign.R
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env Rscript
+
+# This script is used to read in the predictions from CellAssign and assign cell types
```
sjspielman, line 3:
```suggestion
# This script is used to read in the predictions from CellAssign and assign cell types in the annotated RDS file
```

bin/classify_cellassign.R
```diff
@@ -0,0 +1,67 @@
...
+
+# check that input file file exists
+if(!file.exists(opt$input_sce_file)){
+  stop("Missing input SCE file")
+}
+
+if(!file.exists(opt$cellassign_predictions)){
+  stop("Missing CellAssign predictions file")
+}
+
```
sjspielman, line 45: Let's add a few more quick checks. Feel free to rearrange these checks as needed!
FYI, we don't seem to have the "output file ends in rds" check in `classify_SingleR.R`, if you want to add that check in this PR go for it.

(please check my syntax here too!)

```suggestion
# check that reference_name was provided
if (is.null(opt$reference_name)) {
  stop("Missing reference name")
}

# check that output file ends in rds
if(!(stringr::str_ends(opt$output_sce_file, ".rds"))){
  stop("output sce file name must end in .rds")
}
```

bin/classify_cellassign.R
```diff
@@ -0,0 +1,67 @@
...
+)
+
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# check that input file file exists
+if(!file.exists(opt$input_sce_file)){
+  stop("Missing input SCE file")
+}
+
+if(!file.exists(opt$cellassign_predictions)){
```
sjspielman, line 42:
```suggestion
# check that cellassign predictions file was provided
if(!file.exists(opt$cellassign_predictions)){
```

bin/classify_cellassign.R
```diff
@@ -0,0 +1,67 @@
...
+  make_option(
+    opt_str = c("--reference_name"),
+    type = "character",
+    help = "name referring to the marker gene reference used for CellAssign"
+  )
+)
+
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# check that input file file exists
```
sjspielman, line 37:
```suggestion
# check that input file exists
```

bin/classify_cellassign.R
```diff
@@ -0,0 +1,67 @@
...
+# join by barcode to make sure assignments are in the right order
+celltype_assignments <- data.frame(barcode = sce$barcodes) |>
+  dplyr::left_join(celltype_assignments, by = "barcode")
+
+# add into sce
+sce$cellassign_celltype_annotation <- celltype_assignments$celltype
+sce$cellassign_max_prediction <- celltype_assignments$prediction
+
+metadata(sce)$cellassign_predictions <- predictions
+metadata(sce)$cellassign_reference <- opt$reference_name
```
sjspielman, line 67: Missing lines to actually save to RDS below this?

modules/classify-celltypes.nf
```diff
@@ -28,16 +27,17 @@ process classify_singleR {
 
 process predict_cellassign {
   container params.SCPCATOOLS_CONTAINER
+  publishDir "${params.checkpoints_dir}/celltype/${meta.library_id}", mode: 'copy'
```
sjspielman, line 30: 👍 agreed to send to checkpoint!

sjspielman:
> I also updated this only to publish the final SCE object with both annotations. Do we want to continue publishing both?

This is good with me! We might want to make a note of this somewhere to make sure it gets into docs?

bin/classify_cellassign.R
allyhawkins: Yikes... good catch!

modules/classify-celltypes.nf
allyhawkins: Yes this is intentional! We only need the SCE object with both annotations to be published which is the output of `classify_cellassign`.

sjspielman: makes sense, thanks!

allyhawkins:
Thanks for the review @sjspielman! I addressed all of your comments including adding in the output file check for classify singleR. I think I caught all the comment updates but let me know if I missed anything.

bin/classify_cellassign.R
```diff
@@ -0,0 +1,80 @@
...
+    opt_str = c("--reference_name"),
+    type = "character",
+    help = "name referring to the marker gene reference used for CellAssign"
+  )
+)
+
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# check that input file exists
+if(!file.exists(opt$input_sce_file)){
```
sjspielman, line 38:
```suggestion
if (!file.exists(opt$input_sce_file)){
```

bin/classify_cellassign.R
```diff
@@ -0,0 +1,80 @@
...
+  stop("Missing input SCE file")
+}
+
+# check that cellassign predictions file was provided
+if(!file.exists(opt$cellassign_predictions)){
```
sjspielman, line 43:
```suggestion
if (!file.exists(opt$cellassign_predictions)){
```

bin/classify_SingleR.R
```diff
@@ -56,6 +56,11 @@ if(!file.exists(opt$input_sce_file)){
   stop("Missing input SCE file")
 }
 
+# check that output file ends in rds
+if(!(stringr::str_ends(opt$output_sce_file, ".rds"))){
```
sjspielman, line 60:
```suggestion
if (!(stringr::str_ends(opt$output_sce_file, ".rds"))){
```

sjspielman: Suggested spaces here and for other if's, just for tidyverse style

sjspielman (commented):
All looks good to me! I suggested a couple spaces for style FYI.

I just started a run through of this workflow, and I want to see that it finished without error before approving. So I'll come back later today and approve once the cell typing has finished! That said, if you ran the workflow at most recent commit, I can cancel this run and approve now.

allyhawkins:
> I just started a run through of this workflow, and I want to see that it finished without error before approving. So I'll come back later today and approve once the cell typing has finished! That said, if you ran the workflow at most recent commit, I can cancel this run and approve now.

Sorry for the delay, but yes I ran it with the most recent changes, so we should be good to go!

sjspielman (approved):

PR #405, opened 2023-08-15
Add process to cluster cells

### LINKED ISSUE #404: Add process for obtaining cluster assignments
We would like to provide cluster assignments in the SCE objects output from the workflow. These will ultimately be compared to the cell type annotations and shown in the QC report, but we will first need to add them to the objects. Some things we will need to figure out: 

I think we want to do this within the main workflow prior to cell typing. For now that means adding a step after the normalization and dim reduction step that adds in clusters. They would then be returned in the processed SCE object which is currently used as input to the cell type workflow.

sjspielman:
Closes #404 

This PR adds a process to cluster SCEs to the main workflow. This has been successfully tested with `SCPCR000001`. Some notes on this implementation:

- Major caveat about the clustering code _itself_ given some of the discussion going on in #404 - we probably will want to circle back to the exact parameters used, but for now it will do graph-based clustering of some kind at least. (edit - worth noting I don't have `type = "jaccard"` in here, yet?)
- I have ended up using the following variable names, which can be tweaked!
  - Clusters are added to `colData(sce)$clusters`
  - Two metadata variables are added (this may change depending on the final clustering code): `metadata(sce)$cluster_algorithm` and `metadata(sce)$cluster_nn`. 
- While writing the clustering script, I nearly used but ended up not using the new `cluster_sce` function we added to `scpcaTools`, since the arguments for that function might be a tad involved for what we want to do here! That function seemed most suited for use by the integration metrics which it was written for, but we can use it here if that's really important just with some nextflow string wrangling.
- To add clustering to the workflow, I wrote a separate small module with one process to perform clustering. As of now, I have _this process_ sending final RDS files to `publishDir`, and I updated the `post_process_sce` process to instead save these files to a checkpoint directory. As a consequence, the processed sce files will now all have clusters in them. I did this because I didn't think a 4th SCE file with clusters made sense from a memory/storage perspective, but it is kind of a choice that all our processed SCEs now contain clusters. What feelings do we have?

### DIFF
--- a/bin/cluster_sce.R
+++ b/bin/cluster_sce.R
@@ -16,8 +16,7 @@ option_list <- list(
   make_option(
     opt_str = c("-i", "--processed_sce_file"),
     type = "character",
-    default = "~/sce.rds",
-    help = "Path to RDS file that contains the processedsce SCE object to cluster."
+    help = "Path to RDS file that contains the processed sce SCE object to cluster."
   ),
   make_option(
     opt_str = c("--pca_name"),
--- a/bin/cluster_sce.R
+++ b/bin/cluster_sce.R
@@ -28,9 +28,9 @@ option_list <- list(
   make_option(
     opt_str = c("--cluster_algorithm"),
     type = "character",
-    default = "leiden",
+    default = "louvain",
     help = "Clustering algorithm to use. Must be one of the options available in bluster.
-      Default is 'leiden'."
+      Default is 'louvain'."
   ),
   make_option(
     opt_str = c("--nearest_neighbors"),
@@ -52,10 +52,10 @@ opt <- parse_args(OptionParser(option_list = option_list))
 set.seed(opt$seed)
 
 # check and read in SCE file
-if (!file.exists(opt$processed_sce_file)) {
-  stop("Input `sce_file` is missing.")
+if (!file.exists(opt$input_sce_file)) {
+  stop("Input `input_sce_file` is missing.")
 }
-sce <- readr::read_rds(opt$processed_sce_file)
+sce <- readr::read_rds(opt$input_sce_file)
 
 
 # check pca_name is present
@@ -73,6 +73,7 @@ clusters <- bluster::clusterRows(
   pca_matrix,
   bluster::NNGraphParam(
     k = opt$nearest_neighbors,
+    type = "jaccard",
     cluster.fun = opt$cluster_algorithm
   )
 ) 
@@ -84,4 +85,5 @@ metadata(sce)$cluster_algorithm <- opt$cluster_algorithm
 metadata(sce)$cluster_nn <- opt$nearest_neighbors
 
 # export -------------------
-readr::write_rds(sce, opt$processed_sce_file)
+# we are overwriting the input file here:
+readr::write_rds(sce, opt$input_sce_file)
--- a/bin/cluster_sce.R
+++ b/bin/cluster_sce.R
@@ -2,7 +2,7 @@
 
 # Script used to perform clustering on a given SCE object
 # 
-# This script reads in an RDS file containing an SCE, and performed
+# This script reads in an RDS file containing an SCE, and performs
 #  graph-based clustering using the specified algorithm. Cluster identities are 
 #  stored in the SCE's colData slot, and the SCE is written back out to the
 #  original RDS file.
@@ -14,7 +14,7 @@ library(SingleCellExperiment)
 # Set up optparse options
 option_list <- list(
   make_option(
-    opt_str = c("-i", "--processed_sce_file"),
+    opt_str = c("-i", "--input_sce_file"),
     type = "character",
     help = "Path to RDS file that contains the processed sce SCE object to cluster."
   ),
--- a/bin/cluster_sce.R
+++ b/bin/cluster_sce.R
@@ -1,0 +1,87 @@
+#!/usr/bin/env Rscript
+
+# Script used to perform clustering on a given SCE object
+# 
+# This script reads in an RDS file containing an SCE, and performed
+#  graph-based clustering using the specified algorithm. Cluster identities are 
+#  stored in the SCE's colData slot, and the SCE is written back out to the
+#  original RDS file.
+
+# import libraries
+library(optparse)
+library(SingleCellExperiment)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-i", "--sce_file"),
+    type = "character",
+    help = "Path to RDS file that contains the SCE object to cluster."
+  ),
+  make_option(
+    opt_str = c("--pca_name"),
+    type = "character",
+    default = "PCA",
+    help = "Name of the PCA reduced dimensionality representation to perform
+      clustering on."
+  ), 
+  make_option(
+    opt_str = c("--cluster_algorithm"),
+    type = "character",
+    default = "leiden",
+    help = "Clustering algorithm to use. Must be one of the options available in bluster.
+      Default is 'leiden'."
+  ),
+  make_option(
+    opt_str = c("--nn"),
+    type = "integer",
+    default = 15, 
+    help = "Nearest neighbors parameter to set for graph-based clustering."
+  ),
+  make_option(
+    opt_str = c("--seed"),
+    type = "integer",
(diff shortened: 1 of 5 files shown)


sjspielman:
Just want to note this comment from @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/404#issuecomment-1679273937

> All that said, it may be that going with `type="jaccard"` and `cluster.fun="louvain"` is the right move for now to match `seurat`... The extra parameters/bad defaults for Leiden clustering may need some time to work out.

bin/cluster_sce.R
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env Rscript
+
+# Script used to perform clustering on a given SCE object
+# 
+# This script reads in an RDS file containing an SCE, and performed
```
allyhawkins, line 5:
```suggestion
# This script reads in an RDS file containing an SCE, and performs
```

bin/cluster_sce.R
```diff
@@ -0,0 +1,87 @@
...
+#  original RDS file.
+
+# import libraries
+library(optparse)
+library(SingleCellExperiment)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-i", "--processed_sce_file"),
```
allyhawkins, line 17:
```suggestion
    opt_str = c("-i", "--input_sce_file"),
```
I think we want to be consistent with these scripts on how we name the input files. We've been using this pretty consistently.

bin/cluster_sce.R
```diff
@@ -0,0 +1,87 @@
...
+
+# extract the principal components matrix
+pca_matrix <- reducedDim(sce, opt$pca_name)
+
+# cluster with specified algorithm
+clusters <- bluster::clusterRows(
+  pca_matrix,
+  bluster::NNGraphParam(
+    k = opt$nearest_neighbors,
+    cluster.fun = opt$cluster_algorithm
```
allyhawkins, line 76: I think we do want to input `jaccard` here? In downstream analyses we had been using louvain and jaccard together.

modules/cluster-sce.nf
```diff
@@ -0,0 +1,27 @@
...
+    tag "${meta.library_id}"
+    publishDir "${params.results_dir}/${meta.project_id}/${meta.sample_id}", mode: 'copy'
+    input:
+        tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(processed_rds)
+    output:
+        tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(processed_rds)
+    script:
+        """
+        cluster_sce.R \
+          --processed_sce_file ${processed_rds} \
```
allyhawkins, line 17:
```suggestion
          --input_sce_file ${processed_rds} \
```
Just going along with my previous suggestion.

modules/sce-processing.nf
```diff
@@ -167,7 +167,7 @@ process post_process_sce{
     container params.SCPCATOOLS_CONTAINER
     label 'mem_8'
     tag "${meta.library_id}"
-    publishDir "${params.results_dir}/${meta.project_id}/${meta.sample_id}", mode: 'copy'
+    publishDir "${params.checkpoints_dir}/post_processed/${meta.sample_id}", mode: 'copy'
```
allyhawkins, line 170: I don't think we need to publish this to checkpoints. I think we could just get rid of publishing it?

allyhawkins (commented):
Thanks for doing this! This mostly looks good I just had a few small suggestions. 

Also, I agree about not having a fourth file and adding the clusters to the `processed.rds`. That makes the most sense to me at least.

modules/sce-processing.nf
sjspielman: Honestly my main motivation for saving these to checkpoints was to save time debugging clustering and QC report, so I have no real serious argument getting rid of this checkpoint!

bin/cluster_sce.R
sjspielman: Yeah, I'm going to put in louvain+jaccard here, with louvain as default, in the end.

sjspielman: done in f6711f89e6f9c3248ebd715ef8a0db8a49c8a743

modules/sce-processing.nf
sjspielman: Done in abb1ac0

sjspielman:
I've responded to reviews, and re-ran the workflow successfully with `SCPCR000001`. How are we looking now?

sjspielman:
Actually, one more thing to look at - I'm realizing I put 15 for the default nearest neighbors but did not explain why! `bluster`'s default is 10, and [`Seurat`'s default appears to be 20](https://github.com/satijalab/seurat/blob/763259d05991d40721dee99c9919ec6d4491d15e/R/clustering.R#L518). Hence, 15 - this is not _super_ principled, to be fair! 
Let me know if you have any thoughts about this default!

bin/cluster_sce.R
```diff
@@ -0,0 +1,89 @@
...
+#  original RDS file.
+
+# import libraries
+library(optparse)
+library(SingleCellExperiment)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-i", "--input_sce_file"),
```
allyhawkins, line 17: Whomp... I went back to look at something else and I realized I was wrong and was confusing the integration/ cell type workflow with the post processing in this workflow. It looks like we used a combination of `sce_file` and `filtered_sce_file`. So feel free to use `processed_sce_file` if you want. Sorry about that!

bin/cluster_sce.R
```diff
@@ -0,0 +1,89 @@
...
+
+# extract the principal components matrix
+pca_matrix <- reducedDim(sce, opt$pca_name)
+
+# cluster with specified algorithm
+clusters <- bluster::clusterRows(
+  pca_matrix,
+  bluster::NNGraphParam(
+    k = opt$nearest_neighbors,
+    type = "jaccard",
```
allyhawkins, line 76: Should this be an argument? Basically if we switch to `Leiden`, we are still using `jaccard` right? If that's the case then this is fine, but if we think we might change it we should make it an argument.

bin/cluster_sce.R
```diff
@@ -0,0 +1,89 @@
...
+    cluster.fun = opt$cluster_algorithm
+  )
+) 
+# TODO: may wish to modify additional cluster parameters depending on cluster algorithm
```
allyhawkins, line 80: Note to remove this line.

bin/cluster_sce.R
sjspielman: No worries :) 
I think I will revert to `processed_sce_file`, specifically since it's _both_ input and output here.

bin/cluster_sce.R
sjspielman: If we switch to `leiden` I think the answer is _nobody actually knows_, so agree that jaccard should be an argument + param.

jashapiro (commented):
Meant to send the below as a single comment, but apparently clicked the wrong button.

bin/cluster_sce.R
```diff
@@ -0,0 +1,87 @@
...
+  stop("Provided `pca_name` is not present in the SCE object.")
+}
+
+# Perform clustering ----------------
+
+# extract the principal components matrix
+pca_matrix <- reducedDim(sce, opt$pca_name)
+
+# cluster with specified algorithm
+clusters <- bluster::clusterRows(
```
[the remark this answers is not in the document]
sjspielman: Yeah, I was wondering about this actually - I'm glad you commented! I hadn't seen this function before in our code, so I wasn't sure if there was some older conversation I wasn't around for where it was decided to not use this. But, I do appreciate its simplicity here and since you suggested it, I'm going for it!

sjspielman:
@allyhawkins I think this is about ready to go!

bin/cluster_sce.R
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/405/commits/55327f0acfd1a670ca1f3f4606bec01ed259c199, and tested workflow ✅

allyhawkins (approved):
Looks good!

PR #421, opened 2023-08-29
Only save CellAssign celltype column to celltype column

sjspielman:
Closes #420 

Pretty much does what it says! I'm testing this now, and will request review once I have confirmed it works as expected.

PR #422, opened 2023-08-29
Initiate cell type report with summary tables

### LINKED ISSUE #408: Initiate cell type QC report and add summary tables
We need to add a cell type specific section to the current QC report. We should create a separate child report that is added if the object contains cell types. This is similar to how we handle the CITE-seq section of the QC report. 

As outlined in #276, the first section will contain summary tables. There should be one summary table per cell type annotation method that was performed. Each table should include two columns, `cell type` and `total cells`. We should also create a table for submitter annotations if they are available.

sjspielman:
Closes #408

This PR initiates the cell typing report with:

- A brief introduction, with links, to the methods used
- Some tables!
  - For this, I wrote a function that can be used for singler, cellassign, and any submitter annotations which will create a table. I copied `kableExtra` formatting from one of the cite-seq tables to start, but this formatting seemed fine to keep so I didn't tweak it further

The celltyping report is currently displayed as the last child report item, which made sense to me since it feels more "downstream analysis"-y compared to other sections, but let me know what you think of this choice!
I also left some "TODO placeholders" for where we'll eventually need to head with submitter annotations. There's also one TODO question in there that's part of this PR - when I define `has_celltypes`, I rely on the presence of _all four_ celltyping columns - any thoughts about this check are welcome!

I tested this out with `SCPCL000295` (where I manually fixed the celltype column since #421 is still running..), and the report is attached: 
[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12467296/qc_report.html.zip)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -0,0 +1,92 @@
+# Celltype annotation results
```
allyhawkins, line 1:
```suggestion
# Celltype Annotation Summary
```
Keeping with the naming scheme we have in other sections `RNA-seq Experiment Summary`, I think we want to capitalize and also use the word `Summary`.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -0,0 +1,92 @@
...
+The [`BlueprintEncode` reference](https://rdrr.io/github/LTLA/celldex/man/BlueprintEncodeData.html) from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) was used as the reference dataset.
+This reference contains stroma and immune cells identified by Blueprint and Encode.
+* [`CellAssign`](https://github.com/Irrationone/cellassign), which uses a marker-gene based approach.
+Marker genes for celltyping were obtained from [PanglaoDB](https://panglaodb.se/).
+
+Cells annotated as "Unknown cell type" represent cells which could not be confidently identified using the given annotation method. 
+
+If available, these results will be compared to celltype annotations provided by the data's originating research group ("submitter-provided celltype annotations".)
+
+## Celltype annotation summary
```
allyhawkins, line 16:
```suggestion
## Celltype Annotation Statistics
```
Same thing here, just mirroring the naming for the other sections that have the summary tables.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -0,0 +1,92 @@
...
+      # Add percentage column
+      `Percent of cells` =  paste0(round(n/sum(n) * 100, digits = 2), "%"),
+      # set column order in descending order of n, but with unknown at the end
+      `Annotated celltype` = forcats::fct_reorder(`Annotated celltype`, n, .desc=TRUE),
+      `Annotated celltype` = forcats::fct_relevel(`Annotated celltype`, 
+                                                  "Unknown cell type", 
+                                                   after = Inf)
+    ) |>
+    # arrange on `Annotated celltype`
+    dplyr::arrange(`Annotated celltype`) |>
```
allyhawkins, line 58: Aren't you repeating what you already did before when using `forcats::fct_reorder`? I think you could delete the `arrange`, but I haven't tested that.

templates/qc_report/qc_report.rmd
```diff
@@ -101,12 +101,25 @@ if (has_cellhash) {
   modalities <- c(modalities, "Multiplex")
 }
 
-# check for umap, need to be sure that processed_sce exists first
+# check for umap and celltypes, but need to be sure that processed_sce exists first
+# for celltypes, we expect these four columns. 
+# TODO: do we need to check for all 4?
```
allyhawkins, line 106: Another thought here is to separately check for cell assign and singleR cell types? What if for some reason we don't have both (maybe we can't find an appropriate reference to use in both)?

I know we do add some metadata related to cell typing to the annotated objects, but I wonder if we should be including a `has_singler` and a `has_cellassign` directly to the metadata when performing classification? This would save us having to search for columns that we have named elsewhere in the workflow?

templates/qc_report/qc_report.rmd
```diff
@@ -101,12 +101,25 @@ if (has_cellhash) {
...
+  has_celltypes <- all(celltype_cols %in% names(colData(processed_sce)))
+
 } else {
   has_umap <- FALSE
+  has_celltypes <- FALSE
 }
+
+# Note: later, we will also check for submitter-provided annotations
+# for now, we will say this is false
+has_submitter_celltypes <- FALSE
```
allyhawkins, line 122: I think this is fine for now. We also could remove it and come back and incorporate it later, since we haven't yet figured out what this is going to be named, etc.

allyhawkins (commented):
Thanks for getting this started! This mostly looks good and I just left a few comments. I think the biggest thing is deciding how we want to check for the presence of cell types. I think I favor the idea of adding a `has_singler` and `has_cellassign` to the metadata of the SCE object when performing cell type classification, rather than checking for columns being present.

sjspielman:
>  I think I favor the idea of adding a `has_singler` and `has_cellassign` to the metadata of the SCE object when performing cell type classification, rather than checking for columns being present.

I like this idea! Unless you object, I'll get that added in this PR and not separately since it's super quick; will add to the `classify_<method>.R` scripts.

templates/qc_report/celltypes_qc.rmd
sjspielman: Oh yeah, the `arrange` can :100: go.

sjspielman:
Since last review, I've made the following changes.

- Delete that extra `arrange`
- Use new metadata variables `has_singler` and `has_cellassign`, which will only exist if cell typing was performed.
  - Note that these fields simply won't exist otherwise - they won't even be `FALSE`. I am really on the fence for whether we want to add `FALSE` values for these in the main workflow..?
  - For running the QC report here, I directly modified the `processed` object to have these metadata fields since I didn't run the workflow through (CellAssign takes sooo looonnngggg)
- I still have a `has_celltypes` variable derived from those two variables, which is used for logic to show/hide the child QC report
- I added `knitr::asis_output()` for handling conditional text now that we have the option to show either/both of singler or cellassign
- `Celltype` -> `cell type`, as noted in https://github.com/AlexsLemonade/scpca-nf/issues/425#issuecomment-1699281935

Here's the current report: [qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12476925/qc_report.html.zip)

templates/qc_report/celltypes_qc.rmd
```diff
@@ -0,0 +1,108 @@
...
+    singler_celltype_annotation = ifelse(
+      is.na(singler_celltype_annotation),
+      "Unknown cell type", 
+      singler_celltype_annotation
+    ),
+     cellassign_celltype_annotation = ifelse(
+      cellassign_celltype_annotation == "other",
+      "Unknown cell type", 
+      cellassign_celltype_annotation
+    )
```
allyhawkins, line 55: Should these mutation statements be wrapped in `if` statements to check that `has_singler` or `has_cellassign` is true?

allyhawkins (commented):
So I have another idea... sorry to change it up on you, but I think we should change the metadata to explicitly store the cell typing methods that were used. And then if that isn't present it's because cell typing was not performed. I think this is slightly more intuitive.

sjspielman:
> So I have another idea... sorry to change it up on you but I think we should change the metadata to explicitly store the cell typing methods that were used. 

Yeah! I like this more too & addresses weirdness of presence of variable as indicator! Good call.

bin/classify_SingleR.R
```diff
@@ -115,7 +115,8 @@ if (opt$label_name == "label.ont") {
 metadata(sce)$singler_results <- singler_results
 metadata(sce)$singler_reference <- singler_model$reference_name
 metadata(sce)$singler_reference_label <- singler_model$reference_label
-
+# add indicator variable that singler predictions are present
+metadata(sce)$has_singler <- TRUE
```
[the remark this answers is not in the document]
sjspielman: Noting I did this in d09acee but slightly simplified - turns out if the field doesn't exist, it will get created on the fly because sometimes R is friendly about `NULL` values(!) For example:

```
> x <- list()
> x$new_field <- c(x$new_field, "new value")
> x
$new_field
[1] "new value"
```

sjspielman:
Moving right along! I've incorporated the most recent reviews, and the report still compiles as expected.

I also tested if this works as expected when just "singler" is present in `metadata(processed_sce)$celltype_methods`, and indeed 🎉 !

templates/qc_report/qc_report.rmd
```diff
@@ -101,12 +101,26 @@ if (has_cellhash) {
...
 if (has_processed) {
   has_umap <- "UMAP" %in% reducedDimNames(processed_sce)
+  
+  # these metadata fields are only present if celltyping has been performed:
+  has_singler <- "has_singler" %in% names(metadata(processed_sce))
+  has_cellassign <- "has_cellassign" %in% names(metadata(processed_sce))
+  if (has_singler | has_cellassign) {
+    # use overall variable here for QC report logic
+    has_celltypes <- TRUE 
+  }
```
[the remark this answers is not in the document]
sjspielman: done all around in https://github.com/AlexsLemonade/scpca-nf/pull/422/commits/d09acee2b125540632fa6bf6c88ed3db6d2a309c and https://github.com/AlexsLemonade/scpca-nf/pull/422/commits/6824e7aa6dec65536502cc4c21a8ec7ebaa0097c

allyhawkins (approved):
Looks good 🚀

PR #423, opened 2023-08-29
Add heatmaps to celltype QC report

### LINKED ISSUE #412: Include a heatmap comparing cell type annotations to clustering results in cell type report
The fourth section of the report should include a heatmap that directly compares the cell type annotations with both `SingleR` and `CellAssign` to clustering assignments available in the SCE object. 
The heatmap should look similar to the picture below, but right now, we are not including the aneuploid/ diploid status. 

Screenshot 2023-08-23 at 2 27 44 PM

sjspielman:
**⚠️ Stacked on #422**
Closes #412 

This PR continues the QC report by adding a heatmap section. I played around with formatting a lot here, but then decided it's best to stop and get some feedback! Some choices I made to start us off:

- For now I am not showing hierarchical clustering since it's just so busy, but I know the People love their hierarchical clustering. Instead, I have ordered the cell types in the same order they are shown in tables. What's the best route here?
- Using `pheatmap`, I can't actually label the x-axis as "Clusters" _and_ I can't actually move the legend! It's hardcoded on the right-hand side... hm.
- To avoid overlap with cell labels, I moved these to the left but had to make them pretty small to fit. It wouldn't be as challenging to make them fit if they were on the x-axis, but I really like how the sketch for this showed clusters (which is _shared across plots_) on the x-axis, so the text is just small!
- Currently we're using the default palette

One big question is: Is this the right engine for building the heatmap? Should we straight up use `ComplexHeatmap::Heatmap()` instead of converting a `pheatmap`?

Here's an example report: [qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12468002/qc_report.html.zip)

Edit: ~Note that this SCE did not have clustering performed already, so I just used `sample(1:7, ....)` to get some clusters in there to develop the code, so don't worry about the total lack of relationship between clusters and celltypes!~ Clusters in this SCE are now real!


jaclyn-taroni:
> One big question is: Is this the right engine for building the heatmap? Should we straight up use `ComplexHeatmap::Heatmap()` instead of converting a `pheatmap`?

I'm just popping in to say that, based on the sketches of figures we want to include in the report with all their annotations, I'd expect using `{ComplexHeatmap}` to be the move!

sjspielman:
@allyhawkins this is now "again" ready for its first review! I've changed things around from my initial PR:

- Now use `ComplexHeatmap` so we can more flexibly add annotations in the future
  - Bonus, legend placement flexibility!
- Use a CVD-friendly palette; I am extremely _not_ committed to the palette I chose; I just randomly grabbed a diverging CVD-friendly palette to start, but we can figure out the right one (or at least a better one) in review! I wasn't sure about using viridis since that's already all over the place.
- The heatmap now uses hierarchical clustering for ordering, since that's really the standard that The People expect for heatmaps...

templates/qc_report/celltypes_qc.rmd
```diff
@@ -115,3 +117,72 @@ knitr::asis_output("### Submitter-provided cell type annotations\n")
 #  so this variable name `submitter_celltype_annotation` is just a placeholder
 create_celltype_n_table(celltype_df, submitter_celltype_annotation)
 ```
+
+
+## Cell type annotations and clusters
+
+Below, we show heat maps comparing cell type annotations (along the y-axis) to clustering results (along the x-axis).
```
allyhawkins, line 124: We may want to add a statement that the color indicates the number of cells that are found in both the indicated cell type and cluster. I know that seems somewhat obvious given the legend, but I think it could be helpful to users.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -115,3 +117,72 @@ knitr::asis_output("### Submitter-provided cell type annotations\n")
...
+                                            cluster_vector) {
+  
+  # build a matrix with cluster identity as rows and celltype as columns
+  celltype_cluster_mtx <- table(celltype_vector,
+                                cluster_vector) |> 
+    log1p() # log transform for visualization
+    
+  # Define CVD-friendly palette 
+  # TODO: what palette? 
+  heatmap_palette <- RColorBrewer::brewer.pal(n = 7, name = "BrBG")
```
allyhawkins, line 139: I'm not crazy about this color palette. I know we use viridis everywhere, but it is a nice palette and it would keep the coloring consistent throughout the report. Unless you think it looks bad with viridis?

sjspielman: > I'm not crazy about this color palette. 

The reason for that is, probably, because it is very bad.

> I know we use viridis everywhere, but it is a nice palette and it would keep the coloring consistent throughout the report. Unless you think it looks bad with viridis?

I don't necessarily think it will look bad! I just remember, at one point, I made a plot for cite-seq in viridis and during review it was suggested to get rid of viridis so that it wasn't accidentally confusing that viridis meant different things in different plots. I can't immediately find that comment, but tbh I don't think this is a huge concern for a situation like this - especially where we have a legend! 

I'll try adding viridis and/or one of the fun viridis palettes to see which looks heatmap-iest.

jaclyn-taroni: > I made a plot for cite-seq in viridis and during review it was suggested to get rid of viridis so that it wasn't accidentally confusing that viridis meant different things in different plots.

This is a valid concern based on my experience (i.e., reviewer remarks on my own manuscripts)!

sjspielman:
:art: time! Here's the heatmaps in different flavors of viridis. Any thoughts? I like `inferno` the most I think, but honestly I don't really _dislike_ any!

## plasma
plasma


## magma
magma


## inferno
inferno


## cividis
cividis

sjspielman:
woops, tagging @allyhawkins 👆 :art:

allyhawkins:
Ohh so many pretty heatmaps! I'm torn between `plasma` and `inferno`. I like them both. I think the benefit of `inferno` is that the "low" is black and so I think there's a better contrast between that and those that have overlapping cells, but I just think `plasma` is prettier. But I'll let you pick your favorite! Definitely both better than the original option.

sjspielman:
> I think the benefit of `inferno` is that the "low" is black and so I think there's a better contrast between that and those that have overlapping cells, but I just think `plasma` is prettier.

Ok I actually lol'd here, because when I was writing my comment before I _legit almost wrote_: 
> I like `inferno` the most I think, but honestly I don't really dislike any, except _maybe_ `plasma` is a little blinding.

So, I'm going to stick with inferno 😂 😂

allyhawkins (approved):
looks good!

PR #432, opened 2023-08-31
Add SingleR delta median plot to QC report

### LINKED ISSUE #410: Add plot for SingleR delta median score to the cell type report
The third section of the cell type QC report should summarize the metrics reported for cell type annotation. For `SingleR`, we will calculate the delta median score - or the difference between each cell's top and median scores. Then, we want to create a plot that summarizes that information with delta score on the y-axis and cell type on the x-axis. 

sjspielman:
**Stacked on #427**
Closes #410

This PR adds the delta median plot for `SingleR`. Implementation notes:

- I refactored (bonus this is a pun :tada:) some of the earlier QC report code since we'll want to use the annotation factor order for this plot as well. I removed code that sets the factor order from the celltype tables, and instead added a helper function to accomplish this while setting up `celltypes_df`. This way, all downstream code that uses this data frame inherits these levels!
- I made this overall section generally about assessing cell types, and place it immediately after the cell type tables, and before the section with umaps + heatmaps. This way, we get a sense of reliability before diving into plots.
- I'm using a sina plot here, so I had to set a seed. For now, I just put a seed into `qc_report.rmd`, _but_ it might be preferable to use whatever seed is used for the overall workflow and pass that in as a parameter to the report? I'm not sure how much this _really_ matters though for this situation.
- I wrapped the labels that are over 30 characters, and this seems to really help with plot layout.
- Let me know any feedback about my description of delta median too!

Report: [qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12489544/qc_report.html.zip)


allyhawkins:
Thank you for working on this and apologies if this is really annoying, but... I do wonder if we should be doing something similar to what I did with the [ridge plot for CellAssign](https://github.com/AlexsLemonade/sc-data-integration/pull/231). We would plot just the score and label by the top score and then everything else and look at the separation. I don't know if it would work quite as well because it's a score and not a probability, but I think it's worth a shot.

sjspielman:
> Thank you for working on this and apologies if this is really annoying, but... I do wonder if we should be doing something similar to what I did with the https://github.com/AlexsLemonade/sc-data-integration/pull/231.

Not annoying at all! It's been a fun spending time plotting :) Let's see how it looks..

sjspielman:
Here's a super quick side-by-side (well, stacked) comparison of sina vs ridgeplot, just to get a sense:

One one hand, I do like the ridgeplot more, but on the other hand, I'm not entirely it's usable (this may go for `CellAssign` as well...!) - for any categories that have <= 2 cells, nothing gets drawn and that's just how the algorithm works; >=3 points are needed to estimate the distribution.

I wonder if there's a good middle ground we could achieve here, since I really _do_ like the ridgeplot more... Would it it make sense to show both sina + ridgeplot, and/or only show cells types with >=3 cells for the ridgeplot (we'd add text explaining this). 
Very curious to hear your thoughts!

Screenshot 2023-09-01 at 2 25 23 PM

jaclyn-taroni:
> I do wonder if we should be doing something similar to what I did with the [ridge plot for CellAssign](https://github.com/AlexsLemonade/sc-data-integration/pull/231). We would plot just the score and label by the top score and then everything else and look at the separation. I don't know if it would work quite as well because it's a score and not a probability, but I think it's worth a shot.

My interpretation of this comment was to plot the scores themselves, not the median delta values. Perhaps I got that wrong, but if we're going to plot the _median delta_, it's helpful to use a completely different style of plot IMO so folks know they're looking at something quite different.

sjspielman:
> My interpretation of this comment was to plot the scores themselves, not the median delta values. 

Ah no, I think you are right! Let's see..

allyhawkins:
> > My interpretation of this comment was to plot the scores themselves, not the median delta values.
> 
> Ah no, I think you are right! Let's see..


Yes Jackie is correct. I was thinking we would plot the actual scores themselves and then create a plot similar to the one below. 
Screenshot 2023-09-01 at 2 54 59 PM

sjspielman:
I think we want to be mindful of overly-discussing strategies in this PR, mostly because as comments build up things will be become harder to track & review. So, I'm going to open an issue (will come back to this comment and edit w/ a link!) that we can use to discuss visualization strategies, and then we can come back here to continue the PR.

sjspielman:
As discussed in #434, I've updated this to still visualize `delta_median`, but highlighting points that were pruned out. I've updated text in the plot preamble to match what the plot currently shows. Note that this involved a decent bit of wrangling, since we need to plot based on the full labels, not the pruned labels, in order to color by whether a cell was pruned or not. The points are pretty small and possibly tricky to see, but I think this is inevitable when visualizing this many data points (or more!).

[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12538391/qc_report.html.zip)

Screenshot 2023-09-06 at 10 03 56 AM

sjspielman:
@allyhawkins, I can't re-request review here since only comments were left before, so this is my re-request ping :)

templates/qc_report/celltypes_qc.rmd
```diff
@@ -119,9 +125,95 @@ create_celltype_n_table(celltype_df, submitter_celltype_annotation)
...
+
+# Ensure we have no "Unknown cell type" values left:
+if (sum(delta_median_df$celltype == "Unknown cell type") > 0) {
+  stop("Failed to process SingleR data for diagnostic plot.")
+}
+
+
+# remove "Unknown cell type" from levels, and wrap the labels
+new_levels <- levels(delta_median_df$celltype)
+new_levels <- new_levels[-length(new_levels)]
```
allyhawkins, line 184: I'm a little confused what you are doing here? Do you need both or can you just use the first line without the second line since `Unknown cell type` shouldn't be here?

templates/qc_report/celltypes_qc.rmd
```diff
@@ -119,9 +125,95 @@ create_celltype_n_table(celltype_df, submitter_celltype_annotation)
...
+  scale_color_manual(values = c("red", "black")) +
+  guides(
+    color = guide_legend(override.aes = list(size = 1, alpha = 0.9))
+  ) +
+  theme(
+    axis.text.x = element_text(angle = 55, hjust = 1, size = rel(0.85)), 
+    legend.title = element_text(size = rel(0.75)),
+    legend.text = element_text(size = rel(0.75)),
+    legend.position = "bottom"
+  )
```
allyhawkins, line 214: I think we want to add a median point here too? I'm not sure what color though since `red` is being used for the cells that were pruned.

allyhawkins (commented):
This mostly looks good, I just had one clarifying question and a suggestion about adding a median point.

templates/qc_report/celltypes_qc.rmd
sjspielman: I think blue would probably be fine for median. One question though is how this stat should deal with the current grouping. I feel like it would be best if the median _only_ reflected the black points? Any thoughts?

Also, do you think it would be too busy to also make the red points a different shape, like a diamond or something? It might make them easier to spot?

templates/qc_report/celltypes_qc.rmd
sjspielman: Yeah, it's confusing! I realize it can be simplified too. I will add some comments. Here's what's happening:

- Although there is no longer an `"Unknown cell type"` _value_ in the data, that _level_ still exists in the `delta_median_df$celltype` variable
- This doesn't matter for plotting though! One could proceed to just plot, and the x-axis order would be fine. But, it _does_ matter if I want to wrap the labels, since cell type names are very long.
- So, this code was setting up to wrap the labels while also getting rid of the `Unknown` level.
- Looking again with fresh eyes, we really don't need to get rid of the `Unknown` level though! So, I will simplify to this:

```
# add column with ordered levels with wrapped labels for visualization
delta_median_df$annotation_wrapped <- factor(
  delta_median_df$celltype,
  levels = levels(delta_median_df$celltype),
  labels = stringr::str_wrap(levels(delta_median_df$celltype), 30)
)```

File: templates/qc_report/celltypes_qc.rmd (hunk 2 above)
Comment (allyhawkins) [reply templates/qc_report/celltypes_qc.rmd:214]: > I feel like it would be best if the median only reflected the black points? Any thoughts?

This makes sense to me.

> Also, do you think it would be too busy to also make the red points a different shape, like a diamond or something? It might make them easier to spot?

I don't think I would make them a different color and a different shape, that feels like it might be a lot. I might make the median a different shape or a line though.

**Comment by sjspielman** [comment]
I've updated the plot as discussed and simplified that factor code, so this is ready for another look!
[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12541384/qc_report.html.zip)

One important bit: In https://github.com/AlexsLemonade/scpca-nf/pull/432/commits/601087b36b1ebef14ca5a6b8b02b8834de0c6adb, I made some updates which could be reverted. This commit sets things up if we want to pass in the workflow seed to the QC report, for the sina plot layout. But for this to work, we'd need some small changes over in `scpcaTools::generate_qc_report()`.
If we want to take this route for the seed then, two ways forward:
- make `scpcaTools` compatible, then merge this PR
- revert that commit, hardcode the seed for now in this PR. Later, we could circle back with a new PR to set the seed from the workflow seed, after making `scpcaTools` compatible

File: bin/sce_qc_report.R [hunk 3]
```diff
@@ -92,6 +92,12 @@ option_list <- list(
     type = "character",
     default = "vireo",
     help = "Demultiplexing method to use for multiplexed samples. One of `vireo`, `HTOdemux`, or `HashedDrops`"
+  ),
```
allyhawkins, line 95: I do think we want to keep this argument and pass in the seed from the workflow. So I vote we do make the appropriate scpcaTools modifications.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -44,49 +69,30 @@ celltype_df <- colData(processed_sce) |>
...
-    dplyr::mutate(
-     cellassign_celltype_annotation = ifelse(
-      cellassign_celltype_annotation == "other",
-      "Unknown cell type", 
-      cellassign_celltype_annotation
-    )
-  )
-}
-    
+    prepare_annotation_values(cellassign_celltype_annotation)
```
allyhawkins, line 80: I think this needs to be assigned to a variable?

templates/qc_report/celltypes_qc.rmd
```diff
@@ -119,9 +125,105 @@ create_celltype_n_table(celltype_df, submitter_celltype_annotation)
...
+# Prepare SingleR scores for plot
+
+# extract scores into matrix
+singler_scores <- metadata(processed_sce)$singler_result$scores 
+
+# Create data frame for plotting with delta median and the full *non-pruned* cell labels
+delta_median_df <- tibble::tibble(
+  delta_median = rowMaxs(singler_scores) - rowMedians(singler_scores),
+  # Need to grab the non-pruned label for this plot
+  ontology = metadata(processed_sce)$singler_result$labels,
```
allyhawkins, line 161: is this the ontology id or the ontology id label?

jashapiro:
> I've updated the plot as discussed and simplified that factor code, so this is ready for another look! [qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12541384/qc_report.html.zip)
> 
> One important bit: In [601087b](https://github.com/AlexsLemonade/scpca-nf/commit/601087b36b1ebef14ca5a6b8b02b8834de0c6adb), I made some updates which could be reverted. This commit sets things up if we want to pass in the workflow seed to the QC report, for the sina plot layout. But for this to work, we'd need some small changes over in `scpcaTools::generate_qc_report()`. If we want to take this route for the seed then, two ways forward:
> 
> * make `scpcaTools` compatible, then merge this PR
> * revert that commit, hardcode the seed for now in this PR. Later, we could circle back with a new PR to set the seed from the workflow seed, after making `scpcaTools` compatible

You should be able to use the `extra_params` argument to `scpcaTools::generate_qc_report()` to pass in the seed. It is there just so we don't need to update the function every time we make changes to the template!

allyhawkins (commented):
This looks good to me. I will hold off on approving though until we had in the seed argument in scpcaTools.

bin/sce_qc_report.R
```diff
@@ -224,6 +230,7 @@ scpcaTools::generate_qc_report(
   filtered_sce = filtered_sce,
   processed_sce = processed_sce,
   report_template = opt$report_template,
-  output = opt$qc_report_file
+  output = opt$qc_report_file,
+  seed = opt$seed
```
jashapiro, line 234: Implementing my suggestion in https://github.com/AlexsLemonade/scpca-nf/pull/432#issuecomment-1708955195
```suggestion
  extra_params = list(seed = opt$seed)
```

bin/sce_qc_report.R
allyhawkins: Well never mind, I totally forgot we had the `extra_params` option.

templates/qc_report/celltypes_qc.rmd
sjspielman: These are the labels that were actually assigned which are _ontology ids_. I needed to grab this vector since we don't want the pruned labels for this plot. But, then I need to make sure we don't actually use ontology ids in the plot, but the actually cell names. 

All that said, I realize I need to tweak some things here to make sure this works if, for some reason, ontology ids weren't used for singler annotation..

templates/qc_report/celltypes_qc.rmd
```diff
@@ -32,6 +32,31 @@ If available, these results will be compared to cell type annotations provided b
...
+        # cellassign conditon
+        {{annotation_column}} == "other" ~ "Unknown cell type",
+        # otherwise, keep it
+        TRUE ~ {{annotation_column}}
+        
+      ),
+      {{annotation_column}} := forcats::fct_infreq({{annotation_column}}),
+      {{annotation_column}} := forcats::fct_relevel({{annotation_column}},
+                                                    "Unknown cell type",
+                                                    after = Inf)
```
jashapiro, line 55: Minor style thought... Is there a reason this won't work?
```suggestion
    dplyr::mutate(
      {{annotation_column}} := dplyr::case_when(
        # singler condition
        is.na({{annotation_column}}) ~ "Unknown cell type",
        # cellassign conditon
        {{annotation_column}} == "other" ~ "Unknown cell type",
        # otherwise, keep it
        TRUE ~ {{annotation_column}}
      ) |>
        forcats::fct_infreq() |>
        forcats::fct_relevel("Unknown cell type", after = Inf)
```

sjspielman: Ah I see, you are directly piping the `case_when()` into `forcats()` so the whole thing is in `mutate()`? Should work, will try it out! (but probably not via suggestion)

templates/qc_report/celltypes_qc.rmd
sjspielman: Made some changes to this end in https://github.com/AlexsLemonade/scpca-nf/pull/432/commits/6ba63513e8660998cea7e5e6ab876ef8a42d115e (plus bonus `forcats` cleanup code from @jashapiro)

PR #433, opened 2023-09-01
Heatmap for comparing annotations 

### LINKED ISSUE #431: Celltype QC: Add heatmap to directly compare annotations
> We might want to add a heatmap where we explicitly compare assignments between the two methods? I'm just noticing for this library in particular it appears that the group of monocytes are being annotated as monocytes by both methods. We love to see that! But also, it could help users determine how consistent the methods are with their annotations.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpca-nf/pull/427#pullrequestreview-1605044770_

sjspielman:
**Stacked on #432**
Closes #431 

Another day, another heatmap! 

In this PR, I've added another heatmap to directly compare singler and cellassign, assuming both are present. For this, I had to generalize the heatmap function and update calls to the function accordingly.

For the comparison heatmap itself, I placed it into its own section since it doesn't involve a clusters comparison. But, we could have an overall "heatmaps" section if we wanted?

I also tried text wrapping for long annotations, but they just don't line up well with the actual heatmap. Instead, I opted for small font labels and a large plot area to hopefully fit whatever annotations these methods throw at us! To make wrapped labels line up, we'd probably need to use `ggplot()` with `geom_rect()`, but that would make adding additional annotations down the line pretty challenging, and we wouldn't have the hierarchical clustering. Either way, definitely open to other layout ideas etc!

[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12499241/qc_report.html.zip)


allyhawkins (commented):
Honestly, I really like this and I don't have a lot to add. My only comments is how we are naming sections. Do we think we need `Cell type Annotation` appended to everything? I feel like we could name the subsections without it?

sjspielman:
> Honestly, I really like this and I don't have a lot to add.

:tada: 

> Do we think we need `Cell type Annotation` appended to everything? I feel like we could name the subsections without it?

Will de-wordify a bit.

allyhawkins (approved):
Yay for new section headers 🎉

sjspielman:
FYI I'm planning to wait to merge this until after #432 goes in, since I don't want the "files changes" over in that PR to get overly cluttered with this code.

Issue #436, opened 2023-09-06
Store counts for processed object in raw.X


allyhawkins:
According to the CZI schema, the raw counts should be saved to the `X` layer, unless normalized data is also included. In the event that normalized data is present the raw counts should be moved to `raw.X` and the normalized data stored in `X`. 

Once we complete #429 and have saved the normalized counts to `X`, we will need to go into the AnnData objects and move the `counts` layer to `raw.X`. Unfortunately we cannot do this while converting, so we will need a separate process that reads in the hdf5 file and moves the counts. We probably only want to run this process on the `processed.rds` objects.

jashapiro:
Just thinking this through for a sec... Would we expect to do this processing in R via `reticulate` or by writing two hdf5 files and then a separate `python` script to merge the assays? (or does `zellkonverter` write out both assays? In which case we could probably just leave the default output and move things in python if `logcounts` is present?)

allyhawkins:
`zellkonverter` writes out all assays and stores them to the `layers` slot in `AnnData`. For our objects right now, the `counts` assay is stored as `X` and the other assays are present in `layers`. 

My expectation was to do the following: 
1. For normalized objects, store `logcounts` as `X` when writing out the object. 
2. Have a python script that reads in the processed objects only
3. Move the `counts` from `layers` to `raw.X`
4. Save the modified processed object 

Alternatively, we could write them out so they all look the same and then in a separate process read in only the processed objects and save `logcounts` to `X` and then move `counts` to `raw.X`. 
I was originally trying to save having the secondary process, but now it looks like we will need to do some modifications psot writing the object.

allyhawkins:
Closed by #439

PR #439, opened 2023-09-07
Add process to rearrange counts and logcounts in processed AnnData objects

### LINKED ISSUE #429: Save normalized counts in AnnData to `X`
According to the CZI schema, the raw counts should be saved to the `X` layer, unless normalized data is also included. In the event that normalized data is present the raw counts should be moved to `raw.X` and the normalized data stored in `X`. Right now, we are always storing the raw counts in `X` regardless of the presence of normalized counts. According to our [design discussion notes](https://data-lab-knowledge.slab.com/posts/czi-cellxgene-compatibility-ojhdf9pw) about these changes, we would like to make this change, we should just be sure to document it! 

> For filtered and unfiltered counts before adding in normalized counts, do we want to have counts as X or as raw.X?
> filtered and unfiltered store as X and then switch to the normalized counts being X in the normalized object only 

I think the best way to do this is to add a check for the presence of the `logcounts` assay in the SCE object before converting to AnnData. Then we can assign the X assay by adding a variable to [`scpcaTools::sce_to_anndata()`](https://github.com/AlexsLemonade/scpcaTools/blob/main/R/sce_to_anndata.R) that lets us dictate which assay to save as the `X` layer in the AnnData object. Something like: 

```
if("logcounts" %in% assayNames(sce)){
  x_assay <- "logcounts"
} else {
  x_assay <- "counts"
}
```

allyhawkins:
Closes #429 and #436

This PR adds a new process to the workflow for exporting AnnData to rearrange the processed AnnData objects only. For each processed object, we move the `X` assay (counts) to `raw.X` and then move the `logcounts` layer to `X`. 

- This approach means that every AnnData object is saved in the same way when objects are converted from SCE to AnnData. Then, we read in the processed objects only and check if `logcounts` is present in the `layers`. If that's true, then we rearrange. 
- I decided to take the same approach for both RNA and feature processed objects. They are passed in as a list of files and then we check for `logcounts` in each file before doing any rearranging. 
- Should we add some sort of note in the metadata about counts being moved? I did change the `X_name` though so I think that should be sufficient?


allyhawkins:
This is failing checks right now because it can't find `anndata` even though it should be in the image...

modules/export-anndata.nf
```diff
@@ -44,11 +61,21 @@ workflow sce_to_anndata{
       // export each anndata file
       export_anndata(sce_ch)
 
+      processed_anndata_ch = export_anndata.out
+        .filter{ it[2] == "processed"}
```
jashapiro, line 65: rather than filtering in two places, use `branch` to centralize the logic
```suggestion
     anndata_ch = export_anndata.out
        .branch{
          processed: it[2] == "processed"
          other: true
        }
```

modules/export-anndata.nf
```diff
@@ -44,11 +61,21 @@ workflow sce_to_anndata{
...
+
+      // move any normalized counts to X in AnnData
+      move_normalized_counts(processed_anndata_ch)
+
       // combine all anndata files by library id
       // creates anndata channel with [library_id, unfiltered, filtered, processed]
       anndata_ch = export_anndata.out
```
jashapiro, line 72: I can't make the full suggestion here, but replacing everything up to the `.map`
```suggestion
      anndata_ch = anndata_ch.other.mix(move_normalized_counts.out)
```

modules/export-anndata.nf
```diff
@@ -44,11 +61,21 @@ workflow sce_to_anndata{
...
-          meta.library_id, 
+        // remove any files that were processed and went through reorganization
+        .filter{ it[2] != "processed" }
+        // mix with output from moving counts
+        .mix(move_normalized_counts.out)
+        .map{ meta, hdf5_files, file_type -> tuple(
+          meta.library_id,
           meta,
           hdf5_files
         )}
```
jashapiro, line 81: I can't quite comment on it below, but we might want a better comment on what is going on in the `.map` below (I think it could probably be cleaned up with a more explicit map statement.

jashapiro (commented):
I think the main issue here is a `stub` field not being present... I don't think it uses the docker image for stubs. I'm assuming you ran this with real data and it worked?

bin/move_counts_anndata.py
```diff
@@ -0,0 +1,46 @@
...
+
+args = parser.parse_args()
+
+# separate input files
+input_files = args.input_hdf5_file.split(",")
+
+# compile extension regex
+file_ext = re.compile(r"\.hdf5$|.h5$", re.IGNORECASE)
+
+for file in input_files:
```
jashapiro, line 26: I'm going to suggest simplifying the script to take only one file at a time and then looping through the files in the outer process to save nesting splitting, etc.

(I'd also rename the argument to `--anndata_file` or something like that as this is both input and output)

Nextflow is pretty smart about passing in lists of files, so you can do something like (note that the shell variable has to be escaped)
```
script:
  """
  for file in ${processed_hdf5}; do
    move_counts_anndata.py --anndata_file \${file}
  done
  """
```

allyhawkins:
Thanks for looking at this @jashapiro. I added the stub and also changed the script to only take in one input file. I tested the new changes and everything worked. This is ready for another look.

allyhawkins:
I added formatting to this with `black` and then also added the `schema_version` to the AnnData object before exporting. 
Other than the title, that's the only thing that needs to be added to `uns` so thought this was a good time to add it: https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#uns-dataset-metadata

bin/move_counts_anndata.py
```diff
@@ -0,0 +1,50 @@
...
+        "--input_anndata must end in either .hdf5 or .h5 and contain a processed AnnData object."
+    )
+
+# read in anndata
+object = adata.read_h5ad(args.anndata_file)
+
+# if logcounts is present
+if "logcounts" in object.layers.keys():
+    # move X to raw.X by creating the raw object
+    object.raw = object
```
jashapiro, line 41: Does copying this way end up copying _everything_ to `object.raw`? It seems like that would create a lot of extra data. (i.e. is there now `object.raw.layers["logcounts"]`?) Looking at [the docs](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.raw.html), it seems like no, but I just want to confirm.

jashapiro (approved):
LGTM! I had a little question/fear about how `.raw` is created, but I'm pretty sure the answer is what it should be. It just looks a bit strange which is no fault of yours!

bin/move_counts_anndata.py
allyhawkins: No, it looks like it only stores `X` and `var`.

PR #443, opened 2023-09-08
Update SingleR celltyping

### LINKED ISSUE #441: Confirm workflow is robust to different celltyping label
It has occurred to me that, although code has been written with sufficient flexibility to handle using a different reference label besides `label.ont` for `SingleR` celltyping, this may not have been sufficiently tested so there could be some cryptic bugs in there! Best to catch them now!

This issue tracks running the workflow through with a different label, like `label.main`, and debugging as needed.

sjspielman:
Closes #441 

I have tested the celltyping workflow using `label.main` for `SingleR` (including "model training"), and found 2 quick things that needed to be fixed! Everything is working smoothly now, all tested.

- First, this wasn't a bug, but it seemed like the right move. We previously made a parameter called `label_name` for the singler reference label, but this was before we added cellassign. So, I just changed this workflow parameter to be named `singler_label_name` for clarity. I left all R scripts alone.
- Second, this _was_ a bug! The QC report delta median plot x-axis did not the right levels in an ontology-free scenario, so I added levels and now the plot looks as expected.

Report: [qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12562576/qc_report.html.zip)

### DIFF
--- a/build-celltype-ref.nf
+++ b/build-celltype-ref.nf
@@ -40,7 +40,7 @@ process train_singler_models {
       --ref_file ${ref_file} \
       --output_file ${celltype_model} \
       --fry_tx2gene ${t2g_3col_path} \
-      --label_name ${params.label_name} \
+      --label_name ${params.singler_label_name} \
       --seed ${params.seed} \
       --threads ${task.cpus}
     """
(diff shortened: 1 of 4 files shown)


sjspielman:
I fixed the crossbar-not-showing-IQR situation by changing it to a boxplot without whiskers/outliers! I did this because `crossbar` docs were surprisingly cryptic about what they are even showing by default, but I feel very confident in what a boxplot displays! Here's a quick screenshot of that part of the rendered report:
Screenshot 2023-09-08 at 4 47 19 PM

allyhawkins (approved):
LGTM

PR #449, opened 2023-09-13
Account for failed failed model gene variance and pca

allyhawkins:
In processing some data for release, I was getting an error for one sample at the `model_gene_var` step in post processing. This was because there was only 1 cell after filtering (yikes!). We account for a low number of cells and failing normalization, but don't account for that in downstream steps. I added a `try` statement around `model_gene_var` up until calculating the UMAP. 

I also updated the clustering script here to account for objects that don't have any PCA or reduced dimensions present.

### DIFF
--- a/bin/cluster_sce.R
+++ b/bin/cluster_sce.R
@@ -65,30 +65,35 @@ if (!file.exists(opt$processed_sce_file)) {
 }
 sce <- readr::read_rds(opt$processed_sce_file)
 
+# only perform clustering if reduced dimension embeddings are present
+# otherwise just return the object
+if(length(reducedDimNames(sce)) > 0){
 
-# check pca_name is present
-if (!opt$pca_name %in% reducedDimNames(sce)) {
-  stop("Provided `pca_name` is not present in the SCE object.")
-}
+  # check pca_name is present
+  if (!opt$pca_name %in% reducedDimNames(sce)) {
+    stop("Provided `pca_name` is not present in the SCE object.")
+  }
 
-# Perform clustering ----------------
+  # Perform clustering ----------------
 
-# extract the principal components matrix
-clusters <- scran::clusterCells(
-  sce,
-  use.dimred = opt$pca_name,
-  BLUSPARAM = bluster::NNGraphParam(
-    k = opt$nearest_neighbors,
-    type = opt$cluster_weighting,
-    cluster.fun = opt$cluster_algorithm
+  # extract the principal components matrix
+  clusters <- scran::clusterCells(
+    sce,
+    use.dimred = opt$pca_name,
+    BLUSPARAM = bluster::NNGraphParam(
+      k = opt$nearest_neighbors,
+      type = opt$cluster_weighting,
+      cluster.fun = opt$cluster_algorithm
+    )
   )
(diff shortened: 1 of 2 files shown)


bin/cluster_sce.R
```diff
@@ -65,30 +65,35 @@ if (!file.exists(opt$processed_sce_file)) {
 }
 sce <- readr::read_rds(opt$processed_sce_file)
 
+# only perform clustering if reduced dimension embeddings are present
+# otherwise just return the object
+if(length(reducedDimNames(sce)) > 0){
```
jashapiro, line 71: I think it is probably worth warning here, just for log purposes, so I would flip this around:
```suggestion
if(length(reducedDimNames(sce)) == 0) {
  warning("No reduced dimensions present, skipping clustering")
} else {
```

bin/cluster_sce.R
```diff
@@ -65,30 +65,35 @@ if (!file.exists(opt$processed_sce_file)) {
...
 
-# check pca_name is present
-if (!opt$pca_name %in% reducedDimNames(sce)) {
-  stop("Provided `pca_name` is not present in the SCE object.")
-}
+  # check pca_name is present
+  if (!opt$pca_name %in% reducedDimNames(sce)) {
+    stop("Provided `pca_name` is not present in the SCE object.")
+  }
```
jashapiro, line 75: Do we want to use this check instead for the whole block (replace the length test)? If we are going to "pass" when there are no PCs, I don't see why we wouldn't when a particular PC name is missing.

bin/post_process_sce.R
```diff
@@ -247,7 +248,6 @@ processed_sce <- scater::runPCA(
 )
 
 # calculate a UMAP matrix using the PCA results
-try({
   processed_sce <- scater::runUMAP(processed_sce,
     dimred = "PCA"
```
jashapiro, line 252: Minor formatting annoyance that styler doesn't catch for some reason. (If most arguments are on separate lines, the first argument should be too)
```suggestion
  processed_sce <- scater::runUMAP(
    processed_sce,
    dimred = "PCA"
```

bin/post_process_sce.R
```diff
@@ -229,14 +229,15 @@ if (alt_exp %in% altExpNames(processed_sce)) {
 
 # Perform dimension reduction --------------------
 
-# model gene variance using `scran:modelGeneVar()`
-gene_variance <- scran::modelGeneVar(processed_sce)
+try({
```
jashapiro, line 232: We should probably add a check after this try block to see if there are reduced dims and/or HVGs present. If not, print a warning.

bin/post_process_sce.R
```diff
@@ -247,7 +248,6 @@ processed_sce <- scater::runPCA(
 )
 
 # calculate a UMAP matrix using the PCA results
-try({
```
jashapiro, line 250: I think we want two separate `try` blocks here: one for the HVG & PCA, and one for UMAP. If we are able to make a PCA, we would want that even if the UMAP fails. In the current structure, we would basically have already rolled back the `sce` to not having the PCA stored. At least that is how I think it works.

PR #450, opened 2023-09-14
Don't rename UMAPs

### LINKED ISSUE #448: Fix UMAPs in QC report
While testing workflow changes as part of #407, I was sad to see that the QC report failed! Luckily, an easy fix..

What happened was, we recently bumped `scpcaTools` to use `Bioc 3.17`. Turns out, `scater` changed how UMAPs are named in this release - https://code.bioconductor.org/browse/scater/blob/RELEASE_3_17/R/runUMAP.R#L97 . Now, UMAPs are named `"UMAP1"`, etc.

So, this code is now deprecated - no need to rename at all!
https://github.com/AlexsLemonade/scpca-nf/blob/189af37ebeb63771489a3a18df5e7927ada973e6/templates/qc_report/umap_qc.rmd#L62-L65

sjspielman:
Closes #448 

Short and sweet, no longer need to rename now that we are at Bioconductor `3.17`!


allyhawkins (approved):
🚀

PR #451, opened 2023-09-14
Add submitter celltypes

### LINKED ISSUE #407: Add submitter annotations, if available, to SCE objects
If submitter annotations are available, we want to create a process to add a column for submitter annotations to the annotated SCE objects. My initial thought is that we would create a process that takes as input a single SCE object and a tsv file containing all the cell type annotations for a single project. This file would include cell barcodes, cell type annotations, and library id. That file will be used to grab the annotations for the specified library. The output from the process will be the annotated SCE object with a new column, `submitted_annotations`. 

The one thing I haven't quite figured out is how we want to read the tsv file into the workflow. I think we want to include the path to the file in the library metadata and then filter to libraries that contain a submitted file. Then, only those libraries would run through the process of adding submitter annotations, but I'm open to other ideas.

sjspielman:
Closes #407 

This PR adds submitter cell types to the _unfiltered_ SCE object, if available.

### Approach

- For both `generate_unfiltered_sce` & `generate_merged_unfiltered_sce` , we now explicitly pass in the submitter annotations TSV, if available. Otherwise, we send in `NO_FILE.txt`
  - I'm sad to say that we have no elvis operator, b/c the condition to check for is _existence_ of the file, and then if true we need to pass in the _actual file_. 
- I added a script `add_submitter_annotations.R`. This script is called after `generate_unfiltered_sce.R` in the two relevant processes, but **only** if the file is not `NO_FILE.txt`.
	- This script does what you'd expect, and adds submitter annotations into the unfiltered colData
	- Parallel to other cell type methods, I add `"submitter"` to the  `metadata(sce)$celltype_methods` vector, which is where we store indicators that SingleR and/or CellAssign were used. Submitter annotations are not technically isn't a _method_, but this seemed like the right place to store this information (we could change the name of the vector I suppose?)

### But does it work?

It does! I've tested this out under three conditions:

- `SCPCL000001`: RNA-only, no submitter annotations
- `SCPCL000295`: CITE-seq, no submitter annotations
- `SCPCL000495`: RNA-only, **yes** submitter annotations
- None of the submitter cell types we have are for a CITE-seq library, so I didn't test that condition, but I can make some fake cell types for testing if we want? 
 
For review, I'm only going to formally request @allyhawkins here, but tagging in @jashapiro if you have any other nextflow thoughts!


jashapiro:
Can you provide an example of the updated library file? That is needed for the stub, but also should be done in the examples file.

sjspielman:
> Can you provide an example of the updated library file? That is needed for the stub, but also should be done in the examples file.

Just came here to comment that I'm doing the stub now!

main.nf
```diff
@@ -96,6 +96,7 @@ workflow {
       submitter: it.submitter,
       technology: it.technology,
       seq_unit: it.seq_unit,
+      submitter_cell_types_file: it.submitter_cell_types_file,
```
jashapiro, line 99: A while ago we changed from blank entries to `NA`, and I guess I assumed that `.splitCsv()` would handle that, but it does not, which is why the elvis operator doesn't work. I actually think we should translate NAs here at the start of parsing to avoid this and make for better downstream behavior. Other optional fields (`feature_barcode_file`, etc.) should maybe get the same treatment.


```suggestion
      submitter_cell_types_file: it.submitter_cell_types_file == "NA"? it.submitter_cell_types_file : ""
```

The blank string will then allow us to use `file(it[0].submitter_cell_types_file?: empty_file)` later.

sjspielman: > Other optional fields (feature_barcode_file, etc.) should maybe get the same treatment.

I'll open an issue for this to circle back to, since there's already a lot happening here!

modules/sce-processing.nf
```diff
@@ -200,13 +219,18 @@ workflow generate_sce {
...
+    
+    empty_file = file("${projectDir}/assets/NO_FILE.txt")
+ 
     sce_ch = quant_channel
-      .map{it.toList() + [file(it[0].mito_file), file(it[0].ref_gtf)]}
-
+      .map{it.toList() + [file(it[0].mito_file), 
+                         file(it[0].ref_gtf), 
+                         // either submitter cell type files, or empty file if not available
+                         file(it[0].submitter_cell_types_file).exists() ? file(it[0].submitter_cell_types_file) : empty_file
```
allyhawkins, line 229:
```suggestion
                         file(it[0].submitter_cell_types_file)?: file("NO_FILE.txt")
```
I think you could just do this here and I don't think you need a real path.

sjspielman:
Noting the failure for https://github.com/AlexsLemonade/scpca-nf/pull/451/commits/6aa628281d63fb4176ceaefa7c15395c90ca263f was because of a missing comma, fix incoming once I test the code. Worth noting the stub _did_ pass before that commit!

allyhawkins (commented):
I had a few comments, but another thought I had is related to how we are labeling "unclassified cells" vs "unknown cells". We should standardize this across methods probably, but also we may want to distinguish here between a cell that was not found in the submitter cell types (unclassified) vs. a cell type that was NA or unknown in the submitter file (unknown).

modules/sce-processing.nf
sjspielman: First, a real path is definitely needed. `nextflow` does not appreciate files that aren't files. Also, this strategy is changing a little bit, see - https://github.com/AlexsLemonade/scpca-nf/pull/451#discussion_r1326094681
Testing things out now!

bin/add_submitter_annotations.R
```diff
@@ -0,0 +1,77 @@
...
+  library(optparse)
+  library(SingleCellExperiment)
+})
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = c("-u", "--unfiltered_file"),
+    type = "character",
+    help = "path to output unfiltered rds file. Must end in .rds"
+  ),
```
jashapiro, line 17: It probably doesn't matter much, but I don;t think there is anything in this script that makes it specific to unfiltered SCEs. 

I would make the input name more general! 
```suggestion
  make_option(
    opt_str = c("-f", "--sce_file"),
    type = "character",
    help = "path to SingleCellExperiment file to update. Must end in .rds"
  ),
```

bin/add_submitter_annotations.R
```diff
@@ -0,0 +1,77 @@
...
+    barcodes = cell_barcode,
+    submitter_celltype_annotations = cell_type_assignment
+  ) |>
+  # in the event of NA values, change to "Unclassified cell"
+  dplyr::mutate(
+    submitter_celltype_annotations = dplyr::if_else(
+      is.na(submitter_celltype_annotations),
+      "Unclassified cell",
+      submitter_celltype_annotations
+    )
```
jashapiro, line 60: Why not be a bit tidyr?

```suggestion
  # in the event of NA values, change to "Unclassified cell"
  dplyr::mutate(
    submitter_celltype_annotations = tidyr::replace_na(
      submitter_celltype_annotations
      "Unclassified cell"
    ) 
```

Or even:
```suggestion
   # in the event of NA values, change to "Unclassified cell"
    tidyr::replace_na(list(submitter_celltype_annotations = "Unclassified cell"))
```

bin/add_submitter_annotations.R
```diff
@@ -0,0 +1,77 @@
...
+  )
+  
+# Join in submitter cell types
+colData(sce) <- colData(sce) |>
+  as.data.frame() |>
+  dplyr::left_join(
+    submitter_cell_types_df
+  ) |>
```
jashapiro, line 68: Probably worth checking that row count didn't change in case someone submits a table with multiple values for a cell.

bin/add_submitter_annotations.R
```diff
@@ -0,0 +1,77 @@
...
+  # make sure we keep rownames
+  DataFrame(row.names = colData(sce)$barcodes)
+
+# Indicate that we have submitter celltypes in metadata, 
+#  saving in same spot as for actual celltyping
+metadata(sce)$celltype_methods <- "submitter"
```
jashapiro, line 74: This _should_ be the first we add, but we might as well be safe
```suggestion
metadata(sce)$celltype_methods <- c(metadata(sce)$celltype_methods, "submitter")
```

jashapiro (commented):
My main comment here is that we can be a bit more agnositic in the script about whether the file is filtered or not. Just because we are using unfiltered now doesn't mean we always have to! (the suggested change also requires a number of other renaming changes, but not too bad, I hope)

sjspielman:
> I had a few comments, but another thought I had is related to how we are labeling "unclassified cells" vs "unknown cells". We should standardize this across methods probably, but also we may want to distinguish here between a cell that was not found in the submitter cell types (unclassified) vs. a cell type that was NA or unknown in the submitter file (unknown).

More generally, yes we should make sure this is all clear - I will open a separate issue for that!

In terms of _here_, we decided at one point in Slack (see `#scpca` channel) to do what I've done here - 
- `NA` = cells the submitter didn't annotate at all; these cells are not in the submitter TSVs.
- "Unclassified cells" = cells the submitter labeled as `NA`; these cells are in the submitter TSVs.

sjspielman:
Moving right along! I've incorporated @jashapiro's suggestions and tested the workflow again, all running as expected 👍
@allyhawkins, back to you! The `empty_file` sadly must remain, but don't worry, I'm not the biggest fan of it either 😄  (though I do understand why `nextflow` does it that way).

modules/sce-processing.nf
```diff
@@ -224,10 +249,15 @@ workflow generate_merged_sce {
...
-    feature_sce_ch = feature_quant_channel
-      // RNA meta is in the third slot here
-      .map{it.toList() + [file(it[2].mito_file), file(it[2].ref_gtf)]}
 
+    feature_sce_ch = feature_quant_channel 
+      // RNA meta is in the third slot here
+      .map{it.toList() + [file(it[2].mito_file), 
+                         file(it[2].ref_gtf), 
+                         // either submitter cell type files, or empty file if not available
+                         file(it[2].submitter_cell_types_file ?: file("${projectDir}/assets/NO_FILE.txt"))
```
jashapiro, line 258:
```suggestion
                         file(it[2].submitter_cell_types_file ?: empty_file) 
```

modules/sce-processing.nf
```diff
@@ -200,13 +219,18 @@ workflow generate_sce {
     quant_channel
     sample_metafile
   main:
+    
+    empty_file = file("${projectDir}/assets/NO_FILE.txt")
```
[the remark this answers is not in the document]
jashapiro: Since this is used within `file()` statements, I think it can be just:
```suggestion
    empty_file = "${projectDir}/assets/NO_FILE.txt"
```

(the nested file() doesn't _hurt_ exactly, it just isn't necessary)

modules/sce-processing.nf
```diff
@@ -68,6 +78,15 @@ process make_merged_unfiltered_sce{
           --sample_id "${meta.sample_id}" \
           --sample_metadata_file ${sample_metafile} \
           ${params.spliced_only ? '--spliced_only' : ''}
+          
+        # Only run script if annotations are available:
+        if [ ${submitter_cell_types_file.name} != "NO_FILE.txt" ]; then
+          add_submitter_annotations.R \
+            --sce_file ${unfiltered_rds} \
+            --library_id "${meta.library_id}" \
+            --submitter_cell_types_file ${submitter_cell_types_file}
```
jashapiro, line 87: May as well quote them all (particularly the cell types file)
```suggestion
            --sce_file "${unfiltered_rds}" \
            --library_id "${meta.library_id}" \
            --submitter_cell_types_file "${submitter_cell_types_file}"
```

modules/sce-processing.nf
```diff
@@ -23,13 +24,22 @@ process make_unfiltered_sce{
...
           --sample_metadata_file ${sample_metafile} \
           ${params.spliced_only ? '--spliced_only' : ''}
+          
+          
+        # Only run script if annotations are available:
+        if [ ${submitter_cell_types_file.name} != "NO_FILE.txt" ]; then
+          add_submitter_annotations.R \
+            --sce_file ${unfiltered_rds} \
+            --library_id "${meta.library_id}" \
+            --submitter_cell_types_file ${submitter_cell_types_file}
```
jashapiro, line 34: More quotes
```suggestion
          add_submitter_annotations.R \
            --sce_file "${unfiltered_rds}" \
            --library_id "${meta.library_id}" \
            --submitter_cell_types_file "${submitter_cell_types_file}"
```

jashapiro (approved):
LGTM!

jashapiro:
I can't believe I missed this in review: 255a68ab96fc072996e5fae82ddb70a749dfa96a

modules/sce-processing.nf
sjspielman: I would think this needs to be there in full, since `empty_file` is not defined in _this_ subworkflow (it's only in `generate_sce`). But, definitely don't need 2x `file()`.

sjspielman: I'll note that the diff made it hard to see where `empty_file` was actually defined! I ended up removing the 2nd `file()` but keeping the rest.

modules/sce-processing.nf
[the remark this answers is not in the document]
sjspielman: got rid of this `file()`, but then added below on 237 so everybody gets 1 `file()`.
https://github.com/AlexsLemonade/scpca-nf/pull/451/commits/abfaacc69e2a671091ff5492e0d6e88af1040b51

modules/sce-processing.nf
```diff
@@ -200,16 +219,22 @@ workflow generate_sce {
...
+    // used for both submitter cell types & feature barcode files
+    empty_file = "${projectDir}/assets/NO_FILE.txt"
+ 
     sce_ch = quant_channel
-      .map{it.toList() + [file(it[0].mito_file), file(it[0].ref_gtf)]}
-
+      .map{it.toList() + [file(it[0].mito_file), 
+                         file(it[0].ref_gtf), 
+                         // either submitter cell type files, or empty file if not available
+                         file(it[0].submitter_cell_types_file ?: empty_file)
```
allyhawkins, line 230: This is minor, but mirroring what you do below can you just replace the `empty_file` variable with the actual file name? Or is there some nextflow reason you need to define it here, but not below?

allyhawkins (approved):
One small question, but other than that LGTM!

modules/sce-processing.nf
jashapiro: > I'll note that the diff made it hard to see where `empty_file` was actually defined! I ended up removing the 2nd `file()` but keeping the rest.

Pretty sure you define it before the workflow blocks and just have it available to all of them with one definition.

modules/sce-processing.nf
sjspielman: I defined it up here since it's used _twice_ in this sub-workflow, but only once in the other one. I could define it in both sub-workflows, though, if that's preferred!

allyhawkins: oh sorry I missed that you are also using it below in the `map` call. Good to keep it!

sjspielman: np, this PR got big(!)

PR #453, opened 2023-09-14
Incorporate assay ontology, suspension type, and title to AnnData objects

### LINKED ISSUE #419: Incorporate tech_version and seq_unit into cell metadata for AnnData objects
I'm breaking this issue out of #366. In addition to adding metadata from the sample tables to the AnnData objects, we also need to add `tech_version` and `seq_unit` as columns in the AnnData object. 

The `tech_version` is already available as part of the SCE object, but `seq_unit` is not. Do we want to go back and incorporate `seq_unit` into the metadata of the unfiltered SCE object? (Which probably means adding it as an argument to adding metadata to SCE objects in `scpcaTools`? Or do we want to include it as an argument to the AnnData conversion process? 

Also, we will need a way to map the `tech_version` and `seq_unit` to the appropriate ontology term ids. We could include both the human readable value and the ontology term id in the AnnData object, or we could do the conversion prior to making the object so only the ontology term id is present. 

I haven't really decided which options I like better, so I'm just dumping ideas here for now.

### LINKED ISSUE #430: Modify metadata of SCE prior to AnnData conversion to be CZI compliant
We will need to add in the `schema_version` and `title` to the `metadata(sce)` prior to AnnData conversion. 
The `schema_version` should be set to `3.0.0` and the `title` should be a unique description of all the samples in a single project. 
See https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#uns-dataset-metadata

The schema will be easy, but for the title, we should think about how we want to add that into the workflow? Should we add a new column to the sample metadata and pull it from there? Or do we want to read in the project metadata too?

allyhawkins:
Closes #419 
Closes #430 

This PR wraps up the additions of the remaining items to the AnnData objects. 

- I added `assay_ontology_term_id` to the `meta` channel created at the beginning of the script. The other items were already added there so I didn't need to do anything there. 
- There are now new arguments to `generate_unfiltered_sce.R` to incorporate all of the new pieces of metadata into the metadata of the SCE object. 
- The script for converting to AnnData objects that contains the `format_czi` function now includes moving the specific pieces of metadata that are not part of the `sample_metadata` into their own columns in the `colData`.
- I realized previously I had added the schema version in the script that only gets run on processed objects. I moved that to be added before we convert to AnnData objects. 
- When moving the schema, I decided to add a parameter for it. Just in case we ever make changes and want to reformat to fit a different schema, I thought it would be better than hard coding it. 

I tested this and it's working except for the line in the script that does the actual conversion to AnnData is failing. I'm getting an error that there's a bad type in the `subdiagnosis` column. I looked and that column is all `NA_character_`. However, there's another column that is all NA logical. When I remove the `subdiagnosis` column and keep the other NAs that are logical things export correctly. I'm trying to find a way to detect if columns have `NA_character_` and replace them with NA, but I am struggling. I've tried the following code with the `colData`, but all it does is convert each individual element in the column to a logical NA, but the column is still a character, so the error is still thrown. 

```
  coldata_df <- as.data.frame(colData(sce)) |>
    # make all the NA's the same
    dplyr::mutate(across(everything()), anyNA, as.logical)
```

Here's the error for reference: 

Screenshot 2023-09-14 at 11 54 09 AM

Tagging @jashapiro and @sjspielman for any thoughts and help on wrangling NAs.

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -49,6 +49,16 @@ option_list <- list(
     help = "sequencing technology string to store in metadata"
   ),
   make_option(
+    opt_str = c("--assay_ontology_term_id"),
+    type = "character",
+    help = "Experimental Factor Ontology term associated with provided tech_version"
+  ),
+  make_option(
+    opt_str = c("-s", "--seq_unit"),
+    type = "character",
+    help = "sequencing unit string to store in metadata (e.g., cell, nucleus)"
+  ),
+  make_option(
     opt_str = c("--library_id"),
     type = "character",
     help = "library id"
@@ -57,6 +67,11 @@ option_list <- list(
     opt_str = c("--sample_id"),
     type = "character",
     help = "sample id(s). If more than one, separated by commas and/or semicolons."
+  ),
+  make_option(
+    opt_str = c("--project_id"),
+    type = "character",
+    help = "project id"
   ),
   make_option(
     opt_str = c("--sample_metadata_file"),
@@ -111,8 +126,11 @@ unfiltered_sce <- read_alevin(
   include_unspliced = include_unspliced,
   fry_mode = TRUE,
   tech_version = opt$technology,
+  assay_ontology_term_id = opt$assay_ontology_term_id,
+  seq_unit = opt$seq_unit,
   library_id = opt$library_id,
-  sample_id = sample_ids
+  sample_id = sample_ids,
+  project_id = opt$project_id
 )
 
 # read and merge feature counts if present
@@ -122,8 +140,12 @@ if (opt$feature_dir != "") {
     include_unspliced = FALSE,
     fry_mode = TRUE,
     feature_data = TRUE,
+    tech_version = opt$technology,
+    assay_ontology_term_id = opt$assay_ontology_term_id,
+    seq_unit = opt$seq_unit,
     library_id = opt$library_id,
-    sample_id = sample_ids
+    sample_id = sample_ids,
+    project_id = project_id
   )
 
   unfiltered_sce <- merge_altexp(unfiltered_sce, feature_sce, opt$feature_name)
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -145,7 +145,7 @@ if (opt$feature_dir != "") {
     seq_unit = opt$seq_unit,
     library_id = opt$library_id,
     sample_id = sample_ids,
-    project_id = project_id
+    project_id = opt$project_id
   )
 
   unfiltered_sce <- merge_altexp(unfiltered_sce, feature_sce, opt$feature_name)
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -30,12 +30,6 @@ option_list <- list(
     type = "character",
     help = "path to output hdf5 file to store feature counts as AnnData object.
     Only used if the input SCE contains an altExp. Must end in .hdf5 or .h5"
-  ),
-  make_option(
-    opt_str = c("--czi_schema_version"),
-    type = "character",
-    default = "3.0.0",
-    help = "Version number for CZI schema matching formatting of output AnnData objects"
   )
 )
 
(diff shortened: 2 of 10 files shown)


allyhawkins:
> I don't think you want to do any conversion if just some values are NA, but to convert where _all_ are NA you should be able to do:
> 
> ```
> coldata_df <- as.data.frame(colData(sce)) |>
>  dplyr::mutate(across(where(\(x) all(is.na(x))), as.logical))
> ```

That gives the same issue where the column doesn't get converted to a logical and it's still a character, so the same error comes up.

jashapiro:
> * When moving the schema, I decided to add a parameter for it. Just in case we ever make changes and want to reformat to fit a different schema, I thought it would be better than hard coding it.

This is a case where I think we _don't_ want a parameter. The reason is that if we have to change the schema, we will have to change the script to generate the data: it isn't something that we will ever change on the fly, unless we massively rewrite things to support more than one schema.

jashapiro:
> > I don't think you want to do any conversion if just some values are NA, but to convert where _all_ are NA you should be able to do:
> > ```
> > coldata_df <- as.data.frame(colData(sce)) |>
> >  dplyr::mutate(across(where(\(x) all(is.na(x))), as.logical))
> > ```
> 
> That gives the same issue where the column doesn't get converted to a logical and it's still a character, so the same error comes up.

hmm... it works for me in testing: 
Screenshot 2023-09-14 at 1 37 39 PM

sjspielman:
Code works for me too with `SCPCL000001` - 

```
# make main sce czi compliant
sce <- format_czi(sce)

colData(sce) <- colData(sce) |>
  as.data.frame() |> 
  dplyr::mutate(across(where(\(x) all(is.na(x))), as.logical)) |> 
  DataFrame()

# export sce as anndata object
scpcaTools::sce_to_anndata(
  sce,
  anndata_file = opt$output_rna_h5
)
```

allyhawkins:
> I added this fix to scpcaTools, so it should end up redundant here: [AlexsLemonade/scpcaTools#222](https://github.com/AlexsLemonade/scpcaTools/pull/222)

So should I remove it from here? or are we okay with the redundancy?

bin/generate_unfiltered_sce.R
```diff
@@ -122,8 +140,12 @@ if (opt$feature_dir != "") {
     include_unspliced = FALSE,
     fry_mode = TRUE,
     feature_data = TRUE,
+    tech_version = opt$technology,
+    assay_ontology_term_id = opt$assay_ontology_term_id,
+    seq_unit = opt$seq_unit,
     library_id = opt$library_id,
-    sample_id = sample_ids
+    sample_id = sample_ids,
+    project_id = project_id
```
sjspielman, line 148: I don't think this exists? Should it be..

```suggestion
    project_id = opt$project_id
```

bin/sce_to_anndata.R
```diff
@@ -54,16 +54,36 @@ if (!(stringr::str_ends(opt$output_rna_h5, ".hdf5|.h5"))) {
 # CZI 3.0.0 requirements: https://github.com/chanzuckerberg/single-cell-curation/blob/b641130fe53b8163e50c39af09ee3fcaa14c5ea7/schema/3.0.0/schema.md
 format_czi <- function(sce) {
   # add library_id as an sce colData column
+  # need this column to join in the sample metadata with the colData
```
sjspielman, line 57: A+ comment

jashapiro:
> > I added this fix to scpcaTools, so it should end up redundant here: [AlexsLemonade/scpcaTools#222](https://github.com/AlexsLemonade/scpcaTools/pull/222)
> 
> So should I remove it from here? or are we okay with the redundancy?

Probably remove? It makes more sense there, since that is closer to the source of the error.

bin/sce_to_anndata.R
```diff
@@ -54,16 +54,36 @@ if (!(stringr::str_ends(opt$output_rna_h5, ".hdf5|.h5"))) {
...
   sce <- scpcaTools::metadata_to_coldata(sce,
     join_columns = "library_id"
   )
 
+  # modify colData to be AnnData and CZI compliant
+  coldata_df <- colData(sce) |>
+    as.data.frame() |>
+    # make sure all colData columns have the correct NA formatting if all NA
+    # python doesn't like character NA
+    dplyr::mutate(across(where(\(x) all(is.na(x))), as.logical),
```
sjspielman: For legibility, can you enter here? I'm sure I'll mess up the formatting if I try to suggest. Just looking for `across(...` to start on its own line.

bin/sce_to_anndata.R
```diff
@@ -54,16 +54,36 @@ if (!(stringr::str_ends(opt$output_rna_h5, ".hdf5|.h5"))) {
...
+                  # create columns for assay and suspension ontology terms
+                  assay_ontology_term_id = metadata(sce)$assay_ontology_term_id,
+                  suspension_type = metadata(sce)$seq_unit,
```
sjspielman: Might want to add a check to make sure these are not NULL? I don't see how they would be, unless something went wrong in `generate_unfiltered_sce`, but I just love checks! Your call on this, if you think it's way overkill no worries.

sjspielman:
> Probably remove? It makes more sense there, since that is closer to the source of the error.

I think I agree with this, but the rest of that `mutate()` is good!

allyhawkins:
> > Probably remove? It makes more sense there, since that is closer to the source of the error.
> 
> I think I agree with this, but the rest of that `mutate()` is good!

Yes, done in https://github.com/AlexsLemonade/scpca-nf/pull/453/commits/386d10133c71f6d7d074a2058298aec4b735a802

bin/sce_to_anndata.R
```diff
@@ -54,16 +54,36 @@ if (!(stringr::str_ends(opt$output_rna_h5, ".hdf5|.h5"))) {
 # CZI 3.0.0 requirements: https://github.com/chanzuckerberg/single-cell-curation/blob/b641130fe53b8163e50c39af09ee3fcaa14c5ea7/schema/3.0.0/schema.md
 format_czi <- function(sce) {
```
sjspielman, line 55: I wonder if it might be helpful to actually called this function `format_czi_3.0.0`? I don't think it's necessary, and it doesn't hurt to leave it as is, but it's a thought I had. If you like the thought, go for it, and if not, do not!

sjspielman (approved):
LGTM! I left some comments but really nothing big. Note, I didn't test the script myself though, lmk if that would help.

One thing I can't comment directly on - Might be helpful to add a quick comment at the TOP of `sce_to_anndata.R` that code is currently written with `3.0.0` CZI schema in mind.

jashapiro:
> Update: But you won't be able to test/merge this until that goes in and builds the docker image. 🐌

Should be ready

PR #455, opened 2023-09-14
Sync error handling for gene variance and PCA to main

allyhawkins:
So that I can finish processing the Collins data, I want to get the changes that handle the errors seen with post-processing added in #449 into `main` and do a point release. This just incorporates the changes in the post-processing script since the clustering script doesn't exist in `main` yet.


jashapiro (approved):
LGTM!

PR #457, opened 2023-09-14
Add submitter annotations to QC report

### LINKED ISSUE #413: Include a heatmap comparing cell type annotations to submitter annotations in cell type report
The last section of the cell type report should be a similar heatmap to #412, but comparing cell type annotations to submitter annotations. This section will only be included if submitter annotations are available. This should look like the picture below, but again, there is no aneuploid/diploid status right now. 

Screenshot 2023-08-23 at 2 29 08 PM

sjspielman:
Closes #413 
This PR started as just adding a heatmap for _comparing_ with submitter annotations to the QC, but I ended up adding submitter annotations throughout the QC report to mirror `SingleR` and `CellAssign`! The one thing I did _not_ do the same for submitter annotations is convert any of its `NA` values to "Unknown cell type," since those would be cells that the submitter didn't even touch.

For developing this, I had to tweak an SCE manually to have _all_ expected content; the SCEs we have run CellAssign on do not have submitter annotations, so we didn't have an object ready to go. Rather than running CellAssign (I mean, right?!), I used `SCPCL000295` since it has `SingleR` and `CellAssign` but I directly modified the `processed` sce to have all the parts  as follows:
  - Changed UMAP names to `"UMAP1"` to and `"UMAP2"`, to match what the workflow would export now
  - Manually added `"submitter"` as one of the cell type methods in the metadata
  - Manually added `"submitter_celltype_annotations"` as `sample(LETTERS[1:7])`, so dont interpret the submitter plots too much besides the fact that they render! The submitter UMAP, for example, is just confetti 🎨 

I also fiddled a decent amount with the printed text to try and get the right logic depending on which cell type we have or not. Reviewers, please check my `eval = ` in the celltype report chunks!

Here's the report!
[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12612786/qc_report.html.zip)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -18,21 +37,31 @@ knitr::asis_output(
...
 ```
 
 
-Cells annotated as "Unknown cell type" represent cells which could not be confidently identified using the given annotation method.
+<!-- Submitter and at least one inference method --> 
+```{r, eval = has_submitter & (has_singler | has_cellassign)}
+knitr::asis_output(
+  glue::glue(
+  "* Cell type annotations were also provided by the original lab (`submitter-provided`) which generated this data. 
+  In these cell type annotations, the label `Unclassified cell` indicates that submitters did not provide an annotation for the given cell.
```
allyhawkins, line 53: I wonder if it might make more sense to move the statement about unclassified cells to right before the submitter table? Since you have the statement about other unknowns right before the singleR and CellAssign table I feel like that would be easier to follow?

templates/qc_report/celltypes_qc.rmd
```diff
@@ -554,7 +592,7 @@ create_celltype_heatmap(
 )
 ```
 
-```{r, eval = has_submitter_celltypes, fig.height=5, fig.width=7}
+```{r, eval = has_submitter, fig.height=5, fig.width=7}
 knitr::asis_output("### Submitter-provided cluster and cell type heatmap\n")
```
allyhawkins, line 596:
```suggestion
knitr::asis_output("### Submitter-provided cell type and cluster heatmap\n")
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -582,3 +620,37 @@ create_celltype_heatmap(
...
+## Comparison with submitter annotations
+
+This section shows heatmap(s) directly comparing submitter-provided cell type annotations to cell type inferred with the given method(s).
+Again, different annotations may use different names for similar cell types.
+")
+```
+
+
+```{r, eval = has_submitter & has_singler, fig.height=7, fig.width=8}
+
```
allyhawkins, line 635: I feel like it might make sense to have a title above these heatmaps? One for singleR and one for CellAssign? I was confused at first because the legend label for SingleR looks like the title for the CellAssign plot.

allyhawkins (commented):
Mostly looks good, just a few small comments.

sjspielman:
I've responded to reviews as follows:

- Moved "Unclassified cell" sentence to above the submitter table
- Updated the heatmap header as suggested (I did it myself, not _with_ the suggestion)
- For the title - the title in `ComplexHeatmap()` just becomes the _legend_ title (boo), so here's how I handled this:
  - I added an argument to the heatmap function for where the column label should be placed. By default, it's on the _bottom_, but for these two plots I overrode this and placed the column label on the _top_. I added a bit more space with a `<br>` too, and I think they are clearly distinguished now!

Here's the report:
[qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12620777/qc_report.html.zip)

allyhawkins (approved):
LGTM 🎉

PR #458, opened 2023-09-14
Update version number in config and docs

allyhawkins:
Should have done this in #455, PR incoming to make the same changes to `development`.


jashapiro (approved):
⚡

Issue #127, opened 2022-03-11
Check metadata fields as part of the workflow


jashapiro:
With https://github.com/AlexsLemonade/scpca-nf/pull/122, we will be checking for the existence of a metadata file and other workflow-wide settings before starting the workflow itself, in order to hopefully give better error messages.

The metadata file has some specific formatting requirements too, and we ought to be able to check those locally as well.

One idea I had for this was to make the initial check part of the workflow logic with code something like the following:
```nextflow
metafile_ch = Channel.fromPath(run_metafile)
    .splitCsv(header: true, sep: '\t')
    .map{check_meta(it)}
    .branch{ 
        pass: it.errors.size() == 0
        fail: true
    }
```

where `check_meta()` is a function that performs tests on each row, and returns an array of errors. The tests would be for things like making sure the `seq_units` and `technology` contain valid values, that files exist where expected, etc.

The workflow could then continue with the information from rows which passed, and print out errors for the rows that failed. 

Alternatively, there may be a way to make the whole process stop on encountering errors (if there is anything in the `metafile_ch.fail` subchannel).

Not 100% sure of the best way here; there may be some testing needed!

PR #460, opened 2023-09-15
Parse NA in input files

### LINKED ISSUE #452: Update how we handle optional files
A while ago we changed from blank entries to `NA`, and I guess I assumed that `.splitCsv()` would handle that, but it does not, which is why the elvis operator doesn't work. I actually think we should translate NAs here at the start of parsing to avoid this and make for better downstream behavior. **Other optional fields (`feature_barcode_file`, etc.) should maybe get the same treatment.**

```suggestion
submitter_cell_types_file: it.submitter_cell_types_file == "NA"? it.submitter_cell_types_file : ""
```

The blank string will then allow us to use `file(it[0].submitter_cell_types_file?: empty_file)` later.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/451#discussion_r1326094681_
            

NOTE: The **bolded** bit above is the focus of this issue!

jashapiro:
closes #452

Since the .splitCsv() function won't remove NAs for us, we have to do it ourselves. I chose here to write a helper function in the Utils file. Rather than just replacing `NA` with an empty string, it actually makes those values `null`. 

I don't now convert blanks to `null`, because that seemed a step too far. But both `null` and `""` are considered as "falsey", so it should not matter for most use cases. 

Most of these should not require any further adjustments to the workflow: I did "add back" `NA` values in a couple of places to be sure we provided something as an argument, notably for slide info in spatial contexts.

I only parse fields that we allow to be `NA`: required fields _should_ cause failures if they are not present (we could probably be better about catching those and providing meaningful feedback, but that is for another time)


lib/Utils.groovy
```diff
@@ -40,4 +40,15 @@ class Utils {
     def obj = new JsonSlurper().parse(file)
```
sjspielman, line 40: `JsonSlurper` is my new favorite class name

lib/Utils.groovy
```diff
@@ -40,4 +40,15 @@ class Utils {
...
+
+
+  /**
+   * Replace a string with an NA value with null
+   *
+   * @param str A string
+   * @return The input string unless it was NA or a variant thereof, in which case returns null
+   */
+   static def parseNA(str) {
+     str.toLowerCase()in ["na","n/a","nan"]? null : str
```
sjspielman, line 52: unless there is _intentionally_ no space for syntax reasons I'm unaware of, add space - 
 
```suggestion
     str.toLowerCase() in ["na","n/a","nan"]? null : str
```

modules/sce-processing.nf
```diff
@@ -119,7 +119,6 @@ process filter_sce{
         // - barcode file should exist
         // - barcode file should _not_ be the empty file NO_FILE.txt
         adt_present = meta.feature_type == 'adt' &
-          feature_barcode_file.exists() &
```
sjspielman, line 122: 👍 , can you also update the comments right above here?

modules/sce-processing.nf
```diff
@@ -219,25 +218,26 @@ process post_process_sce{
         """
 }
 
+
+// used when a given file is not defined in the below workflows
+empty_file = "${projectDir}/assets/NO_FILE.txt"
```
sjspielman, line 223: 👍

modules/spaceranger.nf
```diff
@@ -24,8 +24,8 @@ process spaceranger{
       --localcores=${task.cpus} \
       --localmem=${task.memory.toGiga()} \
       --image=${image_file} \
-      --slide=${meta.slide_serial_number} \
-      --area=${meta.slide_section}
+      --slide=${meta.slide_serial_number ?: "NA"} \
+      --area=${meta.slide_section ?: "NA"}
```
sjspielman, line 28: Just to confirm you want a string `"NA"` and not an empty string `""` or so. 

I had a _quick_ glance at `spaceranger count` docs, seems like either way (`"NA"` or `""`) this will fail since the flag `--unknown-slide` is needed if this info is missing?

Having not actually run `spaceranger count` before, of course, I trust you know the arguments quite a lot better than I do!

main.nf
```diff
@@ -92,17 +92,17 @@ workflow {
...
-      submitter_cell_types_file: it.submitter_cell_types_file == "NA"? it.submitter_cell_types_file : "",
-      feature_barcode_file: it.feature_barcode_file,
-      feature_barcode_geom: it.feature_barcode_geom,
-      files_directory: it.files_directory,
-      slide_serial_number: it.slide_serial_number,
-      slide_section: it.slide_section,
+      submitter_cell_types_file: Utils.parseNA(it.submitter_cell_types_file),
+      feature_barcode_file: Utils.parseNA(it.feature_barcode_file),
+      feature_barcode_geom: Utils.parseNA(it.feature_barcode_geom),
+      files_directory: Utils.parseNA(it.files_directory),
```
sjspielman, line 103: Seems to me this is really _required_ to not be `NA`? Or, is this is not needed if we are _not_ repeating mapping?

sjspielman (approved):
Seems fine to me, left comments that are mostly for my understanding! Noting I did not run the workflow during review.

Can you also open an issue for this maybe?
> I only parse fields that we allow to be NA: required fields should cause failures if they are not present (we could probably be better about catching those and providing meaningful feedback, but that is for another time

main.nf
sjspielman: ^ "this" = `files_directory`

modules/spaceranger.nf
jashapiro: I did want character NA, as otherwise I would expect `--slide= ` to fail. I actually have no idea what will happen here, as we have never tested with "NA" values, but this reproduces the previous behavior, which was my goal at this point. If we need a future update, I would handle it then.

jashapiro:
> Can you also open an issue for this maybe?

Already exists! https://github.com/AlexsLemonade/scpca-nf/issues/127

PR #463, opened 2023-09-18
Require more than one cell to create an AnnData object

### LINKED ISSUE #461: Discussion: How to handle converting SCE to AnnData with 1 cell present
I recently have run into an error converting any objects that only have 1 cell from SCE to AnnData (which I mean makes sense given its only 1 cell...). I'm attaching the error that I've been getting and noting that I've narrowed it down to a problem with the `zellkonverter` function that is converting from SCE to AnnData. It appears that a list or some other collection type for the column names is needed when creating the AnnData object, rather than a single character vector. This is because in the case of an object with 1 cell the `colnames(sce)` are just a single string rather than a list of strings... 

Screenshot 2023-09-15 at 9 42 26 AM

So we need to decide how to handle this situation. 

1. If SCE's don't have at least 2 cells, we don't convert to AnnData. 
2. We could take this one step further, and if we don't have at least 2 cells after filtering, we just don't even create a processed object... I think that might break more things, though so I'm more hesitant to do this. 
3. There is some secret way to make `zellkonverter` treat the `colnames(sce)` as a list regardless of how many cells are there... but I haven't come up with this answer, and I think outside of changing the function or creating our own AnnData object this is hard to do. 

I would like option 1, but I want to run this by others first and see if others have different ideas. 
@jashapiro @jaclyn-taroni @sjspielman

allyhawkins:
Closes #461 

This PR modifies the workflow so that we only create AnnData objects for SCE objects that have > 1 cell present. One option I thought of was making an optional output for the process that converts SCE to AnnData and only create the output if the number of cells is > 1. However, I was worried about using optional files as output and how that would affect downstream steps. Instead, I used an approach that would allow us to filter the input to the `export_anndata` process based on the number of cells. 

- We already provide the number of `filtered` and `unfiltered` cells in the `<library_id>_metadata.json` file that is returned in the final output. I thought it made sense to also include the number of cells present in the processed object. I added that here. 
- Previously, we had been using the output from post processing as the input to creating the AnnData object. I updated the workflow here to take the output from the clustering process instead. Additionally, I am joining that with the output from the qc report process. This creates a tuple with the `[meta, unfiltered rds, filtered rds, processed rds, qc report, metadata.json]`. This is then the input to the AnnData process. 
- In the workflow for exporting the AnnData, I added a step to read in the `metadata.json` and grab the number of cells for that specific file (e.g., unfiltered, filtered, processed). If the number of cells <= 1 then that rds file doesn't get passed to the conversion step. 

I tested this with both a regular library that should have all the objects converted and the library that was failing conversion with 1 cell. 

Another thing I want to note is that we should make an update to `scpcaTools::sce_to_anndata()` that throws an error if an object only has 1 cell. I will file that as a follow up PR.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -225,8 +225,12 @@ workflow {
   sce_qc_report(cluster_sce.out, report_template_tuple)
 
   // convert SCE object to anndata
-  // do this for everything but multiplexed libraries
-  anndata_ch = post_process_sce.out
+  anndata_ch = cluster_sce.out
+    // join clustering output with qc report data
+    // we need to be able to grab the number of cells for each object from the metadata json (output from sce_qc_report)
+    // tuple of [meta, unfiltered, filtered, processed, qc report, metadata json]
+    .join(sce_qc_report.out, by: 0, failOnDuplicate: true, failOnMismatch: true)
+    // skip multiplexed libraries
     .filter{!(it[0]["library_id"] in multiplex_libs.getVal())}
   sce_to_anndata(anndata_ch)
 
--- a/modules/export-anndata.nf
+++ b/modules/export-anndata.nf
@@ -57,11 +57,20 @@ workflow sce_to_anndata{
       sce_files_ch
     main:
       sce_ch = sce_files_ch
-        // make tuple of [meta, sce_file, type of file]
-        .flatMap{[[it[0], it[1], "unfiltered"],
-                  [it[0], it[2], "filtered"],
-                  [it[0], it[3], "processed"]
+        // make tuple of [meta, sce_file, type of file, metadata.json]
+        .flatMap{[[it[0], it[1], "unfiltered", it[5]],
+                  [it[0], it[2], "filtered", it[5]],
+                  [it[0], it[3], "processed", it[5]]
                  ]}
+        // remove any sce files that don't have enough cells in the sce object
+        // number of cells are stored in each metadata.json file
+        .filter{ Utils.getMetaVal(file(it[3]), "${it[2]}_cells") > 1 }
+        // remove metadata.json file from tuple
+        .map{meta, sce_file, file_type, meta_json -> tuple(
+          meta,
+          sce_file,
+          file_type
+        )}
 
       // export each anndata file
       export_anndata(sce_ch)
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -178,6 +178,7 @@ metadata_list <- list(
   is_multiplexed = multiplexed,
   has_citeseq = has_citeseq,
   has_cellhash = has_cellhash,
+  processed_cells = ncol(processed_sce),
   filtered_cells = ncol(filtered_sce),
   unfiltered_cells = ncol(unfiltered_sce),
   droplet_filtering_method = filtered_sce_meta$filtering_method,


main.nf
```diff
@@ -225,8 +225,12 @@ workflow {
   sce_qc_report(cluster_sce.out, report_template_tuple)
 
   // convert SCE object to anndata
-  // do this for everything but multiplexed libraries
-  anndata_ch = post_process_sce.out
+  anndata_ch = cluster_sce.out
+    // join clustering output with qc report data
+    // we need to be able to grab the number of cells for each object from the metadata json (output from sce_qc_report)
+    // tuple of [meta, unfiltered, filtered, processed, qc report, metadata json]
+    .join(sce_qc_report.out, by: 0, failOnDuplicate: true, failOnMismatch: true)
```
jashapiro, line 232: Rather than joining, should we instead make the qc report process export the rds files as well? then we can just pass it in directly. 

My main concern here is always that I don't trust nextflow joining on metadata. If we modify the metadata in any step, we can break the join. Which is why I have been pulling the library id out for joins like this. But passing the full objects may be easier, and it means we could move the publishing of the rds files from clustering to the QC report step, which maybe makes more sense?

modules/export-anndata.nf
```diff
@@ -57,11 +57,20 @@ workflow sce_to_anndata{
...
                  ]}
+        // remove any sce files that don't have enough cells in the sce object
+        // number of cells are stored in each metadata.json file
+        .filter{ Utils.getMetaVal(file(it[3]), "${it[2]}_cells") > 1 }
+        // remove metadata.json file from tuple
+        .map{meta, sce_file, file_type, meta_json -> tuple(
+          meta,
+          sce_file,
+          file_type
+        )}
```
jashapiro, line 73: This is nice and explicit, but we could do it by just dropping the last element in this case:
```suggestion
        .map{it.dropRight(1)}
```

Or there is actually a method that returns all but the last element, but I think it is really badly named:
```
.map{it.init()}
```

modules/export-anndata.nf
```diff
@@ -57,11 +57,20 @@ workflow sce_to_anndata{
       sce_files_ch
     main:
       sce_ch = sce_files_ch
-        // make tuple of [meta, sce_file, type of file]
-        .flatMap{[[it[0], it[1], "unfiltered"],
-                  [it[0], it[2], "filtered"],
-                  [it[0], it[3], "processed"]
+        // make tuple of [meta, sce_file, type of file, metadata.json]
```
jashapiro, line 60: This comment (or a comment in `take:` should probably also state what the input channel will contain.)

modules/export-anndata.nf
jashapiro: Alternatively, we could update the export_anndata process to expect a metadata json file? Eh, probably better this way.

jashapiro (commented):
This looks good, but I wonder if we might want to restructure just a bit to make the QC report process also the publish process for the rds files. Then we could have the input to the AnnData conversion process already have all of the data we need, and save ourselves a bit of workflow logic. You could even have the qc report process output two tuples, so that we wouldn't need additional logic to drop the qc report as input to the anndata conversion.

allyhawkins:
Thanks for taking a look at this @jashapiro! I think because it was mentioned in https://github.com/AlexsLemonade/scpca-nf/pull/464#issuecomment-1726458336 that we want to keep the AnnData export as the last step, I'm marking this ready for review. 

I updated this so I'm directly passing the rds files through the qc report process rather than joining. I tested this and everything looks good, so this is ready for another look.

modules/qc-report.nf
```diff
@@ -10,7 +10,8 @@ process sce_qc_report{
         tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(processed_rds)
         tuple path(template_dir), val(template_file)
     output:
-        tuple val(meta), path(qc_report), path(metadata_json)
+        tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(processed_rds),
+          path(qc_report), path(metadata_json)
```
jashapiro, line 14: I don't think this is required, but it is kind of a nice option: naming the outputs so you can use only some in a later step.

```suggestion
        tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(processed_rds), path(metadata_json), emit: data
        path qc_report, emit: report
```

Then you can substitute in at the next step `sce_qc_report.out.data` and not be passing around the QC report. In this case it just changes the comment and `it[5]` becomes `it[4]` in the `flatMap{}`, but is maybe a bit more flexible?

jashapiro (commented):
This all looks good, but can you remove the `publishDir` from `cluster_sce` as well, now that publishing of the rds files happens with `sce_qc_report`? 

I also had an optional suggestion to split out the QC report file from the part of the channel that goes downstream, but that one is really up to you.

jashapiro (approved):
LGTM!

Issue #415, opened 2023-08-25
Add cell type workflow into main workflow


allyhawkins:
Right now we have a separate workflow specifically for cell typing. We will need to add the workflow into the main workflow.

allyhawkins:
Just noting that when we do add the cell typing into the main workflow, we should include a parameter for whether or not cell typing should be performed. By default we don't want to do cell typing on every sample that's run through the workflow.

sjspielman:
Question for discussion: What's our preferred approach for passing in _cell type metadata_? We've been using a [project-specific file](https://github.com/AlexsLemonade/ScPCA-admin/blob/a13a93fb7cdacd64f8162affc45994265ff00c90/sample_info/celltype_annotation/scpca-project-celltype-metadata.tsv), but as we don't _require_ a project id to be provided to the overall workflow, I think it would make sense to do one of the following:

- Modify `scpca-library-metadata.tsv` to have expected cell type columns. This way everything just kind of works, but we'd have to have more (optional!) columns. 
- Retain a separate metadata file for cell typing, but also toss library ids and/or run ids into that file, so file would have more rows.

We could also do something fancier and map up which run ids are associated with the project ids in the cell type file, but I do think it's worth it to be more explicit

allyhawkins:
You shouldn't have to do either. I had previously had it in the workflow and all we need to do is tell the main workflow to run the `add_celltypes.nf` workflow. In `add_celltypes.nf` we add in all the library/ project/sample info that we need and filter to the relevant samples based on what's provided. 

I also would still keep `add_celltypes.nf` so we can run it independently like it is. 

This is what I had in there before:
https://github.com/AlexsLemonade/scpca-nf/commit/da47bd7d34f1437470466fbad804afecb7061669

sjspielman:
> This is what I had in there before:

I had forgotten it was once there..! Thanks for the reminder

jashapiro:
Link to fig jam: https://www.figma.com/file/9zH1gXlz9ZlxVrndN4Wy54/cell-type-workflow?type=whiteboard&node-id=0-1&t=fKgV0qjgpXfOkmFo-0

Issue #462, opened 2023-09-15
Ensure SingleR exports a TSV file


sjspielman:
From #446:

> In doing this we will want a way to track the reference versions used for both cell typing methods and only skip when both the output files exist and the reference version matches.

Currently, it's not easy to track whether `SingleR` has been run, since the cell type annotations are directly added into the SCE object. We should update `SingleR` cell typing to first export a TSV file, and then add those contents into the SCE in a second step.

At first, I saw two ways we could do this:
- Add another argument to `classify_SingleR.R` with the TSV file to export
- Create a separate process that first performs cell typing and exports a TSV. Then, have a process which runs an R script to add that information into the SCE object.

I was originally thinking to do the first option, since it's very simple! But, the second option has the benefit of mirroring how we handle `CellAssign`. That said, a whole process for just reading a TSV file and adding to colData seems like a lot of unnecessary S3 file transfers.

So, now I'm thinking we still have two _scripts_ (make a TSV and then add to SCE), but call them _both_ from the same process:
- The process should keep the same name `classify_singleR`
- Then, this process will call two scripts: `classify_SingleR.R` which will export a TSV file of annotations, and `annotate_SingleR.R` which will add to SCE

Moreover, maybe we should take this opportunity and do the same with `CellAssign` before porting things over into `main.nf`. For this, we could have a single process called `classify_cellassign` that runs two scripts: `predict_cellassign.py` and then `annotate_cellassign.R`. Since `CellAssign` takes such a long time, any small time/memory savers like this can't hurt. One change we'd need to make is, we'd have to pass `classify_singler.out` into the `classify_cellassign` process to ensure everything makes it into the same SCE.

Edit: bonus, we probably could use the same R script for add annotations from a TSV, using arguments and `if`s to handle the differences. Give it a general name like `add_annotations_to_sce.R`.

Tagging folks for discussion! @allyhawkins @jashapiro

jashapiro:
I'm not sure I understand the use case, but if we just want to track whether SingleR has been performed for the purposes of the workflow, the easiest way would be to add a metadata field (`meta.has_singler` or something similar). Combining the CellAssign steps into a single process does make some sense to me, but if we are not planning to use the tsv file for anything, I don't think there is any reason to make it.

sjspielman:
[...]

PR #464, opened 2023-09-19
Part 1: Cell typing to main workflow

sjspielman:
Towards #415 

This PR adds the first steps towards adding cell typing into the main workflow. 

First, note 2 tiny recreational revisions:
 - I updated `config/reference_paths.config` & `build-celltype-ref.nf` to rename the param `singler_ref_dir` to `singler_references_dir`, to match the `CellAssign` name. 
- There was a "double comment" (`////`) in `af-features.nf` that I randomly came across, and now it is 1 comment (`//`)

Now, the main attraction. 
- We have some new params, `repeat_celltyping` and `skip_celltyping` both set to `false` as default. For now, I am only using `skip_celltyping`, planning to get back to `repeat_celltyping` in next PR(s).
- The anndata conversion process had to be moved up, since we need the processed hdf5 file for input to `CellAssign`. 
  - Some things in the anndata step are a bit TBD; For now, I'm emitting `move_normalized_counts.out` from it to facilitate heading into cell typing. I know there are other changes planned for anndata (#463), so I left a big "PENDING" comment in that file!
- To prepare for celltyping, I create `celltype_ch` and _branch_ into libraries that will get celltyped and those we'll skip, for now using only `param.skip_celltypes`. This channel only _needs_ meta, processed hdf5s, and processed rds, but to have an easier time collecting all the RDS files back up after cell typing for the QC report, I've also included unfiltered & filtered rds. These files get to come along for the ride throughout celltyping. Originally I did not do this, but after chatting with Ally for a little bit, we decided that this could be cleaner than having a mapping & joining extravaganza back in `main.nf`.
- The celltyping workflow looks more or less the same as it looked on its own, with a couple differences:
  - The R scripts that add celltypes to RDS files now only expect 1 SCE file, which is both the input & output file (`processed_rds`).
  - `meta.has_singler` and `meta.cell_assign` get created in the process that adds cell types to the SCE
  - As mentioned, `unfiltered_rds` and `filtered_rds` go along for the ride to facilitate heading into the QC report next
- After celltyping, mix back up, and off to the QC. I admit, I named this channel `final_sce_ch` as sort of a joke (you're all lucky it's not `final_final_sce_ch`), but I mean, what else should I call it? ;)

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -56,6 +56,14 @@ if (!sample_metafile.exists()) {
   param_error = true
 }
 
+// check for params.celltype_project_metafile, only if we are celltyping
+if (!params.skip_celltyping || params.repeat_celltyping) {
+  if (!file(params.celltype_project_metafile).exists()) {
+    log.error("The 'celltype_project_metafile' file '${params.celltype_project_metafile}' can not be found.")
+  param_error = true
+  }
+}
+
 resolution_strategies = ['cr-like', 'full', 'cr-like-em', 'parsimony', 'trivial']
 if (!resolution_strategies.contains(params.af_resolution)) {
   log.error("'af_resolution' must be one of the following: ${resolution_strategies}")
--- a/main.nf
+++ b/main.nf
@@ -268,7 +268,7 @@ workflow {
    // Bring back the RDS files to create a tuple of:
    // BUT THE METAS WONT MATCH, WILL IT MATTER?
    // [meta, unfiltered, filtered, processed]
-   .join(cluster_sce.out.map{[it[0]["library_id"]] + it, by: 0, failOnDuplicate: true, failOnMismatch: true)
+   .join(cluster_sce.out.map{[it[0]["library_id"]] + it}, by: 0, failOnDuplicate: true, failOnMismatch: true)
    .map{it.drop(1)}
 
   
--- a/main.nf
+++ b/main.nf
@@ -220,51 +220,41 @@ workflow {
 
   // Cluster SCE and export RDS files to publishDir
   cluster_sce(post_process_sce.out)
-  
+
   // convert SCE object to anndata
   // do this for everything but multiplexed libraries
   anndata_ch = cluster_sce.out
     .filter{!(it[0]["library_id"] in multiplex_libs.getVal())}
   sce_to_anndata(anndata_ch)
-  
-  // Prepare for and perform cell type annotation 
+
+  // Prepare for and perform cell type annotation
+  // Create celltype_ch as:
+  // [[meta], processed_hdf5, unfiltered_rds, filtered_rds, processed_rds]
   celltype_ch = sce_to_anndata.out
     .map{[
       it[0], // meta
       it[1]  // processed hdf5
     ]}
     // join with tuple meta, processed rds to create
-    // [meta, processed hdf5, processed rds]
-    .join(cluster_sce.out.map{[it[0], it[3]]})
-   .branch{
-     // TODO: 
-     // later, more checks can go here as needed
-     // do we want a column in the library metadata file for which celltype methods we intend to do?
-     repeat_celltyping: (params.repeat_celltyping
-                         || !it[0]['has_singler']
-                         || !it[0]['has_cellassign'])
-     skip_celltyping: true
-   }
- 
-  annotate_celltypes(celltype_ch.repeat_celltyping)
+    // [meta, processed hdf5, unfiltered_rds, filtered_rds, processed rds]
+    .join(cluster_sce.out, by: 0, failOnDuplicate: true, failOnMismatch: true)
+    .branch{
+      skip_celltyping: params.skip_celltyping
+      perform_celltyping: true
+    }
 
-  // combine celltyped and skipped-celltyped channels
-  all_sce_ch = annotate_celltypes.out.mix(celltype_ch.skip_celltyping) 
-    .map{[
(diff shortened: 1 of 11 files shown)


sjspielman:
more stub things incoming... 🎢

modules/classify-celltypes.nf
```diff
@@ -4,25 +4,20 @@ process classify_singleR {
...
         --label_name ${params.singler_label_name} \
         --seed ${params.seed} \
         --threads ${task.cpus}
       """
-    stub:
-      annotated_rds = "${meta.library_id}_annotated.rds"
-      """
-      touch "${annotated_rds}"
-      """
+
```
allyhawkins, line 20: I think you still need a stub here, otherwise tests will fail: 
```suggestion
  stub:
  """
  # nothing to do since files don't move
  """
```

modules/classify-celltypes.nf
```diff
@@ -57,57 +55,63 @@ process classify_cellassign {
...
     """
     classify_cellassign.R \
-      --input_sce_file ${input_rds} \
-      --output_sce_file ${annotated_rds} \
+      --sce_file ${processed_rds} \
       --cellassign_predictions ${cellassign_predictions} \
       --reference_name ${ref_name}
-
     """
 }
```
allyhawkins, line 69: We probably need one here too.

jashapiro:
Before a careful review, some high level thoughts.

Anndata objects are the largest objects we deal with: it is better if we can avoid copying them around at all, so I would suggest we not move the anndata conversion up. Instead, I would make the first step of the CellAssign process use the script that we have to do the conversion _only_ for the file that needs to be converted for that step. Depending how we add annotations in and things we might do later, we might have to go back to rds files, which makes things confusing. Basically, converting the files is probably cheaper than moving them around, so if we can minimize that, we will come out ahead.

nextflow.config
```diff
@@ -52,6 +52,10 @@ params {
   cluster_weighting = "jaccard" // default weighting scheme for graph-based clustering
   nearest_neighbors = 20 // default nearest neighbors parameter for graph-based clustering
 
+  // Cell type annotation options
+  skip_celltyping = false // by default, we want to perform celltyping
```
allyhawkins, line 56:
```suggestion
  skip_celltyping = true // by default, we don't want to perform celltyping
```
I actually don't think we want to set this default to true. We want to skip cell typing unless we say to do it.

allyhawkins (commented):
Just a few comments to get started, but mostly looks good!

sjspielman:
My guess at this point re-stub is that this failure might be "expected" given that we have booted multiplex from some of the post-processing?

main.nf
```diff
@@ -221,15 +220,42 @@ workflow {
...
+    }
+
+  annotate_celltypes(celltype_ch.perform_celltyping)
+
+  final_sce_ch = celltype_ch.skip_celltyping
+    .map{meta, processed_hd5f, unfiltered_rds, filtered_rds, processed_rds -> tuple(meta,
+                                                                                    unfiltered_rds,
+                                                                                    filtered_rds,
+                                                                                    processed_rds
+                                                                                    )}
```
jashapiro, line 260: This is what I meant by not doing the AnnData conversion early: the processed AnnData we produced didn't have the cell types in it, and won't get them added here, so they won't be complete.

jashapiro:
One more high level thought:

Going back to https://github.com/AlexsLemonade/scpca-nf/issues/462: Do we maybe want the output of the cell typing steps to be a table of cell types that we can mix in at any point?

My thought is that if we have done cell typing but we are running the workflow again there are two options: one is to assume that any upstream changes would invalidate the previous cell typing, and we need to do it again. Another is that we can, in some situations, be "okay" with the old cell typing, and we just want to mix that back in. I don't think we would ever want to do that in "production", but it could be useful for tests.

But if we never want to do that, then we could probably want to skip _everything_ before cell typing in the event that we have cell typing results. 

All of this to say, I think we need to have a bit more discussion before we get the code fully ironed out. I would recommend that discussion happen with flowcharts and diagrams rather than code, because it is much easier to set a plan that way than to try to backsolve from a complex nextflow file.

sjspielman:
> All of this to say, I think we need to have a bit more discussion before we get the code fully ironed out. I would recommend that discussion happen with flowcharts and diagrams rather than code, because it is much easier to set a plan that way than to try to backsolve from a complex nextflow file.

:100:

allyhawkins:
> Going back to https://github.com/AlexsLemonade/scpca-nf/issues/462: Do we maybe want the output of the cell typing steps to be a table of cell types that we can mix in at any point?

> My thought is that if we have done cell typing but we are running the workflow again there are two options: one is to assume that any upstream changes would invalidate the previous cell typing, and we need to do it again. Another is that we can, in some situations, be "okay" with the old cell typing, and we just want to mix that back in. I don't think we would ever want to do that in "production", but it could be useful for tests.

This is why I originally suggested outputting the tsv. Sorry I didn't make it clear that the tsv would be the actual cell type annotations. So we output the cell type annotations from both SingleR and then CellAssign and have a separate process that actually incorporates them into the SCE object (as mentioned in https://github.com/AlexsLemonade/scpca-nf/issues/462#issuecomment-1723475262). I would agree that it does sound like mapping out the new additions would be helpful.

sjspielman:
Closing this out to take a different approach, coming soon, to including celltyping in the main workflow.

PR #466, opened 2023-09-20
Document AnnData output

### LINKED ISSUE #367: Document addition of AnnData output objects
Now that the workflow will create both AnnData and SCE objects, we will want to update the output section of the docs to include the new output files - https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md#output-files. 

We probably also want to update the main description in the README to note that the workflow produces both AnnData and SCE objects.

allyhawkins:
Closes #367 

This PR updates the documentation in this repo to reflect that AnnData objects are now output by the workflow. 

- I updated the main README to note that output is both SCE objects and AnnData. 
- I modified the outputs section in the external instructions to mention that files produced will be both the rds and the hdf5 files. I also updated the expected file output structure and grouped by file type to make things easier to read. 
- I did not include `_adt.hdf5` files in the output structure, but I do mention them in the text. Do we think that's sufficient?
- While working on this I made some changes in the docs to reflect our change to use ADT when talking about CITE-seq.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The workflow processes fastq files from single-cell and single-nuclei RNA-seq sa
 The workflow outputs gene expression data in two formats: as [`SingleCellExperiment` objects](https://www.bioconductor.org/packages/release/bioc/html/SingleCellExperiment.html) and as [`AnnData` objects](https://anndata.readthedocs.io/en/latest/).
 Reads from samples are aligned using selective alignment, to an index with transcripts corresponding to spliced cDNA and to intronic regions, denoted by alevin-fry as `splici`.
 These matrices are filtered and additional processing is performed to calculate quality control statistics, create reduced-dimension transformations, and create output reports.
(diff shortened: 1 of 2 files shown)


sjspielman (commented):
This looks fine to me overall, just a couple very minor phrasing comments. 

My main comment is about the multiplexed libraries and AnnData, since I know that is planned but on hold, right? Maybe we should open an issue to circle back to docs once that is finalized so we can add in the right docs for however that pans out?

allyhawkins:
Thanks for the suggestions. I incorporated the ones you added and then also added cell hashing to the main readme. I think that just hadn't been updated since we added it in so good catch! 
I also added a note to the AnnData output description that no AnnData objects are provided for multiplexed libraries. 
I'm not sure if we really are going to start exporting cell hashed libraries as AnnData just yet, so I'm going to just add a note to the existing issue to make sure we update documentation accordingly if we do make that change.

sjspielman (approved):
🚀

PR #467, opened 2023-09-21
Account for sample metadata as required input to scpca-nf in documentation

### LINKED ISSUE #401: Document sample metadata as input
In #400 we are adding the sample metadata file as a parameter to the workflow. We will need to update the external documentation to include the sample metadata that is required. When we do that we will want to note how to fill in the table when sample information is unknown.

allyhawkins:
Closes #401 
Stacked on #466 

This PR adds documentation throughout the repo documenting the addition of the sample metadata as required input to the workflow. 

- The main change I made was to differentiate between the existing metadata in the documentation, which I now refer to as `run metadata`, and the new metadata for samples, which I refer to as `sample metadata`. I tried to make it clear throughout which is which, but please let me know if I missed any places that still may be confusing! 
- I added making the sample metadata file to the steps on the main readme page and the steps in the overview at the beginning of the external instructions. 
- There is now a new section for `Creating the sample metadata file` in the external instructions. I noted that the only required column is `scpca_sample_id`, and users should include any relevant information about each unique sample. I considered including a table of columns and contents as we have for the run metadata, but that seemed unnecessary with just one column. Alternatively, we could provide suggested or optional columns? 
- I made updates throughout the docs to ensure we account for sample metadata during configuration and defining parameters.
- I added an example sample metadata table and included buttons for viewing rather than links for both run and sample metadata. I wasn't sure if we should show optional columns and how much detail we should provide in the example sample metadata. I'm curious about what reviewers think about what I did end up using in the example. 
- The template config is now updated to include the sample metadata parameter.


examples/README.md
```diff
@@ -16,14 +16,20 @@ We recommend using the example 10X dataset from a [human glioblastoma donor that
...
 
 
-Following download and unzipping of the fastq files, you will need to create a tab-separated values metadata file that looks like the following:
+Following download and unzipping of the fastq files, you will need to create a tab-separated values **run** metadata file that looks like the following:
 
-| scpca_run_id | scpca_library_id | scpca_sample_id | scpca_project_id | technology | seq_unit | sample_reference | files_directory | submitter_cell_types_file |
-| ------------ | ---------------- | --------------- | ---------------- | ---------- | -------- | ---------------- | --------------- | ------------------------ |
-| run01 | library01 | sample01 | project01 | 10Xv3.1 | cell | Homo_sapiens.GRCh38.104 | /path/to/example_fastq_files | /path/to/annotated_cell_types_file
+| scpca_run_id | scpca_library_id | scpca_sample_id | scpca_project_id | technology | assay_ontology_term_id | seq_unit | sample_reference | files_directory | submitter_cell_types_file |
+| ------------ | ---------------- | --------------- | ---------------- | ---------- | -----------------------|-------- | ---------------- | --------------- | ------------------------ |
```
jashapiro, line 22: Most minor alignment update.
```suggestion
| ------------ | ---------------- | --------------- | ---------------- | ---------- | ---------------------- | -------- | ---------------- | --------------- | ------------------------ |
```

examples/README.md
```diff
@@ -16,14 +16,20 @@ We recommend using the example 10X dataset from a [human glioblastoma donor that
...
 
 
-Following download and unzipping of the fastq files, you will need to create a tab-separated values metadata file that looks like the following:
+Following download and unzipping of the fastq files, you will need to create a tab-separated values **run** metadata file that looks like the following:
 
-| scpca_run_id | scpca_library_id | scpca_sample_id | scpca_project_id | technology | seq_unit | sample_reference | files_directory | submitter_cell_types_file |
-| ------------ | ---------------- | --------------- | ---------------- | ---------- | -------- | ---------------- | --------------- | ------------------------ |
-| run01 | library01 | sample01 | project01 | 10Xv3.1 | cell | Homo_sapiens.GRCh38.104 | /path/to/example_fastq_files | /path/to/annotated_cell_types_file
+| scpca_run_id | scpca_library_id | scpca_sample_id | scpca_project_id | technology | assay_ontology_term_id | seq_unit | sample_reference | files_directory | submitter_cell_types_file |
```
jashapiro, line 21: Do we require the `assay_ontology_term_id` now?  I realized it is not in the stub run file, and it should probably get a `parseNA` wrapper in the workflow (which may also require an update in the process that uses it to handle a potentially `null` value).

README.md
```diff
@@ -26,8 +26,9 @@ For all other users, `scpca-nf` can be set up for your computing environment wit
 To run `scpca-nf` on your own samples, you will need to complete the following steps:
 
 1. [Organize your files](external-instructions.md#file-organization) so that each folder contains fastq files relevant to a single sequencing run.
-2. [Prepare a metadata file](external-instructions.md#prepare-the-metadata-file) with one row per library containing all information needed to process your samples.
-3. Set up a [configuration file](external-instructions.md#configuration-files), including the [definition of a profile](external-instructions.md#setting-up-a-profile-in-the-configuration-file), dictating where nextflow should execute the workflow.
+2. [Prepare a run metadata file](external-instructions.md#prepare-the-run-metadata-file) with one row per library containing all information needed to process your samples.
+3. [Prepare a sample metadata file]() with one row per sample containing any relevant metadata about each sample (e.g., diagnosis, age, sex, cell line).
```
jashapiro, line 30:
```suggestion
3. [Prepare a sample metadata file](external-instructions.md#prepare-the-sample-metadata-file) with one row per sample containing any relevant metadata about each sample (e.g., diagnosis, age, sex, cell line).
```

examples/README.md
```diff
@@ -16,14 +16,20 @@ We recommend using the example 10X dataset from a [human glioblastoma donor that
...
+| ------------ | ---------------- | --------------- | ---------------- | ---------- | -----------------------|-------- | ---------------- | --------------- | ------------------------ |
+| run01 | library01 | sample01 | project01 | 10Xv3.1 | EFO:XXX | cell | Homo_sapiens.GRCh38.104 | /path/to/example_fastq_files | /path/to/annotated_cell_types_file
 
 Be sure to enter the **full** path to the directory containing the fastq files in the `files_directory` column, and similarly the full path for the `submitter_cell_types_file` column.
 
+You will also need to create a tab-separated values **sample** metadata file that looks like the following:
```
jashapiro, line 27: I do think we should have a bit more information here. I know the bulk will be in the other document, but we should note that the column values should match those in the run metadata, and maybe show it here with at least two columns?

external-instructions.md
```diff
@@ -134,14 +135,37 @@ The following columns may be necessary for running other data modalities (CITE-s
...
+## Prepare the sample metadata file
+
+Using `scpca-nf` requires a sample metadata file as a TSV (tab separated values) file, where each unique sample that is present in the `scpca_sample_id` column of the run metadata file is a row, and columns contain any relevant sample metadata (e.g., diagnosis, age, sex, cell line).
+For each library that is processed, the corresponding sample metadata will be added to the `SingleCellExperiment` and `AnnData` objects output by the workflow (see the section on [Output files](#output-files)).
+
+At a minimum, all sample metadata tables must contain a column with `scpca_sample_id` as the header.
+The contents of this column should contain all unique sample ids that are present in the `scpca_sample_id` column of the run metadata file.
+
+We encourage you to use standard terminology, such as ontology terms, to describe samples when possible.
+There is no limit to the number of columns allowed for the sample metadata, and you may include as many metadata fields as you please.
```
jashapiro, line 152: I do think giving some suggested columns would be useful here. `age`, `sex`, `tissue` might be a good start? Others you think are common/relevant?

jashapiro (commented):
This looks good; I had a few smallish suggestions. I didn't check that every reference to the run data was there though!

allyhawkins:
> Do we require the assay_ontology_term_id now? I realized it is not in the stub run file, and it should probably get a parseNA wrapper in the workflow (which may also require an update in the process that uses it to handle a potentially null value).

This is a great question. I think we should require it as we are trying to use standard terms, however it's not necessary for the workflow to run. I did make modifications to handle an `NA` value in the case there is not an ontology id provided. I also updated the documentation to state that users should include that column in the run metadata. 

> I do think giving some suggested columns would be useful here. age, sex, tissue might be a good start? Others you think are common/relevant?

I added a statement with suggested columns and then also emphasized the line about the required sample id column. Let me know what you think of this approach? Or should we make a table? 

> I do think we should have a bit more information here. I know the bulk will be in the other document, but we should note that the column values should match those in the run metadata, and maybe show it here with at least two columns?

I expanded on this section a little bit, adding some more information about what is a sample metadata file. 
I also showed some example columns to include with values for the example data that we are using from 10X.

bin/generate_unfiltered_sce.R
```diff
@@ -51,6 +51,7 @@ option_list <- list(
   make_option(
     opt_str = c("--assay_ontology_term_id"),
     type = "character",
+    default = NULL,
```
jashapiro, line 54: In addition to this change, I think we need to modify the following line in `modules/sce-processing.nf` as follows:
```
          --assay_ontology_term_id ${meta.assay_ontology_term_id} \ 
```

to 

```
          ${meta.assay_ontology_term_id? "--assay_ontology_term_id ${meta.assay_ontology_term_id}" : ""} \ 
```



Note that this appears in both `make_unfiltered_sce` and `make_merged_unfiltered_sce`

external-instructions.md
```diff
@@ -65,13 +66,14 @@ nextflow run AlexsLemonade/scpca-nf \
 ```
 
 Where `<path to config file>` is the **relative** path to the [configuration file](#configuration-files) that you have setup and `<name of profile>` is the name of the profile that you chose when [creating a profile](#setting-up-a-profile-in-the-configuration-file).
-This command will pull the `scpca-nf` workflow directly from Github, using the `v0.5.4` version, and run it based on the settings in the configuration file that you have defined.
```
jashapiro, line 68: good catch on this fix

jashapiro (commented):
This looks good to me, but I think we need one more update to the workflow as part of it?

bin/generate_unfiltered_sce.R
allyhawkins: Thanks for catching this. Done in https://github.com/AlexsLemonade/scpca-nf/pull/467/commits/3347c246542cf18143c84e16530fb06dcb59f494

jashapiro (approved):
LGTM!

PR #468, opened 2023-09-22
Include AnnData files in example output

### LINKED ISSUE #369: Update example outputs to reflect AnnData
In the `examples` folder we discuss an example run of the workflow and include a link to example outputs. Once we have completed the addition and modification of the AnnData objects, we will need to update the link to include the AnnData in the output as well.

allyhawkins:
Closes #369 

This PR makes a few internal documentation updates for how to update example output files. As part of this I updated the current example output link to contain the example output with AnnData. 

- I made sure the example metadata files in `s3://scpca-references/example-data` contained the correct columns and have the correct path to the example files. 
- I ran the `development` branch of the workflow. When doing that I found an error in reading in the sample metadata where we expect columns for both `upload_date` and `participant_id`. Since those are not required columns in the sample metadata (see #467) and we only use them internally, I wrapped the modifications of those columns in `if` statements. After doing that the workflow was able to run successfully. 
- I downloaded the new outputs, zipped them up, and re-uploaded them to `s3://scpca-references/example-data/scpca_out.zip` and set the files to public. 
- I made a few changes to the internal documentation to hopefully be more detailed on how the data was generated. 

One thing to note is that the development branch currently has output with clusters, but no cell types yet. So cluster annotations are in the output too, but I think that's okay right now. We may want to re-generate the examples with cell types once that is in the workflow.

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -164,13 +164,13 @@ sample_metadata_df <- readr::read_tsv(opt$sample_metadata_file) |>
   dplyr::mutate(library_id = opt$library_id) 
 
 if("upload_date" %in% colnames(sample_metadata_df)){
-  sample_metadata_df -> sample_metadata_df |>
+  sample_metadata_df <- sample_metadata_df |>
     # remove upload date as we don't provide this on the portal
     dplyr::select(-upload_date)
 }
 
 if("participant_id" %in% colnames(sample_metadata_df)){
-  sample_metadata_df -> sample_metadata_df |>
+  sample_metadata_df <- sample_metadata_df |>
     # rename to donor id for czi compliance
     dplyr::rename("donor_id" = "participant_id")
 }
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -161,11 +161,19 @@ sample_metadata_df <- readr::read_tsv(opt$sample_metadata_file) |>
   dplyr::rename("sample_id" = "scpca_sample_id") |>
   # add library ID as column in sample metadata
   # we need this so we are able to merge sample metadata with colData later
-  dplyr::mutate(library_id = opt$library_id) |>
-  # remove upload date as we don't provide this on the portal
-  dplyr::select(-upload_date) |>
-  # rename to donor id for czi compliance
-  dplyr::rename("donor_id" = "participant_id")
+  dplyr::mutate(library_id = opt$library_id) 
+
+if("upload_date" %in% colnames(sample_metadata_df)){
+  sample_metadata_df -> sample_metadata_df |>
+    # remove upload date as we don't provide this on the portal
+    dplyr::select(-upload_date)
+}
+
+if("participant_id" %in% colnames(sample_metadata_df)){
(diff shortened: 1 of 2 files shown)


bin/generate_unfiltered_sce.R
```diff
@@ -161,11 +161,19 @@ sample_metadata_df <- readr::read_tsv(opt$sample_metadata_file) |>
...
+  sample_metadata_df -> sample_metadata_df |>
+    # remove upload date as we don't provide this on the portal
+    dplyr::select(-upload_date)
+}
+
+if("participant_id" %in% colnames(sample_metadata_df)){
+  sample_metadata_df -> sample_metadata_df |>
+    # rename to donor id for czi compliance
+    dplyr::rename("donor_id" = "participant_id")
+}
```
sjspielman, line 176: Your assignment operators appear to be pointing the wrong way here? I'm not sure I understand how this worked?

I downloaded `scpca_out` and in unfiltered SCE I see this - 
```
> metadata(sce)$sample_metadata
  sample_id optional_column_1 optional_column_2 library_id
1  sample01        diagnosis1              age1  library01
```

But still, you probably mean:

```suggestion
if("upload_date" %in% colnames(sample_metadata_df)){
  sample_metadata_df <- sample_metadata_df |>
    # remove upload date as we don't provide this on the portal
    dplyr::select(-upload_date)
}

if("participant_id" %in% colnames(sample_metadata_df)){
  sample_metadata_df <- sample_metadata_df |>
    # rename to donor id for czi compliance
    dplyr::rename("donor_id" = "participant_id")
}
```

bin/generate_unfiltered_sce.R
```diff
@@ -161,11 +161,19 @@ sample_metadata_df <- readr::read_tsv(opt$sample_metadata_file) |>
...
+if("upload_date" %in% colnames(sample_metadata_df)){
+  sample_metadata_df -> sample_metadata_df |>
+    # remove upload date as we don't provide this on the portal
+    dplyr::select(-upload_date)
+}
+
+if("participant_id" %in% colnames(sample_metadata_df)){
+  sample_metadata_df -> sample_metadata_df |>
+    # rename to donor id for czi compliance
+    dplyr::rename("donor_id" = "participant_id")
```
sjspielman, line 175:
```suggestion
    dplyr::rename(donor_id = participant_id)
```

sjspielman (commented):
Looks good overall, but I'm not sure I fully understand what went out with the assignment operator situation. Maybe worth rerunning the workflow?


> So cluster annotations are in the output too, but I think that's okay right now. We may want to re-generate the examples with cell types once that is in the workflow.

Seems fine to me! If that issue isn't open for later, let's open it.

bin/generate_unfiltered_sce.R
allyhawkins: Yikes this is embarrassing. Thank you for catching it! Those columns aren't present in the example sample metadata because they aren't required, so it didn't actually run that code when I ran the example metadata.

allyhawkins:
Issue filed and assignment operators fixed 🤦‍♀️

sjspielman (approved):
👍

PR #471, opened 2023-09-27
Celltyping to main: SingleR classification

sjspielman:
Towards #415 

This PR begins the process of porting celltyping into the main workflow. For now, I have paired existing code down to only focus on running SingleR and preserving logic in the main workflow. I updated the singler R script and nf process to perform cell typing and export a TSV and RDS file for annotations and the full result, respectively, into checkpoints.

I'm opening this as a _Draft_ because it's still failing the stub. After much hunting down, I still am not seeing where the logic has gone awry... which means it's time for my eyes to take a break and for someone else's fresh eyes to see if you can spot it!
I'm requesting @jashapiro for review since he'll be the one reviewing for the rest of the sprint, but feel free to hunt down logic bugs if you are so moved, @allyhawkins!


modules/classify-celltypes.nf
```diff
@@ -1,76 +1,39 @@
...
+      singler_annotations_tsv = "${meta.library_id}_singler_annotations.tsv"
+      singler_full_results = "${meta.library_id}_singler_full_results.rds"
       """
-      touch "${annotated_rds}"
+      touch "${singler_annotations_tsv}"
+      touch "${singler_full_results}"
       """
 }
 
-process predict_cellassign {
```
allyhawkins, line 28: I think we can leave this in right now and just remove the lines in the workflow that call this process. It might make it easier for later when you go to edit this.

main.nf
```diff
@@ -218,11 +218,28 @@ workflow {
...
+  annotate_celltypes( cluster_sce.out.map{[ it[0], it[3] ]} ) // only send in meta, processed sce
+
+
   // generate QC reports
-  sce_qc_report(cluster_sce.out, report_template_tuple)
+  qc_report_ch = annotate_celltypes.out.map{[ it[0], it[1] ]} // for now, this is meta and processed sce
+    // bring back the other 2 SCEs
+    .concat(
+      // unfiltered, filtered
+      cluster_sce.out.map{[ it[1], it[2] ]}
```
allyhawkins, line 234: I think I would change this to join on the library id, how we do in other places. 
```suggestion
  qc_report_ch = annotate_celltypes.out.map{[ it[0]["library_id], it[0], it[1] ]} // for now, this is meta and processed sce
    // bring back the other 2 SCEs
    .join(
      // unfiltered, filtered
      cluster_sce.out.map{[ it[0]["library_id"], it[1], it[2] ]}, by: 0)
```
I think you could also be more explicit when mapping, but just to give you an idea.

allyhawkins (commented):
I took a quick look and maybe have a suggestion.

test/celltypes/cellassign_ref_file.txt
jashapiro, line 1: These files should probably be placed in `test/references/stub/celltype` to better match other stubs/the organization in s3://scpca-references. Similarly, I think is better to match the expected file extensions, even if they don't really matter; it is another form of documentation.

It is also possible that this is the cause of the stub failures, as the code may be expecting to find them there.

modules/classify-celltypes.nf
jashapiro: I agree! It is okay to have processes that we are not _currently_ using if we expect to use them again soon

main.nf
jashapiro: To add a bit here: the reason to do this is that we might end up modifying ordering when we branch and rejoin. I might also add in `failOnDuplicate:true, failOnMismatch: true`, since we _do_ expect 1:1 matches between the two channels.

config/profile_stub.config
```diff
@@ -12,6 +12,7 @@ params{
   cellhash_pool_file = 'test/stub-pools.tsv'
   outdir = "test/output"
   run_ids = "All"
+  celltype_project_metafile = 'test/celltypes/stub-celltype-project-metadata.tsv'
```
jashapiro, line 15: Moving started just for organization, but also suggesting a slightly different location for this file. (Which we may want to move in the admin folder as well. I'm not sure I understand the current organization of the celltyping files there... I'm not sure why the reference files/scripts are nested in `sample_info`; seems like they should be in a different place and possibly not in `scpca-admin` at all.)
```suggestion
  celltype_project_metafile = 'test/stub-celltype-project-metadata.tsv'
  outdir = 'test/output'
  run_ids = 'All'
```

test/celltypes/cellassign_ref_file.txt
jashapiro: Note that many relative paths are defined in `reference_paths.config`:
https://github.com/AlexsLemonade/scpca-nf/blob/dc259d4e7f56fb42db72b24d5fe10be7daf40d06/config/reference_paths.config#L13-L21

modules/classify-celltypes.nf
sjspielman: I'll be honest, I deleted them from cluttering my eyeballs :) Can restore though!

jashapiro: Code folding is your friend!

jashapiro (commented):
I apparently didn't submit all of my suggestions. Sorry about that!

modules/classify-celltypes.nf
```diff
@@ -1,76 +1,39 @@
...
     input:
         tuple val(meta), path(processed_rds), path(singler_model_file)
     output:
-        tuple val(meta), path(annotated_rds)
+        tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results)
     script:
-      annotated_rds = "${meta.library_id}_annotated.rds"
+
+      singler_annotations_tsv = "${meta.library_id}_singler_annotations.tsv"
+      singler_full_results = "${meta.library_id}_singler_full_results.rds"
```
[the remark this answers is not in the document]
sjspielman: FYI I'll implement these changes myself since my code is well-diverged at this point from what was last pushed, but I'm on it!

sjspielman:
Ok, next round, **still as a draft** - 

1. `stub` is still failing, though it seems to me the stub aspects are in place, but see item 3 here - 
2. The workflow itself _runs all the way through_ for a sample that _does have_ submitter annotations, so that at least is something!
3. The workflow is _failing_ for a sample that _does NOT have_ submitter annotations with an error about an `NA` file heading into `generate_sce:make_unfiltered_sce`, which makes me think it has to do somehow with the `empty_file` changes? I'm running now with `-r development` to check if that causes the same error.

jashapiro (commented):
I think part of the issue now is that not every project has celltype metadata. I presented a couple options for how to handle that... I'm a bit torn on which is better. See below.

main.nf
```diff
@@ -218,11 +218,29 @@ workflow {
...
+    .join(
+      // unfiltered, filtered
+      cluster_sce.out.map{[ it[0]["library_id"], it[1], it[2] ]}, by: 0, failOnDuplicate:true, failOnMismatch: true
+    )
+    .map{it.drop(1)}
+    // Reorder as sce_qc_report expects
+    .map{meta, processed_rds, unfiltered_rds, filtered_rds -> tuple(meta,
+                                                                    unfiltered_rds,
+                                                                    filtered_rds,
+                                                                    processed_rds)}
```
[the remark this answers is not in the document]
jashapiro: Since the `tuple()` is kind of optional, we can make these even a bit simpler? (This makes a list, but it will get converted to a tuple down the line) 
```suggestion
    .map{meta, processed_rds, unfiltered_rds, filtered_rds ->
         [meta, unfiltered_rds, filtered_rds, processed_rds]}
```

config/profile_stub.config
```diff
@@ -20,6 +21,12 @@ params{
   // path to reference files
   ref_json = "test/stub-refs.json"
   ref_rootdir = 'test/references'
+  
+  // paths to celltype reference files
+  celltype_ref_dir = "${params.ref_rootdir}/stub/celltype"
+  singler_models_dir = "${params.celltype_ref_dir}/singler_models"
+  cellassign_ref_dir = "${params.celltype_ref_dir}/cellassign_references"
```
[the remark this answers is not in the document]
sjspielman: `celltype_ref_dir` seems required to me? (unless we reorganize elsewhere)

In the `stub.config`:
https://github.com/AlexsLemonade/scpca-nf/blob/67e175bdbbfb314e0870a1b832d48001bfa87d9e/config/profile_stub.config#L21

Since celltype references for testing are in `test/references/stub/celltype/`, we need to override this from `references.config`:
https://github.com/AlexsLemonade/scpca-nf/blob/67e175bdbbfb314e0870a1b832d48001bfa87d9e/config/reference_paths.config#L10

Frankly, I wasn't sure how things would get defined without also including `singler_models_dir` and `cellassign_ref_dir`, so I popped them in...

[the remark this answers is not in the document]
jashapiro: Ah you are right... but I think the problem is where I told you to put the reference files! They should be in `test/references/celltype/` to properly mirror the S3 bucket.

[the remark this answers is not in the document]
jashapiro: > Since celltype references for testing are in `test/references/stub/celltype/`, we need to override this from `references.config`:

`references.config` needs an update to match S3 anyway? Something isn't aligned there.

sjspielman:
> I think part of the issue now is that not every project has celltype metadata.

This is what I was thinking as well, but if so, shouldn't `AlexsLemonade/scpca-nf -r development` fail too in that circumstance on such a library? And yet it does not fail. (unless a cache got me? checking again now..)

jashapiro:
> This is what I was thinking as well, but if so, shouldn't `AlexsLemonade/scpca-nf -r development` fail too in that circumstance on such a library? And yet it does not fail. (unless a cache got me? checking again now..)

No, because the development branch doesn't have celltyping in it yet?

config/profile_stub.config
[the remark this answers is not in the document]
sjspielman: Off to the whiteboard to disentangle paths!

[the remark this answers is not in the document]
sjspielman: Looks to me like `references.config` matches S3 fine? 

https://github.com/AlexsLemonade/scpca-nf/blob/e0a863c4122ca837da52405ee6431c4e7d7db528/config/reference_paths.config#L4

& 

https://github.com/AlexsLemonade/scpca-nf/blob/e0a863c4122ca837da52405ee6431c4e7d7db528/config/reference_paths.config#L14-L20

With apologies for the terrible quality, I see this current file structure - 

IMG_2764

So, it seems to me that getting rid of `stub/` nested directory as you suggested  - "They should be in `test/references/celltype/` to properly mirror the S3 bucket." - is all to do here. Am I not seeing something?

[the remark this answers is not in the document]
jashapiro: No, my bad... I was looking at `main`

main.nf
```diff
@@ -218,11 +218,29 @@ workflow {
...
+
+
   // generate QC reports
-  sce_qc_report(cluster_sce.out, report_template_tuple)
+  qc_report_ch = annotate_celltypes.out.map{[ it[0]["library_id"], it[0], it[1] ]} // for now, this is meta and processed sce
+    // bring back the other 2 SCEs
+    .join(
+      // unfiltered, filtered
+      cluster_sce.out.map{[ it[0]["library_id"], it[1], it[2] ]}, by: 0, failOnDuplicate:true, failOnMismatch: true
+    )
```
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/471/commits/aca5d9990413f68785720aa635e7ea0e5a19b85c, but want to leave this unresolved

jashapiro:
Another issue I thought of while reviewing https://github.com/AlexsLemonade/ScPCA-admin/pull/668: Not every library has an assigned project id. This is mostly as a way of designating incomplete projects, but if we are `.combine()`ing on project id here, we will lose those libraries at this step. Which may be fine, but we have to be aware of it when producing the outputs. 

I think we can and should put off handling missing project ids to a separate PR: let's get this working for the easy case of complete data first, then add logic later to handle harder cases of mismatches and missing data.

sjspielman:
> I think we can and should put off handling missing project ids to a separate PR
Oh, :100: this PR is already "full enough." Happily all the stub libraries have projects!

sjspielman:
Noting this error was because, again, I forgot that `scpca-meta.json` was not at its most up-to-date following changes in handling NA files...

```
[null] NOTE: Can't stage file file:///home/ccdl/stephanie/scpca-nf/NA -- file does not exist -- Execution is retried (1)
ERROR ~ Unable to re-submit task `generate_sce:make_unfiltered_sce (1)`
```

I am pretty optimistic that after this, it's finally going to form a permanent memory that with certain workflow changes, you gotta manually update the checkpoint `scpca-meta.json` or redo mapping.

test/stub-celltype-project-metadata.tsv
```diff
@@ -1,2 +1,8 @@
-project_id	singler_model_file	cellassign_ref_file	cellassign_ref_name
+project_id	singler_ref_file	cellassign_ref_file	cellassign_ref_name
```
jashapiro, line 1: Update header to match usage:
```suggestion
scpca_project_id	singler_ref_file	cellassign_ref_file	cellassign_ref_name
```

modules/classify-celltypes.nf
```diff
@@ -80,50 +88,31 @@ workflow annotate_celltypes {
         .splitCsv(header: true, sep: '\t')
         .map{[
           project_id = it.scpca_project_id,
```
jashapiro, line 90: We aren't actually using the names here, so we probably shouldn't set them. If we did want to set them, we should do it with colons, not `=`

```suggestion
         it.scpca_project_id, \\ project id
```

modules/classify-celltypes.nf
```diff
@@ -80,50 +88,31 @@ workflow annotate_celltypes {
...
-          singler_model_file = "${params.singler_models_dir}/${it.singler_ref_file}",
-          cellassign_ref_file = "${params.cellassign_ref_dir}/${it.cellassign_ref_file}",
+          singler_model_file = "${params.singler_models_dir}/${parseNA(it.singler_ref_file)}",
+          cellassign_ref_file = "${params.cellassign_ref_dir}/${parseNA(it.cellassign_ref_file)}",
           // add ref name for cellassign since we cannot store it in the cellassign output
           // singler ref name does not need to be added because it is stored in the singler model
-          cellassign_ref_name = it.cellassign_ref_name
+          cellassign_ref_name = Utils.parseNA(it.cellassign_ref_name)
```
jashapiro, line 95:
```suggestion
          Utils.parseNA(it.cellassign_ref_name)
```

config/profile_stub.config
```diff
@@ -20,6 +21,7 @@ params{
   // path to reference files
   ref_json = "test/stub-refs.json"
   ref_rootdir = 'test/references'
+
```
jashapiro, line 24: Turns out you do need these... I have a thought about how to make them update automatically, but I'm not sure it will work, so better to just do it here for now at least
```suggestion
  
  // cell type references directories
  celltype_ref_dir = "${params.ref_rootdir}/celltype"
  // output from save_singler_refs() process
  singler_references_dir = "${params.celltype_ref_dir}/singler_references"
  // output from train_singler_models() process, and input to classify_singler()
  singler_models_dir = "${params.celltype_ref_dir}/singler_models"
  // output from generating cell assign reference matrices, and input to classify_cellassign()
  cellassign_ref_dir = "${params.celltype_ref_dir}/cellassign_references"
```

modules/classify-celltypes.nf
```diff
@@ -80,50 +88,31 @@ workflow annotate_celltypes {
...
-      // create channel grouped_celltype_ch as: [meta, processed sce object, SingleR reference model]
-      // input processed_sce_channel is [meta, unfiltered, filtered, processed]
-      grouped_celltype_ch = processed_sce_channel
+
+      celltype_input_ch = processed_sce_channel
         .map{[it[0]["project_id"]] + it}
         .combine(celltype_ch, by: 0)
         .map{it.drop(1)} // remove extra project ID
+        
+      celltype_input_ch.view()
```
jashapiro, line 104: This will come out later

jashapiro (commented):
Okay, a lot of the pain was maybe another mislabeled stub file column (which may require updating elsewhere?

But I think there were probably a few other little things to make it all work, which I suggested here too. 

I have a version that works locally, and I think I put everything in as suggestions!

test/stub-celltype-project-metadata.tsv
sjspielman: 😭

sjspielman:
✅ ✅ ✅ ✅

main.nf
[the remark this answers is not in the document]
sjspielman: We're all fixed here, so resolving the comment!

modules/classify-celltypes.nf
```diff
@@ -79,51 +87,32 @@ workflow annotate_celltypes {
...
-          cellassign_ref_file = "${params.cellassign_ref_dir}/${it.cellassign_ref_file}",
-          // add ref name for cellassign since we cannot store it in the cellassign output
-          // singler ref name does not need to be added because it is stored in the singler model
-          cellassign_ref_name = it.cellassign_ref_name
+         // project id
+         it.scpca_project_id, 
+         // singler model file
+         Utils.parseNA(it.singler_ref_file) ? file("${params.singler_models_dir}/${it.singler_ref_file}") : null,
+         // cellassign reference file
+         Utils.parseNA(it.cellassign_ref_file) ? file("${params.cellassign_ref_dir}/${it.cellassign_ref_file}") : null,
```
jashapiro, line 95: I used `null` here, but I think we want to use `empty_file` (which probably needs to be defined in this file)
```suggestion
         file(Utils.parseNA(it.singler_ref_file) ? "${params.singler_models_dir}/${it.singler_ref_file}" : empty_file),
         // cellassign reference file
         file(Utils.parseNA(it.cellassign_ref_file) ? "${params.cellassign_ref_dir}/${it.cellassign_ref_file}" : empty_file),
```
And super-minor, but can we drop the `.txt` from `NO_FILE` while we are at it? Or in a separate little PR.

jashapiro: Actually, I take this back! We are going to want to use logic to not run the process at all if the ref is missing, so we want nulls!

jashapiro (approved):
I'm going to approve this on the assumption that the next PR will add some NA values and logic to skip singleR if a reference is missing.

PR #473, opened 2023-09-28
Remove title from cell metadata and move schema to object metadata

allyhawkins:
In working on the docs for AnnData contents, I noticed that we accidentally added the `title` and the `schema_version` to the cell metadata and not the object metadata. They both should be in `metadata(sce)` before converting to AnnData so they get saved in `uns` in the AnnData object. (see https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#uns-dataset-metadata)

I moved the `schema_version`, but I didn't really like the idea of making a duplicate of the project ID and saving it to `title`. It seems silly and hard to explain in the docs. So because we already have `project_id` saved in the metadata, I'm leaving that as is and then removing title for now. As with the `donor_id` change if we submit to CZI we will need to add an actual `title` later anyways.


bin/sce_to_anndata.R
```diff
@@ -88,6 +84,9 @@ format_czi <- function(sce) {
   # remove sample metadata from sce metadata, otherwise conflicts with converting object
   metadata(sce) <- metadata(sce)[names(metadata(sce)) != "sample_metadata"]
 
+  # add schema version
+  metadata(sce)$schema_version = "3.0.0"
+
```
jashapiro, line 89: I'd move this up to the top of the function. That will hopefully remind us to update both the commend and the metadata element at the same time.

jashapiro (approved):
LGTM, with a suggestion to move the chunk setting the schema version up to the to the top to be near the comment about the schema/url.

I don't think this affects the use of `metadata_to_coldata`, since it looks like that only uses `metadata(sce)$sample_metadata`

PR #475, opened 2023-09-29
Drop txt and move channel merge logic

### LINKED ISSUE #474: Drop .txt extension from NO_FILE.txt
From https://github.com/AlexsLemonade/scpca-nf/pull/471#discussion_r1341572256, do what it says

jashapiro:
Closes #474

I also decided to move the logic of dropping and re-adding the unftilered & filtered sces from the channel into the annotation workflow itself. If we reconsidere whether we want to add annotations to filtered sces, we can do that more easily if the files are already there, and it streamlines `main.nf` a bit to make the flow a bit more clear.

### DIFF
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -119,7 +119,7 @@ workflow annotate_celltypes {
       // TODO update below with output channel results:
       export_channel = processed_sce_channel
         .map{[it[0]["library_id"]] + it}
-        // add add in unfiltered and filtered sce files
+        // add in unfiltered and filtered sce files
         .join(sce_files_channel.map{[it[0]["library_id"], it[1], it[2]]},
               by: 0, failOnMismatch: true, failOnDuplicate: true)
         // rearrange to be [meta, unfiltered, filtered, processed]
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -119,7 +119,7 @@ workflow annotate_celltypes {
       // TODO update below with output channel results:
       export_channel = processed_sce_channel
         .map{[it[0]["library_id"]] + it}
-        // add add in unprocessed and filtered sce files
+        // add add in unfiltered and filtered sce files
         .join(sce_files_channel.map{[it[0]["library_id"], it[1], it[2]]},
               by: 0, failOnMismatch: true, failOnDuplicate: true)
         // rearrange to be [meta, unfiltered, filtered, processed]
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -81,14 +81,17 @@ process classify_cellassign {
 }
 
 workflow annotate_celltypes {
-    take: processed_sce_channel
+    take: sce_files_channel // channel of meta, unfiltered_sce, filtered_sce, processed_sce
     main:
+      // get just the meta and processed sce
+      processed_sce_channel = sce_files_channel.map{[it[0], it[3]]}
+
       // channel with celltype model and project ids
       celltype_ch = Channel.fromPath(params.celltype_project_metafile)
         .splitCsv(header: true, sep: '\t')
         .map{[
          // project id
-         it.scpca_project_id, 
+         it.scpca_project_id,
          // singler model file
          Utils.parseNA(it.singler_ref_file) ? file("${params.singler_models_dir}/${it.singler_ref_file}") : null,
          // cellassign reference file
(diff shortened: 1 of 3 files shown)


modules/classify-celltypes.nf
```diff
@@ -103,16 +106,26 @@ workflow annotate_celltypes {
...
+        .map{meta, processed_rds, singler_model, cellassign_model, cellassign_ref_name ->
+             [meta, processed_rds, singler_model]}
 
       // perform singleR celltyping and export TSV
       classify_singleR(singler_input_ch)
 
-    // temporary during development
-    emit: celltype_input_ch
+      // add back in the unchanged sce files
+      // TODO update below with output channel results:
```
sjspielman, line 119: what do you mean "output channel" here? is this just a placeholder/reminder to update handling output as we continue developing things?

assets/NO_FILE
sjspielman, line 1: Just noting that indeed this is the right thing to do! https://nextflow-io.github.io/patterns/optional-input/

sjspielman (commented):
Overall looks good to me, but I have a couple small organizational thoughts..

- Do we want to define `empty_file` at the top of `main.nf` instead of buried in `sce-processing.nf`? I don't think we'll need to use it for cell typing given the `null`s, though, so this doesn't seem necessary _now_ (maybe later, who knows what features may come).
- What about making a global string variable for `"NO_FILE"` rather than hardcoding that string everywhere. We [don't expect this file to change names](https://nextflow-io.github.io/patterns/optional-input/), but I am always anxious about hardcoding and typos!

modules/classify-celltypes.nf
sjspielman: back to say I apparently glossed over the fact that this is starting with `processed_sce_channel`, so yes reasonable TODO!

jashapiro: Yes, it is a placeholder for when we have the actual annotated sces!

jashapiro:
> Overall looks good to me, but I have a couple small organizational thoughts..
> 
> * Do we want to define `empty_file` at the top of `main.nf` instead of buried in `sce-processing.nf`? I don't think we'll need to use it for cell typing given the `null`s, though, so this doesn't seem necessary _now_ (maybe later, who knows what features may come).
> * What about making a global string variable for `"NO_FILE"` rather than hardcoding that string everywhere. We [don't expect this file to change names](https://nextflow-io.github.io/patterns/optional-input/), but I am always anxious about hardcoding and typos!

I'm nervous about defining it as a global because then we have to do extra work to pass it down to `include`d files: note how we have to explicitly pass some variables down with `addParams:`

One thought I did have was to move some of those variables into an importable lib like `Utils` (or just `Utils`), but I'm not sure whether those libs have access to `${ProjectDir}`, which we need to define the file location.

So I think the best move is to leave things as they are for now!

jashapiro:
> * What about making a global string variable for `"NO_FILE"` rather than hardcoding that string everywhere. We [don't expect this file to change names](https://nextflow-io.github.io/patterns/optional-input/), but I am always anxious about hardcoding and typos!

Also note that any change would require us to add a lot of boilerplate! We are sometimes using `NO_FILE` in the context of a shell script, so we would have to replace `"NO_FILE"` with something like `"${Util.no_file}"`? Which seems way more prone to typos!

sjspielman (approved):
👍

sjspielman:
> We are sometimes using NO_FILE in the context of a shell script, so we would have to replace "NO_FILE" with something like "${Util.no_file}"? Which seems way more prone to typos!

good point!! this would indeed be worse.

modules/classify-celltypes.nf
```diff
@@ -103,16 +106,26 @@ workflow annotate_celltypes {
...
+      export_channel = processed_sce_channel
+        .map{[it[0]["library_id"]] + it}
+        // add add in unprocessed and filtered sce files
```
jashapiro, line 122:
```suggestion
        // add add in unfiltered and filtered sce files
```

modules/classify-celltypes.nf
```diff
@@ -103,16 +106,26 @@ workflow annotate_celltypes {
...
       // perform singleR celltyping and export TSV
       classify_singleR(singler_input_ch)
 
-    // temporary during development
-    emit: celltype_input_ch
+      // add back in the unchanged sce files
+      // TODO update below with output channel results:
+      export_channel = processed_sce_channel
+        .map{[it[0]["library_id"]] + it}
+        // add add in unfiltered and filtered sce files
```
sjspielman, line 122:
```suggestion
        // add in unfiltered and filtered sce files
```

sjspielman: if we're fixing this comment anyways..

PR #476, opened 2023-09-29
Add CellAssign process to main workflow

sjspielman:
Towards #415 

This PR takes the next steps towards celltyping moving into the main workflow:

- I added branching to create two channels in the subworkflow, `celltype_input_ch.run` and `celltype_input_ch.skip`, based on whether or not the singler file value is null (is there a better way to check if a value is `null` here? I couldn't find it!)
  - However, I am not currently checking if the file exists. This probably needs to happen somewhere before going into the process, but wanted to get feedback on this branching first before upgrading it.
- I then prepared a channel for heading into CellAssign, and updated the CellAssign process itself. I included the same type of branch here as well - skip the cellassign if its file value is null. I again have no checks for the file existing and I am also not checking the reference name. Do we think I should?
- I have written the code with the SingleR output files coming along for the right into the CellAssign process, but I left a couple TODO comments about where we would change code to have that not happen. In that case, we'd join these files back up with cell assign output for input to the final process that will add annotations to the SCE, in the next PR.
- A couple more TODOs for subsequent PRs are sprinkled in there, too. Let me know if anything about them is unclear.
- (Edit) - I also updated the stub celltype file to have some `NA`s, to ensure the flow of the skipped branches gets "tested".

### DIFF
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -113,10 +113,11 @@ workflow annotate_celltypes {
         }
       
       // create input for singleR: [meta, processed, SingleR reference model]
-      singler_input_ch = celltype_input_ch
+      singler_input_ch = celltype_input_ch.run
         .map{meta, processed_rds, singler_model, cellassign_model, cellassign_ref_name ->
              [meta, processed_rds, singler_model]}
-
+             
+             
       // perform singleR celltyping and export TSV
       classify_singleR(singler_input_ch)
 
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -101,12 +101,17 @@ workflow annotate_celltypes {
          Utils.parseNA(it.cellassign_ref_name)
         ]}
 
-
+      // create cell typing channel: [meta, processed_rds, singler_model, cellassign_model, cellassign_ref_name]
       celltype_input_ch = processed_sce_channel
         .map{[it[0]["project_id"]] + it}
         .combine(celltype_ch, by: 0)
         .map{it.drop(1)} // remove extra project ID
-
+        // we only run celltyping for rows with a singler model file
+        .branch{
+          skip: it[2] == null
+          run: true
+        }
+      
       // create input for singleR: [meta, processed, SingleR reference model]
       singler_input_ch = celltype_input_ch
         .map{meta, processed_rds, singler_model, cellassign_model, cellassign_ref_name ->
@@ -114,6 +119,8 @@ workflow annotate_celltypes {
 
       // perform singleR celltyping and export TSV
       classify_singleR(singler_input_ch)
+
+      // TODO: mix celltyping results back up with `celltype_input_ch.skip`
 
       // add back in the unchanged sce files
       // TODO update below with output channel results:
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -143,8 +143,13 @@ workflow annotate_celltypes {
               by: 0, failOnMismatch: true, failOnDuplicate: true)
        .map{meta, processed_rds, singler_annotations_tsv, singler_full_results, cellassign_ref ->
           [meta, processed_rds, cellassign_ref]}
+       // branch if the ref is null. This code assumes the reference name exist if the filename exists!
+       .branch{
+          skip: it[2] == null
+          run: true
+        }    
       
-     // classify_cellassign(cellassign_ch) // output: meta, processed_rds, cellassign_predictions, cellassign_ref_name
+     // classify_cellassign(cellassign_ch.run) // output: meta, processed_rds, cellassign_predictions
 
       // TODO mix celltyping results back up with `celltype_input_ch.skip`
 
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -39,18 +39,18 @@ process classify_cellassign {
     publishDir (
         path: "${params.checkpoints_dir}/celltype/${meta.library_id}",
         mode:  'copy',
-        pattern: "*{.tsv,.json}" // Only the prediction matrix (tsv) and meta
+        pattern: "*{_predictions.tsv,.json}" // Only the prediction matrix (tsv) and meta
     )
   label 'mem_32'
   label 'cpus_12'
   input:
-    tuple val(meta), path(processed_rds), path(cellassign_ref)
+    tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results), path(cellassign_ref)
   output:
-    tuple val(meta), path(processed_rds), path(cellassign_predictions_tsv)
+    tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results), path(cellassign_predictions_tsv)
   script:
     processed_hdf5 = "${meta.library_id}_processed.hdf5"
-    cellassign_predictions = "${meta.library_id}_predictions.tsv"
-
+    cellassign_predictions_tsv = "${meta.library_id}_predictions.tsv"
+    
     """
     # Convert SCE to AnnData
     sce_to_anndata.R \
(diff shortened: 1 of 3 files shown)


modules/classify-celltypes.nf
```diff
@@ -34,33 +34,48 @@ process classify_singleR {
...
 
 
-process predict_cellassign {
+process classify_cellassign {
   container params.SCPCATOOLS_CONTAINER
-  publishDir "${params.checkpoints_dir}/celltype/${meta.library_id}", mode: 'copy'
+    publishDir (
+        path: "${params.checkpoints_dir}/celltype/${meta.library_id}",
+        mode:  'copy',
+        pattern: "*{_predictions.tsv,.json}" // Only the prediction matrix (tsv) and meta
```
jashapiro, line 42: Noting that we don't actually write the metadata here, but we should.  Missed this in the singleR step to.

To do this there are a couple of options: the easiest is probably to actually adjust the output so we are writing out a folder rather than the individual files. This actually makes a few things easier downstream, and removes a bunch if definitions... I will follow with suggestions implementing a version of this.

The first is to remove the pattern (sorry glob!)
```suggestion
```
Ugh, because of how I commented you will need to remove the comma on the previous line

modules/classify-celltypes.nf
```diff
@@ -101,20 +116,45 @@ workflow annotate_celltypes {
...
         .map{[it[0]["project_id"]] + it}
         .combine(celltype_ch, by: 0)
         .map{it.drop(1)} // remove extra project ID
-
+        // we only run celltyping for rows with a singler model file
+        // branch here so we have meta and processed sce in the .skip
+        .branch{
+          skip: it[2] == null
+          run: true
+        }
```
jashapiro, line 129: I think we probably want this to be done separately for singler and for cellassign? So I would move this down to be part of the `singler_input_ch` definition, and then mix it back before going to cellassign step, repeating the logic.

Now I'm going back to my earlier thought about when to assign the files and use `NO_FILE`, because we could pretty easily use that for this branch logic, and then just pass all files to the singler process.

Pseudocode below: This assumes that there are no nulls, just `file()` results

```
  .branch{
    skip: it[2].name == "NO_FILE"
    run: true
  }

  classify_singleR(singler_input_ch.run)
  
cellassign_input_ch = classify_singleR.out
   // add on blank file for skipped singleR results and mix back in
  .mix(singler_input_ch.skip.map{it.asList() + [ file(empty_file) ] )
  .branch{
    skip: it[3].name == "NO_FILE"
    run: true
  }

```

modules/classify-celltypes.nf
```diff
@@ -34,33 +34,48 @@ process classify_singleR {
...
   input:
-    tuple val(meta), path(processed_hdf5), path(cellassign_reference_mtx), val(ref_name)
+    tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results), path(cellassign_ref)
   output:
-    tuple val(meta), path(cellassign_predictions), val(ref_name)
+    tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results), path(cellassign_predictions_tsv)
   script:
-    cellassign_predictions = "${meta.library_id}_predictions.tsv"
+    processed_hdf5 = "${meta.library_id}_processed.hdf5"
+    cellassign_predictions_tsv = "${meta.library_id}_predictions.tsv"
```
jashapiro, line 52: I'm actually going to not use this at all, but instead follow the `salmon` mode of an output file with _constant_ file names. The library ID is already in the publishing path. We could do either way though really.

modules/classify-celltypes.nf
```diff
@@ -34,33 +34,48 @@ process classify_singleR {
...
+        pattern: "*{_predictions.tsv,.json}" // Only the prediction matrix (tsv) and meta
+    )
   label 'mem_32'
   label 'cpus_12'
   input:
-    tuple val(meta), path(processed_hdf5), path(cellassign_reference_mtx), val(ref_name)
+    tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results), path(cellassign_ref)
   output:
-    tuple val(meta), path(cellassign_predictions), val(ref_name)
+    tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results), path(cellassign_predictions_tsv)
```
jashapiro, line 49: I'm aspirationally changing the singler annotations output here to assume you have changed that process as well to be similar. Note that the cellassign path is now a constant, which we will create below

```suggestion
    tuple val(meta), path(processed_rds), path(singler_dir), path("cellassign")
```

The updated `input:` would be:
```
tuple val(meta), path(processed_rds), path(singler_dir), path(cellassign_ref)
```

modules/classify-celltypes.nf
```diff
@@ -34,33 +34,48 @@ process classify_singleR {
...
   script:
-    cellassign_predictions = "${meta.library_id}_predictions.tsv"
+    processed_hdf5 = "${meta.library_id}_processed.hdf5"
+    cellassign_predictions_tsv = "${meta.library_id}_predictions.tsv"
+    
     """
+    # Convert SCE to AnnData
```
jashapiro, line 55: create an output directory
```suggestion
    # create output directory
    mkdir -p cellassign

    # Convert SCE to AnnData
```

modules/classify-celltypes.nf
```diff
@@ -34,33 +34,48 @@ process classify_singleR {
...
+    sce_to_anndata.R \
+        --input_sce_file ${processed_rds} \
+        --output_rna_h5 ${processed_hdf5} 
+    
+    # Run CellAssign
     predict_cellassign.py \
       --input_hdf5_file ${processed_hdf5} \
-      --output_predictions ${cellassign_predictions} \
-      --reference ${cellassign_reference_mtx} \
+      --output_predictions ${cellassign_predictions_tsv} \
```
jashapiro, line 63: Constant path for output
```suggestion
      --output_predictions cellassign/predictions.tsv \
```

modules/classify-celltypes.nf
```diff
@@ -34,33 +34,48 @@ process classify_singleR {
...
+      --reference ${cellassign_ref} \
       --seed ${params.seed} \
       --threads ${task.cpus}
     """
```
jashapiro, line 67:
```suggestion
      --threads ${task.cpus}
      
    # write out metadata for tracking
    echo ${echo ${Utils.makeJson(meta)} > cellassign/scpca-meta.json
    """
```

modules/classify-celltypes.nf
```diff
@@ -34,33 +34,48 @@ process classify_singleR {
...
   stub:
-    cellassign_predictions = "${meta.library_id}_predictions.tsv"
+    processed_hdf5 = "${meta.library_id}_processed.hdf5"
+    cellassign_predictions_tsv = "${meta.library_id}_predictions.tsv"
     """
-    touch "${cellassign_predictions}"
+    touch "${cellassign_predictions_tsv}"
+    touch "${processed_hdf5}"
```
jashapiro, line 73: You'll want to update the stub to create the output folder as well

jashapiro (commented):
I started with some suggestions, but I ended up realizing we were missing something, which requires a bit more of a rewrite.

I'm recommending changing the output to be a folder, which means a few different things, discussed below. We might want to do that first as a separate PR for the singleR to test things out, even though I made all my suggestions here.

Oh, and checking for `null`: `!thing` is usally sufficient (and would be here)
It is `true` if thing is `null`, as null is treated as `false` in logic, unlike in R (as is `0` or `""` or `[]`, but you usually want those to be treated as `false` as well)

jashapiro:
Ignore most of this review and look at #498 first.

sjspielman:
Given changes in #477 & #478, I'm going to close out this PR and open a fresh one so that the PR is shorter without all the above "deprecated" comments. Bonus, no need to ever find out what conflict horrors await when merging those changes into this branch 😬

PR #477, opened 2023-09-30
Use output folder for singler

jashapiro:
In my comments on https://github.com/AlexsLemonade/scpca-nf/pull/476, I proposed a bunch of change, and I decided that it would actually be easier for me to just demonstrate and make some of the changes that I was thinking about, so I did that.

The main goal here was to change the output of the singleR process to be a _folder_ with results rather than individual files. This folder also contains the `scpca-meta.json` file. This solves a couple of problems: rather than having to track a bunch of files separately, we just put the output files we want in the folder,  and then pass that along as a single unit. It also gives us a place to put `scpca-meta.json` where we won't be overwriting it or having to give it different names depending on the output. 

For some reason, maybe my bad memory, I thought we wanted to just pass everything through the processes as simply as possible, so I added the cellassign refs to the singler process inputs & outputs so we can got to the next step without a join. (We could still drop the singleR ref before the cellassign process: it is larger than the others so that might be worth it)

Relatedly, I added a `singler_output_ch` mostly just to demonstrate how the mixing would go, with adding an `empty_file` for the skipped singler results. I also chose different names for the branches in anticipation of adding a 3rd branch, `has_singler` or similar, that would just fetch the existing results.

To make all of that easier, I followed as similar pattern that we used in `af-rna.nf` and elsewhere, where we had previously had output directories and step skipping: https://github.com/AlexsLemonade/scpca-nf/blob/2b3fb3cf4f39ba08851465de55b0d231e8059e4b/modules/af-rna.nf#L107-L111

Essentially we are now defining the output directories at the start of the subworkflow and adding them into the meta object so we can access them later, both within the process, as I do now, and as we implement step skipping. 

Finally, for testing purposes, I changed one of the values in the stub file to be NA, and it did seem to work.

### DIFF
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -9,14 +9,13 @@ process classify_singler {
     label 'mem_8'
     label 'cpus_4'
     input:
-        tuple val(meta), path(processed_rds),
-              path(singler_model_file),
-              path(cellassign_reference_file), val(cellassign_reference_name)
+      tuple val(meta), path(processed_rds),
+            path(singler_model_file),
+            path(cellassign_reference_file), val(cellassign_reference_name)
     output:
-        tuple val(meta), path(processed_rds),
-              path(singler_model_file),
-              path(cellassign_reference_file), val(cellassign_reference_name),
-              path(singler_dir)
+      tuple val(meta), path(processed_rds),
+            path(cellassign_reference_file), val(cellassign_reference_name),
+            path(singler_dir)
     script:
       singler_dir = file(meta.singler_dir).name
       """
@@ -112,7 +111,7 @@ workflow annotate_celltypes {
          Utils.parseNA(it.cellassign_ref_name)
         ]}
 
-      // create input for typing: [meta, processed, SingleR model file, cellassign reference file, cellassign reference name]
+      // create input for typing: [meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name]
       celltype_input_ch = processed_sce_channel
         .map{[it[0]["project_id"]] + it}
         .combine(celltype_ch, by: 0)
@@ -137,7 +136,9 @@ workflow annotate_celltypes {
 
       // singleR output channel
       singler_output_ch = singler_input_ch.missing_ref
-        .map{it.asList() + [file(empty_file)]} // add empty file for missing output
+        // drop singler model and add empty file for missing output (match classify_singler output)
+        .map{meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name ->
+             [meta, processed_sce, cellassign_reference_file, cellassign_reference_name, file(empty_file)]}
         // add in channel outputs
         .mix(classify_singler.out)
 
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -1,35 +1,40 @@
 
-process classify_singleR {
+process classify_singler {
     container params.SCPCATOOLS_CONTAINER
-    publishDir (
-        path: "${params.checkpoints_dir}/celltype/${meta.library_id}",
-        mode:  'copy',
-        pattern: "*{_singler.rds,.tsv,.json}" // Everything except processed rds
-    )
+    publishDir "${meta.celltype_publish_dir}", mode: 'copy'
     label 'mem_8'
     label 'cpus_4'
     input:
-        tuple val(meta), path(processed_rds), path(singler_model_file)
+        tuple val(meta), path(processed_rds),
+              path(singler_model_file),
+              path(cellassign_reference_file), val(cellassign_reference_name)
     output:
-        tuple val(meta), path(processed_rds), path(singler_annotations_tsv), path(singler_full_results)
+        tuple val(meta), path(processed_rds),
+              path(singler_model_file),
+              path(cellassign_reference_file), val(cellassign_reference_name),
+              path(singler_dir)
     script:
-      singler_annotations_tsv = "${meta.library_id}_singler_annotations.tsv"
-      singler_full_results = "${meta.library_id}_singler.rds"
+      singler_dir = file(meta.singler_dir).name
       """
+      # create output directory
+      mkdir "${singler_dir}"
+
       classify_SingleR.R \
-        --sce_file ${processed_rds} \
-        --singler_model_file ${singler_model_file} \
-        --output_singler_annotations_file ${singler_annotations_tsv} \
-        --output_singler_results_file ${singler_full_results} \
+        --sce_file "${processed_rds}" \
+        --singler_model_file "${singler_model_file}" \
+        --output_singler_annotations_file "${singler_dir}/singler_annotations.tsv" \
+        --output_singler_results_file "${singler_dir}/singler_results.rds" \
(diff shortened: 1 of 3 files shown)


modules/classify-celltypes.nf
```diff
@@ -93,27 +103,45 @@ workflow annotate_celltypes {
...
+        }
+
+      // perform singleR celltyping and export results
+      classify_singler(singler_input_ch.do_singler)
+
+      // singleR output channel
+      singler_output_ch = singler_input_ch.missing_ref
+        // drop singler model and add empty file for missing output (match classify_singler output)
+        .map{meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name ->
+             [meta, processed_sce, cellassign_reference_file, cellassign_reference_name, file(empty_file)]}
```
sjspielman, line 141: Noting that here `empty_file` <-> `path(singler_dir)`. See my overall comment for whether `singler_dir` needs to actually be outputted. If not, would change to - 

```suggestion
        .map{meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name ->
             [meta, processed_sce, cellassign_reference_file, cellassign_reference_name]}
```

modules/classify-celltypes.nf
```diff
@@ -93,27 +103,45 @@ workflow annotate_celltypes {
...
+          missing_ref: it[2].name == "NO_FILE"
+          do_singler: true
+        }
+
+      // perform singleR celltyping and export results
+      classify_singler(singler_input_ch.do_singler)
+
+      // singleR output channel
+      singler_output_ch = singler_input_ch.missing_ref
+        // drop singler model and add empty file for missing output (match classify_singler output)
```
sjspielman, line 139: classify_singler output? singler_dir?

sjspielman (commented):
Overall I like this approach of using individual directories for the checkpoint files, thanks for doing that!
Related, regarding nextflow efficiency, is it maybe more efficient for nextflow to grab files from checkpoint directories (for the forthcoming `add_celltypes` process where we read annotations back in and add to SCE) vs passing them around as `path()`s? Or is it about the same?

One thing I'm not following is why `singler_dir` is being output from `classify_singler` but it was previously also added to `meta` - this seems redundant? Is this what you meant to return? There's one other comment I left that is directly about this too.

modules/classify-celltypes.nf
```diff
@@ -93,27 +103,45 @@ workflow annotate_celltypes {
...
 
-
+      // create input for typing: [meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name]
       celltype_input_ch = processed_sce_channel
         .map{[it[0]["project_id"]] + it}
         .combine(celltype_ch, by: 0)
         .map{it.drop(1)} // remove extra project ID
+        // add celltype_publish_dir to the meta object for later use
+        .map{
+          it[0].celltype_publish_dir = "${params.checkpoints_dir}/celltype/${it[0].library_id}";
```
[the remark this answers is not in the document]
jashapiro: They could be just returns, but not commas. 

We are passing in a multi-line statement: first modifying `it` then returning all of `it` in the next line. I find that within a `map` statement or other anonymous function/lambda/closure it is helpful to add the explicit semicolons.

modules/classify-celltypes.nf
jashapiro: I mean the output of the `classify_singler` process, which is a tuple of [meta, processed, cellassign reference, cellassign reference name, and singler directory]. 

When we put things in `meta`, we are just putting in the path as a string. When we pass them along as part of the channel (as a `path` output element), we actually pass along an object that instructs Nextflow that it needs to stage those files for the next process it encounters. In addition, having the output be a `path` means that it will be copied to as part of `publishDir` This is why we need it as an actual output, not just in meta.

jashapiro:
> Related, regarding nextflow efficiency, is it maybe more efficient for nextflow to grab files from checkpoint directories (for the forthcoming `add_celltypes` process where we read annotations back in and add to SCE) vs passing them around as `path()`s? Or is it about the same? (edit - this is more of a question than a comment!)

I think either way could be about the same... I'm not sure I completely follow, but perhaps we could avoid some extra files being passed in by adding the reference paths to the `meta` dictionary and then getting them as required for the individual steps. 

The thought would be that we could update the following chunk:
```
      celltype_input_ch = processed_sce_channel
        .combine(celltype_ch, by: 0)
        // add celltype_publish_dir to the meta object for later use
        .map{
          it[0].celltype_publish_dir = "${params.checkpoints_dir}/celltype/${it[0].library_id}";
          it[0].singler_dir = "${it[0].celltype_publish_dir}/${it[0].library_id}_singler";
          it[0].cellassign_dir = "${it[0].celltype_publish_dir}/${it[0].library_id}_cellassign";
          it
        }
```

to also include updating `it[0].singler_reference`, `it[0].cellassign_reference`, etc. and then drop those _files_  from the output that we pass along in the channel. So we would just pass along `[meta, proccessed_sce]`.

Then we could add `.map{it.toList() + file(it[0].singler_reference)` when creating `singler_input_ch`, and similar for `cellassign_input_ch`, which might be a bit simpler, but we would still want to carry along the singler output when going into 
 the cellassign process. I can pull together a version of that if you think it is something you want to see.

modules/classify-celltypes.nf
sjspielman: Got it. To paraphrase for myself:

- We place it in `meta` so we have the string pre-defined for use when working with the directory
- Explicitly popping this as a `path` in the output is the only thing that causes nextflow to bring it along for the ride as an actual directory with files we'll want to use. If we were to just refer to the meta string later on, we wouldn't get its published contents.

sjspielman:
> I can pull together a version of that if you think it is something you want to see.

No, it's fine - my question was actually related to this answer you gave me in https://github.com/AlexsLemonade/scpca-nf/pull/477#discussion_r1342674619. I was thinking that why pass around `path()` values if everything is stored in directories we have saved in `meta` (as strings!), and the answer is because that's how nextflow rolls.

modules/classify-celltypes.nf
jashapiro: Yes. Regarding your second point: we _might_ get the published contents, but there are no guarantees. When the copying from the workdir to the publishdir happens is something that nextflow makes guarantees about, so if we want the output of a step down the line, we need to explicitly pass those files/directories along as an output in the channel.

sjspielman: > we _might_ get the published contents, but there are no guarantees.

cool cool.

sjspielman (approved):
👍 

Just noting that depending on what monsters e**merge** (im not sorry for writing this) when merging this into my branch, I may close the open PR in #476 and start fresh. In fact now that I write this, that might be the right move anyways since many of those comments in #476 are now deprecated and long PRs are hard to follow!

jashapiro:
> 👍
> 
> Just noting that depending on what monsters e**merge** (im not sorry for writing this) when merging this into my branch, I may close the open PR in #476 and start fresh. In fact now that I write this, that might be the right move anyways since many of those comments in #476 are now deprecated and long PRs are hard to follow!

This seems like the right move. Hang on a bit for one more option.

PR #479, opened 2023-10-02
Add CellAssign process

sjspielman:
Towards #415 

The next step is running cell assign, done here! I think given all the nice organizational changes in #476 & #477 this is pretty straightforward for review. The one additional thing to note is that I realized we probably need to be including `singler_reference_name` in the same way that we include `cellassign_reference_name`. This wasn't previously needed since we had been doing singler celltyping _and adding to SCE object_ in the same process, where the reference name was accessible from the trained singler model itself. Now that we will be adding singler & cellassign to the final step all at once in the last process, we'll need to grab both reference names for input.

I also have one discussion-y comment to bring up at this point - my memory from our figjam session 🎸  was that we had decided not to keep using indicator variables in meta, e.g. `meta.has_singler=true` (edit lowercase true, this isn't R). I am not seeing any notes that explicitly say this though, so let's make sure I am remembering correctly! We can/should certainly still add relevant information to _SCE metadata_ as we had been doing, but with our current celltyping workflow structure I don't really see the need for anything in the json metadata unless I am missing something?

Noting the next PR will write the process & R script for adding celltypes into the processed SCE.

### DIFF
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -18,7 +18,8 @@ option_list <- list(
   make_option(
     opt_str = c("--singler_model_file"),
     type = "character",
-    help = "path to file containing a single model generated for SingleR annotation"
+    help = "path to file containing a single model generated for SingleR annotation. 
+            File name is expected to be in form: <model name>_model.rds."
   ),
   make_option(
     opt_str = c("--output_singler_annotations_file"),
@@ -55,7 +56,7 @@ if (!file.exists(opt$sce_file)) {
   stop("Missing SCE file")
 }
 
-# check that output files have the righr extensions
+# check that output files have the right extensions
 if (!(stringr::str_ends(opt$output_singler_results_file, ".rds"))) {
   stop("output SingleR result file name must end in .rds")
 }
@@ -63,10 +64,19 @@ if (!(stringr::str_ends(opt$output_singler_annotations_file, ".tsv"))) {
   stop("output SingleR annotations file name must end in .tsv")
 }
 
-# check that references all exist
+# check that reference exists and filename is properly formatted
 singler_model_file <- opt$singler_model_file
 if (!file.exists(singler_model_file)) {
   stop(glue::glue("Provided model file {singler_model_file} is missing."))
+}
+if (!(stringr::str_ends(singler_model_file, "_model.rds"))) {
+  stop(glue::glue("Provided model file {singler_model_file} must end in .rds."))
+}
+
+# get & check reference name
+reference_name <- stringr::str_remove(singler_model_file, "_model.rds")
+if (reference_name == "") {
+  stop(glue::glue("Provided model file name must be formatted as `<model_name>_model.rds`"))
 }
 
 # set up multiprocessing params
@@ -90,6 +100,10 @@ singler_results <- SingleR::classifySingleR(
   BPPARAM = bp_param
 )
 
+# add reference name as metadata in singler_results DataFrame
+metadata(singler_results)$reference_name <- reference_name
+
+
 # export results
 
 # first, a stand-alone tsv of annotations with both pruned and full labels
@@ -102,6 +116,7 @@ readr::write_tsv(
 )
 
 # next, the full result to a compressed rds
+
 readr::write_rds(
   singler_results,
   opt$output_singler_results_file,
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -70,11 +70,11 @@ if (!file.exists(singler_model_file)) {
   stop(glue::glue("Provided model file {singler_model_file} is missing."))
 }
 if (!(stringr::str_ends(singler_model_file, "_model.rds"))) {
-  stop(glue::glue("Provided model file {singler_model_file} must end in .rds."))
+  stop(glue::glue("Provided model file {singler_model_file} must end in '_model.rds.'"))
 }
 
 # get & check reference name
-reference_name <- stringr::str_match(singler_model_file, "(\\w+)_model.rds")[2]
+reference_name <- stringr::str_remove(singler_model_file, "_model.rds$")
 if (reference_name == "") {
   stop(glue::glue("Provided model file name must be formatted as `<model_name>_model.rds`"))
 }
@@ -116,7 +116,6 @@ readr::write_tsv(
 )
 
 # next, the full result to a compressed rds
-
 readr::write_rds(
   singler_results,
   opt$output_singler_results_file,
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -74,7 +74,7 @@ if (!(stringr::str_ends(singler_model_file, "_model.rds"))) {
 }
 
 # get & check reference name
-reference_name <- stringr::str_remove(singler_model_file, "_model.rds")
+reference_name <- stringr::str_match(singler_model_file, "(\\w+)_model.rds")[2]
 if (reference_name == "") {
   stop(glue::glue("Provided model file name must be formatted as `<model_name>_model.rds`"))
 }
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -48,9 +48,9 @@ process classify_cellassign {
   label 'mem_32'
   label 'cpus_12'
   input:
-      tuple val(meta), path(processed_rds), path(cellassign_reference_file)
+    tuple val(meta), path(processed_rds), path(cellassign_reference_file)
   output:
-      tuple val(meta.library_id), path(cellassign_dir)
+    tuple val(meta.library_id), path(cellassign_dir)
   script:
     cellassign_dir = file(meta.cellassign_dir).name
 
@@ -63,12 +63,12 @@ process classify_cellassign {
     # Convert SCE to AnnData
     sce_to_anndata.R \
         --input_sce_file ${processed_rds} \
-        --output_rna_h5 ${processed_hdf5} 
+        --output_rna_h5 processed.hdf5 
         
     # Run CellAssign
     predict_cellassign.py \
-      --input_hdf5_file ${processed_hdf5} \
-      --output_predictions ${cellassign_predictions_tsv} \
+      --input_hdf5_file processed.hdf5 
+      --output_predictions ${cellassign_dir}/cellassign_predictions.tsv \
       --reference ${cellassign_ref} \
       --seed ${params.seed} \
       --threads ${task.cpus}
@@ -196,9 +196,6 @@ workflow annotate_celltypes {
         // add in cell assign results
         .join(cellassign_output_ch, by: 0, failOnMismatch: true, failOnDuplicate: true)
         .map{it.drop(1)} // remove library_id
-        // bring out the reference names
-        .map{meta, processed_sce, singler_dir, cellassign_dir -> 
-          [meta, processed_sce, singler_dir, meta.singler_reference_name, cellassign_dir, meta.cellassign_reference_name]}
 
 
       // Next PR:
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -38,33 +38,57 @@ process classify_singler {
 }
 
 
-process predict_cellassign {
+process classify_cellassign {
   container params.SCPCATOOLS_CONTAINER
-  publishDir "${params.checkpoints_dir}/celltype/${meta.library_id}", mode: 'copy'
+    publishDir (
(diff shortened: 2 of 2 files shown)


jashapiro (commented):
Looks good. I suggested a few changes: 

- Using static filenames within the "bash script" of the process. Since these files are isolated and never "escape" the process, it isn't worth assigning them library-specific names.
- Similarly, for checkpoint files, it doesn't seem worth the effort to give them special names: the directory they are in will already encode specificity to the library
- Removing the `singler_reference_name` variable. We didn't have it before because it is already in the `.rds` files. 
  - Speaking of the `.rds` files: If we keep this structure where we are not passing the singler results directly along to cellassign, we might want to trim the output `singler_results.rds ` file in `classify_SingleR.R` output to save space. This would be a separate PR, and I would probably want to make trimming an option at the script level (which we would invoke in the process).
- Finally: no need to add extra values to the input channel for `assign_celltypes`: As long as they are in `meta`, we can access that from within the process.

jashapiro:
> I also have one discussion-y comment to bring up at this point - my memory from our figjam session 🎸 was that we had decided not to keep using indicator variables in meta, e.g. `meta.has_singler=true` (edit lowercase true, this isn't R). I am not seeing any notes that explicitly say this though, so let's make sure I am remembering correctly! We can/should certainly still add relevant information to _SCE metadata_ as we had been doing, but with our current celltyping workflow structure I don't really see the need for anything in the json metadata unless I am missing something?

The presence of a reference file(s) for singleR and/or cellassign will be enough to say whether that processing has happened. There should not be any need for a separate boolean value.

jashapiro:
> * Removing the `singler_reference_name` variable. We didn't have it before because it is already in the `.rds` files.

I realized that I am not 100% sure whether the reference name _is_ in the output `.rds` object. It was there when we included the full SCE object, but maybe not now? I'd still say it is better if we can include it in the object itself, rather than creating an extra `meta` variable and associated column that we would have to add to the input table.

sjspielman:
A couple thoughts on reference names - 

1. The singler trained model name _and_ the cellassign reference names are both present in their respective file names, .e.g. `BlueprintEncodeData_model.rds` & `PanglaoDB-blood.tsv`. So, since these filenames are in `meta`, I guess we don't need to pass either the singler or cellassign reference names in separately. We could extract this info from filenames with some string manipulation. (Not loving that we have `_` vs `-`, though..). 
3. In terms of adding the singler model name to the singler object saved in`rds`, I'm not sure how that would play out - my idea was to keep that as the exact structure that singler exports so that people can actually use it if they want to with singler's native functions, e.g. heatmap. It's a `DataFrame`, so we'd have to add a column with model name which should be fine unless there is crazy strict checking in singler's functions. But, if we do my point in 1^ , then this isn't needed.

sjspielman:
> I realized that I am not 100% sure whether the reference name is in the output .rds object. It was there when we included the full SCE object, but maybe not now? 

It is not in the output rds from the singler projcess at this point:
https://github.com/AlexsLemonade/scpca-nf/blob/fb5f3bf1d9dad7533046cf05230a5cf0b3ddabaf/modules/classify-celltypes.nf#L25

modules/classify-celltypes.nf
```diff
@@ -100,31 +121,34 @@ workflow annotate_celltypes {
          it.scpca_project_id,
          // singler model file
          Utils.parseNA(it.singler_ref_file) ? "${params.singler_models_dir}/${it.singler_ref_file}" : null,
+         // singler reference name 
+         Utils.parseNA(it.singler_ref_name),
```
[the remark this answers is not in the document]
sjspielman: just noting i'll circle back to this suggestion after final.final discussion in the PR!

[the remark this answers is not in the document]
jashapiro: > The reason we didn't have this before is because it is encoded in the rds file of the SingleR model. Since the cellassign input is just a tsv, we don't have that info.

I was a bit confused: the reference name was being parsed out of the file name: https://github.com/AlexsLemonade/scpca-nf/blob/67e175bdbbfb314e0870a1b832d48001bfa87d9e/bin/classify_SingleR.R#L78

Looking back at https://github.com/AlexsLemonade/scpca-nf/pull/402, it seems like we added that variable to the table mostly for convenience, but we could have parsed it out of the file name. It seems like we are using a shorthand at the moment, but the full file name (less `.tsv`) should probably be included as the name (i.e. "PanglaoDB-blood" rather than "blood"). In nextflow this can be done with `file("path/my_model.tsv").baseName` which strips both the directory and the extension.

And fun addition, if we want to do the same string parsing for the singler model files, which end with "_model.rds", we could do: `file("path/my_model.tsv").baseName - '_model'` which is so odd, but it works. If you want to be safer, use a little regex instead: `file("path/my_model.tsv").baseName - ~'_model$'`. Yes, you can make a regex by just adding a ~ before a quote. Groovy is so strange, but kind of cool sometimes.

sjspielman:
> But we can add it quite trivially, as DataFrame objects have metadata slots just like the SCE. So we could add in something like metadata(singler_result)$singler_reference <- singler_reference_name to store the reference name along with the results itself. 

Right, I would just want to make sure this doesn't cause singler's native functions to break, which it really shouldn't. Let me double check though...

jashapiro:
> It is possible for the `meta` object to become disconnected from the file name/reference that was actually used if we are not careful, but I expect that the `repeat_celltyping` logic should handle that when we add it.

One way to handle this would be to (optionally) pass in the `spca-meta.json` files that are in the cellassign and singler directories

modules/classify-celltypes.nf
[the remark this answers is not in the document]
jashapiro: > just noting i'll circle back to this suggestion after final.final discussion in the PR!

My current suggestion is to punt this to a separate PR: it isn't really part of getting the cellassign process done.

sjspielman:
> As long as we are adding in metadata, I'm fine with passing in the reference names for both cellassign and singler directly as an argument to the assign script. The only thing we have to be cautious about (and we have to do this anyway) is that we need to make sure we are passing in the correct value.

Feels like this could serve as an extra check as well - we'd want to make sure the singler reference name that gets passed in is the same as the name recorded in the singler DataFrame's metadata.

sjspielman:
Changes since last time:

- https://github.com/AlexsLemonade/scpca-nf/pull/479/commits/9de11fc74436ccf2ae5229bf38f55fd158bdf39b: I got rid of all tracking of the reference names in favor of "regexing them," if you will..., out of the reference/model file names directly.
- https://github.com/AlexsLemonade/scpca-nf/pull/479/commits/09565d13607fefecf064070d7db20b97e0ecaf3a: I updated the SinglerR script to add the reference name as DataFrame metadata, with a couple checks to go with it

bin/classify_SingleR.R
```diff
@@ -55,19 +56,28 @@ if (!file.exists(opt$sce_file)) {
...
 }
+if (!(stringr::str_ends(singler_model_file, "_model.rds"))) {
+  stop(glue::glue("Provided model file {singler_model_file} must end in .rds."))
+}
+
+# get & check reference name
+reference_name <- stringr::str_remove(singler_model_file, "_model.rds")
+if (reference_name == "") {
+  stop(glue::glue("Provided model file name must be formatted as `<model_name>_model.rds`"))
+}
```
sjspielman, line 80: Hm, maybe this isn't the right stringr function. It's fine for pulling out the model name, but not for _checking_ the model name. Maybe we want `string::str_match()`?

bin/classify_SingleR.R
```diff
@@ -55,19 +56,28 @@ if (!file.exists(opt$sce_file)) {
...
 }
 
-# check that references all exist
+# check that reference exists and filename is properly formatted
 singler_model_file <- opt$singler_model_file
 if (!file.exists(singler_model_file)) {
   stop(glue::glue("Provided model file {singler_model_file} is missing."))
 }
+if (!(stringr::str_ends(singler_model_file, "_model.rds"))) {
+  stop(glue::glue("Provided model file {singler_model_file} must end in .rds."))
```
jashapiro, line 73:
```suggestion
  stop(glue::glue("Provided model file {singler_model_file} must end in '_model.rds.'"))
```

bin/classify_SingleR.R
```diff
@@ -55,19 +56,28 @@ if (!file.exists(opt$sce_file)) {
...
+}
+
+# get & check reference name
+reference_name <- stringr::str_match(singler_model_file, "(\\w+)_model.rds")[2]
+if (reference_name == "") {
+  stop(glue::glue("Provided model file name must be formatted as `<model_name>_model.rds`"))
+}
```
jashapiro, line 80: I think the `str_remove` is more clear and simpler.

Your already checked above that the file had the right ending, so the second check here will only be triggered if the full file name was "_model.rds". 
```suggestion
reference_name <- stringr::str_remove(singler_model_file, "_model.rds$")
if (reference_name == "") {
  stop(glue::glue("Provided model file name must be formatted as `<model_name>_model.rds`"))
}
```

bin/classify_SingleR.R
```diff
@@ -102,6 +116,7 @@ readr::write_tsv(
 )
 
 # next, the full result to a compressed rds
+
```
jashapiro, line 119:
```suggestion
```

modules/classify-celltypes.nf
```diff
@@ -101,30 +120,27 @@ workflow annotate_celltypes {
...
-        // current contents: [project_id, meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name]
+        // current contents: [project_id, meta, processed_sce, singler_model_file, cellassign_reference_file]
         // add values to meta for later use
-        .map{ project_id, meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name ->
+        .map{ project_id, meta, processed_sce, singler_model_file, cellassign_reference_file ->
           meta.celltype_publish_dir = "${params.checkpoints_dir}/celltype/${meta.library_id}";
           meta.singler_dir = "${meta.celltype_publish_dir}/${meta.library_id}_singler";
           meta.cellassign_dir = "${meta.celltype_publish_dir}/${meta.library_id}_cellassign";
-          meta.singler_model_file = singler_model_file;
+          meta.singler_model_file = singler_model_file;
```
jashapiro, line 136:
```suggestion
          meta.singler_model_file = singler_model_file;
```

modules/classify-celltypes.nf
```diff
@@ -133,23 +149,50 @@ workflow annotate_celltypes {
           missing_ref: it[2].name == "NO_FILE"
           do_singler: true
         }
+      
 
       // perform singleR celltyping and export results
       classify_singler(singler_input_ch.do_singler)
-
+
```
jashapiro, line 156:
```suggestion

```

modules/classify-celltypes.nf
```diff
@@ -101,30 +120,27 @@ workflow annotate_celltypes {
...
         // add values to meta for later use
-        .map{ project_id, meta, processed_sce, singler_model_file, cellassign_reference_file, cellassign_reference_name ->
+        .map{ project_id, meta, processed_sce, singler_model_file, cellassign_reference_file ->
           meta.celltype_publish_dir = "${params.checkpoints_dir}/celltype/${meta.library_id}";
           meta.singler_dir = "${meta.celltype_publish_dir}/${meta.library_id}_singler";
           meta.cellassign_dir = "${meta.celltype_publish_dir}/${meta.library_id}_cellassign";
-          meta.singler_model_file = singler_model_file;
+          meta.singler_model_file = singler_model_file; 
           meta.cellassign_reference_file = cellassign_reference_file;
-          meta.cellassign_reference_name = cellassign_reference_name;
```
jashapiro, line 122: I assume that by removing this here, you plan to add some code to `add_celltypes_to_sce` to pull out the cellassign reference name?

jashapiro (approved):
This looks fine, with a few small suggestions (some just whitespace). 
Any concerns related to tracking reference file names can be settled when we are actually using that part.

modules/classify-celltypes.nf
sjspielman: Yes, that was the idea!

sjspielman:
Just for posterity...

- stub failed for https://github.com/AlexsLemonade/scpca-nf/pull/479/commits/3c2715cbb808124be2b41b4b6b5a38ec393b6aed, but none of these changes really should have caused that?
- Locally, stub ran fine for me at that commit!
- I retriggered the CI stub check and now it's passing 🤷‍♀️

Issue #447, opened 2023-09-11
Move everything but UMAPs and tables for cell type results to a separate report


allyhawkins:
The QC report is a little long with all the cell type information, so we should pull out some of the results from cell typing into a separate report. We want to keep the section and include only the tables and the UMAPs. Then, the rest of the report will get moved to it's own file that gets rendered separately. 

We can render this file within the same script that we use to render the full QC report. We just need to modify the process to include two output html files. 

We will also want to link the full cell type report within the cell type section of the QC report.

sjspielman:
I'm going to actually close this out and it split this up into 2 issues:

- split up the qc report: #481
- update the process to: #482

Issue #371, opened 2023-07-06
Update QC report tables


sjspielman:
From this [comment](https://github.com/AlexsLemonade/scpca-nf/pull/364#discussion_r1254599659)

> Oh, one more comment, which I don't want to make hold this up: When we have a lot of ADTs, this table gets kind of unreasonably long. We could limit this to a set number of ADTs, or collapse the table, but I'm not really sure what the best option is here. Maybe a JS table like https://glin.github.io/reactable/ ? 

We should take some time to revisit tables overall, including but not limited to excessively long tables. Another lightweight (aka no additional dependencies) approach could be `DT::datatable()`, but we should probably explore a couple options.

sjspielman:
Another table update we could use, from https://github.com/AlexsLemonade/scpca-nf/pull/483#discussion_r1344520068:

> If we are reusing this style frequently across different table types, it might make sense as a separate function that we define at the top and reuse as needed. (I think we duplicate this a lot)

We reuse a lot of kable formatting so we might benefit from a function to handle a lot of that.

sjspielman:
`reactable` might give us files that are too big, if `react.js` is coming along for the ride! Something to keep our 👀 on if we go that direction.

sjspielman:
Noting that `DT` has a somewhat uncertain future with a new maintainer incoming, it seems - 

https://yihui.org/en/2024/01/bye-rstudio/

> So please do not panic—our existing R packages will still be maintained. The only exception is the DT package, which is not included in the contract, and Posit plans to find a new maintainer for it. Before that happens, I might still be able to help (time permitting), but I cannot promise.

sjspielman:
We're back 1.5 years later! I did a little poking around at how the table package affects document size. I made a small notebook that reads a merged SCE and prints (tabsetted) 3 cell type tables. I made a version of this notebook with the following table engines, and they led to the final document sizes:

* `knitr::kable()`: 627 Kb  - _this is not interactive_
* `DT::datatable()`: 841 Kb
* `reactable::reactable()`:  1.1 Mb

To me, the `reactable` one looks the best, but `DT` is more lightweight for an interactive table.

[compare-tables.zip](https://github.com/user-attachments/files/21469769/compare-tables.zip)

sjspielman:
> I think the zip you attached is empty?

My bad! 

[compare-tables.zip](https://github.com/user-attachments/files/21472729/compare-tables.zip)

allyhawkins:
[...]

PR #483, opened 2023-10-03
Split up cell types QC report

### LINKED ISSUE #481: Separate cell types qc report
_Broken out of #447_

> The QC report is a little long with all the cell type information, so we should pull out some of the results from cell typing into a separate report. We want to keep the section and include only the tables and the UMAPs. Then, the rest of the report will get moved to it's own file that gets rendered separately.
> 
> We will also want to link the full cell type report within the cell type section of the QC report.

The first step here is to create a second Rmd for the "supplementary cell types" but splitting up the existing QC report.
Since there will be a decent amount of code reuse, we should also move cell type functions out of the QC report and into a separate utils file.

sjspielman:
Closes #481 

This PR splits up the cell type report:
- In the main report, only count tables and UMAPs (including coloring by clusters) are present
- There is now a supplementary report `templates/celltypes_report.rmd` that contains diagnostic plots and all the heatmaps. This report is linked from the main QC report, using the relative link as it would appear in the ScPCA download.
- Since we have a lot of reused code now, _and_ since there were a bunch of functions scattered throughout the celltype subreport, I moved all the functions into a new file `templates/qc_utils/celltype_functions.R`, and added roxygen comments.

Notable places for feedback:

- File organization and naming! I placed the celltype report in`templates/` since there was an integration report there too, and similarly this is where I placed `qc_utils/`. I am not tied to these location choices if we have other good ideas here
- Is content in the right place? I went off of notes recorded in #447, but I was out the day those plans were discussed, so let's ensure I interpreted things correctly.
- Maybe we should beef up text in the supplementary report? What else might it need?


templates/qc_utils/celltype_functions.R
```diff
@@ -0,0 +1,223 @@
...
+                                  # Legend parameters
+                                  heatmap_legend_param = list(
+                                    title = "Log(Number of cells)",
+                                    title_position = "leftcenter-rot",
+                                    legend_height = unit(4, "cm")
+                                  )
+  )
+  # draw with legend on left for spacing
+  ComplexHeatmap::draw(heat, heatmap_legend_side = "left")
+}
```
sjspielman, line 223:
```suggestion
}

```

templates/qc_utils/celltype_functions.R
```diff
@@ -0,0 +1,223 @@
...
+prepare_annotation_values <- function(df, annotation_column) {
+  df |>
+    dplyr::mutate(
+      {{ annotation_column }} := dplyr::case_when(
+        # singler condition
+        is.na({{ annotation_column }}) ~ "Unknown cell type",
+        # cellassign conditon
+        {{ annotation_column }} == "other" ~ "Unknown cell type",
+        # otherwise, keep it
+        TRUE ~ {{ annotation_column }}
```
jashapiro, line 57: updated case_when syntax
```suggestion
        .default = {{ annotation_column }}
```

templates/qc_utils/celltype_functions.R
```diff
@@ -0,0 +1,223 @@
...
+    ) |>
+    # kable formatting
+    knitr::kable(align = "r") |>
+    kableExtra::kable_styling(
+      bootstrap_options = "striped",
+      full_width = FALSE,
+      position = "left"
+    ) |>
+    kableExtra::column_spec(2, monospace = TRUE)
+}
```
jashapiro, line 93: I might leave the `kable` formatting out. If we are reusing this style frequently across different table types, it might make sense as a separate function that we define at the top and reuse as needed. (I think we duplicate this a lot)

templates/qc_utils/celltype_functions.R
```diff
@@ -0,0 +1,223 @@
...
+    dplyr::mutate(
+      celltypes = celltype_df[[glue::glue("{celltype_method}_celltype_annotation")]] |>
+        forcats::fct_lump_n(
+          n_celltypes,
+          other_level = "Other cell type"
+        )
+    ) |>
+    # finally, rename temporary `celltypes` column to the provided method
+    dplyr::rename({{ celltype_method }} := celltypes)
+}
```
jashapiro, line 119: I don't understand this function. Why are there two data frames? Why not just manupulate the cell types within the data frame that already exists? Here we are relying on the two dataframes to be parallel without any checks or joins?

My guess is that this can be reduced to just the `mutate` call... not sure why there is a rename here rather than direct assignment?

templates/qc_utils/celltype_functions.R
jashapiro, line 223: Until we have things settled, I would probably rather have the heatmap function inline where the plot is being displayed. (I generally feel this way about plotting functions: I'd rather see the full plot code as much as possible near the plot).

Minor formatting thing as well: move first argument onto the next line to prevent having such an heavily entabbed chunk.

templates/qc_utils/celltype_functions.R
```diff
@@ -0,0 +1,223 @@
...
+
+
+#' Create `celltype_df` data frame for use in cell type QC reports
+#'
+#' @param processed_sce The processed sce object
+#' @param has_singler Boolean for whether SingleR annotations are present
+#' @param has_cellassign Boolean for whether CellAssign annotations are present
+#'
+#' @return `celltype_df` with column of cell types, as factors, for each annotation method
+create_celltype_df <- function(processed_sce, has_singler, has_cellassign) {
```
jashapiro, line 11: Looking at the lump function and my complaint with it: why don't we put the UMAP columns in as part of this function?

[`scuttle::makePerCellDF()`](https://rdrr.io/github/LTLA/scuttle/man/makePerCellDF.html) can do this for us, and in fact could probably replace most of this function, with the excpetion of the `prepare_annotation_values`. Having some extra columns around is probably not going to hurt us.

templates/qc_utils/celltype_functions.R
```diff
@@ -0,0 +1,223 @@
...
+      barcode,
+      clusters,
+      contains("singler"),
+      contains("cellassign"),
+      contains("submitter")
+    )
+
+  if (has_singler) {
+    celltype_df <- celltype_df |>
+      prepare_annotation_values(singler_celltype_annotation)
```
jashapiro, line 27: where does `singler_celltype_annotation` come from? I'm assuming it is a column name, but we should probably check that it exists rather than trusting `has_singler`. In fact, we could just do that check and remove the extra arguments?

jashapiro (commented):
A few initial thoughts, some of which I can probably get over:

- My first thought is that I don't really like using `source()` to define functions, as I think it can make for extremely confusing namespaces. But I can probably get over that. (I looked into some ideas for creating a namespace... one idea here: https://community.rstudio.com/t/is-it-dangerous-to-create-a-namespace-for-functions-in-a-script-without-a-package/91141 But this is probably too much effort for my own discomfort.)
  - I would try to limit the `source`d file to only the functions that are used in more than one place, and try to avoid code that is about formatting (kable & plotting) as I think the indirection makes it harder to do the live adjustments we are probably going to need during revisions.
  - When using sourced files, those files should contain all required `library()` imports, even if they are potentially redundant. (In this case, SingleCellExperiment, most likely). This is probably controversial, as it could result in overwriting of functions in a non-transparent way, but that is really more being mad at `source()` again.
- The `celltypes_report.rmd` file should probably live inside `qc_report(s)` (possibly with a bit of renaming... `main_qc_report.rmd`?)
  - This is because when we are passing files to a process, it is easier to pass the directory that contains all needed files and know that the structure will be preserved, vs passing in each of the component files that we will need, and worrying that nextflow might rename them. 
  - Since we need the `qc_utils` directory that should also live within `qc_report` _or_ be just a bare file in that directory.

- As I expect we will want to give the celltype report a dynamic name, we probably want to add a parameter with that value to the QC report and use that for the link.

Finally, in response to your last question: yes, I think we will want to expand the celltype report with a bit more explanation in a few places, but I think that can wait for separate PRs.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -55,89 +55,20 @@ knitr::asis_output(
 )
 ```
 
+For additional information about cell typing, including diagnostic plots and/or heatmap comparisons among annotations (if available), please refer to the [supplementary cell type QC report](celltypes_report.html).
```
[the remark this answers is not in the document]
sjspielman: Great point, thanks!

templates/qc_utils/celltype_functions.R
```diff
@@ -0,0 +1,223 @@
...
+#' @return `celltype_df` with column of cell types, as factors, for each annotation method
+create_celltype_df <- function(processed_sce, has_singler, has_cellassign) {
+  celltype_df <- colData(processed_sce) |>
+    as.data.frame() |>
+    # barcodes to a column
+    tibble::rownames_to_column(var = "barcode") |>
+    # keep only cell name, celltyping, and clusters
+    dplyr::select(
+      barcode,
+      clusters,
```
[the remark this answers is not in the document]
sjspielman: It's the clusters - 
https://github.com/AlexsLemonade/scpca-nf/blob/fb5f3bf1d9dad7533046cf05230a5cf0b3ddabaf/bin/cluster_sce.R#L87-L97

Was this your question?

templates/qc_utils/celltype_functions.R
sjspielman: > If we are reusing this style frequently across different table types, it might make sense as a separate function that we define at the top and reuse as needed. (I think we duplicate this a lot)

:100:^:100:

templates/qc_utils/celltype_functions.R
sjspielman: So yeah.. It used to be more similar to what you describe, but I changed in response to an earlier review I received: https://github.com/AlexsLemonade/scpca-nf/commit/a674f4a854b94936233730e3e04ded2205274f5c

templates/qc_utils/celltype_functions.R
sjspielman: It currently doesn't exist anymore after some changes to `classify_SingleR.R`, but previously it came from the script `classify_SingleR.R`.  The `has_singler` variable is defined in `qc_report.rmd` based on whether `singler` is present in the sce metadata; see the last lines of this code excerpt (have to scroll). 

Here's the most recent version of `classify_SingleR.R` that saved the annotations to SCE, and I had been thinking this same code would end up in the forthcoming process/R script to add the celltypes - 
https://github.com/AlexsLemonade/scpca-nf/blob/b20fcb1174686e73ad3857b7e01e98ff86f83697/bin/classify_SingleR.R#L92-L120


So that column should definitely exist, unless it got filled with NULL. I wonder if the right place for that check is actually in the forthcoming R script that will add annotations to the SCE?

templates/qc_utils/celltype_functions.R
jashapiro: That seems to have been before piping `celltype_df[[]]` into `forcats`; But my real problem is the two data frames. No need for that!

templates/qc_utils/celltype_functions.R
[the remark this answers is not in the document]
jashapiro: Yes, but I actually also have a problem with it being named `$clusters`: It should be `$cluster` unless we are assigning one cell to multiple clusters.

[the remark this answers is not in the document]
sjspielman: Ok, I can open a separate issue for that!

templates/qc_utils/celltype_functions.R
jashapiro: I don't understand what you are saying above? The column doesn't exist? Or it should exist? Either way, my point was that I would expect the `singler_celltype_annotation` column to only exist in the case that `has_singler` is true, but that makes that information redundant: we could have a simpler function with no `has_singler` argument if we do something like this:
```
if("singler_celltype_annotation" %in% names(.)){
      prepare_annotation_values(singler_celltype_annotation)
}
```

Alternatively, we could make `prepare_annotation_values` smarter and have it do nothing if the column doesn't exist.

templates/qc_utils/celltype_functions.R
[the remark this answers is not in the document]
sjspielman: fyi https://github.com/AlexsLemonade/scpca-nf/issues/484

templates/qc_utils/celltype_functions.R
sjspielman: > I don't understand what you are saying above?

Just showing where that column was created. It doesn't exist _today_, but it did exist when this code was originally written so I expect it to exist again when we finish celltyping. **punchline: it SHOULD exist once singler cell types are added into the SCE.** 

Yes, there is redundancy, but `has_singler` is also used more generally to control logic in the QC report. So, while we don't have to use `has_singler` _here_, I don't want to get rid of that variable. We could have column checks in `prepare_annotation_values` instead, and I wonder if you have some tidyverse fanciness in mind re- "smarter"; all I can think of are moving some `if`s into the function directly.

jashapiro: Right, I'm just for making the function call simpler: I wasn't saying that `has_singler` isn't useful in other contexts!

No fanciness, just moving the column check to the `prepare_annotation_values` to remove redundancy. If we really want to go whole hog, you could make `prepare_annotation_values` take a vector of column names and process them all together (for the column names that exist).

sjspielman:
Ok, we're getting some stochastic stub errors https://github.com/AlexsLemonade/scpca-nf/pull/479#issuecomment-1745592839

templates/qc_utils/celltype_functions.R
sjspielman: for now I'm splitting out the `kable` code into its own function, but more generally, see https://github.com/AlexsLemonade/scpca-nf/issues/371#issuecomment-1747011384.

sjspielman:
@jashapiro this should be ready for another look! I've made quite a few changes in response to review so I won't list them all here, but the commits are each very focused 💪 to serve as that "change list." Noting there's also a for loop that made it in for this round but I think it was a judiciously-used one, let me know if you have other thoughts!

templates/qc_report/celltypes_qc.rmd
```diff
@@ -516,149 +273,10 @@ plot_umap(umap_df,
...
 ```{r eval = has_submitter, message=FALSE, warning=FALSE}
 plot_umap(umap_df,
-  submitter,
+  submitter_celltype_annotation_lumped,
   "Cell types",
   legend_nrow = 4
 ) +
   ggtitle("UMAP colored by submitter-provided annotations") +
   scale_color_brewer(palette = "Dark2")
 ```
```
jashapiro, line 282: Minor, but I think this should probably be first?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -0,0 +1,434 @@
...
+delta_median_df <- tibble::tibble(
+  delta_median = rowMaxs(singler_scores) - rowMedians(singler_scores),
+  # Need to grab the non-pruned label for this plot
+  full_labels = metadata(processed_sce)$singler_result$labels,
+  # TRUE pruned values ==> low confidence
+  # so, negate for this variable:
+  confident = !is.na(metadata(processed_sce)$singler_result$pruned.labels)
+)
+
+# If ontologies were used for `full_labels`, we'll need to map back to cell names
```
jashapiro, line 100:
```suggestion
# If ontologies were used for `full_labels`, we'll need to map back to cell type names
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -0,0 +1,434 @@
...
+
+# extract scores into matrix
+singler_scores <- metadata(processed_sce)$singler_result$scores
+
+# Create data frame for plotting with delta median and the full *non-pruned* cell labels
+delta_median_df <- tibble::tibble(
+  delta_median = rowMaxs(singler_scores) - rowMedians(singler_scores),
+  # Need to grab the non-pruned label for this plot
+  full_labels = metadata(processed_sce)$singler_result$labels,
+  # TRUE pruned values ==> low confidence
```
jashapiro, line 95:
```suggestion
  # if pruned.labes are NA ==> low confidence
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -0,0 +1,434 @@
...
+        celltype = celltype_df$singler_celltype_annotation
+      ) |> dplyr::distinct()
+    ) |>
+    dplyr::select(-full_labels)
+} else {
+  # otherwise, full_labels already contain what we want to plot, so just rename it
+  delta_median_df <- delta_median_df |>
+    dplyr::rename(celltype = full_labels)
+  # still need to add levels though:
+  # unknown will still be present, but it's not in the data, so it won't matter
```
jashapiro, line 118: This comment makes me think I am still quite confused about what the manipulations that are going on here really are... What is unknown here? The cell type? What is the data that it's not in?

jashapiro: Oh, you mean "unknown" will be a factor level?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -0,0 +1,434 @@
...
+  levels(delta_median_df$celltype) <- levels(celltype_df$singler_celltype_annotation)
+}
+
+# Ensure we have no "Unknown cell type" values left:
+if (sum(delta_median_df$celltype == "Unknown cell type") > 0) {
```
jashapiro, line 123:
```suggestion
if (any(delta_median_df$celltype == "Unknown cell type")) {
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -1,5 +1,125 @@
...
+  for (method in methods) {
+    colname <- glue::glue("{method}_celltype_annotation")
+    if (colname %in% names(df)) {
+      lumped_colname <- glue::glue("{colname}_lumped")
+        df[[lumped_colname]] <- forcats::fct_lump_n(
+          df[[colname]],
+          n_celltypes,
+          other_level = "Other cell type"
+        )
+    }
```
jashapiro, line 71: So, this is fine, but maybe something like:
```
mutate(across(
  ends_with("_celltype_annotation"),
  \(x) forcats::fct_lump_n(x, n_celltypes, other_level = "Other cell type"),
  .names = "{.col}_lumped"
))

**Review by jashapiro — APPROVED** [review]
This looks fine to me. 

I probably would have a few more suggestions if I went through every line, but for the general task of breaking things up, it should be fine. 

But, I couldn't really handle the for loop, so I made a very tidy suggestion of an alternative.

File: templates/qc_report/qc_utils/celltype_functions.R [hunk 15]
```diff
@@ -0,0 +1,63 @@
...
+#' @return `celltype_df` with column of cell types, as factors, for each annotation method
+create_celltype_df <- function(processed_sce) {
+  celltype_df <- processed_sce |>
+    scuttle::makePerCellDF(use.dimred = "UMAP") |>
+    # only keep columns of interest & rename UMAP columns
+    dplyr::select(
+      barcodes,
+      clusters,
+      UMAP1 = UMAP.1,
+      UMAP2 = UMAP.2,
```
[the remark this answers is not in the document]
sjspielman: Oh this is a good point, I will confirm I wasn't using outdated RDS files for testing this

templates/qc_report/celltypes_qc.rmd
```diff
@@ -1,5 +1,125 @@
...
+      `Percent of cells`
+    ) 
+}
+
+#' Format tables of cell type counts as kable
+#'
+#' @param df Data frame to format
+#'
+#' @return kable table of cell type counts
+format_celltype_n_table <- function(df) {
```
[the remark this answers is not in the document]
sjspielman: I'd favor circling back when we _really_ upgrade tables - https://github.com/AlexsLemonade/scpca-nf/issues/371

templates/qc_report/qc_utils/celltype_functions.R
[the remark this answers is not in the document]
sjspielman: Ok, this seems to be necessary _because_ of  `scuttle::makePerCellDF()`.

- Recent code change due to the BioC 3.17 bump means UMAPs are now actually named `UMAP1` and `UMAP2`: https://github.com/AlexsLemonade/scpca-nf/pull/450/files
- But, `scuttle::makePerCellDF()` appears to rename to use periods (see line 96): https://code.bioconductor.org/browse/scuttle/blob/devel/R/makePerCellDF.R#L96
  - That said based on the code, I might have thought these names would end up as `UMAP1.1`? 

I'm not 100% sure this won't break later, but if that does happen it will be easy to fix, so maybe leave it..

templates/qc_report/celltypes_qc.rmd
sjspielman: I'd like to keep things parallel, so if we move this first (but still after clusters?) I'll also move the count table and heatmap to be first in their respective sections.

templates/qc_report/qc_utils/celltype_functions.R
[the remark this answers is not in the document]
jashapiro: maybe `rename_with(\(x) stringr::str_replace(x, "^UMAP\\.", "UMAP")` to handle either way?

templates/qc_report/celltypes_qc.rmd
jashapiro: Makes sense. I just happened to notice it here! (Just to be explicit, even though I think we are on the same page: the thing I am commenting on is the having the submitter labels first)

sjspielman: I thought we were on the same page but now I'm not sure 😂 

> having the submitter labels first

I intepretted this to mean the actual plot, right? Not some kind of axis labeling?

jashapiro: Yes. The plot. Basically I think we should put any display of labels we have from submitters before the automated* labels from our workflow. Which is why it makes sense for the count tables and heatmaps to be parallel, as you said.

*Yes, we can still put clusters first... which are also automated.

sjspielman: same page ✅

sjspielman:
Feeling good about #487 with all these green checks!

PR #490, opened 2023-10-05
Add ontology/celltype mapping to SingleR

### LINKED ISSUE #485: Rethink SingleR process
While working towards the process to add celltypes to the SCE as part of #415, I found some spots where I think we need to rethink:

- First, I don't think we need to export a separate TSV file of annotations if we're already exporting the _whole_ singler result object! That information is all in the full object, and there's no need for a separate file. Seems more efficient to just retain the DataFrame and grab what we need from it when adding to SCE.

- Second, we need to retrieve some deleted information SingleR process & R code. Originally I removed this code that maps ontology labels to cell types names since that process was no longer adding to the SCE; I figured I'd bring that code back for the add celltypes process: https://github.com/AlexsLemonade/scpca-nf/blob/b20fcb1174686e73ad3857b7e01e98ff86f83697/bin/classify_SingleR.R#L92-L110
  However, this code _requires_ having access to the singler model file - `singler_model$cell_ontology_df`. There are two immediate ways I can think of to handle this:
  - Make `singler_model_file` an input to the add celltypes process (and necessarily the label name)
  - Or, _do not cancel_ the separate TSV file for singler annotations as I suggested above, but instead keep it and do the ontology/cell names mapping in the classify singler script. Then, the TSV will have this mapping taken care of and and we can simply join it into the coldata. I think I prefer this option?

- A related point is that we could also consider _enforcing_ using `label.ont` reference labels with SingleR, but this makes me nervous in case we end up having to use something other than celldex one day. Currently this is just a default workflow parameter.

CC @jashapiro

sjspielman:
Closes #485 

This PR gets ontology mapping back into the SingleR script and updates the script's exports:
- We still export the TSV, but with with another column for ontology IDs
- That data frame is added to the metadata of the singler result. This is somewhat different from the suggestion to add `singler_model$cell_ontology_df` to the metadata (https://github.com/AlexsLemonade/scpca-nf/issues/485#issuecomment-1746927716), but seemed like it might be more efficient for later on in case we use it? That said, it does start to bloat this object quite a bit with redundant information, so I'm not 100% sure in my decision. Eh?
- I had seen one spot when working on #483 that I meant to tweak, did it here. I realize the full labels are in the colData directly, so just grab them without dipping into the metadata - https://github.com/AlexsLemonade/scpca-nf/commit/71cea81a593232b09c8290dae645ebe7a0ed494f

### DIFF
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -22,6 +22,12 @@ option_list <- list(
             File name is expected to be in form: <model name>_model.rds."
   ),
   make_option(
+    opt_str = c("--singler_label_name"),
+    type = "character",
+    default = "label.ont",
+    help = "label used when building the SingleR reference"
+  ),
+  make_option(
     opt_str = c("--output_singler_annotations_file"),
     type = "character",
     help = "path to output TSV file that will store the SingleR annotations. Must end in .tsv"
@@ -100,20 +106,41 @@ singler_results <- SingleR::classifySingleR(
   BPPARAM = bp_param
 )
 
-# add reference name as metadata in singler_results DataFrame
+# add reference name to singler_results DataFrame metadata
 metadata(singler_results)$reference_name <- reference_name
 
 
-# export results
-
-# first, a stand-alone tsv of annotations with both pruned and full labels
-readr::write_tsv(
-  tibble::tibble(
-    barcode = rownames(singler_results),
-    pruned_labels = singler_results$pruned_labels
-  ),
-  opt$output_singler_annotations_file
+# create data frame of annotations
+annotations_df <- tibble::tibble(
+  barcode = rownames(singler_results),
+  singler_celltype_annotation = singler_results$labels,
 )
+
+# map ontology labels to cell type names, as needed
+if (opt$singler_label_name == "label.ont") {
+  
+  # end up with columns: barcode, singler_celltype_annotation, singler_celltype_ontology
+  annotations_df <- annotations_df |>
+    dplyr::left_join(
+      # column names: ontology_id, ontology_cell_names
+      singler_model$cell_ontology_df, 
+      by = c("singler_celltype_annotation" = "ontology_id")
+    ) |> 
+    # rename columns
+    dplyr::rename(
+      singler_celltype_ontology = singler_celltype_annotation,
+      singler_celltype_annotation = ontology_cell_names
+    )
+  
+} 
+
+# add annotations_df to singler_results DataFrame metadata, just in case
+metadata(singler_results)$annotations_df <- annotations_df
+
+# export results ---------------
+
+# first, a stand-alone tsv of annotations
+readr::write_tsv(annotations_df, opt$output_singler_annotations_file)
 
 # next, the full result to a compressed rds
 readr::write_rds(
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -22,12 +22,6 @@ option_list <- list(
             File name is expected to be in form: <model name>_model.rds."
   ),
   make_option(
-    opt_str = c("--singler_label_name"),
-    type = "character",
-    default = "label.ont",
-    help = "label used when building the SingleR reference"
-  ),
-  make_option(
     opt_str = c("--output_singler_annotations_file"),
     type = "character",
     help = "path to output TSV file that will store the SingleR annotations. Must end in .tsv"
@@ -113,17 +107,19 @@ metadata(singler_results)$reference_name <- reference_name
 # create data frame of annotations
 annotations_df <- tibble::tibble(
   barcode = rownames(singler_results),
-  singler_celltype_annotation = singler_results$labels,
+  singler_celltype_annotation = singler_results$pruned.labels,
 )
 
 # map ontology labels to cell type names, as needed
-if (opt$singler_label_name == "label.ont") {
-  
+# we can tell if ontologies were used because this will exist:
+cell_ontology_df <- singler_model$cell_ontology_df
+if (!is.null(cell_ontology_df)) {
+
   # end up with columns: barcode, singler_celltype_annotation, singler_celltype_ontology
   annotations_df <- annotations_df |>
     dplyr::left_join(
       # column names: ontology_id, ontology_cell_names
-      singler_model$cell_ontology_df, 
+      cell_ontology_df, 
       by = c("singler_celltype_annotation" = "ontology_id")
     ) |> 
     # rename columns
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -127,10 +127,11 @@ if ("cell_ontology_df" %in% names(singler_model)) {
       singler_celltype_annotation = ontology_cell_names
     )
   
+  # add cell_ontology_df to singler_results DataFrame metadata
+  metadata(singler_results)$cell_ontology_df <- singler_model$cell_ontology_df
 } 
 
-# add annotations_df to singler_results DataFrame metadata, just in case
-metadata(singler_results)$annotations_df <- annotations_df
+
 
 # export results ---------------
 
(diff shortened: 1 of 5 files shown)


bin/classify_SingleR.R
```diff
@@ -100,21 +106,42 @@ singler_results <- SingleR::classifySingleR(
...
+    # rename columns
+    dplyr::rename(
+      singler_celltype_ontology = singler_celltype_annotation,
+      singler_celltype_annotation = ontology_cell_names
+    )
+  
+} 
+
+# add annotations_df to singler_results DataFrame metadata, just in case
+metadata(singler_results)$annotations_df <- annotations_df
```
jashapiro, line 138: This does seem quite redundant, as the data is pulled out of the data frame only to be added back with different labels. I really do think just adding the `cell_ontology_df` is the right move here to keep things compact but allow later translation if needed (which it isn't for this workflow, as we are keeping both in the table, which I think is good).

bin/classify_SingleR.R
```diff
@@ -21,6 +21,12 @@ option_list <- list(
     help = "path to file containing a single model generated for SingleR annotation. 
             File name is expected to be in form: <model name>_model.rds."
   ),
+  make_option(
+    opt_str = c("--singler_label_name"),
+    type = "character",
+    default = "label.ont",
+    help = "label used when building the SingleR reference"
+  ),
```
jashapiro, line 29: For the purposes of this script, I think it makes more sense for this to be a flag rather than a specific label, as the label name is not part of the output object. So this is a very specific case for something that might be a more general use, and doesn't quite signal why we have this option.

```suggestion
  make_option(
    opt_str = c("--singler_uses_ontology"),
    action = "store_true",
    default = FALSE,
    help = "Indicates that the SingleR reference labels in the model are ontology ids. SingleR model object should include `cell_ontology_df`."
  ),
```

You can then test that the second sentence there is true, and error if not.

bin/classify_SingleR.R
```diff
@@ -100,21 +106,42 @@ singler_results <- SingleR::classifySingleR(
...
-  ),
-  opt$output_singler_annotations_file
+# create data frame of annotations
+annotations_df <- tibble::tibble(
+  barcode = rownames(singler_results),
+  singler_celltype_annotation = singler_results$labels,
 )
 
+# map ontology labels to cell type names, as needed
+if (opt$singler_label_name == "label.ont") {
```
jashapiro, line 120: If you take my previous suggestion
```suggestion
if (opt$singler_uses_ontology) {
```

modules/classify-celltypes.nf
```diff
@@ -22,6 +22,7 @@ process classify_singler {
       classify_SingleR.R \
         --sce_file "${processed_rds}" \
         --singler_model_file "${singler_model_file}" \
+        --singler_label_name ${params.singler_label_name} \
```
jashapiro, line 25: Completing the suggestions.
```suggestion
        ${params.singler_label_name == "label.ont" ? "--singler_uses_ontology" : ""} \
```

bin/classify_SingleR.R
jashapiro: > You can then test that the second sentence there is true, and error if not.

Actually... the presence of `cell_ontology_df` is something we can check for and just eliminate the option altogether!

bin/classify_SingleR.R
jashapiro: Better suggestion!
```suggestion
if ("cell_ontology_df" %in% names(singler_model)) {
```

nextflow.config
```diff
@@ -52,6 +52,9 @@ params {
   cluster_weighting = "jaccard" // default weighting scheme for graph-based clustering
   nearest_neighbors = 20 // default nearest neighbors parameter for graph-based clustering
 
+  // Cell type annotation options
+  singler_label_name = "label.ont" // celldex reference label used for SingleR cell tyoe annotation
```
jashapiro, line 56: Originally just a typo fix, but if this is only really used in the reference workflow by my updated suggestions, maybe this?
```suggestion
  singler_label_name = "label.ont" // celldex reference label used for SingleR reference building
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -86,11 +86,10 @@ The red overlayed boxes represent the median ± interquartile range (IQR), speci
 # extract scores into matrix
 singler_scores <- metadata(processed_sce)$singler_result$scores
 
-# Create data frame for plotting with delta median and the full *non-pruned* cell labels
+# Create data frame for plotting with delta median and the full cell labels
```
jashapiro, line 89: So this is or isn't pruned? "full" is quite ambiguous, so I'd rather have a different word... Maybe just `all_labels` to distinguish from pruned? or `unpruned_labels` to be very explicit?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -86,11 +86,10 @@ The red overlayed boxes represent the median ± interquartile range (IQR), speci
...
 delta_median_df <- tibble::tibble(
   delta_median = rowMaxs(singler_scores) - rowMedians(singler_scores),
-  # Need to grab the non-pruned label for this plot
-  full_labels = metadata(processed_sce)$singler_result$labels,
+  full_labels = processed_sce$singler_celltype_annotation,
```
jashapiro, line 92: I would have assumed that `processed_sce$singler_celltype_annotation` was being filled with pruned labels (that step is still not done, right?), which is why you were going back to the original labels for this plot?

jashapiro (commented):
I have two alternative suggestions, but in the interests of simplicity, I like my second one way better!

Since `singler_model$cell_ontology_df` only exists in the case of an ontology-based label, we can use the presence of that slot to decide whether or not to translate, with no need for a separate option or flag!

I also think your second change is maybe premature? We need to see what we are putting into `processed_sce`, but I suspect it will be pruned labels?

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: I'm actually pretty sure I was wrong to do this! I think in my mind I was thinking maybe we would include both full and pruned in the SCE, but then I promptly didn't code that up and did this anyways. So, I'm going to revert it.

sjspielman:
> I also think your second change is maybe premature? We need to see what we are putting into processed_sce, but I suspect it will be pruned labels?

Ah, woops!! I was planning to put _both_ pruned and full labels in there as a point of discussion, but looks like that didn't happen! The main plan has been to present pruned labels, and they can access the full labels from the full singler object in the SCE's metadata.

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: yeah, it was wrong. Not enough lunch.

modules/classify-celltypes.nf
sjspielman: noting unneeded since https://github.com/AlexsLemonade/scpca-nf/pull/490/commits/a854b7ae21030350aff8c865f882d2e910952af0

bin/classify_SingleR.R
sjspielman: noting unneeded since https://github.com/AlexsLemonade/scpca-nf/pull/490/commits/a854b7ae21030350aff8c865f882d2e910952af0

bin/classify_SingleR.R
sjspielman: done - https://github.com/AlexsLemonade/scpca-nf/pull/490/commits/a4d4b006b09a2c57623f8281045c86a7e7ca6650

jashapiro (approved):
So cute and little now!

sjspielman:
_And_ still passing checks 🚀

Issue #480, opened 2023-10-03
Standardize formatting for nextflow files


jashapiro:
Right now there is a bit of a mix of styles in the nextflow files in this repository. Most prominently, we have a mix of two-space and four-space tabs, so we should standardize on one or the other (I prefer 2 here, despite using 4 in Python, I guess because whitespace in nextflow isn't syntactically meaningful). A quick pass through the nextflow files could get this done in a single PR: We should be able to review it with `ignore whitespace` to verify that there are no code changes in that PR.

Looking forward a bit more, we should write up a bit of a style guide/editor setup instructions. I think this would include the use of `black` for Python and `styler`/tidyverse style for R scripts.

We might also include some examples of the way we want to handle `.map{}` functions and the like, `tuple()` vs. `[]`, etc.
These later changes might be more incremental in implementation, but it might be nice to have some guidelines written down. 

I tried to look for existing guidelines in the [nf-core](https://github.com/nf-core/) repositories but largely came up short (their [code formatting](https://nf-co.re/docs/contributing/code_formatting) section is mostly focused on `prettier` and `black`, neither of which actually format nextflow or groovy code). There are obviously conventions but not "guidelines" that I could find.

jashapiro:
When writing the style guide, we might include some notes on groovy oddities, like the fact that variables are global by default, even if defined in a function or closure (like `.map{}`). To avoid this, use `def` to keep the scope local.

jashapiro:
closed by #546

PR #491, opened 2023-10-06
Def all internal variable

jashapiro:
With the knowledge gained in https://github.com/AlexsLemonade/scpca-nf/pull/487, I decided it wouldn't be a bad idea to go through this repo and make sure that all variables declared in a function or `.map{}` statement was defined only in that scope using `def`.

The ones here were not causing problems that we know of, but it seems like a good idea, partly as a bit of reminder that those `def` statements actually do something (keep scope local).

The most common changes are in the definitions of paths for step-skipping, where I now make an explicit copy of the input  object and then modify the copy.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -130,11 +130,11 @@ workflow {
   // read in metadata with all organisms to create references for
   ref_ch = Channel.fromPath(params.ref_metadata)
     .splitCsv(header: true, sep: '\t')
-    .map{[
-      reference_name = "${it.organism}.${it.assembly}.${it.version}",
-      // grab file paths for each organism
-      ref_paths[reference_name]
-    ]}
+    .map{
+      def reference_name = "${it.organism}.${it.assembly}.${it.version}";
+      // reference name & reference file paths for each organism
+      [reference_name, ref_paths[reference_name]]
+    }
     .map{it +[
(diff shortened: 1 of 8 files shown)


sjspielman (approved):
LGTM! I checked all the `.nf` files and I don't see any other spots where `it` is being modified ✅ 

The one question I have is why we don't need `def memory...` throughout `process_base.config`? 

More generally, I wonder if we'd benefit from a teeny bit of docs about remembering to do this. Would this be appropriate to somehow add to `CONTRIBUTING.md`? Either way, I don't see that as necessary for this PR, but maybe a future one.

jashapiro:
> The one question I have is why we don't need `def memory...` throughout `process_base.config`?

Good question! I think this is to do with how the config is processed: While it does contain code, I everything gets done in some kind of eval process such that all of those parameters end up defined in within config config objects. In this case since the `memory` setting is part of a `process` block, that is where it ends up: as `process.memory`.  But I don't know how that config is processed exactly, and based on some of the docs, there are probably additional quirks.

jashapiro:
> More generally, I wonder if we'd benefit from a teeny bit of docs about remembering to do this. Would this be appropriate to somehow add to `CONTRIBUTING.md`? Either way, I don't see that as necessary for this PR, but maybe a future one.

Maybe some kind of tips/practice doc. I think this might be a good thing to tag onto #480

PR #492, opened 2023-10-06
Rename cluster column

### LINKED ISSUE #484: Change cluster column name
Currently we are naming the cluster column `clusters`: 
https://github.com/AlexsLemonade/scpca-nf/blob/fb5f3bf1d9dad7533046cf05230a5cf0b3ddabaf/bin/cluster_sce.R#L88

In https://github.com/AlexsLemonade/scpca-nf/pull/483#discussion_r1344614615, @jashapiro pointed out that this may imply existence of multiple clusters _per cell_, which ain't it. 

>  It should be $cluster unless we are assigning one cell to multiple clusters.

So, let's rename change to `cluster`; this will involve some QC report changes as well.

sjspielman:
**Stacked on #491**

Closes #484, and does what it says! Recommended to hide whitespace during review...

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -234,7 +234,7 @@ clusters_plot <- plot_umap(
   cluster,
   "Cluster"
 ) +
-  ggtitle("UMAP colored by clusters")
+  ggtitle("UMAP colored by cluster identity")
 
 # Determine palette based on number of levels.
 # If we have <=8, we can use a CVD-friendly palette (they generally don't have more than 8 colors).
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -24,7 +24,7 @@ create_celltype_n_table <- function(df, celltype_column) {
       `Annotated cell type` = {{ celltype_column }},
       `Number of cells` = n,
       `Percent of cells`
-    ) 
+    )
 }
 
 #' Format tables of cell type counts as kable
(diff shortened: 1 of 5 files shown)


sjspielman:
In retrospect this probably didn't need to be stacked, but here we are..

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -351,7 +351,7 @@ create_celltype_heatmap <- function(
 knitr::asis_output("### Submitter-provided cell type and cluster heatmap\n")
 create_celltype_heatmap(
   x_vector = celltype_df$submitter_celltype_annotation,
-  y_vector = celltype_df$clusters,
+  y_vector = celltype_df$cluster,
   y_label = "Clusters"
```
jashapiro, line 355: Presumably the label here too (and below)?
```suggestion
  y_label = "Cluster"
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -231,15 +231,15 @@ umap_df <- lump_celltypes(celltype_df)
 ```{r message=FALSE, warning=FALSE}
 clusters_plot <- plot_umap(
   umap_df,
-  clusters,
+  cluster,
   "Clusters"
```
jashapiro, line 235:
```suggestion
  "Cluster"
```

jashapiro (commented):
Looks good, but I think you missed one: 
I think you missed one: https://github.com/AlexsLemonade/scpca-nf/blob/5f98088bb6dd69e74639a13f2991ff9e64dda30f/templates/qc_report/utils/celltype_functions.R#L21-L29

I also suggested changing axis labels. I think that makes sense if the elements are individual cells, but I might be wrong.

templates/qc_report/utils/celltype_functions.R
```diff
@@ -20,7 +20,7 @@ create_celltype_df <- function(processed_sce) {
     # only keep columns of interest
     dplyr::select(
       barcodes,
```
jashapiro, line 22: This bothers me too, but I think it partly comes from somebody naming it elsewhere, so I am okay with leaving it.

jashapiro (approved):
👍🏼

templates/qc_report/utils/celltype_functions.R
sjspielman: You know I can't exactly pinpoint where this happened, so I actually wonder if this is a bug... I'm going to an open another issue to figure it out, probably circling back once all the cell typing is in the main workflow.

sjspielman: > so I actually wonder if this is a bug

To elaborate, I've been using some "testing SCEs" to develop this section of the report while cell typing is in progress, and maybe I made a test SCE incorrectly and therefore popped in a bad column name. But I'm not 100& sure, hence something to revisit.

PR #496, opened 2023-10-09
Render cell type report from sce_qc_report process

### LINKED ISSUE #482: Update qc report process to export separate cell types report
_Broken out of #447_

> We can render this file within the same script that we use to render the full QC report. We just need to modify the process to include two output html files.

We will need to update the process that generates a QC report to also generate the supplemental cell type QC report if cell types are present.

sjspielman:
Closes #482 

This PR adds functionality to also render the supplemental cell types report, if cell types are provided. I implemented this as discussed in comments in #482. I think code is fairly straightforward for review, but let me know if anything is confusing!

Caveats:
-  I did not check that `ggridges` and `ComplexHeatmap` are installed (which would mirror the qc report code in `scpcaTools`), but can add if we really want it!
- `stub` is failing with an error I can't pin down.. The file `qc_report` is defined based on the library id (`qc_report = "${meta.library_id}_qc.html"`) within the process, suggesting (???) that maybe the `library_id` isn't unique, which is a whole thing if true 😬  But hopefully this is just an uninformative error message that might be solved with some fresh eyes!

  ```
  ERROR ~ Error executing process > 'sce_qc_report (1)'

  Caused by:
    Process `sce_qc_report` input file name collision -- There are multiple input files for each of the following file names: qc_report
  ```

- Something I wasn't actually 100% sure about is syntax for checking if a groovy variable is true from bash, which is relevant in the stub. For now I have the following (`== "true"`), which mayyy or may not be correct syntax!
  ```
  if [ ${has_celltypes} == "true"]; 
       touch ${celltype_report}
  fi
  ```


jashapiro (commented):
I am not sure if the error you are seeing is caused by passing in the `report_template_dir` twice, but I kind of suspect it is. The double input that you see as `qc_report` is probably referring to the last bit of that path, not the output variable name.

So I made a suggestion to just pass in the whole thing as a single tuple, which I think will solve the bug and simplify things a bit.

I also suggested how to output the cell type report optionally (so as not to upset nextflow), and little thought about nulls and defaults in R.

sjspielman:
> I am not sure if the error you are seeing is caused by passing in the `report_template_dir` twice, but I kind of suspect it is. 

Actually, I had been passing two separate paths in (e.g.`celltype_report_template_dir` defined as same path value), Simplifying to just pass in `report_template_dir` twice was something I tried as part of debugging... But maybe reverting and combining with some of these other changes it will appease nextflow! The length-3 tuple makes more sense, too.

jashapiro:
> Actually, I had been passing two separate paths in (e.g.`celltype_report_template_dir` defined as same path value), 

Right, I phrased poorly: I meant that you were passing in the same path _value_ for two different variables (or `path` objects).

modules/qc-report.nf
```diff
@@ -9,19 +9,27 @@ process sce_qc_report{
     input:
         tuple val(meta), path(unfiltered_rds), path(filtered_rds), path(processed_rds)
         tuple path(template_dir), val(template_file)
+        tuple path(celltype_template_dir), val(celltype_template_file)
     output:
         tuple val(meta), path(unfiltered_out), path(filtered_out), path(processed_out), path(metadata_json), emit: data
-        path qc_report, emit: report
+        path qc_report, emit: report // TODO: CELLTYPE REPORT OUTPUT?
```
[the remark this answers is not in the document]
sjspielman: seems right... https://github.com/nextflow-io/nextflow/discussions/1980#discussioncomment-558777

bin/sce_qc_report.R
```diff
@@ -45,6 +54,11 @@ option_list <- list(
     type = "character",
     help = "path to QC report output file"
   ),
+  make_option(
+    opt_str = c("--celltype_report_file"),
+    type = "character",
+    help = "path to supplemental cell type QC report output file. Only considered if not NULL."
```
sjspielman, line 60:
```suggestion
    help = "path to supplemental cell type QC report output file. Only considered if not empty string."
```

bin/sce_qc_report.R
```diff
@@ -233,14 +247,45 @@ if (multiplexed) {
...
+  extra_params = list(
+    seed = opt$seed,
+    # this will only be used if cell types exist
+    celltype_report = opt$celltype_report_file
+  )
 )
+
+
+# render supplemental cell types report, if needed
+if (!(is.null(opt$celltype_report_file))) {
```
sjspielman, line 274:
```suggestion
if (opt$celltype_report_file != "") {
```

sjspielman:
Well, we probably shouldn't be referring to `params.perform_celltyping` yet, so that's fixed!

But now it doesn't like this logic statement (with or without the surrounding parentheses). Possibly another `null` problem..


`has_celltypes = meta.submitter_cell_types_file | meta.singler_model_file | meta.cellassign_reference_file`

```

ERROR ~ Error executing process > 'sce_qc_report (1)'

Caused by:
  No signature of method: java.lang.String.or() is applicable for argument types: (null) values: [null]
Possible solutions: is(java.lang.Object), tr(java.lang.CharSequence, java.lang.CharSequence), trim(), md5(), any(), grep() -- Check script './modules/qc-report.nf' at line: 7
```

bin/sce_qc_report.R
```diff
@@ -14,6 +16,13 @@ option_list <- list(
     default = NULL,
     help = "path to rmd template file for report"
   ),
+  make_option(
+    opt_str = c("--celltype_report_template"),
+    type = "character",
+    help = "path to template file for supplemental cell types rmd report.
+    Only used if `celltype_report_file` is not NULL."
```
jashapiro, line 23:
```suggestion
    Only used if `celltype_report_file` is not empty."
```

bin/sce_qc_report.R
```diff
@@ -45,6 +54,12 @@ option_list <- list(
     type = "character",
     help = "path to QC report output file"
   ),
+  make_option(
+    opt_str = c("--celltype_report_file"),
+    type = "character",
+    default = "",
+    help = "path to supplemental cell type QC report output file. Only considered if not empty string."
```
jashapiro, line 61:
```suggestion
    help = "Path to supplemental cell type QC report output file. Only considered if not empty string."
```

bin/sce_qc_report.R
```diff
@@ -14,6 +16,13 @@ option_list <- list(
     default = NULL,
     help = "path to rmd template file for report"
   ),
+  make_option(
+    opt_str = c("--celltype_report_template"),
+    type = "character",
```
jashapiro, line 21: I like defaults, even when they don't really matter.
```suggestion
    type = "character",
    default = NULL,
```

bin/sce_qc_report.R
```diff
@@ -233,14 +248,45 @@ if (multiplexed) {
...
+  
+  # check that the template file exists
+  if (!file.exists(opt$celltype_report_template)) {
+    stop("Supplemental cell types report template not found.")
+  }
+  
+  # render report
+  rmarkdown::render(
+    input = opt$celltype_report_template,
+    output_file = opt$celltype_report_file,
```
jashapiro, line 285: A few safety things copied over from `generate_qc_report`.
My recollection is that there are some weird decisions about where output files are rendered (in the template directory?) so we have to be explicit about both the output file and input files.

And while we are at it, we might as well use tempdirs for everything beforehand and a fresh environment.
```suggestion
    output_file = basename(opt$celltype_report_file),
    output_dir = dirname(opt$celltype_report_file),
    intermediates_dir = tempdir(),
    knit_root_dir = tempdir(),
    envir = new.env(),
```

modules/qc-report.nf
```diff
@@ -58,11 +69,18 @@ process sce_qc_report{
         processed_out = "${meta.library_id}_processed.rds"
         qc_report = "${meta.library_id}_qc.html"
         metadata_json = "${meta.library_id}_metadata.json"
+        
+        // TODO: add `params.perform_celltyping & ` when implemented
+        has_celltypes = meta.submitter_cell_types_file || meta.singler_model_file || meta.cellassign_reference_file
```
jashapiro, line 74: We should make sure we are testing this with at least one stub project that does not have any of these values.

modules/qc-report.nf
```diff
@@ -58,11 +69,18 @@ process sce_qc_report{
         processed_out = "${meta.library_id}_processed.rds"
         qc_report = "${meta.library_id}_qc.html"
         metadata_json = "${meta.library_id}_metadata.json"
+        
+        // TODO: add `params.perform_celltyping & ` when implemented
```
jashapiro, line 73: Any reason not to just add this now? Maybe not, as we might want to add some things to the test action at the same time, but it might good to just get the logic done all at once here.
```suggestion
        // TODO: add `params.perform_celltyping && (...) ` when implemented
```

modules/qc-report.nf
```diff
@@ -58,11 +69,18 @@ process sce_qc_report{
...
+        has_celltypes = meta.submitter_cell_types_file || meta.singler_model_file || meta.cellassign_reference_file
+        celltype_report = "${meta.library_id}_celltype-report.html" // rendered HTML
+
         """
         touch ${unfiltered_out}
         touch ${filtered_out}
         touch ${processed_out}
         touch ${qc_report}
+        ${has_celltypes ? "touch ${celltype_report}" : ""}
```
jashapiro, line 82: Since it is hard to check this with GHA, it is probably worth confirming that the files were actually created & published in a local stub run

jashapiro (commented):
Looks good, with a few minor safety and other suggestions. 

I'd make sure we are doing a few additional tests, and some of them local just for easier confirmation.

bin/sce_qc_report.R
sjspielman: I left this lowercase for consistent style since all other help strings here are lowercase. Can make all uppercase though?

modules/qc-report.nf
sjspielman: I think since we have to make a couple more associated changes I might prefer to wait, and instead add to that issue to do these TODOs. Eg, I imagine the stub may have to change to include `--perform_celltyping` for testing, and I don't want that to get muddled in this PR.

modules/qc-report.nf
sjspielman: Ok, indeed. This line is definitely not working as expected, which in retrospect makes sense this is of groovy code in a shell block; the code does not execute whatever the true/false result is.

Looking though an output directory, aptly named called `outtt`...


#### Do we have all the expected checkpoint files from `checkpoints/celltype`?

_yes_. the stub is currently only exported JSON files.

```
├── STUBL01
│   ├── STUBL01_cellassign
│   │   └── scpca-meta.json
│   └── STUBL01_singler
│       └── scpca-meta.json
├── STUBL02
│   └── STUBL02_cellassign
│       └── scpca-meta.json
├── STUBL05
│   ├── STUBL05_cellassign
│   │   └── scpca-meta.json
│   └── STUBL05_singler
│       └── scpca-meta.json
├── STUBL06
│   ├── STUBL06_cellassign
│   │   └── scpca-meta.json
│   └── STUBL06_singler
│       └── scpca-meta.json
├── STUBL10
│   ├── STUBL10_cellassign
│   │   └── scpca-meta.json
│   └── STUBL10_singler
│       └── scpca-meta.json
├── STUBL11
│   ├── STUBL11_cellassign
│   │   └── scpca-meta.json
│   └── STUBL11_singler
│       └── scpca-meta.json
└── STUBL12
    ├── STUBL12_cellassign
    │   └── scpca-meta.json
    └── STUBL12_singler
        └── scpca-meta.json
```


#### How about expected results files?

_NO_. Here's an abbreviated output from `results/`:

Something else to note here is that for projects without CITEseq, we're also getting `_null.hdf5` files. @allyhawkins do you expect this at this stage of development?

```
├── STUBP01
│   └── STUBS01
│       ├── STUBL01_filtered.rds
│       ├── STUBL01_filtered_null.hdf5
│       ├── STUBL01_filtered_rna.hdf5
│       ├── STUBL01_metadata.json
│       ├── STUBL01_processed.rds
│       ├── STUBL01_processed_null.hdf5
│       ├── STUBL01_processed_rna.hdf5
│       ├── STUBL01_qc.html
│       ├── STUBL01_unfiltered.rds
│       ├── STUBL01_unfiltered_null.hdf5
│       └── STUBL01_unfiltered_rna.hdf5
├── STUBP02
│   ├── STUBP02_bulk_metadata.tsv
│   ├── STUBP02_bulk_quant.tsv
│   └── STUBS02
│       ├── STUBL02_filtered.rds
│       ├── STUBL02_filtered_CITEseq.hdf5
│       ├── STUBL02_filtered_rna.hdf5
│       ├── STUBL02_metadata.json
│       ├── STUBL02_processed.rds
│       ├── STUBL02_processed_CITEseq.hdf5
│       ├── STUBL02_processed_rna.hdf5
│       ├── STUBL02_qc.html
│       ├── STUBL02_unfiltered.rds
│       ├── STUBL02_unfiltered_CITEseq.hdf5
│       └── STUBL02_unfiltered_rna.hdf5
```


#### Attempted solutions

```
# This ends up ALWAYS evaluating to true, and we do end up with cell type HTMLs. So, at least the `emit` part is working!!

## shell part:
# I've tried testing whether this is equal to all flavors of "true" or "1", etc., and it leads to an error each time
if [[ ${has_celltypes} ]]; then
  touch ${celltype_report}
fi
```

```
# this made `touch` sad to not touch anything and threw an error

## groovy part:
celltype_report = has_celltypes ? "${meta.library_id}_celltype-report.html" : ""

## shell part:
touch ${celltype_report}
```


```
# hackish solution for testing/exploring - no error, but also no celltype report in the output

## groovy part:
celltype_report = has_celltypes ? "${meta.library_id}_celltype-report.html" : "removeme.txt"

## shell part:
touch ${celltype_report}
rm -f removeme.txt

```

jashapiro: > Ok, indeed. This line is definitely not working as expected, which in retrospect makes sense this is groovy code in a shell block; the code does not execute whatever the true/false result is.

The fact that it is groovy code is not the issue: that substitution and the execution of the groovy code is happening _before_ it is submitted as a shell script. Something else must be going on.

jashapiro: Ah, I think the reason this is "failing" is that what is going into the `sce_qc_report` file at the moment is not coming out of the actual  annotation workflow, since we are short-circuiting that subworkflow. So nothing has annotation files in meta, so nothing is getting a `true` value for `has_celltypes`

sjspielman: I just tested this, and the cell type report was present (and not present when set to `false`) as expected, so this is consistent with your comment ☝️ 

```

        has_celltypes = true // and then I tested with false!
        celltype_report = "${meta.library_id}_celltype-report.html" // rendered HTML

        """
        ${has_celltypes ? "touch ${celltype_report}" : ""}
        """

```

modules/qc-report.nf
sjspielman: I ended up adding a whole new stub project with full NAs for cell typing (submitter and annotation methods), and you'll also see that I added some missing run files that didn't make in when they should have.

jashapiro:
Pushing a fix to the `_null.hdf5` issue, which is presumably just the stub. cc @allyhawkins

sjspielman:
Confirming that on a local run, https://github.com/AlexsLemonade/scpca-nf/pull/496/commits/e368a073cf3caaa373e608077bebd2a80ed575da cleans up null files:


```
stephanie:~/scpca-nf$ ls outtt/results/STUBP01/STUBS01/
STUBL01_filtered.rds       STUBL01_processed.rds       STUBL01_unfiltered.rds
STUBL01_filtered_rna.hdf5  STUBL01_processed_rna.hdf5  STUBL01_unfiltered_rna.hdf5
STUBL01_metadata.json      STUBL01_qc.html
```

This directory has nulls before in https://github.com/AlexsLemonade/scpca-nf/pull/496#discussion_r1352860771

modules/classify-celltypes.nf
```diff
@@ -198,14 +198,14 @@ workflow annotate_celltypes {
...
+     //  export_channel = processed_sce_channel
+    //     .map{[it[0]["library_id"]] + it}
+     //    // add in unfiltered and filtered sce files
+    //     .join(sce_files_channel.map{[it[0]["library_id"], it[1], it[2]]},
+    //           by: 0, failOnMismatch: true, failOnDuplicate: true)
+         // rearrange to be [meta, unfiltered, filtered, processed]
+    //     .map{library_id, meta, processed_sce, unfiltered_sce, filtered_sce ->
+    //         [meta, unfiltered_sce, filtered_sce, processed_sce]}
 
     emit: sce_files_channel
```
jashapiro, line 210: I think it may be good to uncomment/update this chunk as part of this PR: Testing with this (which has the augmented metadata) does give the expected results for the QC report (at least for me).
```suggestion
      export_channel = celltype_input_ch
        .map{[it[0]["library_id"]] + it}
        // add in unfiltered and filtered sce files
        .join(sce_files_channel.map{[it[0]["library_id"], it[1], it[2]]},
              by: 0, failOnMismatch: true, failOnDuplicate: true)
        // rearrange to be [meta, unfiltered, filtered, processed]
        .map{library_id, meta, processed_sce, unfiltered_sce, filtered_sce ->
            [meta, unfiltered_sce, filtered_sce, processed_sce]}

    emit: export_channel
```

sjspielman:
> Testing with this (which has the augmented metadata) does give the expected results for the QC report (at least for me).

To confirm, do you mean you _are_ seeing a cell type HTML being outputted where appropriate at https://github.com/AlexsLemonade/scpca-nf/pull/496/commits/655162c149542fcf8f1370d4e41f97bf7e64a444?

I am still not seeing it (noting I deleted the cache and previous output directory first before a local re-run after that commit). I would have thought that, if the output channel is being created from `celltype_input_ch`, then nextflow might skip the actual singler and cellassign steps as unnecessary, which is consistent with what I'm seeing, e.g.:

```
├── STUBP01
│   └── STUBS01
│       ├── STUBL01_filtered.rds
│       ├── STUBL01_filtered_rna.hdf5
│       ├── STUBL01_metadata.json
│       ├── STUBL01_processed.rds
│       ├── STUBL01_processed_rna.hdf5
│       ├── STUBL01_qc.html
│       ├── STUBL01_unfiltered.rds
│       └── STUBL01_unfiltered_rna.hdf5
```

main.nf
```diff
@@ -225,12 +225,13 @@ workflow {
...
-
-
+  annotate_celltypes(cluster_sce.out)
+  
   // generate QC reports
-  sce_qc_report(annotate_celltypes.out, report_template_tuple)
+  sce_qc_report(
+    annotate_celltypes.out, 
+    report_template_tuple
+  )
```
jashapiro, line 234: Since we are back to just the two items, let's keep it compact.
```suggestion
  sce_qc_report(annotate_celltypes.out, report_template_tuple)
```

jashapiro (approved):
I'm not sure we needed to add a whole other project for the testing, but it doesn't hurt either. We probably should add an NA for just cellassign too though, as we have for singler. (Arguably that condition is more important, as we are planning to do singler for all projects). I'd just modify `STUBP03` in `stub-celltype-project-metadata.tsv`?

jashapiro:
> > Testing with this (which has the augmented metadata) does give the expected results for the QC report (at least for me).
> 
> To confirm, do you mean you _are_ seeing a cell type HTML being outputted where appropriate at [655162c](https://github.com/AlexsLemonade/scpca-nf/commit/655162c149542fcf8f1370d4e41f97bf7e64a444)?
> 
> I am still not seeing it (noting I deleted the cache and previous output directory first before a local re-run after that commit). I would have thought that, if the output channel is being created from `celltype_input_ch`, then nextflow might skip the actual singler and cellassign steps as unnecessary, which is consistent with what I'm seeing, e.g.:
> 
> ```
> ├── STUBP01
> │   └── STUBS01
> │       ├── STUBL01_filtered.rds
> │       ├── STUBL01_filtered_rna.hdf5
> │       ├── STUBL01_metadata.json
> │       ├── STUBL01_processed.rds
> │       ├── STUBL01_processed_rna.hdf5
> │       ├── STUBL01_qc.html
> │       ├── STUBL01_unfiltered.rds
> │       └── STUBL01_unfiltered_rna.hdf5
> ```

I am indeed getting a `STUBL01_celltype-report.html` in that directory on a fresh  of the current branch, which is what I expected.

Nextflow only looks "forward", so the fact that the output channel doesn't use the cell assignment results will not affect whether those processes run (if it did, we would have seen nothing at all run when outputting the input channel!).

PR #501, opened 2023-10-10
Add in changes for v0.6.0 to main

allyhawkins:
This PR syncs main up with the AnnData changes and any cell type changes that occurred prior to actually running the cell type workflow in `main.nf`. All commits between the previous release and 9/28 are included and then I picked out a few commits between 9/28 and today that are relevant: 

1. Changes from #473 
2. Changes from #487 
3. Commit https://github.com/AlexsLemonade/scpca-nf/pull/496/commits/e368a073cf3caaa373e608077bebd2a80ed575da

I did a test run and things worked successfully, but that's with `edge` for `scpcaTools`. I'm going to do a release of `scpcaTools` since we now have new functions that handle adding in metadata before and then update the version before doing a release of `scpca-nf`.


allyhawkins:
Okay I've now updated this to include the version bump in both `scpca-nf` and `scpcaTools`. This should be ready for review now.

bin/sce_qc_report.R
```diff
@@ -140,16 +146,20 @@ processed_sce_meta <- metadata(processed_sce)
 sample_ids <- unlist(stringr::str_split(opt$sample_id, ",|;")) |> sort()
 
 # check for multiplexing
-multiplexed = if (length(sample_ids) > 1){TRUE} else {FALSE}
+multiplexed <- if (length(sample_ids) > 1) {
+  TRUE
+} else {
+  FALSE
+}
```
jashapiro, line 153: I don't know if this is updated in the main branch, but it should be!

```suggestion
multiplexed <- length(sample_ids) > 1
```

main.nf
```diff
@@ -206,15 +214,21 @@ workflow {
 
   // **** Post processing and generate QC reports ****
   // combine all SCE outputs
-  // Make channel for all library sce files & run QC report
+  // Make channel for all library sce files
   all_sce_ch = sce_ch.no_genetic.mix(genetic_demux_sce.out)
   post_process_sce(all_sce_ch)
 
+  // Cluster SCE and export RDS files to publishDir
```
jashapiro, line 221: We don't publish here anymore
```suggestion
  // Cluster SCE
```

main.nf
```diff
@@ -206,15 +214,21 @@ workflow {
...
+  cluster_sce(post_process_sce.out)
+
   // generate QC reports
```
jashapiro, line 224:
```suggestion
  // generate QC reports and publish .rds files with SCEs.
```

jashapiro (approved):
I think this all looks fine, with a few minor changes that I noticed as I scanned through. Some of these are things that should to be updated in `development` as well (or instead of requiring a separate test here) and are hopefully pretty small adjustments.

PR #503, opened 2023-10-11
Process for adding cell types to SCE object

### LINKED ISSUE #495: Update process for adding annotations to SCE object
We will need to update the existing `add_celltypes_to_sce` process within the `annotate_celltypes` workflow to take as input both the sce object, singleR results, singleR annotations, cell assign annotations, and cell assign predictions matrix. 

The process should run a script that does the following: 

- adds singleR annotations to `colData` 
- adds singleR object to `metadata`
- adds CellAssign annotations to `colData`
- adds CellAssign prediction mtx to `metadata`
- adds a note about which cell type methods were used to the `metadata`

When doing this we will also need to update the output of the `annotate_celltypes` workflow to include meta, unfiltered, filtered, and processed rds. We want the `annotate_celltypes.out` to be passed directly as input to the qc report process in the main workflow.

allyhawkins:
Closes #495 

This PR updates the previous process and script that we had for incorporating cell type annotations into the processed SCE object. 

- I updated the actual Nextflow process and workflow to take as input the processed rds, the`singler_dir`, and the`cellassign_dir`. To account for potentially empty results from `SingleR` or `CellAssign`, I made the inputs to the R script optional and added a check for `singler_present` and `cellassign_present` to the process.
- I am also filtering out any libraries missing cell type results for both `SingleR` and `CellAssign`. Otherwise, we get a conflicting file error since we have two inputs that are the exact same (two copies of empty file). 
- The script we used previously only incorporated the annotations from `CellAssign`. So, I modified this script to take in both `CellAssign` and `SingleR` results and add the annotations to the SCE object. 
  - For `SingleR`, I decided to just read in the `SingleR` object that we output in `classify_singler`. If the`SingleR` object is provided, then I pull out the pruned labels and then cell ontology if present and add them to the SCE object. I also add the `SingleR` results to the metadata.  One thing to note here is that we previously were outputting both the annotations tsv and the `SingleR` results as a checkpoint. But I don't think we need to have both. If we are outputting the entire `SingleR` object then we can directly grab the annotations from that object. THis would mean we would take out all of the wrangling of labels in  the `classify_singler.R` script and have it in the `add_celltypes_to_sce.R` script. I'm curious what others think about this approach.
  - For `CellAssign`, I'm reading in the prediction matrix.  If CellAssign predictions are provided, then I add those annotations to the SCE object and also add the prediction matrix to the metadata. 

I'm running a test run now, but there was a syntax error in the `classify_cellassign` process that I had to fix. So it's taking a little bit of time to get through everything. I did run the stubs locally and everything looked good.


jashapiro (commented):
I left a number of not very large suggestions below. 

The issue about passing in `NO_FILE` more than once is interesting... We are lucky enough to only be passing in two files at this point, but it is interesting to consider we might have to include more such files at some point. I wonder if rather than filtering we should start to account for that possibility now, and create a separate `NO_FILE.1` file. Or we could be even more specific with `NO_FILE.cellassign` and `NO_FILE.singler` Then we could change the checks to be something like `"${singler_dir.baseName}" != "NO_FILE"`. 

I still think it makes sense to skip the process entirely if there are no annotations to be added, so this is fine for now, but it is something we may need in the future if we are passing in something like 1 of 3 annotation files. The main reason to implement it now is so we don't have to figure out the pattern later.

jashapiro:
I'm curious about why you moved the functions in eb114eb? I think the logic of having them there is that those functions are used in both the main QC report and the supplemental, but are they not getting loaded properly for some reason?

allyhawkins:
> I'm curious about why you moved the functions in https://github.com/AlexsLemonade/scpca-nf/commit/eb114ebb6148315d0f8f38115c2058f8119311cd? I think the logic of having them there is that those functions are used in both the main QC report and the supplemental, but are they not getting loaded properly for some reason?

Yes it looks like Nextflow is having trouble finding them. I had just moved them for now to test that the workflow runs, and then was going to circle back to why that might be the case. I didn't realize that they were in the supplemental cell type report too, so either way I'm getting the same error loading the file right now. 
Screenshot 2023-10-12 at 8 38 48 AM

One thought I had was to make the functions file an argument to the script that renders the template. That way we can specify the file path relative to the `$projectDir` in the workflow.

jashapiro:
> Yes it looks like Nextflow is having trouble finding them. I had just moved them for now to test that the workflow runs, and then was going to circle back to why that might be the case. I didn't realize that they were in the supplemental cell type report too, so either way I'm getting the same error loading the file right now. Screenshot 2023-10-12 at 8 38 48 AM
> 
> One thought I had was to make the functions file an argument to the script that renders the template. That way we can specify the file path relative to the `$projectDir` in the workflow.

Yeah, that makes some sense... I wasn't worried about it because we had no trouble with the included "child" files, but I guess that `source`d files are handled differently from Rmarkdown includes. I think what we may want to do is to use that functionality to include files, so rather than including the functions directly, we can replace the `source()` command with blocks like

````
```{r, child='celltype_functions.Rmd'}
```
````

(I think these do have to be Rmd docs, but maybe the raw R will work?)

**Comment by allyhawkins** [comment]
This should be ready for review now. I addressed your earlier comments and I changed using `source` for the `utils` to using a child report. Note that I did have to change that file to `.rmd` instead of using `R`. 

The other thing I noticed while working on this was that mapping wasn't being skipped. It looks like when the new `meta` definitions happened when defining the publish directories, we were still using `it` to refer to a directory that we were adding to `meta`. So it was returning `null` for the directory. I fixed this to use `meta` and then also changed to using `meta` throughout defining the paths since `meta` is a copy of `it` and I think that's easier to follow personally.

File: modules/af-features.nf
Comment (jashapiro) [inline modules/af-features.nf:1]: Thanks for catching this. I agree with the change to just using `meta` everywhere. I think I did that in the first implementation, but then for some reason thought I shouldn't. But I was obviously wrong in that thought!

File: bin/add_celltypes_to_sce.R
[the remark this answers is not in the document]
Comment (jashapiro) [reply bin/add_celltypes_to_sce.R:1]: I checked this time... 👍🏼

File: modules/classify-celltypes.nf [hunk 1]
```diff
@@ -63,11 +63,11 @@ process classify_cellassign {
     # Convert SCE to AnnData
     sce_to_anndata.R \
         --input_sce_file "${processed_rds}" \
-        --output_rna_h5 processed.hdf5
+        --output_rna_h5 "processed.hdf5"
```
jashapiro, line 66: Quotes aren't really needed here (we know that there are no spaces or other special characters in the file path), but they don't hurt.

modules/classify-celltypes.nf
```diff
@@ -184,21 +194,32 @@ workflow annotate_celltypes {
...
       assignment_input_ch = processed_sce_channel
         .map{[it[0]["library_id"]] + it}
         // add in singler results
         .join(singler_output_ch, by: 0, failOnMismatch: true, failOnDuplicate: true)
         // add in cell assign results
         .join(cellassign_output_ch, by: 0, failOnMismatch: true, failOnDuplicate: true)
         .map{it.drop(1)} // remove library_id
+        .branch{
+          // pull out libraries that actually have at least 1 type of annotations
+          add_celltypes: (it[2].name != "NO_FILE") || (it[3].name != "NO_FILE")
```
jashapiro, line 207: Just adding this for future flexibility in the file names (allows adding extenstions)
```suggestion
          add_celltypes: (it[2].baseName != "NO_FILE") || (it[3].baseName != "NO_FILE")
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -1,11 +1,13 @@
 # Cell type Annotation Summary
```
jashapiro, line 2: Probably worth adding a comment about the chunk below.
```suggestion

<!-- Import shared functions for cell type plotting -->
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -34,10 +34,15 @@ theme_set(
       strip.background = element_rect(fill = "transparent")
     )
 )
+```
+
+
```
jashapiro, line 39:
```suggestion
<!-- Import shared functions for cell type plotting -->
```

jashapiro (commented):
This looks good to me, but I thought I would send along my comments while you work out the oddness of the `child` functions

allyhawkins:
Okay it looks like Nextflow doesn't like nested child reports, but when we load in the functions in the main qc report it works as expected. So now this should be good to go!

templates/qc_report/celltypes_supplemental_report.rmd
jashapiro: Since this report is not a child of the main QC report, we still need these function to be a child here, right?

templates/qc_report/main_qc_report.rmd
```diff
@@ -547,6 +547,10 @@ The raw counts from all cells that remain after filtering low quality cells (RNA
 ```
 
 <!-- Next section only included if celltype annotations are present -->
```
jashapiro, line 549:
```suggestion
<!-- Next sections only included if celltype annotations are present -->
```

jashapiro (approved):
Looks good, but we do need those functions in the supplemental report too, right? As long as that works, we should be good to go.

templates/qc_report/celltypes_supplemental_report.rmd
allyhawkins: Ah yes, you're right. Thanks for catching that!

PR #506, opened 2023-10-12
Add perform_celltyping parameter

### LINKED ISSUE #494: Add flag perform_celltyping
_Related to #415_

By default, we do not want to perform cell typing. We should include a flag `perform_celltyping` with a default of `false` (or the inverse, `skip_celltyping` with a default of `true`), and then use an if statement to govern whether to enter into the celltypes process at all.

Figjam for reference: https://www.figma.com/file/9zH1gXlz9ZlxVrndN4Wy54/cell-type-workflow?type=whiteboard&node-id=0-1&t=fKgV0qjgpXfOkmFo-0

allyhawkins:
Closes #494 
Stacked on #503 

Here I'm adding in the flag to specify whether or not we want to run the cell typing part of the workflow with `--perform_celltyping`. As we discussed, this is set to `false` by default. This means to annotate cell types when running the workflow, you need to use the flag. 

There wasn't really anything too involved here since the output of `annotate_celltypes` already matches the output of `cluster_sce`. 
I tested this with the stub workflow both with and without `--perform_celltyping` and the expected processes were skipped when the flag was absent. I'm doing a test run with an actual library just to be sure everything is good. 
Edit: The tests finished so flag works as expected!

### DIFF
--- a/modules/qc-report.nf
+++ b/modules/qc-report.nf
@@ -19,15 +19,17 @@ process sce_qc_report{
         metadata_json = "${meta.library_id}_metadata.json"
         workflow_url = workflow.repository ?: workflow.manifest.homePage
         workflow_version = workflow.revision ?: workflow.manifest.version
-        
+
         // names for final output files
         unfiltered_out = "${meta.library_id}_unfiltered.rds"
         filtered_out = "${meta.library_id}_filtered.rds"
         processed_out = "${meta.library_id}_processed.rds"
-        
+
         // check for cell types
-        // TODO: add `params.perform_celltyping && (...)` when implemented
-        has_celltypes = (meta.submitter_cell_types_file || meta.singler_model_file || meta.cellassign_reference_file)
+        // only provide report template if cell typing was performed and either singler or cellassign was used
+        has_celltypes = (params.perform_celltyping
+                        && (meta.singler_model_file || meta.cellassign_reference_file)
+                        )
         celltype_report = "${meta.library_id}_celltype-report.html" // rendered HTML
         celltype_template_path = "${template_dir}/${celltype_template_file}" // template input
 
@@ -69,9 +71,10 @@ process sce_qc_report{
         processed_out = "${meta.library_id}_processed.rds"
         qc_report = "${meta.library_id}_qc.html"
         metadata_json = "${meta.library_id}_metadata.json"
-        
-        // TODO: add `params.perform_celltyping && (...)` when implemented
-        has_celltypes = meta.submitter_cell_types_file || meta.singler_model_file || meta.cellassign_reference_file
+
+        has_celltypes = (params.perform_celltyping
+                        && (meta.singler_model_file || meta.cellassign_reference_file)
+                        )
         celltype_report = "${meta.library_id}_celltype-report.html" // rendered HTML
 
         """
@@ -80,7 +83,7 @@ process sce_qc_report{
         touch ${processed_out}
         touch ${qc_report}
         ${has_celltypes ? "touch ${celltype_report}" : ""}
-        
+
         echo '{"unfiltered_cells": 10, "filtered_cells": 10, "processed_cells": 10}' > ${metadata_json}
         """
 }
--- a/modules/qc-report.nf
+++ b/modules/qc-report.nf
@@ -27,9 +27,7 @@ process sce_qc_report{
 
         // check for cell types
         // only provide report template if cell typing was performed and either singler or cellassign was used
-        has_celltypes = (params.perform_celltyping
-                        && (meta.singler_model_file || meta.cellassign_reference_file)
-                        )
+        has_celltypes = params.perform_celltyping && (meta.singler_model_file || meta.cellassign_reference_file)
         celltype_report = "${meta.library_id}_celltype-report.html" // rendered HTML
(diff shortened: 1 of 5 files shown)


jashapiro:
A couple more things to do with this PR, based on things that happened last week:

https://github.com/AlexsLemonade/scpca-nf/blob/6fd98de381367f068536a8e1f4018e81649b7451/modules/qc-report.nf#L29-L30

and: 

https://github.com/AlexsLemonade/scpca-nf/blob/6fd98de381367f068536a8e1f4018e81649b7451/modules/qc-report.nf#L73-L74

We just want to skip the supplemental report if we don't do cell typing. Though maybe we want to keep it if there are submitter cell types? I'm sort of rethinking what the exact logic should be, but skipping the extended report seems reasonable enough.

allyhawkins:
Sorry I totally missed those things. I'll add them in and then re-request review. 

> We just want to skip the supplemental report if we don't do cell typing. Though maybe we want to keep it if there are submitter cell types? I'm sort of rethinking what the exact logic should be, but skipping the extended report seems reasonable enough.

I think definitely skipping the extended report, but I actually think keeping the cell typing with just submitter annotations is a good idea. To me, that's being added in separate from the SingleR and CellAssign cell types, so I think we want to be able to show those even if we didn't do the other methods.

allyhawkins:
Okay I added in some logic to make sure that the main section of the cell type report is there if submitter annotations are there, regardless of if other cell type methods have been performed or not. Then, the supplemental report is only there if the `--perform_celltyping` is used and either `CellAssign` or `SingleR` has been run. 

Here's an example of a report with submitter cell types, but no other cell types. 
[SCPCL000495_qc.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/12895058/SCPCL000495_qc.html.zip)

modules/qc-report.nf
```diff
@@ -69,9 +71,10 @@ process sce_qc_report{
         processed_out = "${meta.library_id}_processed.rds"
         qc_report = "${meta.library_id}_qc.html"
         metadata_json = "${meta.library_id}_metadata.json"
-        
-        // TODO: add `params.perform_celltyping && (...)` when implemented
-        has_celltypes = meta.submitter_cell_types_file || meta.singler_model_file || meta.cellassign_reference_file
+
+        has_celltypes = (params.perform_celltyping
+                        && (meta.singler_model_file || meta.cellassign_reference_file)
+                        )
```
jashapiro, line 77: Just simplifying this a bit, because it _shouldn't_ affect the logic, but if it does...
```suggestion
        has_celltypes = params.perform_celltyping && (meta.singler_model_file || meta.cellassign_reference_file)
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -171,7 +171,15 @@ knitr::asis_output(
...
-For additional information about cell typing, including diagnostic plots and/or heatmap comparisons among annotations (if available), please refer to the [supplementary cell type QC report](`r params$celltype_report`).
+
+<!-- At least one cell type annotation method required for the supplemental report -->
+```{r, eval = has_singler | has_cellassign}
+knitr::asis_output(
+  glue::glue("
+  For additional information about cell typing, including diagnostic plots and/or heatmap comparisons among annotations (if available), please refer to the [supplementary cell type QC report](`r params$celltype_report`).
+ ")
+)
+```
```
jashapiro, line 182: Is there a reason we are wrapping here and above rather than using the chunk options? I think `knitr::asis_output()` is useful if we have a mix of outputs, but kind of distracting if the entire chunk is `asis`
```suggestion
````{r, eval = has_singler | has_cellassign, result = 'asis'}
glue::glue("
  For additional information about cell typing, including diagnostic plots and/or heatmap comparisons among annotations (if available), please refer to the [supplementary cell type QC report](`r params$celltype_report`).
 ")
```
````

File: modules/qc-report.nf [hunk 3]
```diff
@@ -19,15 +19,17 @@ process sce_qc_report{
...
         processed_out = "${meta.library_id}_processed.rds"
-        
+
         // check for cell types
-        // TODO: add `params.perform_celltyping && (...)` when implemented
-        has_celltypes = (meta.submitter_cell_types_file || meta.singler_model_file || meta.cellassign_reference_file)
+        // only provide report template if cell typing was performed and either singler or cellassign was used
+        has_celltypes = (params.perform_celltyping
+                        && (meta.singler_model_file || meta.cellassign_reference_file)
+                        )
```
jashapiro, line 32: if it works
```suggestion
        has_celltypes = params.perform_celltyping && (meta.singler_model_file || meta.cellassign_reference_file)
```

jashapiro (approved):
This looks good, but I had a couple minor formatting suggestions (assuming they work as I feel like they should).

templates/qc_report/celltypes_qc.rmd
allyhawkins: Honestly, the only reason I did it this way was to mirror how the other sections were. But we can change to using `result='asis'` throughout instead of `knitr::asis_output`?

allyhawkins:
Just noting that I decided to switch all the chunks that were just text to use `results='asis'` rather than `knitr::asis_output` and rendered the report to test that things still looked as expected. Going to go ahead and merge this.

PR #510, opened 2023-10-16
Parameter for skipping cell typing if results already exist

### LINKED ISSUE #446: Skip cell typing if results files already exist
After we complete #415, we will want to add an option to skip the actual cell type classification for `SingleR` and `CellAssign` if the output already exists. This should save us time whenever we are re-processing samples. Similar to `--repeat_mapping`, except `--repeat_celltyping`. 

In doing this we will want a way to track the reference versions used for both cell typing methods and only skip when both the output files exist and the reference version matches. 
For PanglaoDB references we can pull the date from the file name. For `celldex` reference versions we should match up the package version.

allyhawkins:
Closes #446 

This PR adds a new parameter to the workflow, `--repeat_celltyping`, that acts in a similar fashion to `--repeat_mapping`. 

- The default is set to `false`. This means that if the results from either `SingleR` or `CellAssign` already exist in the `checkpoints` folder for a specific library, then those processes will be skipped. 
- If you use the `--repeat_celltyping` option, then regardless of whether the results exist or not, they will be repeated. 
- This argument is only relevant if you are using `--perform_celltyping`, otherwise, cell typing will be skipped for all samples anyway. 
- I kept in the check for the missing reference, so if there's no reference, cell typing is still skipped as it was previously. 

A few other things: 

- I found a discrepancy between the existing `SingleR` results and the processed SCE object due to a difference in the number of cells. The cell numbers are only off by three cells for the example I looked at, but I think this is probably due to some filtering differences we have seen before. Instead of directly adding a new column with the annotations, I updated the `add_celltypes_to_sce.R` script to join by the barcodes, which is safer anyway.  
- When digging into the mismatches, I noticed we had an error in how we set the seed in `classify_singleR.R`, so I fixed that here. 
- This currently works other than generating the supplemental report, which is dependent on `ggforce` being added in. I'm still trying to figure out why the PR for adding that in is failing tests and image building... see https://github.com/AlexsLemonade/scpcaTools/pull/226

### DIFF
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -18,7 +18,7 @@ option_list <- list(
   make_option(
     opt_str = c("--singler_model_file"),
     type = "character",
-    help = "path to file containing a single model generated for SingleR annotation. 
+    help = "path to file containing a single model generated for SingleR annotation.
             File name is expected to be in form: <model name>_model.rds."
   ),
   make_option(
@@ -44,7 +44,7 @@ opt <- parse_args(OptionParser(option_list = option_list))
 # Set up -----------------------------------------------------------------------
 
 # set seed
-set.seed(opt$random_seed)
+set.seed(opt$seed)
 
 # check that input file file exists
 if (!file.exists(opt$sce_file)) {
--- a/bin/classify_SingleR.R
+++ b/bin/classify_SingleR.R
@@ -22,11 +22,6 @@ option_list <- list(
             File name is expected to be in form: <model name>_model.rds."
   ),
   make_option(
-    opt_str = c("--output_singler_annotations_file"),
-    type = "character",
-    help = "path to output TSV file that will store the SingleR annotations. Must end in .tsv"
-  ),
-  make_option(
     opt_str = c("--output_singler_results_file"),
     type = "character",
     help = "path to output RDS file that will store the SingleR results object. Must end in .rds"
@@ -59,9 +54,6 @@ if (!file.exists(opt$sce_file)) {
 # check that output files have the right extensions
 if (!(stringr::str_ends(opt$output_singler_results_file, ".rds"))) {
   stop("output SingleR result file name must end in .rds")
-}
-if (!(stringr::str_ends(opt$output_singler_annotations_file, ".tsv"))) {
-  stop("output SingleR annotations file name must end in .tsv")
 }
 
 # check that reference exists and filename is properly formatted
@@ -103,40 +95,7 @@ singler_results <- SingleR::classifySingleR(
 # add reference name to singler_results DataFrame metadata
 metadata(singler_results)$reference_name <- reference_name
 
-
-# create data frame of annotations
-annotations_df <- tibble::tibble(
-  barcode = rownames(singler_results),
-  singler_celltype_annotation = singler_results$pruned.labels,
-)
-
-# map ontology labels to cell type names, as needed
-# we can tell if ontologies were used because this will exist:
-if ("cell_ontology_df" %in% names(singler_model)) {
-
-  # end up with columns: barcode, singler_celltype_annotation, singler_celltype_ontology
-  annotations_df <- annotations_df |>
-    dplyr::left_join(
-      # column names: ontology_id, ontology_cell_names
-      singler_model$cell_ontology_df, 
-      by = c("singler_celltype_annotation" = "ontology_id")
-    ) |> 
-    # rename columns
-    dplyr::rename(
-      singler_celltype_ontology = singler_celltype_annotation,
-      singler_celltype_annotation = ontology_cell_names
-    )
-  
-  # add cell_ontology_df to singler_results DataFrame metadata
-  metadata(singler_results)$cell_ontology_df <- singler_model$cell_ontology_df
-} 
-
-
-
 # export results ---------------
-
-# first, a stand-alone tsv of annotations
-readr::write_tsv(annotations_df, opt$output_singler_annotations_file)
 
 # next, the full result to a compressed rds
 readr::write_rds(
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -173,7 +173,7 @@ workflow annotate_celltypes {
       // singleR output channel: [library_id, singler_results]
       singler_output_ch = singler_input_ch.skip_singler
         // provide existing singler results dir for those we skipped
-        .map{[it[0]["library_id"], file(it[0].singler_dir, type: 'dir')}
+        .map{[it[0]["library_id"], file(it[0].singler_dir, type: 'dir')]}
         // add empty file for missing ref samples
         .mix(singler_input_ch.missing_ref.map{[it[0]["library_id"], file(empty_file)]} )
         // add in channel outputs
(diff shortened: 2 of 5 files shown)


modules/classify-celltypes.nf
```diff
@@ -149,6 +148,8 @@ workflow annotate_celltypes {
           meta.cellassign_dir = "${meta.celltype_publish_dir}/${meta.library_id}_cellassign";
           meta.singler_model_file = singler_model_file;
           meta.cellassign_reference_file = cellassign_reference_file;
+          meta.singler_results_file = "${meta.singler_dir}/singler_results.rds";
+          meta.cellassign_predictions_file = "${meta.cellassign_dir}/cellassign_predictions.tsv"
```
jashapiro, line 152: I was going to comment that looking for the directory should be sufficient, but I seem to recall we have been burned by that in the past, so this is probably better.

modules/classify-celltypes.nf
```diff
@@ -158,18 +159,24 @@ workflow annotate_celltypes {
       singler_input_ch = celltype_input_ch
         // add in singler model or empty file
         .map{it.toList() + [file(it[0].singler_model_file ?: empty_file)]}
-        // skip if no singleR model file
+        // skip if no singleR model file or if singleR results are already present
         .branch{
-          missing_ref: it[2].name == "NO_FILE"
+          skip_singler: (!params.repeat_celltyping && file(it[0].singler_results_file).exists())
+                        || it[2].name == "NO_FILE"
```
jashapiro, line 165: I think it may make sense to keep these as three separate branches.
```suggestion
          skip_singler: !params.repeat_celltyping && file(it[0].singler_results_file).exists()
          missing_ref: it[2].name == "NO_FILE"
```

modules/classify-celltypes.nf
```diff
@@ -158,18 +159,24 @@ workflow annotate_celltypes {
...
+
       // singleR output channel: [library_id, singler_results]
-      singler_output_ch = singler_input_ch.missing_ref
-        .map{[it[0]["library_id"], file(empty_file)]}
+      singler_output_ch = singler_input_ch.skip_singler
+        // provide existing singler results dir for those we skipped and empty file for those missing reference
+        .map{[
+          it[0]["library_id"],
+          file(it[0].singler_results_file).exists() ? file(it[0].singler_dir, type: 'dir') : file(empty_file)
+          ]}
```
jashapiro, line 179: Keeping the separate branches, we can do a two step mix.
```suggestion
      singler_output_ch = singler_input_ch.skip_singler
        // provide existing singler results dir for those we skipped
        .map{[it[0]["library_id"], file(it[0].singler_dir, type: 'dir')}
        // add empty file for missing ref samples
        .mix(singler_input_ch.missing_ref.map{[it[0]["library_id"], file(empty_file)]} )
```

modules/classify-celltypes.nf
```diff
@@ -179,7 +186,8 @@ workflow annotate_celltypes {
         .map{it.toList() + [file(it[0].cellassign_reference_file ?: empty_file)]}
         // skip if no cellassign reference file or reference name is not defined
         .branch{
-          missing_ref: it[2].name == "NO_FILE"
+          skip_cellassign: (!params.repeat_celltyping && file(it[0].cellassign_predictions_file).exists())
+                           || it[2].name == "NO_FILE"
```
jashapiro, line 190: Same changes as above would make sense here, if you want to do those.

bin/classify_SingleR.R
```diff
@@ -60,9 +55,6 @@ if (!file.exists(opt$sce_file)) {
 if (!(stringr::str_ends(opt$output_singler_results_file, ".rds"))) {
   stop("output SingleR result file name must end in .rds")
 }
-if (!(stringr::str_ends(opt$output_singler_annotations_file, ".tsv"))) {
-  stop("output SingleR annotations file name must end in .tsv")
-}
```
jashapiro, line 65: I'm debating whether we should leave this for debugging purposes... But removing the duplicated code seems worth doing.

jashapiro (commented):
Looks good. I made a suggestion to keep three separate branches, which moves a bit of logic around, but should be functionally the same. I'm suggesting it mostly because that was what I had been planning with the `missing_ref` branch naming... that there would be a separate branch when we were skipping.

bin/classify_SingleR.R
allyhawkins: Why would we keep this if we don't even provide an annotations file anymore? We are only outputting the rds file?

modules/classify-celltypes.nf
allyhawkins: Ah I did not know you could do three branches, which is why I had combined them. But this makes sense!

bin/classify_SingleR.R
jashapiro: I was thinking we would output both (keeping all the code you removed in this script, not just this line). I don't think we should need to though.

allyhawkins: Okay, I feel like if we are outputting the entire object then we don't need to also output the annotations as a separate file, which is why I moved this to the script where we add the annotations to the processed object in https://github.com/AlexsLemonade/scpca-nf/pull/503. But if you think there's a compelling reason to also output the annotations, I can add the code back in.

nextflow.config
```diff
@@ -39,6 +39,7 @@ params {
   publish_fry_outs = false // alevin-fry outputs are not published by default. Use `--publish_fry_outs` to publish these files to the `checkpoints` folder.
   spliced_only = false // include only spliced reads in counts matrix, by default both unspliced and spliced reads are totaled and found in `counts` asasy of returned SingleCellExperiment object
   perform_celltyping = false // specify whether or not to incorporate cell type annotations
+  repeat_celltyping = false // if cell type annotations alread exist, skip cell type classification with SingleR and CellAssign
```
sjspielman, line 42:
```suggestion
  repeat_celltyping = false // if cell type annotations already exist, skip cell type classification with SingleR and CellAssign
```

sjspielman (commented):
just looking things over to catch up and found a typo!

allyhawkins:
So I was testing this with the new updates to `scpcaTools`, and I'm getting an error converting SCE to AnnData objects that have the SingleR object stored in the metadata, similar to the issue we previously had with the miQC object. I think we have two options, and before I implement anything, I'm curious what others think. 

1. We can update the `sce_to_anndata` function in `scpcaTools` to check for the `singler_results` in the metadata before conversion. Then we remove the object and print out a warning that the SingleR results were removed. This is exactly how we handle the miQC model. 
2. We can update the script we use in `scpca-nf` to prep objects prior to converting to AnnData to remove the SingleR results prior to running `sce_to_anndata`. 

I tend to favor option 1 since it will catch any objects getting converted both in and out of the workflow. Tagging @jashapiro and @sjspielman for any thoughts or opinions on how to proceed.

allyhawkins:
> I tend to favor option 1 as well. I wonder if we can do a more general check to remove anything that isn't a standard R vector from the metadata and print the name of the thing that was removed?

Yes, I think we should try that!

jashapiro:
> I tend to favor option 1 as well. I wonder if we can do a more general check to remove anything that isn't a standard R vector from the metadata and print the name of the thing that was removed?

I think we may be able to use `is.object()` to test each item and remove anything that is an object from the metadata.

A very simple implementation would be:
```
removed_metadata <- metadata(sce) |> purrr::keep(is.object) |> names()
metadata(sce) <- metadata(sce) |> purrr::discard(is.object)
```
But we would probably want to get the names of the thi

jashapiro (approved):
LGTM

Issue #512, opened 2023-10-17
Main report - Fix bugs and formatting in cell type tables


allyhawkins:
We found a few bugs/ issues with formatting of the cell type tables that need to be addressed: 

- [ ] `NA` in submitter annotations should be "Unclassified cell" 
- [ ]  Make order of tables based on percentage of cells annotated as each cell type. Keep Unknown/ Unclassified cells as last row 
- [ ] Present the cell ontology labels rather than cell ontology term ids for the `SingleR` table

While working on this, it might be a good idea to account for submitters using something similar to unclassified cells with their annotations. If we can, we want to avoid two rows that both refer to unclassified cells.

sjspielman:
Noting this existing related issue `Standardize how we refer to unknown cell types` https://github.com/AlexsLemonade/scpca-nf/issues/454

allyhawkins:
Closed by #528

PR #522, opened 2023-10-20
Add label type for SingleR refs to annotated SCE 

allyhawkins:
Based on https://github.com/AlexsLemonade/scpca-docs/pull/166#discussion_r1367258141, this PR adds back in tracking the reference label type for SingleR. When building the SingleR model se specify the label type to use. That label type is then stored in the metadata of the model object (see https://github.com/AlexsLemonade/scpca-nf/blob/934df294bbfe2a01d8ffdb94ba9f5d6b432a547d/bin/train_SingleR.R#L134).

So here I take the label name that's stored in the reference model object and store it in the results object in `classify_SingleR.R`. Then in `add_celltypes_to_sce.R`, I get the label type from the results object and add it to the metadata of the annotated SCE object. I also used it to check whether or not to map the annotations to the ontology label.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -62,6 +62,9 @@ if(!is.null(opt$singler_results)){
   }
 
   singler_results <- readr::read_rds(opt$singler_results)
+  
+  # get label type from metadata of singler object
+  label_type <- metadata(singler_results)$reference_label
 
   # create a tibble with annotations and barcode
   # later we'll add the annotations into colData by joining on barcodes column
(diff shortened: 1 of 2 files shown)


sjspielman (approved):
LGTM, thanks for doing this!

One other change that you _could_ make is here - check for the `"label.ont"` reference label in the processed metadata instead of the df in the model object. https://github.com/AlexsLemonade/scpca-nf/blob/934df294bbfe2a01d8ffdb94ba9f5d6b432a547d/templates/qc_report/celltypes_supplemental_report.rmd#L114

Either way, don't need to see this again!

allyhawkins:
> One other change that you could make is here - check for the "label.ont" reference label in the processed metadata instead of the df in the model object.

I need to make some changes to get the correct labels on plots since right now they are still showing the ontology ids, so I think we should address this as we make updates to each of the sections, starting with making sure the tables are using the right labels in https://github.com/AlexsLemonade/scpca-nf/issues/512.

Issue #514, opened 2023-10-17
Cell type report: Add introduction and sessionInfo


sjspielman:
Implement the following changes to the cell type report:

- [x] Add Introductory text on "why do we have this report to begin with"?
  - Be sure to explain the uncertainty inherent in cell type annotations
- [x] Frame report around comparing annotations methods & agreement as a qualitative indicator of reliability
  - This is why we do multiple methods 
- [x] Change title to "supplemental cell type report" (~annotation~)
- [x] `sessionInfo()`!

sjspielman:
Closed by #539

PR #523, opened 2023-10-20
Update introduction of main cell type report

### LINKED ISSUE #511: Main report - Update cell type introduction
We need to update some of the text that is present in the introduction of the cell type section in the main QC report: 

- [x] Make sure that description of the annotations is in the same order that is used throughout the report. Submitter, SingleR, CellAssign
- [x] Add in citations for SingleR and CellAssign

allyhawkins:
Closes #511 

This PR does some reorganization of the introduction for the cell type section that's in the main report. 

- I simplified a lot of the eval statements to have the same intro regardless of what methods were used. Then I just check for each individual cell type being present before printing out the bullet point for that method. 
- I added in the refs for SingleR and CellAssign. 
- The order is now submitter, SingleR, CellAssign

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -117,8 +117,8 @@ plot_umap <- function(
 
 <!-- Background depends on which cell types are present in the SCE. -->
 
-This section details results from performing cell type annotation:
-Cell type annotations were obtained in the following way: 
+This section details results from performing cell type annotation.
+Cell type annotations were obtained in the following way(s): 
 
 <!-- Submitter -->
 ```{r, eval = has_submitter, results='asis'}
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -117,7 +117,7 @@ plot_umap <- function(
 
 <!-- Background depends on which cell types are present in the SCE. -->
 
-This section details results from performing cell type annotation.
+The plots and tables included here detail the results from performing cell type annotation.
 Cell type annotations were obtained in the following way(s): 
 
 <!-- Submitter -->
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -115,48 +115,32 @@ plot_umap <- function(
 
 ```
 
-
 <!-- Background depends on which cell types are present in the SCE. -->
 
+This section details results from performing cell type annotation:
+Cell type annotations were obtained in the following way: 
 
-<!-- Submitter _only_ -->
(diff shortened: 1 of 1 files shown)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -115,53 +115,37 @@ plot_umap <- function(
 
 ```
 
-
 <!-- Background depends on which cell types are present in the SCE. -->
 
+This section details results from performing cell type annotation:
+Cell type annotations were obtained in the following way:
```
sjspielman, line 121:
```suggestion
Cell type annotations were obtained in the following way(s): 
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -115,53 +115,37 @@ plot_umap <- function(
 
 ```
 
-
 <!-- Background depends on which cell types are present in the SCE. -->
 
+This section details results from performing cell type annotation:
```
sjspielman, line 120: I'm not 100% sure about keeping this sentence as it is currently worded because we are planning to include this as a child rmd in the supplementary cell type report. But, this is clearly the right sentence to have at the top of this section as it appears in the main report!

sjspielman, line 120: But, make it a period if we do keep it - 
```suggestion
This section details results from performing cell type annotation.
```

sjspielman (commented):
All looks good to me, just not sure what to do about the opening sentence which makes perfect sense for _this_ report, but the wording seems off for inclusion in the supplementary report. 
I don't think (?) there's a built in environmental variable or so to check if something is a child report or not, but we could add like a param or something to check if it's a child or not, and decide whether to print this sentence or one that is more along the lines of "here we're showing you results from the main qc report." A param seems a little heavy-duty for this though, so any other ideas are great to hear!

templates/qc_report/celltypes_qc.rmd
```diff
@@ -115,53 +115,37 @@ plot_umap <- function(
 
 ```
 
-
 <!-- Background depends on which cell types are present in the SCE. -->
 
+This section details results from performing cell type annotation.
```
allyhawkins, line 120:
```suggestion
The plots and tables included here detail the results from performing cell type annotation.
```
How about something more generic like this? I think the key here is removing the mention of "this section" and then it can be used in both reports. That being said, I do think when we address https://github.com/AlexsLemonade/scpca-nf/issues/514, we may be changing the introduction of the supplemental report regardless. So it might make sense to separate out the intro statements here to only have them in the main report and not include in the supplemental. I think that's a decision we should make when we get the supplemental intro done.

sjspielman (approved):
Approving assuming you accept your own sentence suggestion!

PR #524, opened 2023-10-26
Parse NAs when reading in json files

allyhawkins:
When processing projects with `v0.6.0`, I noticed that projects that don't contain a submitter cell types file were throwing an error when trying to find the submitter cell types file designated as `NA`. This is because if we skip mapping, we read in the meta from `scpca-meta.json`, rather than using the `meta` object we create from `scpca-library-metadata.tsv`. We had previously only been handling the NAs when reading in the metadata file, not the json file. This PR makes the fix to handle NA's when reading in json too. 

I tested this with a run that doesn't have a submitter cell types file and it worked successfully. I also am filing this to `main`, because I want to create a new release with these changes only. I will file the same PR to the `development` branch to keep things in sync after this is approved.


jashapiro (approved):
LGTM!

PR #526, opened 2023-10-26
Remove extra publish step

allyhawkins:
When running through some samples and performing cell typing, I noticed that we were still publishing the `_annotated_processed.rds` file to the results directory. 
We don't want to do that and only want to publish the final SCE objects in the process that creates the QC report.


jashapiro (approved):
LGTM!

Issue #470, opened 2023-09-27
Update celltype reference building to better track reference versions


sjspielman:
We've discussed strategies for tracking whether cell typing might need to be repeated in the main workflow. One aspect to track is versions of the cell typing references. We track panglao's version as the _date_ embedded in the reference filename (essentially our copy of the database):
https://github.com/AlexsLemonade/scpca-nf/blob/3c8a0b1ff84d493bb5a239a0b56712a0a5a2d0ee/config/reference_paths.config#L24

We should also record a package version for celldex. Look through the code now, we really don't have an equivalent celldex file to that panglaodb file for recording this information, since celldex is of course just there in an R package, not a stand-alone TSV. 

I see two overall kinds of approaches here:

1. We should actually get the version info into the specific _model/marker gene files_. For example:
    - `cellassign_references/PanglaoDB-blood.tsv` --> `cellassign_references/PanglaoDB-27_Mar_2020-blood.tsv` 
    - `singler_models/BlueprintEncodeData_model.rds` --> `singler_models/BlueprintEncodeData-<celldex version>-model.rds`

2. Or, modify the `celltype_ref_source` column in https://github.com/AlexsLemonade/scpca-nf/blob/3c8a0b1ff84d493bb5a239a0b56712a0a5a2d0ee/references/celltype-reference-metadata.tsv to be in the format `method_version`, e.g. `panglaodb-27_Mar_2020` and `celldex_<version>`.  Or, we could add a `celltype_ref_version` column with the version info. 

Tagging @allyhawkins @jashapiro for thoughts/discussion!

allyhawkins:
I think we should modify the `celltype_ref_source` column to include the versions as you suggested in 2. But then I think we should also include the versions in the reference files that get output. That way the reference version is captured in the filename and we can use that to check for repeating cell typing.

jashapiro:
> * `cellassign_references/PanglaoDB-blood.tsv` --> `cellassign_references/PanglaoDB-27_Mar_2020-blood.tsv`

Please use ISO8601:
`cellassign_references/PanglaoDB-blood_2020-03-27.tsv`

Also: I (following Jenny Bryan) prefer to use `-` to separate words, `_` for "fields". We maybe haven't used this 100% consistently, but I'm trying.
So probably: `singler_models/BlueprintEncodeData_<celldex version>_model.rds`

[...]

PR #528, opened 2023-10-27
Fixes to cell type tables

### LINKED ISSUE #512: Main report - Fix bugs and formatting in cell type tables

### LINKED ISSUE #527: Don't plot UMAP if all cell types are NA
When going through some of the cell type reports that I generated when testing processing a project, I noticed that we have cases where all of the cells get annotated as NA for submitter annotations. This happened because that project has a submitter cell types file, but the cells in that specific library weren't annotated. 

In the report, we see a table where all the annotations are NA and then an empty UMAP. We should probably add some logic only to include the plots for submitter annotations as long as all of the cells aren't NA. I think if they are all NA for submitter we should just disregard submitter annotations for that library and treat it as if we don't have submitter annotations.

allyhawkins:
Closes #512 
Closes #527 

This PR makes the necessary adjustments to fix some of the formatting and display of the tables that are mentioned in #512. 

- In the `create_celltype_df` function, I added a check for submitter cell types and set NA to "Unclassified cell". 
- I also made sure to add in the re-leveling so that the table for submitter cell types is in order of percentage of cells with "Unclassified cell" last. 
- I also fixed a small typo that meant we weren't setting the cell ontology labels as the annotations, which is what we want to do. That should now fix the situation of seeing the cell ontology ids as the axis labels in the plots. 
- The last thing I did was just add a check that all submitter cell types are not NA. If that's the case, then `has_submitter` is set to `FALSE`. No more blank UMAPs or heatmaps. 

Note that because I'm running CellAssign right now I haven't been able to fully test this with Nextflow, but I did test this locally.

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -138,9 +138,15 @@ glue::glue(
 
 
 ```{r, eval = has_cellassign, results='asis'}
+tissue_type <- stringr::word(
+  metadata(processed_sce)$cellassign_reference,
+  -1,
+  sep = "-"
+)
+
 glue::glue(
   "* Cell types were annotated using [`CellAssign`](https://github.com/Irrationone/cellassign), which uses a marker-gene based approach ([Zhang _et al._ 2019](https://doi.org/10.1038/s41592-019-0529-1)).
-  Marker genes associated with `{metadata(processed_sce)$cellassign_reference}` tissue were obtained from [PanglaoDB](https://panglaodb.se/).
+  Marker genes associated with `{tissue_type}` tissue were obtained from [PanglaoDB](https://panglaodb.se/).
   "
 )
 ```
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -132,7 +132,7 @@ glue::glue(
 ```{r, eval = has_singler, results='asis'}
 glue::glue(
   "* Cell types were annotated using [`SingleR`](https://bioconductor.org/packages/release/bioc/html/SingleR.html), which uses a reference-based approach ([Looney _et al._ 2019](https://doi.org/10.1038/s41590-018-0276-y)).
-  The `{metadata(sce)$singler_reference}` reference dataset, obtained from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) package, was used for cell typing."
+  The `{metadata(processed_sce)$singler_reference}` reference dataset, obtained from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) package, was used for cell typing."
 )
 ```
 
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -132,7 +132,7 @@ glue::glue(
 ```{r, eval = has_singler, results='asis'}
 glue::glue(
   "* Cell types were annotated using [`SingleR`](https://bioconductor.org/packages/release/bioc/html/SingleR.html), which uses a reference-based approach ([Looney _et al._ 2019](https://doi.org/10.1038/s41590-018-0276-y)).
-  The `{metadata(processed_sce)$singler_reference}` reference dataset, obtained from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) package, was used for cell typing."
+  The `{metadata(processed_sce)$singler_reference}` reference dataset, obtained from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html), was used for cell typing."
 )
 ```
 
--- a/templates/qc_report/main_qc_report.rmd
+++ b/templates/qc_report/main_qc_report.rmd
@@ -114,7 +114,8 @@ if (has_processed) {
 
   has_singler <- "singler" %in% metadata(processed_sce)$celltype_methods
   has_cellassign <- "cellassign" %in% metadata(processed_sce)$celltype_methods
-  has_submitter <- "submitter" %in% metadata(processed_sce)$celltype_methods
+  has_submitter <- "submitter" %in% metadata(processed_sce)$celltype_methods &
+    !all(is.na(processed_sce$submitter_celltype_annotation)) # make sure they aren't all NA
 
   # If at least 1 is present, we have cell type annotations.
   has_celltypes <- any(has_singler, has_cellassign, has_submitter)
--- a/templates/qc_report/utils/celltype_functions.rmd
+++ b/templates/qc_report/utils/celltype_functions.rmd
@@ -32,14 +32,7 @@ create_celltype_df <- function(processed_sce) {
   
   if("submitter_celltype_annotation" %in% names(celltype_df)){
     celltype_df <- celltype_df |> 
-      dplyr::mutate(submitter_celltype_annotation = dplyr::if_else(
-        is.na(submitter_celltype_annotation),
-        "Unclassified cell", 
-        submitter_celltype_annotation
-      ) |>
-        forcats::fct_infreq() |>
-        forcats::fct_relevel("Unclassified cell", after = Inf)
-      )
+      prepare_annotation_values(submitter_celltype_annotation, is_submitter = TRUE) 
   }
 
   if ("singler_celltype_annotation" %in% names(celltype_df)) {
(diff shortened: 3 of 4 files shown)


sjspielman:
> Note that because I'm running CellAssign right now I haven't been able to fully test this with Nextflow, but I did test this locally.

Do you have a locally-run report you can attach here?

> The last thing I did was just add a check that all submitter cell types are not NA. If that's the case

Just generally thinking about when/how this would happen in the first place. The circumstance I can think of is if we end up filtering out _all_ the cells that the submitter annotated, so only `NA` labels are left in the processed object. Are there other situations I'm not immediately seeing?

templates/qc_report/main_qc_report.rmd
```diff
@@ -114,7 +114,8 @@ if (has_processed) {
 
   has_singler <- "singler" %in% metadata(processed_sce)$celltype_methods
   has_cellassign <- "cellassign" %in% metadata(processed_sce)$celltype_methods
-  has_submitter <- "submitter" %in% metadata(processed_sce)$celltype_methods
+  has_submitter <- "submitter" %in% metadata(processed_sce)$celltype_methods &
```
sjspielman, line 117: 👍

templates/qc_report/utils/celltype_functions.rmd
```diff
@@ -29,10 +29,22 @@ create_celltype_df <- function(processed_sce) {
...
+  if("submitter_celltype_annotation" %in% names(celltype_df)){
+    celltype_df <- celltype_df |> 
+      dplyr::mutate(submitter_celltype_annotation = dplyr::if_else(
+        is.na(submitter_celltype_annotation),
+        "Unclassified cell", 
+        submitter_celltype_annotation
+      ) |>
+        forcats::fct_infreq() |>
+        forcats::fct_relevel("Unclassified cell", after = Inf)
+      )
```
sjspielman, line 42: I wonder if we should modify and use the `prepare_annotation_values()` function. One way to do that is below (also I un-hardcoded some strings as part of it). 

What do you think? 

We could still take this conceptual approach without the specific way I have implemented it; `fct_relevel()` for example will work even if provided a string that isn't present but it will toss out a warning which will need to be suppressed. Using `fct_string` means no warning will be applied.

```
prepare_annotation_values <- function(df, annotation_column, is_submitter) {

  submitter_string <- "Unclassified cell" # use for unclassified submitter cells 
  method_string  <- "Unknown cell type" # use for unclassified singler and cellassign cells
  fct_string <- ifelse(is_submitter, submitter_string, method_string) # which string to consider when reordering the factor

  df |>
    dplyr::mutate(
      {{ annotation_column }} := dplyr::case_when(
        # submitter condition - HAS TO BE FIRST since these are also NA which is the singler check
        is_submitter  ~ submitter_string,
        # singler condition
        is.na({{ annotation_column }})~ method_string,
        # cellassign conditon
        {{ annotation_column }} == "other" ~ method_string,
        # otherwise, keep it
        .default = {{ annotation_column }}
      ) |>
        # order column by number of cells
        forcats::fct_infreq() |>
        # relevel so the applied unknown is at the end
        forcats::fct_relevel(fct_string, after = Inf)
       ## OR this will work but with a warning so have to suppress them from chunk
        forcats::fct_relevel(c(submitter_string, method_string), after = Inf)
    )
```

sjspielman: Oh, and can default `is_submitter=FALSE` in the function, and set to true when calling submitters.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -132,7 +132,7 @@ glue::glue(
 ```{r, eval = has_singler, results='asis'}
 glue::glue(
   "* Cell types were annotated using [`SingleR`](https://bioconductor.org/packages/release/bioc/html/SingleR.html), which uses a reference-based approach ([Looney _et al._ 2019](https://doi.org/10.1038/s41590-018-0276-y)).
-  The `{metadata(sce)$singler_reference}` reference dataset, obtained from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) package, was used for cell typing."
+  The `{metadata(processed_sce)$singler_reference}` reference dataset, obtained from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html) package, was used for cell typing."
```
sjspielman, line 135: we [missed this](https://en.wikipedia.org/wiki/Major_Major_Major_Major) before 
```suggestion
  The `{metadata(processed_sce)$singler_reference}` reference dataset, obtained from the [`celldex` package](http://bioconductor.org/packages/release/data/experiment/html/celldex.html), was used for cell typing."
```

sjspielman (commented):
A couple small comments from looking at the report you posted in https://github.com/AlexsLemonade/scpca-nf/pull/529#issuecomment-1783234093 . One I left as a suggestion (too many "package"s), but the other I can't comment on directly. 

On L143, we have:
```
 Marker genes associated with `{metadata(processed_sce)$cellassign_reference}` tissue were obtained from [PanglaoDB](https://panglaodb.se/).
```

But `{metadata(processed_sce)$cellassign_reference}` still contains the `Panglao-` prefix. We need to remove that for pretty printing purposes, but I wonder where we should do that! Can either be in the report just for printing, or back in `add_celltypes_to_sce.R` to change how we save it in the final SCE.

allyhawkins:
> Just generally thinking about when/how this would happen in the first place. The circumstance I can think of is if we end up filtering out all the cells that the submitter annotated, so only NA labels are left in the processed object. Are there other situations I'm not immediately seeing?

Replying to this comment first, while I go through the other comments. 
I added a check for not all submitter cell types being NA, because some libraries in SCPCP000005 were never annotated in the first place. Because submitter cell types are added on a project level (we have one metadata file for cell types per project) all the cell types will be assigned as NA for a library that the submitter never annotated.

sjspielman:
> Because submitter cell types are added on a project level

Project level! Yes, thanks!

allyhawkins:
> But {metadata(processed_sce)$cellassign_reference} still contains the Panglao- prefix. We need to remove that for pretty printing purposes, but I wonder where we should do that! Can either be in the report just for printing, or back in add_celltypes_to_sce.R to change how we save it in the final SCE.

I updated this to extract the actual tissue type from the name, removing the PanglaoDB in https://github.com/AlexsLemonade/scpca-nf/pull/528/commits/c57e64c2a0c63e8b686fea0d155fd09cd0b77e89. 

> I wonder if we should modify and use the prepare_annotation_values() function. One way to do that is below (also I un-hardcoded some strings as part of it).

Thanks for the suggestion here! I updated the function to consider submitter annotations rather than directly modifying them in `create_celltype_df`. I did test to make sure that everything was named correctly after implementing those changes in https://github.com/AlexsLemonade/scpca-nf/pull/528/commits/b734f96f75da15b641055c1ac914f1e22b3ad5b6. 

This should be ready for another round of review @sjspielman.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -132,15 +132,21 @@ glue::glue(
...
 ```
 
 
 ```{r, eval = has_cellassign, results='asis'}
+tissue_type <- stringr::word(
+  metadata(processed_sce)$cellassign_reference,
+  -1,
+  sep = "-"
+)
+
```
sjspielman, line 146: I'm of two minds here... (although really maybe i am of one mind and playing devil's advocate with the other).

1. This is nicely general if we use other references besides PanglaoDB one day. We'll want to add some documentation that ensures future references are of the format `<databasename>-<tissuename>` moving forward, somewhere here. I envision a _super quick_ addition of "if/when we add future references use format `<databasename>-<tissuename>`".
https://github.com/AlexsLemonade/scpca-nf/blob/4c6d9a24469bcd7c069ecf02e74d4f314984a369/internal-instructions.md?plain=1#L167-L174

3. That said, _do we_ anticipate using something besides PanglaoDB? If we absolutely really really don't, then this could be simplified to a `str_remove()`. But, I am more in favor of keeping what you have here since I like to keep things flexible!

sjspielman (approved):
LGTM!

I left one main comment about some docs we might want to add given some reasonable coding choices you made. If you disagree with my thoughts, let me know! The punchline of what I'm (probably) recommending means you'll want to add a little more docs in `internal-instructions.md`, but I don't need to necessarily see again unless you want to tag me back in!

templates/qc_report/celltypes_qc.rmd
allyhawkins: I am somewhat hesitant to add the documentation in at this point, since I envision changes in the cell type reference file naming as part of #470. So I think I'll just comment on that issue with a link to the line here that I altered. We may need to make further adjustments anyways.

sjspielman: very reasonable!

PR #529, opened 2023-10-27
Fix UMAPs in main cell type section of report

### LINKED ISSUE #513: Main report - Cell type UMAP adjustments
There are a few small adjustments that need to be made to the UMAPs in the cell type section: 

- [ ] Capitalize `Jaccard` and `Louvain` in the text
- [ ] Move the statement about cell types in the UMAPs to be below the cluster UMAP 
- [ ] Make sure that `NA` is renamed to `Unknown` and included when counting the top cell types. 
- [ ] Change `other cell types` to `All remaining cell types`

allyhawkins:
Closes #513 
Stacked on #528 

This PR makes the small adjustments to the UMAPs that were outlined in #513. 

- I made sure that the methods for clustering and weighting were capitalized. One thing to note is that we are grabbing these from the metadata of the processed object. Should we store them in the object as capitalized instead of what I did here? 
- I moved the description of the cell type UMAPs to be immediately before the first cell type UMAP rather than grouped with the description of the cluster UMAP. 
- Part of the fixes in #528 actually helped account for all the NA's being renamed and counted when grouping the cell types together. 
- I updated the name of the "other cell types" group to "All remaining cell types" and updated the text accordingly.

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -199,7 +199,7 @@ create_celltype_n_table(celltype_df, cellassign_celltype_annotation) |>
 ## UMAPs
 
 In this section, we show UMAPs colored by clusters. 
-Clusters were calculated using the graph-based `r stringr::str_to_sentence(metadata(processed_sce)$cluster_algorithm)` algorithm with `r stringr::str_to_sentence(metadata(processed_sce)$cluster_weighting)` weights.
+Clusters were calculated using the graph-based `r metadata(processed_sce)$cluster_algorithm` algorithm with `r metadata(processed_sce)$cluster_weighting` weighting.
 
 
 ```{r}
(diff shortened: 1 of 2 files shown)


allyhawkins:
@sjspielman, I'm including a copy of the main report with all fixes from #528 and this PR here. One thing to note is that I don't have an object currently that has the correct ontology labels because that was a change in the actual nextflow workflow.

sjspielman (commented):
LGTM! I only had a few small textual comments. 

Only reason I'm not approving now is because of potential future changes from your comment - 

> Should we store them in the object as capitalized instead of what I did here?

I think it doesn't make a difference whether users received upper or lower case strings, _and_ I do things it's good to keep report information as similar as we can* to how we store info in the SCE. So, I have a slight preference for storing as uppercase in the object (so change how we store them in the first place), but really it is only slight.

* I am not advocating for changing our cell labels though, I don't think we should not replace with "Unknown cell type" etc in the SCE. (unless someone convinces me otherwise?!)

allyhawkins:
> I think it doesn't make a difference whether users received upper or lower case strings, and I do things it's good to keep report information as similar as we can* to how we store info in the SCE. So, I have a slight preference for storing as uppercase in the object (so change how we store them in the first place), but really it is only slight.

I think I agree about keeping things consistent, so I made the change directly when adding the algorithm/ weighting to the SCE object rather than when printing in the cell type report. 

I also addressed your other smaller comments, so this is ready for another review @sjspielman.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -232,6 +228,13 @@ if (length(levels(umap_df$cluster)) <= 8) {
 }
 ```
 
+
+Next, we show UMAPs colored by cell types.
+A separate UMAP is shown for each cell typing method used.
```
sjspielman, line 233: miscellaneous note that I bet we need to firm up this language. 
"Cell type annotation method"?
"Cell typing method"? etc.
 
I'll open a separate issue.

sjspielman (approved):
🚀

Issue #515, opened 2023-10-17
Cell type report: Include celltypes_qc.rmd in cell type report


sjspielman:
At the top of the cell type report, we should include the `celltypes_qc.rmd` subreport as a child at the top of the cell type report (probably after the introduction to be added as part of #514?).

sjspielman:
Note that both the supp report and the child celltype report define - `celltype_df <- create_celltype_df(processed_sce)`. When calling the subreport, we should remove the `celltype_df` definition from the supp report since it will be duplicated.

sjspielman:
Closed by #539

PR #531, opened 2023-10-30
Establish updated heatmap styling

sjspielman:
Towards #516 

This PR takes the next steps in updating heatmaps:

Edit - the report! 
[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13207665/celltypes_supplemental_report.html.zip)

- Heatmaps now come first 
  - note that the introductory parts are still forthcoming - #514 & #515
- I removed the previous function we had been using for making heatmaps. For now, I've commented out the code that used that function to compare singler/cellassign to submitter. The next PR will clear this out using the strategy we land on in this PR.
- Heatmap styling changes:
  - I have updated the color scheme to something less dramatic - a standard yellow and red, but this can certainly change!! (it's super yellow now which is partly b/c these clusters + submitter annotations are simulated data for development)
  - No longer show dendrograms and order by frequency. The order is happily directly inherited from previous code setting factor levels before heading into matrix construction. 
- I changed the submitter label to simply be "Submitter" in the interest of spacing.
- I added some brief text about jaccard. Can (should?) be expanded!!

### a couple points of discussion
- To size the heatmap, I set the height of the viewport to 2.5x the number of heatmaps shown. I think this works ok?
- I placed the shared legend on the bottom. I did this for a few reasons:
  -  Simplifies the approach to choose which heatmap's legend we want to keep. We can arbitrarily grab the first one and use that legend (otherwise, we need a whole logic statement to choose which legend based on which cell types are present, but I have this written on the side if we want to pop it in). 
  - The title for the legend is long and horizontal looks nicer to me than vertical did.
  - Gives us more horizontal room for long cell type names
- I implemented some padding around the method labels on the right to, well, get them padded! Otherwise letters will slightly overlap with the heatmap itself. I modified [`ht_opt`](https://rdrr.io/github/jokergoo/ComplexHeatmap/man/ht_opt.html) for this. But....
  - This proved a bit weird, though not tricky - for reasons I don't totally understand, setting that padding seems to require me to actually load the `ComplexHeatmap` package; I got an "object not found" error when I tried to scope with `::`. Any ideas for how to get this global padding parameter working without loading the package, I'm all ears!
 

### what's next?

I'd like to get the heatmap code into a function so we can re-use it for the next heatmap comparing to submitter annotations, but let's get some of the styling down first in this PR. I plan to functionalize it in the next PR.

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -161,16 +161,16 @@ col_fun <- circlize::colorRamp2(c(0, 1), colors = c("yellow", "red"))
 keep_legend_name <- names(jaccard_cluster_matrices)[1]
 
 ###############################################################################
-# BELOW FAILS WITH ERROR:  - Error: object 'ComplexHeatmap' not found
+# BELOW FAILS WITH ERROR:  - Error: object 'ComplexHeatmap' not found...????
 #ComplexHeatmap::ht_opt[["TITLE_PADDING"]]<- unit(0.5, "cm")
 
 # BELOW WORKS!
-library(ComplexHeatmap)
-ht_opt[["TITLE_PADDING"]]<- unit(0.5, "cm")
+suppressPackageStartupMessages(library(ComplexHeatmap))
+ht_opt[["TITLE_PADDING"]] <- unit(0.5, "cm")
 ###############################################################################
 
 
-# TODO: We might want to make a wrapper for this, to use for direct comparison heatmaps? Circle back.
+# TODO: make this into a function so can re-use for 2nd group of heatmaps (still forthcoming)
 heatmap_list <- jaccard_cluster_matrices |> 
   purrr::imap(\(mat, name){
     ComplexHeatmap::Heatmap(
@@ -206,7 +206,8 @@ heatmap_list <- jaccard_cluster_matrices |>
 # Render the heatmap list
 ComplexHeatmap::draw(
   heatmap_list, 
-  heatmap_legend_side = "bottom")
+  heatmap_legend_side = "bottom" 
+)
 ```
 
 
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -185,7 +185,7 @@ heatmap_list <- jaccard_cluster_matrices |>
       heatmap_legend_param = list(
         title = "Jaccard similarity index",
         direction = "horizontal",
-        legend_width = unit(4, "cm")
+        legend_width = unit(1.5, "in")
       ),
       # Retain only 1 legend in the final plot      
       show_heatmap_legend = name == keep_legend_name,
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -112,8 +112,16 @@ celltype_df <- create_celltype_df(processed_sce)
 
 ## Heatmaps
 
-Below, we show heat maps comparing cell type annotations (along the y-axis) to clustering results (along the x-axis).
-Heatmap colors represent the log number of cells present in both the given cell type and cluster.
+Below, we show heat maps comparing each set of cell type annotations to clustering results.
+Heatmap colors represent the Jaccard similarity index between each grouping.
+
+This metric is bounded by 0 and 1:
+
+* A value of 1 indicates complete overlap between categories.
+The categories being compared contain the same set of cells.
+* A value of 0 indicates no overlap between categories. 
+No cells that appear in one category appear in the other.
+
 
 ```{r}
 # Calculate all jaccard matrices of interest for input to heatmap
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -154,21 +154,14 @@ heatmap_height <- length(jaccard_cluster_matrices) * 2.5
 ```
 
 ```{r, fig.height = heatmap_height, fig.width = 7, warning = FALSE}
-# Define color ramp for shared use in the heatmap - TO BE DISCUSSED!
-col_fun <- circlize::colorRamp2(c(0, 1), colors = c("yellow", "red"))
+# Define color ramp for shared use in the heatmap
+col_fun <- circlize::colorRamp2(c(0, 1), colors = c("white", "darkslateblue"))
 
 # We only want one shared legend in the end, arbitrarily grab the first one
 keep_legend_name <- names(jaccard_cluster_matrices)[1]
 
-###############################################################################
-# BELOW FAILS WITH ERROR:  - Error: object 'ComplexHeatmap' not found...????
-ComplexHeatmap::ht_opt(TITLE_PADDING = grid::unit(0.5, "cm"))
-
-# BELOW WORKS!
-suppressPackageStartupMessages(library(ComplexHeatmap))
-ht_opt[["TITLE_PADDING"]] <- unit(0.5, "cm")
-###############################################################################
-
+# Set heatmap padding option
+ComplexHeatmap::ht_opt(TITLE_PADDING = grid::unit(0.2, "in"))
 
 # TODO: make this into a function so can re-use for 2nd group of heatmaps (still forthcoming)
 heatmap_list <- jaccard_cluster_matrices |> 
@@ -176,6 +169,7 @@ heatmap_list <- jaccard_cluster_matrices |>
     ComplexHeatmap::Heatmap(
       t(mat), # transpose because matrix rows are in common & we want a vertical arrangement
       col = col_fun,
+      border = TRUE, # each heatmap gets its own outline
       ## Row parameters
       cluster_rows = FALSE,
       row_title = name, # each heatmap gets its own title
@@ -188,15 +182,15 @@ heatmap_list <- jaccard_cluster_matrices |>
       column_title = glue::glue("Cluster"),
       column_names_side = "top",
       column_names_rot = 0,
-      column_names_gp = grid::gpar(fontsize = 12),
+      column_names_gp = grid::gpar(fontsize = 8), # match row name size
       ## Legend parameters
-    heatmap_legend_param = list(
-      title = "Jaccard similarity index",
-      direction = "horizontal",
-      legend_width = unit(4, "cm")
-    ),
-    # Retain only 1 legend in the final plot      
-    show_heatmap_legend = name == keep_legend_name,
+      heatmap_legend_param = list(
+        title = "Jaccard similarity index",
+        direction = "horizontal",
+        legend_width = unit(4, "cm")
+      ),
+      # Retain only 1 legend in the final plot      
+      show_heatmap_legend = name == keep_legend_name,
     )
   }) |>
   # concatenate vertically into HeatmapList object
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -108,16 +108,23 @@ has_submitter <- "submitter" %in% metadata(processed_sce)$celltype_methods
 celltype_df <- create_celltype_df(processed_sce)
 ```
 
-## Comparison to clusters
-                   
-This section displays heatmaps comparing cell type annotations to cell clusters, which were calculated using the graph-based `r metadata(processed_sce)$cluster_algorithm` algorithm with `r metadata(processed_sce)$cluster_weighting` weighting.
 
-We use the [Jaccard similarity index](https://en.wikipedia.org/wiki/Jaccard_index) to compare pairs of labels from clusters and cell type annotations.
+## Cell label comparison plots
+
+This section displays heatmaps comparing cell labels from various methods. 
+
+We use the [Jaccard similarity index](https://en.wikipedia.org/wiki/Jaccard_index) to display the agreement between between pairs of labels from different methods.
 
 The Jaccard index reflects the degree of overlap between the two labels and ranges from 0 to 1. 
 
 * If the labels are assigned to identical sets of cells, the Jaccard index will be 1.
 * If the labels are assigned to completely non-overlapping sets of cells, the Jaccard index will be 0.
+
+
+### Comparison with unsupervised clustering
+
+Here we show the labels from unsupervised clustering compared against cell type annotations. 
+Cluster assignment was performed using the `r metadata(processed_sce)$cluster_algorithm` algorithm.
 
 
 ```{r}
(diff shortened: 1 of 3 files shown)


jashapiro:
Can you please add an image of a rendered heatmap in a comment here? Don't need a full report, I don't think.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -141,98 +148,69 @@ if (has_cellassign) {
...
+```{r, fig.height = heatmap_height, fig.width = 7, warning = FALSE}
+# Define color ramp for shared use in the heatmap - TO BE DISCUSSED!
+col_fun <- circlize::colorRamp2(c(0, 1), colors = c("yellow", "red"))
+
+# We only want one shared legend in the end, arbitrarily grab the first one
+keep_legend_name <- names(jaccard_cluster_matrices)[1]
+
+###############################################################################
+# BELOW FAILS WITH ERROR:  - Error: object 'ComplexHeatmap' not found...????
+#ComplexHeatmap::ht_opt[["TITLE_PADDING"]]<- unit(0.5, "cm")
```
jashapiro, line 165: Did you try this form?
```suggestion
ComplexHeatmap::ht_opt(TITLE_PADDING = grid::unit(0.5, "cm"))
```

sjspielman:
Here's the stand-alone heatmap!
Screenshot 2023-10-30 at 1 15 29 PM

templates/qc_report/celltypes_supplemental_report.rmd
jashapiro: A secondary comment here: `fig.height` is in inch units so maybe we should stick with that throughout?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -141,98 +148,69 @@ if (has_cellassign) {
...
-    column_names_rot = column_names_rotation,
-    column_names_gp = grid::gpar(fontsize = column_font_size),
-    row_dend_side = "left",
-    row_title_side = "right",
-    row_title = x_label,
-    row_names_gp = grid::gpar(fontsize = row_font_size),
-    # Legend parameters
+```{r, fig.height = heatmap_height, fig.width = 7, warning = FALSE}
+# Define color ramp for shared use in the heatmap - TO BE DISCUSSED!
+col_fun <- circlize::colorRamp2(c(0, 1), colors = c("yellow", "red"))
```
jashapiro, line 158: Since we don't have NA values (I don't think we can?) I would probably favor starting with white and going to a nice dark blue or purple here. Something like what is shown in the bottom plot here: https://towardsdatascience.com/consensus-clustering-f5d25c98eaf2

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -141,98 +148,69 @@ if (has_cellassign) {
...
     heatmap_legend_param = list(
-      title = "Log(Number of cells)",
-      title_position = "leftcenter-rot",
-      legend_height = unit(4, "cm")
+      title = "Jaccard similarity index",
+      direction = "horizontal",
+      legend_width = unit(4, "cm")
+    ),
+    # Retain only 1 legend in the final plot      
+    show_heatmap_legend = ifelse(name == keep_legend_name, TRUE, FALSE),
```
jashapiro, line 199: No need for `ifelse` if just T/F!
```suggestion
    show_heatmap_legend = name == keep_legend_name,
```

jashapiro (commented):
Some initial comments... I have to drive for a bit, but I will probably add a few more when I get home.

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: The sheer number of times I have corrected something like this on homeworks....for me to go do it myself is...impressive!

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -113,15 +112,23 @@ celltype_df <- create_celltype_df(processed_sce)
 
 ## Heatmaps
 
-Below, we show heat maps comparing cell type annotations (along the y-axis) to clustering results (along the x-axis).
-Heatmap colors represent the log number of cells present in both the given cell type and cluster.
+Below, we show heat maps comparing each set of cell type annotations to clustering results.
```
jashapiro, line 115: I added "unsupervised" here, but I actually think part of this sentence needs to be moved down, as we want to start general and then have this sentence for the first set of heatmaps, then a different one for the ones below (comparing to submitter, etc.)

```suggestion
Below, we show heat maps comparing each set of cell type annotations to unsupervised clustering results.
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -113,15 +112,23 @@ celltype_df <- create_celltype_df(processed_sce)
...
+Heatmap colors represent the Jaccard similarity index between each grouping.
+
+This metric is bounded by 0 and 1:
+
+* A value of 1 indicates complete overlap between categories.
+The categories being compared contain the same set of cells.
+* A value of 0 indicates no overlap between categories. 
+No cells that appear in one category appear in the other.
```
jashapiro, line 123: I'm trying here to clarify that the comparisons are always between labels in different methods, but since they are not all cell typing methods, it is a bit hard to word. Here is an attempt.

```suggestion
Heatmap colors represent the Jaccard similarity index between pairs of labels from different methods.
The Jaccard index reflects the degree of overlap between the two labels, and ranges from 0 to 1. 

* If the labels are assigned to identical sets of cells, the Jaccard index will be 1
* If the labels are assigned to completely non-overlapping sets of cells, the Jaccard index will be 0
* 
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -141,98 +148,69 @@ if (has_cellassign) {
...
+      row_title_gp = grid::gpar(fontsize = 10), 
+      row_title_side = "right",
+      row_names_side = "left",
+      row_names_gp = grid::gpar(fontsize = 8),
+      ## Column parameters
+      cluster_columns = FALSE,
+      column_title = glue::glue("Cluster"),
+      column_names_side = "top",
+      column_names_rot = 0,
+      column_names_gp = grid::gpar(fontsize = 12),
```
jashapiro, line 191: Even though we have more space here, I think probably we want to keep just one font size for rows and columns?

jashapiro (commented):
A few more wording comments... which I am not really happy with, because I always describing cluster/label comparison really hard.

templates/qc_report/celltypes_supplemental_report.rmd
allyhawkins, line 123: I think I might reorganize this a bit. We are using the Jaccard index for both this heatmap and all the other heatmaps right? So I think I would start with a general statement about the section as a whole - that we are interested in comparing cell type annotations to clusters and to each other. Then I would say that we are using the Jaccard Index to measure the similarity between these different groups. Then you have an intro sentence before each heatmap saying what exactly is being compared and used to calculate the Jaccard index for that heatmap.

allyhawkins:
Just noting that I'm in the middle of reviewing this, so if you could wait to accept Josh's comments until I finish so I don't lose my comments, that would be great!

templates/qc_report/celltypes_supplemental_report.rmd
allyhawkins: I like this idea of using white for close to 0 and a dark color for 1.

allyhawkins (commented):
I think this looks generally good, I didn't have too much to add on top of Josh's comments.

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: > Since we don't have NA values (I don't think we can?) 

Confirming we can't, 🐛's aside! All the `NA`s will have been converted at this point to strings.

jashapiro: I meant for the Jaccard calculation, which better not be strings!

sjspielman: they better not! 😱

sjspielman:
Vote on using a border? I think probably yes(?), in case of lots of white (again this is fake data) - 
Screenshot 2023-10-30 at 4 15 37 PM
Screenshot 2023-10-30 at 4 18 15 PM

This color is `darkslateblue` and I really like it; I will need a compelling argument to switch it up ;)

allyhawkins:
+1 for the border! And I like the colors!

jashapiro:
+1 on the border. Color looks good to me, but maybe also try `midnightblue` just to compare? (Basically the first single hue option from https://colorbrewer2.org/#type=sequential&scheme=Blues&n=9)

sjspielman:
`midnightblue` looks _strikingly_ similar..! I knit this several times to be sure it was right since I could barely tell the difference! In the end, the `midnightblue` is _slightly_ darker but looks mostly the same as the color I liked, and given the ~increased contrast I think it's the move.
Screenshot 2023-10-30 at 5 12 10 PM

sjspielman:
Just one more - this is `#08306b` <-- oooh github so fancy
Screenshot 2023-10-30 at 5 16 16 PM

jashapiro:
> Just one more - this is `#08306b` <-- oooh github so fancy

The `darkslateblue` looks better to me... The way it made the colors with my choice is too grey in the middle.

I also started playing around just a bit, and was looking at `circlize::colorRamp2(c(0, 1), hcl_palette = "Blues 3", reverse = TRUE)`, which is pretty good, but doesn't start quite at white. (The other HCL palettes can be seen at https://blog.r-project.org/2019/04/01/hcl-based-color-palettes-in-grdevices/)

sjspielman:
This could be a way to go with `Blues` and get white to start 🎨  
```
mycolors <- c("#ffffff", hcl.colors(n = 5, palette = "Blues", rev =TRUE)[2:5])
col_fun <- circlize::colorRamp2(c(0, 0.25, 0.5, 0.75, 1), colors = mycolors)
```

Screenshot 2023-10-31 at 8 38 13 AM

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -141,154 +145,92 @@ if (has_cellassign) {
...
+      cluster_columns = FALSE,
+      column_title = glue::glue("Cluster"),
+      column_names_side = "top",
+      column_names_rot = 0,
+      column_names_gp = grid::gpar(fontsize = 8), # match row name size
+      ## Legend parameters
+      heatmap_legend_param = list(
+        title = "Jaccard similarity index",
+        direction = "horizontal",
+        legend_width = unit(4, "cm")
```
jashapiro, line 188: For consistent units
```suggestion
        legend_width = unit(1.5, "in")
```

sjspielman:
### Current status

- I updated some of the Jaccard text in the spirit of suggestions (but I not directly accept them hence "spirit"; too many moving parts so did it myself).
- I liked @allyhawkins's idea to have an overall organization of first we say comparisons are nice, then Jaccard, and then show each of the heatmaps. But, I struggled to get the "top-level" wording right since -
  - we will always be comparing to clusters, but...
  - we will only sometimes be comparing to submitter
  - we will only sometimes be comparing singler and cellassign to one another

So, as a _concept_, I _stubbed out_ some organization (more text, and obviously heatmaps!, to come) where each "type of heatmap" is in its own header. I don't really like that the Jaccard is buried only in the first header though, so this is just a concept to be able to say "yes expand on this" or "no back up for different organization."

Also right now we have the `blues` strategy I posted in the heatmap above, again can revert to `darkslateblue` but now this one is growing on me..!

### When to move on to the next PR

Let's agree on the right approach for organizing these heatmaps and their associated text, and if we have that + heatmap aesthetics, let's move this along. In next PR, I'll make the heatmap code a function and make some more heatmaps under the organization we decide.



### Here's a screenshot of relevant section of the report
<br>
Screenshot 2023-10-31 at 9 08 17 AM

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -109,19 +108,24 @@ has_submitter <- "submitter" %in% metadata(processed_sce)$celltype_methods
 celltype_df <- create_celltype_df(processed_sce)
 ```
 
+## Comparison to clusters
```
jashapiro, line 111: Here is my stab at how to organize the text with subsections. 

```
## Cell label comparison plots

This section displays heatmaps comparing cell type labels from various methods. 

We use the [Jaccard similarity index](https://en.wikipedia.org/wiki/Jaccard_index) to display the agreement between between pairs of labels from different methods.

...

### Comparison with unsupervised clustering

Here we show the results from unsupervised clustering compared against cell type annotations. 
Cluster assignment was performed using the `r metadata(processed_sce)$cluster_algorithm` algorithm.
```
(I left off weighting here: I think that is more detail than we need in this report, and it does not capture the full array of parameters which would really be needed to compare to other clustering results)

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -141,154 +145,92 @@ if (has_cellassign) {
...
-  x_label = "Submitter-provided annotations",
-  y_label = "SingleR annotations",
-  y_title_location = "top",
-  column_names_rotation = 55,
-  column_font_size = 8
+```{r, eval = has_submitter & (has_cellassign | has_singler)}
+methods_string <- dplyr::case_when(
+  has_cellassign & has_singler ~ glue::glue("`SingleR` and `CellAssign`"),
+  has_cellassign  ~ glue::glue("`CellAssign`"),
+  has_singler  ~ glue::glue("`SingleR`"),
```
jashapiro, line 213: I don't think `glue` is doing anything here?

(Also made order consistent)
```suggestion
  has_singler & has_cellassign ~ "`SingleR` and `CellAssign`",
  has_singler  ~ "`SingleR`",
  has_cellassign  ~ "`CellAssign`",
```

jashapiro (commented):
I tried my own hand at what I was thinking for organization Ally and I had suggested (my original suggestion may have been unclear, but I think we meant the same thing). I agree that it is a bit strange that there will be cases where there is only one subsection (just comparison of clustering to singler, presumably), but I think that is okay...

The heatmap itself looks good to me. Either color scheme is fine!

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: > I don't think `glue` is doing anything here?

I thought it might be good for printing these as code-styled when used elsewhere, in particular since one has an "and" which we don't want to be code-style, but maybe if it's inside a `glue::glue()` later it will work without that here! Will test and Do the Right Thing™️.

templates/qc_report/celltypes_supplemental_report.rmd
allyhawkins: +1 for this! This is similar to what I had envisioned when I wrote that comment.

allyhawkins: Another thought, we could tackle the heatmap aesthetics in this PR, turn it into a function in the next PR, and then add in the text as a separate PR. That might help focus the remaining reviews.

sjspielman: > Another thought, we could tackle the heatmap aesthetics in this PR, turn it into a function in the next PR, and then add in the text as a separate PR. That might help focus the remaining reviews.

yeah, i'm good with this. I want to get this text in now, but we can polish it up later!

allyhawkins (approved):
LGTM! For the record, I like the `darkslateblue`!

jashapiro (approved):
LGTM!

PR #532, opened 2023-10-30
Save singleR reference label and ontology data frame to singleR results

allyhawkins:
In working on processing some projects through CellAssign, I was running into an error when adding the cell types to the SCE object. 
Screenshot 2023-10-30 at 11 48 39 AM

I went back and realized that we weren't properly storing the name of the reference label in the SingleR results file. Also, we didn't pass on the `cell_ontology_df` from the model to the results. We need that to map cell ontology IDs. 

One other thing to note is that I had to separate out modifying the `colData` and replacing the `colData`. Nextflow was complaining about the `colData` function.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -81,7 +81,7 @@ if(!is.null(opt$singler_results)){
     colData(sce) <- annotations_df |>
       dplyr::left_join(
         # column names: ontology_id, ontology_cell_names
-        singler_results$cell_ontology_df,
+        metadata(singler_results)$cell_ontology_df,
         by = c("singler_celltype_annotation" = "ontology_id")
       ) |>
       # rename columns
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -95,10 +95,11 @@ if(!is.null(opt$singler_results)){
   }
 
   # add annotations to colData
-  colData(sce) <- colData(sce) |>
+  new_coldata <- colData(sce) |>
     as.data.frame() |>
     dplyr::left_join(annotations_df, by = c("barcodes")) |>
     DataFrame(row.names = colData(sce)$barcodes)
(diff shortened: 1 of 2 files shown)


bin/add_celltypes_to_sce.R
```diff
@@ -95,10 +95,11 @@ if(!is.null(opt$singler_results)){
   }
 
   # add annotations to colData
-  colData(sce) <- colData(sce) |>
+  new_coldata <- colData(sce) |>
     as.data.frame() |>
     dplyr::left_join(annotations_df, by = c("barcodes")) |>
     DataFrame(row.names = colData(sce)$barcodes)
+  colData(sce) <- new_coldata
```
sjspielman, line 102: i dont understand why nextflow would complain about this other than it likes to complain, but sure why not.

bin/classify_SingleR.R
```diff
@@ -95,7 +95,10 @@ singler_results <- SingleR::classifySingleR(
 # add reference name to singler_results DataFrame metadata
 metadata(singler_results)$reference_name <- reference_name
 # add label name to metadata
-metadata(singler_results)$reference_label <- metadata(singler_model)$reference_label
+metadata(singler_results)$reference_label <- singler_model$reference_label
+# save cell ontology table to results
+# if this doesn't exist it will just be NULL
```
sjspielman, line 100: we probably dont need this comment although it is true! i dont mind leaving it though if you really want it!

sjspielman (approved):
👍

bin/classify_SingleR.R
allyhawkins: I think this is more a reminder to myself if things break later and I'm trying to figure out why.

bin/add_celltypes_to_sce.R
allyhawkins: For context, this was the error I was seeing, but could not replicate it locally. 

Screenshot 2023-10-30 at 12 35 33 PM

PR #533, opened 2023-10-30
Update tag to v0.6.1

allyhawkins:
Since we recently fixed an error with processing samples without cell types from submitters (see https://github.com/AlexsLemonade/scpca-nf/pull/524), we need to tag a new release. This just updates the tag to v0.6.1 throughout.


jashapiro (approved):
LGTM

PR #534, opened 2023-10-30
Sync updates to v0.6.1 to development

allyhawkins:
No new changes. Syncing #533 to the `development` branch.


jashapiro (approved):
We'll be updating these again when dev goes in, so this is probably not really needed, but it's fine anyway.

PR #536, opened 2023-11-01
Add remaining heatmaps

sjspielman:
Ever towards #516 

More heatmaps, and please enjoy the branch name, and if you don't enjoy the branch name, it means you still need to see _Zoolander_.

### Changes in this PR

1. Heatmap function has been implemented
2. Although I use the heatmap function for the first two heatmaps, I don't for the third (directly comparing singler-cellassign) since it is not a list of stacked heatmaps.
    - If we want, I can create a 2nd function to just make a heatmap from 1 matrix and call it from the current function. This would modularize the code further; let me know if we want to see this happen! I didn't do it (yet?) since this heatmap seemed sufficiently different that it might be fine to not use the function.
3. I realized that I have been developing this with unrealistic submitter labels - they have just been single letters! We need to allow for longer names, so I updated my test data to use long names to see what needed to change. Changes were:
    - I moved heatmap column labels to the bottom, but kept the column title at the top for spacing.
    - I wrote a function to more dynamically determine the right font size based on the label `nchar`
    - I allow for column labels to now be at an angle
4. Smaller change, I updated text to say "Jaccard ~similarity~ index" all around. Shorter and consistent with the wikipedia link.

For this round, I'm attaching the full report since there are really 3 heatmaps to look at given the changes I've made.
[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13228985/celltypes_supplemental_report.html.zip)

Again, we expect 1 more PR with textual updates (note I may bundle those with #514 text updates).

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -124,16 +124,18 @@ draw_heatmap <- function(heatmap, legend_side = "bottom") {
 #' @param labels_font_size Font size to use for rows and column labels.
 #' @param keep_legend_name The name to use in the legend
 #' @param column_names_angle Angle for column names. Default is 0.
-#' @param col_fun Color function for the heatmap palette. Default is `heatmat_col_fun`.
+#' @param col_fun Color function for the heatmap palette. Default is `heatmap_col_fun`.
 #' 
 #' @return A ComplexHeatmap object
-create_single_heatmap <- function(mat,
-                                  row_title, 
-                                  column_title, 
-                                  labels_font_size,
-                                  keep_legend_name,
-                                  column_names_angle = 0,
-                                  col_fun = heatmap_col_fun) {
+create_single_heatmap <- function(
+  mat,
+  row_title, 
+  column_title, 
+  labels_font_size,
+  keep_legend_name,
+  column_names_angle = 0,
+  col_fun = heatmap_col_fun
+) {
   
   heat <- ComplexHeatmap::Heatmap(
     t(mat), # transpose because matrix rows are in common & we want a vertical arrangement
@@ -159,7 +161,7 @@ create_single_heatmap <- function(mat,
       direction = "horizontal",
       legend_width = unit(1.5, "in")
     ),
-    # Retain only 1 legend in the final plot      
+    # only keep legends that match `keep_legend_name`
     show_heatmap_legend = row_title == keep_legend_name,
   )
   
@@ -279,13 +281,13 @@ Cluster assignment was performed using the `r metadata(processed_sce)$cluster_al
 # Calculate all jaccard matrices of interest for input to heatmap
 jaccard_cluster_matrices <- available_celltypes |> 
   stringr::str_to_lower() |>
+  purrr::set_names(available_celltypes) |>
   purrr::map(\(name) {
     make_jaccard_matrix(
       celltype_df, 
       "cluster", 
       glue::glue("{name}_celltype_annotation")
-    )}) |>
-  purrr::set_names(available_celltypes)
+    )}) 
 
 n_celltypes <- jaccard_cluster_matrices |>
   purrr::map(colnames) |> 
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -93,7 +93,7 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
 ```
 
 
-<!-- Define variables, options, and function for plotting heatmap from list of matrices --> 
+<!-- Define variables, options, and functions for plotting heatmaps --> 
 ```{r}
 # Define color ramp for shared use in the heatmap
 heatmap_col_fun <- circlize::colorRamp2(c(0, 1), colors = c("white", "darkslateblue"))
@@ -101,64 +101,102 @@ heatmap_col_fun <- circlize::colorRamp2(c(0, 1), colors = c("white", "darkslateb
 # Set heatmap padding option
 ComplexHeatmap::ht_opt(TITLE_PADDING = grid::unit(0.2, "in"))
 
-#' Function to plot a ComplexHeatmap
+
+#' Draw a ComplexHeatmap
+#'
+#' @param heatmap A heatmap or list of heatmaps to draw
+#' @param legend_side Side legend appears on, default bottom
+draw_heatmap <- function(heatmap, legend_side = "bottom") {
+  # Render the heatmap or list of heatmaps
+  ComplexHeatmap::draw(
+    heatmap, 
+    heatmap_legend_side = legend_side
+  )
+}
+
+
+#' Create a ComplexHeatmap from a matrix
+#'
+#' @param mat Matrix to create heatmap from.
+#' @param row_title Label for row title.
+#' @param column_title Label for column title.
+#' @param labels_font_size Font size to use for rows and column labels.
+#' @param keep_legend_name The name to use in the legend
+#' @param column_names_angle Angle for column names. Default is 0.
+#' @param col_fun Color function for the heatmap palette. Default is `heatmat_col_fun`.
+#' 
+#' @return A ComplexHeatmap object
+create_single_heatmap <- function(mat,
+                                  row_title, 
+                                  column_title, 
+                                  labels_font_size,
+                                  keep_legend_name,
+                                  column_names_angle = 0,
+                                  col_fun = heatmap_col_fun) {
+  
+  heat <- ComplexHeatmap::Heatmap(
+    t(mat), # transpose because matrix rows are in common & we want a vertical arrangement
+    col = col_fun,
+    border = TRUE, # each heatmap gets its own outline
+    ## Row parameters
+    cluster_rows = FALSE,
+    row_title = row_title, # each heatmap gets its own title
+    row_title_gp = grid::gpar(fontsize = 10), 
+    row_title_side = "right",
+    row_names_side = "left",
+    row_names_gp = grid::gpar(fontsize = labels_font_size),
+    ## Column parameters
+    cluster_columns = FALSE,
+    column_title = column_title,
+    column_title_gp = grid::gpar(fontsize = 10), 
+    column_names_side = "bottom",
+    column_names_rot = column_names_angle,
+    column_names_gp = grid::gpar(fontsize = labels_font_size),
+    ## Legend parameters
+    heatmap_legend_param = list(
+      title = "Jaccard index",
+      direction = "horizontal",
+      legend_width = unit(1.5, "in")
+    ),
+    # Retain only 1 legend in the final plot      
+    show_heatmap_legend = row_title == keep_legend_name,
+  )
+  
+  return(heat)
+}    
+      
+#' Function to plot a vertically-stacked ComplexHeatmap from a list of matrices
 #'
 #' @param matrix_list List of matrices to plot in a vertical layout. 
 #' @param column_title Title to use for columns, shared among all heatmaps
-#' @param column_names_angle Angle for column names
 #' @param labels_font_size Font size to use for rows and column labels.
-#' @param col_fun Color function for the heatmap palette. Default is `heatmat_col_fun`
+#' @param column_names_angle Angle for column names. Default is 0.
+#' @param col_fun Color function for the heatmap palette. Default is `heatmat_col_fun`.
 #'
-#' @return Draws a ComplexHeatmap, but does not explicitly return anything
-plot_heatmap <- function(matrix_list,
-                         column_title, 
-                         column_names_angle = 0,
-                         labels_font_size,
-                         col_fun = heatmap_col_fun) {
+#' @return A list of ComplexHeatmap objects
+create_heatmap_list <- function(matrix_list,
+                              column_title, 
+                              labels_font_size,
+                              column_names_angle = 0,
+                              col_fun = heatmap_col_fun) {
   
   
   # We only want one shared legend in the end, arbitrarily grab the first one
   keep_legend_name <- names(matrix_list)[1]
 
   heatmap_list <- matrix_list |> 
-    purrr::imap(\(mat, name){
-      ComplexHeatmap::Heatmap(
-        t(mat), # transpose because matrix rows are in common & we want a vertical arrangement
-        col = col_fun,
-        border = TRUE, # each heatmap gets its own outline
-        ## Row parameters
-        cluster_rows = FALSE,
-        row_title = name, # each heatmap gets its own title
-        row_title_gp = grid::gpar(fontsize = 10), 
-        row_title_side = "right",
-        row_names_side = "left",
-        row_names_gp = grid::gpar(fontsize = labels_font_size),
-        ## Column parameters
-        cluster_columns = FALSE,
-        column_title = column_title,
-        column_names_side = "bottom",
-        column_names_rot = column_names_angle,
-        column_names_gp = grid::gpar(fontsize = labels_font_size),
-        ## Legend parameters
-        heatmap_legend_param = list(
-          title = "Jaccard index",
-          direction = "horizontal",
-          legend_width = unit(1.5, "in")
-        ),
-        # Retain only 1 legend in the final plot      
-        show_heatmap_legend = name == keep_legend_name,
-      )
-    }) |>
+    purrr::imap(
+      create_single_heatmap, 
+      column_title, 
+      labels_font_size,
+      keep_legend_name,
+      column_names_angle = column_names_angle,
+      col_fun = col_fun
+    ) |>
     # concatenate vertically into HeatmapList object
     purrr::reduce(ComplexHeatmap::`%v%`) 
   
-  
-  # Render the heatmap list
-  ComplexHeatmap::draw(
-    heatmap_list, 
-    heatmap_legend_side = "bottom"
-  )
-  
+  return(heatmap_list)
 }
 
 
@@ -197,6 +235,9 @@ find_label_size <- function(input_labels,
 # define library and sce object
 library_id <- params$library
 processed_sce <- params$processed_sce
+
+library_id <- "SCPCL000295" 
+processed_sce <- readRDS(glue::glue("{library_id}_processed.rds"))
 
 # check for annotation methods
 has_singler <- "singler" %in% metadata(processed_sce)$celltype_methods
@@ -255,17 +296,15 @@ if (has_cellassign) {
 
 # set height based on number of matrices
 heatmap_height <- length(jaccard_cluster_matrices) * 2.5
-
-
-
 ```
 
 ```{r, fig.height = heatmap_height, fig.width = 8.5, warning = FALSE}
-plot_heatmap(
-  jaccard_cluster_matrices, 
-  column_title = "Clusters", 
-  labels_font_size = find_label_size(jaccard_cluster_matrices)
-)
+jaccard_cluster_matrices |>
+  create_heatmap_list(
+    column_title = "Clusters", 
+    labels_font_size = find_label_size(jaccard_cluster_matrices)
+  ) |>
+  draw_heatmap()
 ```
 
 
@@ -315,12 +354,14 @@ heatmap_height <- n_celltypes / 4.5
 ```
 
 ```{r, fig.height = heatmap_height, fig.width = 8.5, warning = FALSE}
-plot_heatmap(
-  jaccard_submitter_matrices, 
-  column_title = "Submitted-provided annotations",
-  labels_font_size = find_label_size(jaccard_submitter_matrices),
-  column_names_angle = 45
-)
+# create_heatmap(
+#   jaccard_submitter_matrices, 
+#   column_title = "Submitted-provided annotations",
+#   labels_font_size = find_label_size(jaccard_submitter_matrices),
+#   column_names_angle = 45
+# )
+# 
+# draw_heatmap
 ```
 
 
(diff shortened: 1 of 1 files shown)


templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -154,57 +253,18 @@ if (has_cellassign) {
...
-# Render the heatmap list
-ComplexHeatmap::draw(
-  heatmap_list, 
-  heatmap_legend_side = "bottom" 
+```{r, fig.height = heatmap_height, fig.width = 8.5, warning = FALSE}
+plot_heatmap(
+  jaccard_cluster_matrices, 
+  column_title = "Clusters", 
+  labels_font_size = find_label_size(jaccard_cluster_matrices)
 )
```
allyhawkins, line 268:
```suggestion
heatmap_list <- purrr::map(jaccard_cluster_matrices, \(jaccard_matrix, matrix_name) {
plot_heatmap(
  jaccard_matrix, 
  matrix_name,
  column_title = "Clusters", 
  labels_font_size = find_label_size(jaccard_cluster_matrices)
)} |> 
    # concatenate vertically into HeatmapList object
    purrr::reduce(ComplexHeatmap::`%v%`) 
  
  
  # Render the heatmap list
  ComplexHeatmap::draw(
    heatmap_list, 
    heatmap_legend_side = "bottom"

```
Then you can do this here. The only thing that I don't think this catches is dealing with the legend to keep. Can you go into the ComplexHeatmap object and remove legends like you can with ggplot?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,106 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
 
+<!-- Define variables, options, and function for plotting heatmap from list of matrices --> 
+```{r}
+# Define color ramp for shared use in the heatmap
+heatmap_col_fun <- circlize::colorRamp2(c(0, 1), colors = c("white", "darkslateblue"))
+
+# Set heatmap padding option
+ComplexHeatmap::ht_opt(TITLE_PADDING = grid::unit(0.2, "in"))
+
+#' Function to plot a ComplexHeatmap
```
jashapiro, line 104: Noting this is underselling! This function plots a list of matrices in a joined matrix!

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -226,7 +286,41 @@ knitr::asis_output(
...
+  n_celltypes <- n_celltypes + length(unique(celltype_df$singler_celltype_annotation))
+}
+if (has_cellassign) {
+  jaccard_submitter_matrices[["CellAssign"]] <- make_jaccard_matrix(
+    celltype_df, 
+    "submitter_celltype_annotation", 
+    "cellassign_celltype_annotation"
+  )
+  n_celltypes <- n_celltypes + length(unique(celltype_df$cellassign_celltype_annotation))
+}
```
allyhawkins, line 311:
```suggestion
names <- c(dplyr::if_else(has_singler, "SingleR", NA),
           dplyr::if_else(has_cellassign, "CellAssign", NA)) |> 
  na.omit()

jaccard_submitter_matrices <- names |> 
  purrr::map( \(name) {
     lower_name <- stringr::str_to_lower(name)
    make_jaccard_matrix(celltype_df, "submitter_celltype_annotation", glue::glue("{lower_name}_celltype_annotation")} |> purrr::set_names(names)
```
I think you could simplify this so that you are not duplicating the call to calculating the jaccard matrix. 
Another thing to note here is that you will over write the `n_celltypes` variable if you have both SingleR and CellAssign. I think we want to handle that separately from calculating the jaccard matrix.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -237,9 +331,58 @@ knitr::asis_output("
...
 ")
+
+# Set plot dimensions based on number of SingleR types
+heatmap_height <- length(levels(celltype_df$singler_celltype_annotation)) / 3.5
 ```
 
 
+```{r, eval = has_cellassign & has_singler, fig.height = heatmap_height, fig.width = 8.5}
+# Calculate jaccard matrix
+singler_cellassign_matrix <- make_jaccard_matrix(
```
allyhawkins, line 342: Just note that if you take my suggestion about the function you will need to update this chunk to use that function.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,106 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
+        row_title = name, # each heatmap gets its own title
+        row_title_gp = grid::gpar(fontsize = 10), 
+        row_title_side = "right",
+        row_names_side = "left",
+        row_names_gp = grid::gpar(fontsize = labels_font_size),
+        ## Column parameters
+        cluster_columns = FALSE,
+        column_title = column_title,
+        column_names_side = "bottom",
+        column_names_rot = column_names_angle,
```
allyhawkins, line 140: Personally, I prefer a 90 degree label because I think it's easier to line up which cell type is in which column.

allyhawkins: I think if you are worried about cell type names being too long, you could wrap them.

allyhawkins (commented):
My main comment here is that I think you should make the function just create one heatmap and then do any combining before drawing the heatmap outside of the function. The actual base of the heatmap is the same, the only difference is that we have stacked vs. not stacked. Because the stacked heatmap requires a list, it would be easy to create that list using `purrr::map` with a function that returns a single heatmap.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,106 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
+  heatmap_list <- matrix_list |> 
+    purrr::imap(\(mat, name){
+      ComplexHeatmap::Heatmap(
+        t(mat), # transpose because matrix rows are in common & we want a vertical arrangement
+        col = col_fun,
+        border = TRUE, # each heatmap gets its own outline
+        ## Row parameters
+        cluster_rows = FALSE,
+        row_title = name, # each heatmap gets its own title
+        row_title_gp = grid::gpar(fontsize = 10),
```
jashapiro, line 132: Can we set the `column_title_gp` to be the same? It is strange that they are different sizes. (They also seem to be padded differently, but maybe that is a function of the font size difference).

jashapiro (commented):
I think this looks good overall, but some comments below. 

* I think I would move the single heatmap function out to reuse as much as possible.
* I like "Jaccard similarity index" when we first describe it, but I am fine with "Jaccard index" for subsequent use. I don't think anyone will be confused.

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: 💪

sjspielman:
Next round:

- I wrote functions as requested, but in a slightly different way! There are now three functions:
  - 1 makes a standalone heatmap
  - 1 makes a list of heatmaps, in part by calling ☝️ 
  - 1 very simply _draws_ a heatmap
- I updated up some plot styling and cleaned up some other code spots as requested in review
- I'm now using @allyhawkins's suggested (and better!) `purrr::map()` approach for building up matrices

The one thing that I'm really not sure how to handle is dynamically setting the plot size itself. Stacked vs single heatmaps seem to have some difference in how height pans out, and in part because of this I can't tell whether the strategy I'm using is actually going to generalize well across reports. 
I'm thinking one way to handle the sizing could be to directly manipulate the heatmap dimensions itself and then set the plot size accordingly, but with some padding for vertical text? Blog post - https://jokergoo.github.io/2020/05/11/set-cell-width/height-in-the-heatmap/

Any thoughts on sizing are definitely appreciated! 

Report version as of now:
[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13243785/celltypes_supplemental_report.html.zip)

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,147 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
+#'
+#' @param heatmap A heatmap or list of heatmaps to draw
+#' @param legend_side Side legend appears on, default bottom
+draw_heatmap <- function(heatmap, legend_side = "bottom") {
+  # Render the heatmap or list of heatmaps
+  ComplexHeatmap::draw(
+    heatmap, 
+    heatmap_legend_side = legend_side
+  )
+}
```
allyhawkins, line 116: I'm not sure this really needs to be a function since you're just calling an already builtin function?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,147 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
+
+#' Create a ComplexHeatmap from a matrix
+#'
+#' @param mat Matrix to create heatmap from.
+#' @param row_title Label for row title.
+#' @param column_title Label for column title.
+#' @param labels_font_size Font size to use for rows and column labels.
+#' @param keep_legend_name The name to use in the legend
+#' @param column_names_angle Angle for column names. Default is 0.
+#' @param col_fun Color function for the heatmap palette. Default is `heatmat_col_fun`.
```
allyhawkins, line 127:
```suggestion
#' @param col_fun Color function for the heatmap palette. Default is `heatmap_col_fun`.
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,147 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
+    column_names_side = "bottom",
+    column_names_rot = column_names_angle,
+    column_names_gp = grid::gpar(fontsize = labels_font_size),
+    ## Legend parameters
+    heatmap_legend_param = list(
+      title = "Jaccard index",
+      direction = "horizontal",
+      legend_width = unit(1.5, "in")
+    ),
+    # Retain only 1 legend in the final plot
```
allyhawkins, line 162:
```suggestion
    # only keep legends that match `keep_legend_name`
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,147 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
+#' @param labels_font_size Font size to use for rows and column labels.
+#' @param column_names_angle Angle for column names. Default is 0.
+#' @param col_fun Color function for the heatmap palette. Default is `heatmat_col_fun`.
+#'
+#' @return A list of ComplexHeatmap objects
+create_heatmap_list <- function(matrix_list,
+                              column_title, 
+                              labels_font_size,
+                              column_names_angle = 0,
+                              col_fun = heatmap_col_fun) {
```
allyhawkins, line 182:
```suggestion
create_heatmap_list <- function(matrix_list,
                                column_title, 
                                labels_font_size,
                                column_names_angle = 0,
                                col_fun = heatmap_col_fun) {
```
spacing

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -129,89 +277,37 @@ Cluster assignment was performed using the `r metadata(processed_sce)$cluster_al
...
-# Render the heatmap list
-ComplexHeatmap::draw(
-  heatmap_list, 
-  heatmap_legend_side = "bottom" 
-)
+```{r, fig.height = heatmap_height, fig.width = 8.5, warning = FALSE}
+jaccard_cluster_matrices |>
+  create_heatmap_list(
+    column_title = "Clusters", 
+    labels_font_size = find_label_size(jaccard_cluster_matrices)
```
allyhawkins, line 304: I wonder if you could use `purrr::flatten()` to get a list of all colnames in the list of data frames. Then you can use that as input to the `find_label_size` function and remove what's in the `if(is_list)` chunk and just start with a list of all column names.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -226,20 +322,90 @@ knitr::asis_output(
...
+  "cellassign_celltype_annotation"
+)
+
+labels_font_size <- find_label_size(
+  c(
+    rownames(singler_cellassign_matrix), 
+    colnames(singler_cellassign_matrix)
+  ), 
+  is_list = FALSE
+)
```
allyhawkins, line 396:
```suggestion
labels_font_size <- find_label_size(
  c(
    rownames(singler_cellassign_matrix), 
    colnames(singler_cellassign_matrix)
  )
)
```
If you are able to figure out flattening the list as input, then you could remove the `is_list` here.

allyhawkins (commented):
Thanks for making these changes! Definitely easier to follow. I think I would still remove the `draw_heatmap` function since you are really just using an existing function there and not doing anything else special. 

Also in regards to heatmap size, I'm having trouble opening the zip file that you sent so can't comment on if what you're doing will turn out okay.

jashapiro (commented):
Getting there! I had a few additional comments, but largely agree with things Ally said, and also couldn't open the zip file.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -226,20 +322,90 @@ knitr::asis_output(
...
+singler_cellassign_matrix <- make_jaccard_matrix(
+  celltype_df, 
+  "singler_celltype_annotation",
+  "cellassign_celltype_annotation"
+)
+
+labels_font_size <- find_label_size(
+  c(
+    rownames(singler_cellassign_matrix), 
+    colnames(singler_cellassign_matrix)
```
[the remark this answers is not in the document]
sjspielman: Yeah, this clearly isn't right. I know I had a good reason for doing this at one point but it entirely escapes me now. Plus, this probably explains why I wasn't getting the "right dimensions" for this plot..!

Either way, I think we what _should_ be doing is considering:
1. The number of cell types along the y axis (row names)
2. The nchar of cell types along the x axis (column names)

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -226,20 +322,90 @@ knitr::asis_output(
...
 knitr::asis_output("
 ### Comparison between `SingleR` and `CellAssign` annotations
 
 This section displays a heatmap directly comparing `SingleR` and `CellAssign` cell type annotations.
 Note that due to different annotations references, these methods may use different names for similar cell types.
 ")
+
+# Set plot dimensions based on number of SingleR cell types
+n_celltypes <- length(levels(celltype_df$singler_celltype_annotation))
+heatmap_height <- n_celltypes/4 + 2
```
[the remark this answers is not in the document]
sjspielman: No idea!!! For some reason 6 _really_ squished this plot in ways that it did not squish others, and 4 didn't. As noted in https://github.com/AlexsLemonade/scpca-nf/pull/536#issuecomment-1791424482 I'm not a fan of how I've set up sizing. Something is still missing. Let's circle back after I get the other reviews in.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -129,89 +277,37 @@ Cluster assignment was performed using the `r metadata(processed_sce)$cluster_al
...
+    )}) |>
+  purrr::set_names(available_celltypes)
+
+n_celltypes <- jaccard_cluster_matrices |>
+  purrr::map(colnames) |> 
+  unlist() |> 
+  length()
+  
+# set height based on number of celltypes, and add on 2x padding
+# there are no long vertical layers here to accommodate
```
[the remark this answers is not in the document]
sjspielman: Ah, typo - "there are no long vertical **labels** here to accommodate" (cluster numbers along the axis, not long cell type names)

sjspielman:
Sorry for bad report! I think I just added a `.zip` extension instead of actually zipping it 😂 😭 
I'll post one when I re-request review next!

sjspielman:
I've gotten all reviews in, so next we just need to figure out the Right™️ way to determine heatmap height. Getting there!!

Here's a _confirmed-working_ `zip` file!
[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13253956/celltypes_supplemental_report.html.zip)

templates/qc_report/celltypes_supplemental_report.rmd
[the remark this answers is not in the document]
jashapiro: > Either way, I think we what _should_ be doing is considering:
> 
> 1. The number of cell types along the y axis (row names)
> 2. The nchar of cell types along the x axis (column names)

That sounds right to me. Maybe a little function to do the work, like the font size function, that takes a list of rownames and colnames and goes from there.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -129,89 +256,38 @@ Cluster assignment was performed using the `r metadata(processed_sce)$cluster_al
...
+      glue::glue("{name}_celltype_annotation")
+    )}) 
+
+unique_celltypes <- jaccard_cluster_matrices |>
+  purrr::map(colnames) |> 
+  unlist() 
+  
+# set height based on number of celltypes, and add on 2x padding
+# there are no long vertical labels here to accommodate, only cluster names (ints)
+heatmap_height <- length(unique_celltypes)/6 + 2*heatmap_padding
```
allyhawkins, line 275:
```suggestion
heatmap_height <- length(all_celltypes)/6 + 2*heatmap_padding
```
I don't know if this is the "right" way to do this, but I do think it's an okay way. If you are using cell types to determine the height then ideally, it should scale as the number of cell types change. You could try with adding in like 10 more submitter types and see if it still looks okay? But honestly it's probably fine and if we notice things being weird after running some samples we can always revisit this.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -226,21 +302,85 @@ knitr::asis_output(
...
+    make_jaccard_matrix(
+      celltype_df, 
+      "submitter_celltype_annotation", 
+      glue::glue("{name}_celltype_annotation")
+    )
+  }) |>
+  purrr::set_names(available_celltypes)
+
+# how many unique cell types?
+unique_celltypes <- jaccard_submitter_matrices |>
```
allyhawkins, line 324:
```suggestion
all_celltypes <- jaccard_submitter_matrices |>
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -226,21 +302,85 @@ knitr::asis_output(
...
+  purrr::map(colnames) |> 
+  unlist() 
+
+# Set plot dimensions based on number of cell types plus:
+## 1x padding
+## 2" for long labels
+heatmap_height <- length(unique_celltypes)/6 + heatmap_padding + 2
```
allyhawkins, line 331:
```suggestion
heatmap_height <- length(all_celltypes)/6 + heatmap_padding + 2 
```

allyhawkins (approved):
I think this looks good, and I'm okay with how you are determining the height. It's going to be hard to know for sure until we run a few samples and look at how the plots look once you change the number of cell types and clusters, etc.

sjspielman:
>  It's going to be hard to know for sure until we run a few samples and look at how the plots look once you change the number of cell types and clusters, etc.

I've been playing around with this for most of the morning so far, and I think you are almost certainly right that we'll need to circle back to this. Let me send over a function I've been working on just to make things more consistent, and I'll also open another issue to circle back and fix sizing once we have more (& realistic!) examples to work with.

sjspielman:
@allyhawkins I just pushed over the function I wrote which we can circle back to, for a quick peek from you?
Since my code diverged from what you reviewed, I swapped `unique_celltypes` -> `all_celltypes` on my own. 
One other change - use `stringr::str_length()` instead of `nchar()` just in case the vector that gets passed in is a factor and not a character vector.

All here: https://github.com/AlexsLemonade/scpca-nf/pull/536/commits/07bf56a4fe7dcfdda5fef4b2d9ed7b7b2b0a82b2

sjspielman:
Note - https://github.com/AlexsLemonade/scpca-nf/issues/541

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -94,6 +93,157 @@ make_jaccard_matrix <- function(celltype_df, colname1, colname2){
...
+
+  heatmap_list <- matrix_list |> 
+    purrr::imap(
+      create_single_heatmap, 
+      column_title, 
+      labels_font_size,
+      keep_legend_name,
+      column_names_angle = column_names_angle,
+      col_fun = col_fun
+    ) |>
```
jashapiro, line 185: Since the tidyverse people say they are now preferring lambda functions, I converted this, and put the `...` in for where I would do that. 
```suggestion
    purrr::imap(
      \(mat, name) create_single_heatmap(
        mat,
        name,
        column_title, 
        labels_font_size,
        keep_legend_name,
        col_fun,
        ...
      )
    ) |>
```

jashapiro (approved):
LGTM. I just had one real suggestion, which is to not to rename the ComplexHeatmaop `column_names_rot` variable in the wrapper functions. I even went so far as to suggest using `...` to pass it along, but that is up to you.

Also a couple more little future-proofing suggestions, but I don't think I should need to see this again.

allyhawkins (approved):
👍

PR #540, opened 2023-11-03
Make sure that AnnData conversion is skipped for any library with cell hashing

allyhawkins:
When running the multiplexed project through the workflow, I noticed that one of the libraries was still getting converted to AnnData, even though it has a cell hashing library. This then threw an error in the actual conversion process which we want to avoid. This library was still around because even though it has a cell hashing library, it's not actually multiple samples, so it doesn't get categorized as a `multiplex_libs`. So then it's not filtered out. 

Here, I created a separate list of all libraries that have cell hashing even if they aren't multiplexed. I then use that to filter before converting to AnnData. I tested this with the problem library and AnnData conversion was successfully skipped.

### DIFF
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -55,10 +55,10 @@ if (!(stringr::str_ends(opt$output_rna_h5, ".hdf5|.h5"))) {
 # sure that the sce that is getting converted to AnnData is compliant with CZI
 # CZI 3.0.0 requirements: https://github.com/chanzuckerberg/single-cell-curation/blob/b641130fe53b8163e50c39af09ee3fcaa14c5ea7/schema/3.0.0/schema.md
 format_czi <- function(sce) {
-  
+
   # add schema version
   metadata(sce)$schema_version = "3.0.0"
-  
+
   # add library_id as an sce colData column
   # need this column to join in the sample metadata with the colData
   sce$library_id <- metadata(sce)$library_id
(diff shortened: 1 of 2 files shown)


jashapiro:
While this is one solution, I wonder if we might actually want to keep this library? If so, we could modify the script to remove the cellhash data before converting it to AnnData? It is only one library, so either way is probably fine, but I can't think of a reason _not_ to make an anndata file here, except that it might be the only library in the project that has it.

allyhawkins:
> While this is one solution, I wonder if we might actually want to keep this library? If so, we could modify the script to remove the cellhash data before converting it to AnnData? It is only one library, so either way is probably fine, but I can't think of a reason _not_ to make an anndata file here, except that it might be the only library in the project that has it.

Yea I think this is a good point. There are actually three other libraries that are just single-cell, so it wouldn't be the only single-cell library. I can play around with removing it before conversion. I haven't had a chance to really dig into the error, so I will do that.  I think it's because it's trying to convert the `altExp` that contains the cell hash data. So if that's the case, then we can just remove that. Just for context, here's the error: 

Screenshot 2023-11-03 at 1 47 54 PM

jashapiro:
I think that error may well be what happens if there are multiple assays... But it is pretty inscrutable!

allyhawkins:
> I think that error may well be what happens if there are multiple assays... But it is pretty inscrutable!

`zellkonverter` just ignores anything that's an `altExp`, so we convert the main experiment and any `altExp`'s separately. 
This error is from specifically trying to convert the cell hashing experiment. It's the same error we were getting with only one cell, but here we just have one HTO. So I think we need to add a check within the `scpcaTools` function that the number of rows in the experiment getting converted is > 1. Similar to how we check the number of cells: 
https://github.com/AlexsLemonade/scpcaTools/blob/ca04ccd958898e81c530b92182d9f55c28cc2d95/R/sce_to_anndata.R#L32-L34

Then in  the `sce_to_anndata.R` script in `scpca-nf`, we can check the number of rows in the `altExp` object and only convert if the number of rows > 1. This would stop any single-sample multiplexed samples from producing an AnnData object and catch any potential future experiments with only one ADT tag. I don't anticipate that actually happening, but you honestly never know at this point.

allyhawkins:
@jashapiro I went ahead and added a check for the number of rows in the `altExp` here. I didn't add a check for the number of rows in the main experiment because we always have the same number of genes. I can add it if you think it's necessary. I tested this with the library that was having issues and things worked as expected.

jashapiro (approved):
LGTM!

>  I didn't add a check for the number of rows in the main experiment because we always have the same number of genes. I can add it if you think it's necessary. 

Seems fine. If we ever get only a single gene, something is very wrong and probably worth the workflow failing!

PR #548, opened 2023-11-08
Update the reference name to include `-compartment`

### LINKED ISSUE #537: Change reference names stored in reference metadata
We have decided that `blood` or `muscle` is not totally reflective of what's included when building the PanglaoDB references. So we need to change the `ref_name` to be `<organ>_compartment` in the reference metadata file stored in `references`.

allyhawkins:
Closes #537 

This was pretty straight forward. I updated the name for the PanglaoDB reference to be `<organ>-compartment` as we discussed.


sjspielman (approved):
👍

PR #551, opened 2023-11-09
Track cell type reference version in reference files 

### LINKED ISSUE #470: Update celltype reference building to better track reference versions

allyhawkins:
Closes #470 

This PR makes adjustments to the `build-celltype-ref.nf` workflow to produce reference files that contain the reference version in the filename. I made the following changes: 

- The reference files for both SingleR and CellAssign are now in the format `<ref_name>_<ref_source>_<version>.<ext>`. 
- The `celldex` version is grabbed in the `save-celldex-refs.R` script by grabbing the package version. I then replace the `.` in the version string with `-`, because I feel like `.`'s in  filenames are bad. 
- I adjusted for the fact that we won't know the full filename that is output by `save-celldex-refs.R`, by changing the output file argument in the script to `ref_file_prefix`. Then `version.rds` is appended to that prefix before exporting the reference file.
- For PanglaoDB, I changed the reference file name to have the date in ISO8601. I was concerned that PanglaoDB might do unknown things in how they name their files in the future, so I didn't want to assume the date pattern would always be the same. I updated the internal instructions to account for this change in the filename. 
- Changing to ISO8601 allows me to use a regex to grab the date from the filename in the workflow and directly store that as the version. 
- I noticed that we were accidentally publishing the tx2gene file when we passed it through the SingleR processes. This meant we got an extra ref file in our ref directory that shouldn't be there. I made some adjustments to stop that from happening. 

The last thing I did here, which I'm open to removing, is add a stub run for this workflow to our GHA for stubs. I also think it's helpful for development to be able to use this workflow with the `-stub` profile we have locally. So I made a few adjustments to account for that. I do think we should do the same thing with the `build-index.nf`, but I didn't want to overcrowd this PR. If others are okay with adding this in, then I'll file an issue for that. 

Lastly, I left the original files in `s3://scpca-references` for now, but once I make the next adjustments to the actual workflow, then I'll remove those. The new updated files with the new file names should already be there.


bin/save_singler_refs.R
```diff
@@ -12,15 +12,25 @@ option_list <- list(
     help = "name associated with celldex reference, e.g., HumanPrimaryCellAtlasData"
   ),
   make_option(
-    opt_str = c("--ref_file"),
+    opt_str = c("--ref_file_prefix"),
     type = "character",
-    help = "path to save reference file"
+    help = "prefix to use for naming saved reference file.
+      All files will be saved with the following name: <ref_file_prefix>_celldex_version.rds"
```
sjspielman, line 18: Two suggestions - 

- seems like `celldex` is already part of `ref_file_prefix`
- adding `v` to match how you code it up below (or you can leave off the `v` all around, I don't think we need it? up to you)

```suggestion
      All files will be saved with the following name: <ref_file_prefix>_v<celldex_version>.rds"
```

bin/save_singler_refs.R
```diff
@@ -12,15 +12,25 @@ option_list <- list(
...
+celldex_version <- packageVersion("celldex") |>
+  as.character() |>
+  # replace . in version string to avoid . in filename
+  stringr::str_replace_all("\\.", "-")
+
+# construct output ref file name
+ref_output_file <- glue::glue("{opt$ref_file_prefix}_v{celldex_version}.rds")
+
 # check that provided ref name is in celldex package
 if (!opt$ref_name %in% ls("package:celldex")) {
```
sjspielman, line 35: Just a note that this _doesn't work_ unless `celldex` is actually loaded, and it is! We may want to add a comment at `library(celldex)` at the top that it must be actually loaded and we can't get away with `::`, for future us...

bin/save_singler_refs.R
sjspielman, line 1: One general comment that I'm not sure about. In the nextflow code we are a much less prescriptive that the reference source is `celldex` via use of `ref_source` variable. 

So, I wonder if we want to rename this script to `save_celldex_refs.R`. I would leave the cellassign script as is (though I kind of want it to be named `save_` and not `generate_` for consistency, but I can certainly live with it as is) since it's not necessarily specific to PanglaoDB. But, our singler script is quite specific to the celldex reference.

build-celltype-ref.nf
```diff
@@ -61,7 +63,10 @@ process generate_cellassign_refs {
   output:
     path ref_file
   script:
-    ref_file="${ref_source}-${ref_name}.tsv"
+    // get ref version from filename
+    // this requires the date stored in the filename to be in ISO8601 format
+    ref_version = (marker_gene_file =~ /.+([0-9]{4}\-[0-9]{2}\-[0-9]{2}).tsv/)[0][1]
```
sjspielman, line 68: We could get a bit more specific (and even more specific than this too, but seems overkill after these additions)...

```suggestion
    ref_version = (marker_gene_file =~ /PanglaoDB_markers_(2[0-9]{3}\-[0-9]{2}\-[0-9]{2}).tsv/)[0][1]
```

sjspielman (commented):
Overall looks good! Left some small comments and a couple "take it or leave it" comments.

Also noting I did not actually run this workflow through but based on S3 timestamps, looks like you did and it worked! Let me know if you want me to run through as well.

A couple additional comments:

1. 💯 for stub and to stubbifying the main workflow index building in a separate PR
2. Your stub model files all have hardcoded paths to where they live on your computer in them....make them empty, right?

bin/save_singler_refs.R
sjspielman: why is this so red!!!!!!!

bin/save_singler_refs.R
jashapiro: An alternative that should always work: `ls(getNamespace("celldex"))`

bin/save_singler_refs.R
allyhawkins: I think the original goal was to try and be as general as we can be with SingleR, even though we are only currently supporting `celldex` references. We could theoretically update the script to use references from the `scRNAseq` package. To do that we would just add a variable for package to this script. So I think we should leave it as is and just make sure to comment throughout that only `celldex` is supported right now.

bin/save_singler_refs.R
```diff
@@ -12,15 +12,25 @@ option_list <- list(
...
 opt <- parse_args(OptionParser(option_list = option_list))
 
+# get celldex package version
+celldex_version <- packageVersion("celldex") |>
+  as.character() |>
+  # replace . in version string to avoid . in filename
+  stringr::str_replace_all("\\.", "-")
+
+# construct output ref file name
+ref_output_file <- glue::glue("{opt$ref_file_prefix}_v{celldex_version}.rds")
```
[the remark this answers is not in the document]
allyhawkins: I am torn on including the v... I don't love it but I also don't hate it. Maybe to mirror CellAssign where we just have the date with no v then I'll remove it.

build-celltype-ref.nf
allyhawkins: I thought about that but I wanted to account for whatever future file name Panglao might use. The only thing I think we should for sure do is change the date.

allyhawkins:
> Your stub model files all have hardcoded paths to where they live on your computer in them....make them empty, right?

I think they got created like that from running the workflow. It looks like we don't keep the other output files stored in the repo, so I removed them. 

I removed the `v` before the celldex version, updated how we get the ref names for `celldex`, and then also took `ref_gtf` out of the tuple as was suggested. This should be ready for another round of review.

internal-instructions.md
```diff
@@ -130,8 +130,8 @@ To find all possible organs, see the `organs` column of `PanglaoDB_markers_27_Ma
 This file is required as input to the `build-celltype-ref.nf` workflow, which will create all required cell type references for performing cell type annotation from the main workflow.
```
sjspielman, line 130: Can't comment directly, but the line _right above_ this still has an outdated Panglao file name

sjspielman (approved):
LGTM!! Just one more spot to update the Panglao file name

PR #552, opened 2023-11-09
Adjust processes to not modify inputs

### LINKED ISSUE #480: Standardize formatting for nextflow files

### LINKED ISSUE #505: Export a new file in process for moving logcounts in AnnData
Just noting that we should update this to create a separate output file for nextflow reasons (it doesn't like us to alter inputs). However, we can save this for the next set of updates.
```suggestion
    tuple val(meta), path(processed_hdf5), val(file_type)
  script:
    """
    for file in ${processed_hdf5}; do
      move_counts_anndata.py \
        --anndata_file \${file}
    done
    """
  stub:
     """
     # nothing to do since files don't move
     """
```

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/501#discussion_r1355648111_

This is the process in `export_anndata` that moves the `logcounts` to the `X` slot if present. This should be updated to have a unique output file.

jashapiro:
Closes #480
Closes #505

Here I am changing inputs and outputs for a few different processes to make them different files and/or directories. 
In some cases this was accomplished by modifying the underlying scripts to write out new files rather than overwriting the inputs, which seems safer anyway.

In `fry-quant` processes, I modified the `alevin-fry` steps to copy over relevant `alevin` files into the new directory (keeping the input rad directory separate), rather than deleting `.rad` and `.bin` files.

The other big change is that in `modules/export-anndata.nf` I combined the moving of `logcounts` into the same process as the original anndata export, which saves any renaming and simplifies the workflow a bit (no `mix` needed, and no double-publishing of files). We had originally kept that separate with the idea that R and python could be different, but they are using the same docker image at this point, so it seemed easier to just combine them.

I have done initial checks to be sure the workflow still works as expected, but I am still running the demux version, just for completeness.

There are also a few little doc updates, mostly because I kept using the wrong option names when testing!

### DIFF
--- a/bin/cluster_sce.R
+++ b/bin/cluster_sce.R
@@ -20,6 +20,11 @@ option_list <- list(
       Must contain a PCA matrix to calculate clusters from."
   ),
   make_option(
+    opt_str = c("-o", "--output_sce_file"),
+    type = "character",
+    help = "Output path for clustered SCE file. Must end in .rds"
+  ),
+  make_option(
     opt_str = c("--pca_name"),
     type = "character",
     default = "PCA",
@@ -92,5 +97,4 @@ if (!opt$pca_name %in% reducedDimNames(sce)) {
 }
 
 # export -------------------
-# we are overwriting the `processed_sce_file` here
-readr::write_rds(sce, opt$processed_sce_file)
+readr::write_rds(sce, opt$output_sce_file)
--- a/bin/cluster_sce.R
+++ b/bin/cluster_sce.R
@@ -70,6 +70,10 @@ if (!file.exists(opt$processed_sce_file)) {
 }
 sce <- readr::read_rds(opt$processed_sce_file)
 
+if (!stringr::str_ends(opt$output_sce_file, ".rds")) {
+  stop("`output_sce_file` must end in .rds")
+}
+
 # only perform clustering if reduced dimension embeddings are present
 # otherwise just return the object
 if (!opt$pca_name %in% reducedDimNames(sce)) {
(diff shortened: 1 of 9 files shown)


allyhawkins (approved):
All the changes look good. As long as your test runs completed successfully then good to go 👍

bin/cluster_sce.R
```diff
@@ -19,6 +19,11 @@ option_list <- list(
     help = "Path to RDS file that contains the processed SCE object to cluster.
       Must contain a PCA matrix to calculate clusters from."
   ),
+  make_option(
+    opt_str = c("-o", "--output_sce_file"),
+    type = "character",
+    help = "Output path for clustered SCE file. Must end in .rds"
+  ),
```
sjspielman, line 26: should we add a check that this ends in `rds`?

sjspielman: similar to how you did it in `bin/post_process_sce.R`, probably

PR #553, opened 2023-11-09
Fix submitter unknown labels

### LINKED ISSUE #454: Standardize how we refer to unknown cell types
This was brought up by @allyhawkins in https://github.com/AlexsLemonade/scpca-nf/pull/451#pullrequestreview-1627139453 - 

> [A]nother thought I had is related to how we are labeling "unclassified cells" vs "unknown cells". We should standardize this across methods probably.

Indeed, we should. Here's what we _currently_ have:

### Submitter
- Cells that the submitter did not annotate at all are labeled as `NA`.
- Cells that the submitter attempted to annotate but could not define are labeled as `"Unclassified cell"`

### `SingleR`
- In the pruned labels, cells that could not be confidently annotated are labeled `NA`. 
- In the QC report, we show these in figures/tables as `"Unknown cell type"`.

### `CellAssign`
- Cells that do _not_ have the marker genes are labeled as `"other"`.
- In the QC report, we show these in figures/tables as `"Unknown cell type"`.

sjspielman:
Closes #454 

This PR updates the submitter `NA` labeling situation as we discussed in #454.

### Overview of changes

1. In `bin/add_submitter_annotations.R`, I ensure all columns read in are `character`, and use `"Submitted-excluded"` for any cells not present in the submitter TSV.
2. In `templates/qc_report/celltypes_qc.rmd`, I updated wording accordingly.
3. In `templates/qc_report/utils/celltype_functions.rmd` I split up the function that prepares cell type factors into two functions: a) one for automated methods (singler and cellassign), and b) one for submitter only (doesn't need to be a function for reuse, but sure helps for legibility). 
     -  The new function `prepare_submitter_annotation_values()` moves all unknown levels - from a variety of choices! - to the end, but with `"Submitted-excluded"` always as the `final.final` level. 

### Spots to discuss

- You'll see some `TODO`s in `bin/add_submitter_annotations.R` about something we forgot - the submitters might provide us with ontology annotations, and those need to go in also!
  - That should not be part of this PR, but we'll need an issue for it. Something to also decide whether this blocks a cell type release. My vote is not to block.
-  Any feedback on the choice in `templates/qc_report/utils/celltype_functions.rmd` to allow more than 1 unknown? The drawback is it feels a little silly, but the benefit is that it lets people really people to their heart's content.
  - Related, am I missing any unknown-y strings?
 
### Ugh

**edit** oh it's probably a meta json thing...

I have been trying to test this with a different submitter TSV file that has some unknowns added in, but I can't get nextflow to accept my parameter overriding. Something must be cached somewhere and I need to stop trying for now. But, I have tested all this code (script and rmd) interactively and they work as expected.  

Been try to run with:
```
op run -- nextflow run main.nf -profile ccdl,batch --run_ids SCPCL000495 -with-tower -resume --run_metafile ~/scpca-nf/scpca-library-metadata-TEST.tsv
```

Where the submitter TSV column in `~/scpca-nf/scpca-library-metadata-TEST.tsv` points to a local file with all epithelial replaced with `"na"` and all fibroblasts replaced with `"unknown"`.

So, better to just get this under review already and test after revisions..


templates/qc_report/utils/celltype_functions.rmd
```diff
@@ -30,61 +30,114 @@ create_celltype_df <- function(processed_sce) {
...
+  # used for unknowns - strings which might be "unknown"-y
+  unknown_strings <- c("na", "n/a", "<na>", "<n/a>", "null", "unknown", " ")
+  
+  # define variations here rather than lowering the submitter annotations themselves
+  unknown_strings <- c(
+    unknown_strings,
+    stringr::str_to_upper(unknown_strings), 
+    stringr::str_to_title(unknown_strings)
+  ) |> 
+    unique()
```
jashapiro, line 107: I'd go the other way; it is safer that assuming you have all the capitalization variants: use a mutate to lower case everything, and maybe strip all non-characters?

Something like this:

```
unknown_strings <- c("na", "null", "unknown", "")

# remove non-word characters and lower case
normalized_levels <- submitter_levels |>
  stringr::str_replace_all("\\W+", "") |>
  stringr::str_to_lower()

unknown_levels <- submitter_levels[normalized_levels %in% unknown_strings]
```

templates/qc_report/utils/celltype_functions.rmd
```diff
@@ -30,61 +30,114 @@ create_celltype_df <- function(processed_sce) {
...
+  new_submitter_levels <- c(
+    submitter_levels[!(submitter_levels %in% unknown_levels)], 
+    unknown_levels
+  )
+    
+  # finally, apply the new levels
+  df$submitter_celltype_annotation <- df$submitter_celltype_annotation |>
+    # just in case, remove any current levels
+    as.character() |>
+    factor(levels = new_submitter_levels)
```
jashapiro, line 138: I think you can do all of this with just:
```suggestion
# Move unknown levels to the end in order of frequency, in followed by submitter-excluded
df$submitter_celltype_annotation <- df$submitter_celltype_annotation |>
  forcasts::fct_relevel(unknown_levels, after = Inf)
```

jashapiro (commented):
Looks good, with a few suggestions:
* Explicitly keeping the NA values as strings in `read_tsv`
* Modifying the unknown finding to be more robust
* more 🐈

bin/add_submitter_annotations.R
```diff
@@ -34,41 +34,73 @@ if (!(stringr::str_ends(opt$sce_file, ".rds"))) {
...
   )
 
-# Check rows before sending back into the SCE object
-if (nrow(submitter_cell_types_df) != ncol(sce)) {
+# Perform some checks before sending back into the SCE object
+
+# There should not be any  _logical_ NAs here ("NA" etc is allowed )
+if (any(is.na(coldata_df$submitter_celltype_annotation))) { 
+  stop("There are NA values in the submitter annotation column. This column should be all character.")
+}
```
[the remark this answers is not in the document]
sjspielman: yes this is pure paranoia! I can remove it

templates/qc_report/utils/celltype_functions.rmd
sjspielman: I swear I tried this and it didn't work w/ >1 level to relocate... but I just tried it again and it worked fine, so who knows what I did in testing..! Thanks for re-surfacing

```
> iris$Species |> fct_relevel(c("setosa", "virginica"), after=Inf) -> species
> levels(species)
[1] "versicolor" "setosa"     "virginica"
```

sjspielman:
> more 🐈

@jashapiro is there a `purrr::` comment you maybe meant to leave, or shall I add some 🐈 into comments? 🙃

templates/qc_report/utils/celltype_functions.rmd
sjspielman: Nice!! 

```
> submitter_levels <- c("cell type", "Unknown", "<N/A>", "stephanie", "null")
>   unknown_strings <- c("na", "nan", "null", "unknown",  "") 
>   submitter_normalized_levels <- submitter_levels |>
+     stringr::str_replace_all("\\W+", "") |>
+     stringr::str_to_lower()
> 
> submitter_levels[submitter_normalized_levels %in% unknown_strings]
[1] "Unknown" "<N/A>"   "null"   
```

bin/add_submitter_annotations.R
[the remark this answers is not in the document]
sjspielman: 😬 https://github.com/AlexsLemonade/scpca-nf/pull/553/commits/5d0998d3eefef28f974ede49384d1c1d88d9ddd3

jashapiro (commented):
Now it is my turn for a bit of paranoia. Since the point of the join is to get rows into the same order as the coldata, I don't like that there might be any modifications of the data frame after that point. So I had a few suggestions for how that might be avoided below. If one of those works, we should be good to go.

sjspielman:
We're finally all debugged and running through with a test file of submitter cell types, which had all of `na`, `NA`, and `Unknown` labels added in for testing. Your paranoia in updating the check to `identical()` was well-placed - https://github.com/AlexsLemonade/scpca-nf/pull/553/commits/d970bc694f81850de81cb075dd73c8894c091743

Screenshot 2023-11-13 at 4 04 31 PM

jashapiro:
> We're finally all debugged and running through with a test file of submitter cell types, which had all of `na`, `NA`, and `Unknown` labels added in for testing. 

Just to clarify, in this case, the original file does not actually have all three types? The reason I ask is because I would probably be okay with normalizing `"na"` and `"NA"` (and `"n/a"` and `<N/A>`) but if this isn't something we are seeing in real data (yet), I would push that off for a while.

sjspielman:
> Just to clarify, in this case, the original file does not actually have all three types? 

The original file has all defined cell types - no `NA`s at all. So, I had to make some kind of test file anyways and figured might as well make it fancy... We aren't seeing anything that needs to be harmonized as you describe yet, so indeed a later problem.

sjspielman:
And we're sorted (is this a pun?):

Screenshot 2023-11-13 at 4 24 23 PM

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -264,7 +264,18 @@ available_celltypes <- c(
...
+
+# These variables need to be defined for edge conditions when certain cell types
+#  aren't available:
+#  When we have chunks with `eval=<has_method>`, knitr seems to need all other 
+#  variables used in chunk params defined, including plot dimension variables 
+#  which would have only been created in chunks which also had an associated 
+#  eval=<has_method>`. Since those chunks won't have been eval'd if FALSE, we 
+#  need to define these variables to prevent errors. BUT THEY ARE NOT USED WITH
+#  THESE VALUES.
+plot_height <- 1; heatmap_height <- 1
```
jashapiro, line 278: The styler will complain...
```suggestion
plot_height <- 1
heatmap_height <- 1
```

templates/qc_report/utils/celltype_functions.rmd
```diff
@@ -30,61 +30,107 @@ create_celltype_df <- function(processed_sce) {
...
-prepare_annotation_values <- function(
-    df,
-    annotation_column,
-    is_submitter = FALSE) {
-  submitter_string <- "Unclassified cell" # use for unclassified submitter cells
-  method_string <- "Unknown cell type" # use for unclassified singler and cellassign cells
-  fct_string <- ifelse(is_submitter, submitter_string, method_string) # which string to consider when reordering the factor
+prepare_automated_annotation_values <- function(
+  df,
+  annotation_column) {
```
jashapiro, line 64: Avoiding later restyling. Apparently [tidyverse style is double-indented functions](https://style.tidyverse.org/functions.html#long-lines-1).
```suggestion
    df,
    annotation_column) {
```

jashapiro (approved):
LGTM, with a couple very little style changes that I noticed only because I have been playing with stylers. But you don't actually have to fix these now... they will get fixed the next time somebody with `pre-commit` turned on modifies these files!

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: Good point, let me actually go ahead and get my pre-commit installed in the directory now that that PR went in!

sjspielman:
> But you don't actually have to fix these now... they will get fixed the next time somebody with pre-commit turned on modifies these files!

Did it anyways and who knows, maybe the styler will still find ways to complain about it later ;)

jashapiro:
> Did it anyways and who knows, maybe the styler will still find ways to complain about it later ;)

I'm sure it will... Those were just the ones I noticed scanning through quickly!

PR #554, opened 2023-11-10
Add pre-commit to the repo

jashapiro:
To reduce some development cycles, I thought we might add [pre-commit](https://pre-commit.com/) to this repo, mostly for spell checking, but also for code styling where we can.

The basic idea here is that _if you want to use it_, you can install `pre-commit` on your machine, then run `pre-commit install` in your github repo. After that point, when you make a commit, it will run through some quick tests and styling to make sure everything is up to snuff.

If the stylers or spell checker find any problems, your commit will be aborted, and you can review any changes made by the stylers, fix any spelling errors, etc. With this set up, you should not have to wait for github to tell you that there were spelling mistakes!

Note that the first time you run this it will take a bit of time to set up all of the environments it needs, but subsequent runs should be lickity-split.

The checks I added are:
- spell check
  - I originally thought this would use a public hook, but because they insisted on a particular location for the dictionary file, I decided to use our existing script.
  - To ease use with our existing spell check script, I made a few changes
    - First, if there are files passed to the script as arguments, it only checks those files (if they are Rmd files), and assumes we are in a pre-commit context (`pre-commit` passes the changed files to the command)
    - The pre-commit context also enables automatic updating of the dictionary. So now if there is a spelling error found, you do have to delete the incorrect word from the dictionary as part of your fix. But if it is a "good" word, you just have to commit the updated dictionary. I _think_ I like this workflow (it is what the package I would have used would have done), but if we want to leave the dictionary updates as manual, I can undo that part easily.

- stylers and linters
  - ruff for python; styles the same way as black but faster (and does linting)
  - styler for R and Rmd
  - prettier for yaml and markdown files
  - add trailing newlines to appease github

There are of course more things we could add, but this seemed like a good place to start. (I didn't add `lintr` because we currently fail many checks in that, and I didn't feel quite ready to set up all the exceptions we would want... `lintr` and the tidyverse are not the best of friends.)

Oh, and in doing this I caught a typo that had made it into our dictionary 🙀.


allyhawkins (approved):
This looks good to me and I'm good with the decisions that you made here. Are you planning on running it through with all files and committing all the style changes in a separate PR?

jashapiro:
> This looks good to me and I'm good with the decisions that you made here. Are you planning on running it through with all files and committing all the style changes in a separate PR?

The way it works is that it will run on every file as it is committed, so things will get run as modifications happen. No need to change everything at once.

allyhawkins:
> The way it works is that it will run on every file as it is committed, so things will get run as modifications happen. No need to change everything at once.

Ahh okay. I saw there was an option to check all the files, so wanted to see if you were planning to do that or not. Makes sense!

CONTRIBUTING.md
```diff
@@ -74,4 +73,18 @@ For R code, we try to follow [`tidyverse` style conventions](https://style.tidyv
...
+## Pre-commit hooks
+
+For convenience, we have included a set of [pre-commit hooks](https://pre-commit.com/) that can be used to automatically format code according to the above specifications, as well as to spellcheck and check for other common errors.
+
+To use these hooks, install the `pre-commit` package according to your favorite method (`pip install pre-commit` or `conda install pre-commit`), then run `pre-commit install` in the `scpca-nf` directory.
+This will install the hooks in the `.git/hooks` directory, and they will be run automatically when you commit changes.
+If any of the hooks fail, the commit will be aborted, and you will need to fix the errors and re-commit.
+
+Notably, the spellcheck hook will report spelling errors, but will also add any words it finds to the dictionary file.
+This is convenient for many cases (where the word is real but unknown), but be sure to remove truly misspelled words from the dictionary file before committing, or they will not be caught in the future!
```
sjspielman, line 90: ah yes indeed!

components/dictionary.txt
```diff
@@ -42,9 +42,9 @@ scpca
 ScPCA
 transcriptome
 transcriptomic
-trancriptomics
 TSV
 UMAP
 UMI
 uri
 Visium
+VSCode
```
sjspielman, line 50: Hm, there does seem to be a space in `VS Code` - https://en.wikipedia.org/wiki/Visual_Studio_Code
I might just get `VS` in the dictionary and update text accordingly?

scripts/spell-check.R
```diff
@@ -1,28 +1,55 @@
...
+
+# if there are arguments, check those files, otherwise check all markdown & rmd files
+if (length(arguments) > 0) {
+  precommit <- TRUE
+  files <- arguments[grepl("\\.(Rmd|md|rmd)$", arguments)]
+} else {
+  precommit <- FALSE
+  # The only files we want to check are R Markdown and Markdown files
+  files <- list.files(pattern = "\\.(Rmd|md|rmd)$", recursive = TRUE, full.names = TRUE)
+}
```
sjspielman, line 19: Maybe define a pattern string variable to use in both spots here?

scripts/spell-check.R
```diff
@@ -1,28 +1,55 @@
...
-dictionary <- readLines(file.path(root_dir, 'components', 'dictionary.txt'))
+dict_file <- file.path(root_dir, "components", "dictionary.txt")
+dictionary <- readLines(dict_file)
 
 # Add emoji to dictionary
-dictionary <- c(dictionary, spelling::spell_check_text("⚠️")$word)
-
-# The only files we want to check are R Markdown and Markdown files
-files <- list.files(pattern = '\\.(Rmd|md|rmd)$', recursive = TRUE, full.names = TRUE)
+dictionary_plus <- c(dictionary, spelling::spell_check_text("⚠️")$word)
```
sjspielman, line 29: yesss my Panglao `blood-plus` naming scheme made it in in its own way 🎉

scripts/spell-check.R
```diff
@@ -1,28 +1,55 @@
...
+    cat("The following spelling errors were found:\n")
+    print(data.frame(spelling_errors))
+  }
 
-# Save spell errors to file temporarily
-readr::write_tsv(spelling_errors, 'spell_check_errors.tsv')
+  # Update dictionary for future use
+  updated_dict <- union(dictionary, spelling_errors$word)
+  # remove empty strings
+  updated_dict <- updated_dict[updated_dict != ""]
```
sjspielman, line 46: Is there a world where we'd get fully empty strings of `"\\s+"`? If so, maybe replace with more flexible regex `\\s*`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+          - readr
+          - rprojroot
+          - spelling
+          - tidyr
+      - id: forbid-to-commit
+        name: Don't commit common R artifacts
+        entry: Cannot commit .Rhistory, .RData, .Rds or .rds.
+        language: fail
+        files: '\.(Rhistory|RData|Rds|rds)$'
+        # `exclude: <regex>` to allow committing specific files
```
sjspielman, line 19: can this be case insensitive for `rds`?

also what about `html`? we usually don't want it, but i'm not sure we will never want it. hmm.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - id: ruff
+        # Run the formatter.
+      - id: ruff-format
+  - repo: https://github.com/lorenzwalthert/precommit
+    # R styling and linting
+    rev: v0.3.2.9025
+    hooks:
+      - id: style-files
+        args: [--style_pkg=styler, --style_fun=tidyverse_style]
+      # - id: lintr #skip R linting for now...
```
sjspielman, line 34: noting you could open a future issue for this

sjspielman (approved):
> add trailing newlines to appease github

🤩 

LGTM! Left some comments but nothing I need to see again unless you want me to

Just one question/point of clarification - if the spell check adds words to the dictionary, presumably it won't automagically commit those, right? This situation would arise alongside a spell check fail, and we'd check the updated dictionary for accuracy and then commit it again ourselves. Yes?

scripts/spell-check.R
jashapiro: No, this is just a weird case where it includes an empty string in the world list due to a formatting choice the package makes, I think. We have never seen spaces in the word list.

jashapiro:
> Just one question/point of clarification - if the spell check adds words to the dictionary, presumably it won't automagically commit those, right?

It won't. Your commit will fail, and the dictionary will be modified, but the modifications won't be staged. And pre-commit is even smart enough to temporarily check out the old dict so it only checks against any changes you have actually staged.

CONTRIBUTING.md
```diff
@@ -74,4 +73,18 @@ For R code, we try to follow [`tidyverse` style conventions](https://style.tidyv
...
+If you use Visual Studio Code, you can install the [`black` extension](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter).
+
+As an alternative to `black`, you can also use [`ruff`](https://docs.astral.sh/ruff/) for formatting and linting; it follows the same code style conventions.
+`ruff` is also available as a [VSCode extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff).
+
+## Pre-commit hooks
+
+For convenience, we have included a set of [pre-commit hooks](https://pre-commit.com/) that can be used to automatically format code according to the above specifications, as well as to spellcheck and check for other common errors.
+
+To use these hooks, install the `pre-commit` package according to your favorite method (`pip install pre-commit` or `conda install pre-commit`), then run `pre-commit install` in the `scpca-nf` directory.
```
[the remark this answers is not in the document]
jashapiro: Leaving off, as this is an implementation detail that is not really relevant to the usage of the tool.

jashapiro (commented):
Noting that the styler made the emoji more opaque by changing to the code points, but I couldn't figure out how to avoid this... It seems to be baked into the way styler works.

PR #556, opened 2023-11-13
Skip cell typing if cell type reference versions match 

### LINKED ISSUE #538: Check that reference file is the same when skipping cell typing
Now that we will be tracking versions for both `SingleR` and `CellAssign` references, we will want to check that the reference file is the same when skipping mapping. To do this, we should check that `meta.singler_ref_file` is equivalent to the `singler_ref_file` stored in `scpca-meta.json`. Similar to how we check the `ref_assembly` for mapping. This should be done for both `SingleR` and `CellAssign`. 

We will also want to add human-readable values to help track the version used for cell typing in `checkpoints/celltype/scpca-meta.json`, such as `celltype_source` and `celltype_version`.

allyhawkins:
Closes #538 

The main purpose of this PR is to incorporate a check to make sure the reference file is the same when skipping either SingleR or CellAssign. I also made the following adjustments: 

- I renamed the `celltypes_publish_dir` to `celltypes_checkpoint_dir`. This made more sense to me since we are storing checkpoints, not final results. 
- I grabbed the SingleR and CellAssign reference versions from the filenames and pass them to the script that adds the cell type annotation to the SCE object. Here, I chose to store both the source and version together as either `singler_ref_version` or `cellassign_ref_version` in the metadata of the SCE object. Rather than storing the source and version separately, it made sense to me to have one string. How do people feel about this?


sjspielman (commented):
> Rather than storing the source and version separately, it made sense to me to have one string. How do people feel about this?

I think this does make sense. It seems like the way you have implemented this (am I following correctly??) is that the value saved in the metadata _is the reference file name_ without the extension. I like the ease of matching things up between reference file names and the metadata - no need to grab different things out of metadata for it.

The one thing I'm not really sure about is _where_ this should get handled. Nice work with string manipulation in nextflow, but I wonder if we actually want to do this in the R script? Since we already have `singler_model_file` and `cellassign_reference_file` in the meta, we could get away with just passing those into the R script and parse items out accordingly. 
Then, for `SingleR`, we wouldn't strictly need this line anymore, since we could parse out the reference name and version from the model file.
  ```
   metadata(sce)$singler_reference <- metadata(singler_results)$reference_name
  ```

Basically I feel like R debugging >>> nextflow debugging, so moving things like this into R and reducing the number of variables created in nextflow might be nice. But, I'd also want to hear what @jashapiro thinks about my idea, since true to form, I am vacillating!

--- 

Other than any ^ changes we might want to pursue, this LGTM!

allyhawkins:
> The one thing I'm not really sure about is where this should get handled. Nice work with string manipulation in nextflow, but I wonder if we actually want to do this in the R script? Since we already have singler_model_file and cellassign_reference_file in the meta, we could get away with just passing those into the R script and parse items out accordingly.

I think when I started working on this I was thinking I needed to check the version, not the filename, so started doing it in Nextflow. I also kind of liked having explicit variables for the file vs the reference. In case we ever were to change the file format, we would need to change the code in the script. But that's probably the case for Nextflow too. So I'm good with this change and will go ahead and do that. 

> With the easier manipulation in R, I think I would prefer storing the sources and versions as separate strings, just for the logical separation, but I don't feel too strongly about it.

I agree with this so I think I'll separate them.

allyhawkins:
Okay I've moved all of the file name manipulation into the R script, rather than storing the version and source information in the `meta` object. I also split up the source and version to store them separately in the metadata of the SCE object. 

The last thing I did was fix the supplemental cell type report to skip anything related to submitter cell types if no submitter cell types are present. It looks like there were a few chunks missing `has_submitter`.

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman, line 1: thanks for catching these!

bin/add_celltypes_to_sce.R
```diff
@@ -26,15 +26,22 @@ option_list <- list(
     type = "character",
     help = "path to rds file containing SingleR results object"
   ),
+  make_option(
+    opt_str = c("--singler_model_file"),
+    type = "character",
+    help = "Name of file containing a single model generated for SingleR annotation.
+            File name is expected to be in form: <ref_name>_<source>_<version>_model.rds."
```
sjspielman, line 33:
```suggestion
            File name is expected to be in form: `<ref_name>_<source>_<version>_model.rds`."
```

sjspielman: So github's aggressive red clearly has something to go with `>`. So aggressive!

bin/add_celltypes_to_sce.R
```diff
@@ -26,15 +26,22 @@ option_list <- list(
...
     type = "character",
     help = "path to tsv file containing the prediction matrix returned by running CellAssign"
   ),
   make_option(
-    opt_str = c("--cellassign_ref_name"),
+    opt_str = c("--cellassign_ref_file"),
     type = "character",
-    help = "name of reference used for CellAssign"
+    help = "Name of marker by cell type reference file used with CellAssign.
+            File name is expected to be in form: <ref_name>_<source>_<version>.tsv"
```
sjspielman, line 44:
```suggestion
            File name is expected to be in form: `<ref_name>_<source>_<version>.tsv`"
```

bin/add_celltypes_to_sce.R
```diff
@@ -60,6 +67,10 @@ if (!is.null(opt$singler_results)) {
     stop("Missing singleR results file")
   }
 
+  if (is.null(opt$singler_model_file)) {
+    stop("singleR model filename must be provided")
```
sjspielman, line 71:
```suggestion
    stop("SingleR model filename must be provided")
```

bin/add_celltypes_to_sce.R
```diff
@@ -60,6 +67,10 @@ if (!is.null(opt$singler_results)) {
     stop("Missing singleR results file")
```
sjspielman, line 67:
```suggestion
    stop("Missing SingleR results file")
```

bin/add_celltypes_to_sce.R
```diff
@@ -60,6 +67,10 @@ if (!is.null(opt$singler_results)) {
...
   }
 
+  if (is.null(opt$singler_model_file)) {
+    stop("singleR model filename must be provided")
+  }
+
```
sjspielman, line 73: Can you add here a check for `cellassign_ref_file` too?

bin/add_celltypes_to_sce.R
```diff
@@ -98,10 +109,21 @@ if (!is.null(opt$singler_results)) {
     DataFrame(row.names = colData(sce)$barcodes)
   colData(sce) <- new_coldata
 
+  # get reference name, source and version
+  singler_ref_info <- basename(opt$singler_model_file) |>
+    # select everything before the model extension
```
sjspielman, line 114: pipe-ier?

```suggestion
  singler_ref_info <- opt$singler_model_file |>
    basename() |>
    # select everything before the model extension
```

bin/add_celltypes_to_sce.R
```diff
@@ -98,10 +109,21 @@ if (!is.null(opt$singler_results)) {
...
+    stringr::word(1, sep = "_model.rds") |>
```
sjspielman, line 115:
```suggestion
    stringr::word(1, sep = "_model\\.rds") |>
```

bin/add_celltypes_to_sce.R
sjspielman: Ah, I see the check below! Let's move it up to the top and fail sooner.

bin/add_celltypes_to_sce.R
```diff
@@ -98,10 +109,21 @@ if (!is.null(opt$singler_results)) {
     DataFrame(row.names = colData(sce)$barcodes)
   colData(sce) <- new_coldata
 
+  # get reference name, source and version
```
sjspielman, line 112: Actually, maybe let's make this a function with an `extension` argument where you pass in arguments `_model.rds` or `.tsv`? Then you can have just one line here and below for cellassign to define `<method>_ref_info` list.

sjspielman (commented):
Almost there!! Aside from just small text comments (like `singleR` -> `SingleR`) and a smidge of reorganization (move all checks to the top), the main thought I have is that it might be good to define a little function to parse both singler and cellassign model names into the list of values to add to the SCE metadata. Then, less code in the body of the script.

bin/add_celltypes_to_sce.R
allyhawkins: No, because we only check for that if cell assign results are present. That's why it's in that chunk.

allyhawkins:
I added in a little function to create the ref info list from a filename. I did not move the checks though because I only want to check for each of the filenames if the results for that cell type annotation were provided. So I want them in their respective if statements.

bin/add_celltypes_to_sce.R
```diff
@@ -26,20 +26,42 @@ option_list <- list(
...
+    help = "Name of marker by cell type reference file used with CellAssign.
+            File name is expected to be in form: `<ref_name>_<source>_<version>.tsv`"
   )
 )
 
 opt <- parse_args(OptionParser(option_list = option_list))
 
+# Set up -----------------------------------------------------------------------
+
+get_ref_info <- function(ref_filename, extension) {
```
sjspielman, line 52: roxygen?

```suggestion
#' Get reference information from a reference file
#' 
#' @param ref_filename reference file name to parse
#' @extension file extension to consider during parsing
#' 
#' @return list of reference file components to include in SCE metadata
get_ref_info <- function(ref_filename, extension) {
```

sjspielman (approved):
LGTM, just needs a few more docs and I think my suggestion about wraps it up!

PR #502, opened 2023-10-11
Add celltyping to external instructions

sjspielman:
Closes #499 

This PR adds external cell type documentation. There is a decent amount here, so I expect >1 rounds of review! (reminder that I am out this Thursday+Friday, which may affect when reviewers want to schedule their time to review this!).

I have added documentation throughout for adding submitter annotations and for performing cell typing, including an example celltype metadata file. Any organizational or wording feedback let me know! There is also one `TODO` in there about whether we want to say "want a different organ for cellassign? let us know," so please make sure to comment there.

Worth noting that the internal docs are still under review (#500) and this PR is not stacked; we may want to add more relative links between these docs files, but I'd like to wait until #500 is all set first. Any more relative links can be added later in this PR or in a subsequent PR.

While working on these docs, I had some thoughts about how we have set up the cell type annotation metadata file - specifically, do _actually_ want this file to include the columns `singler_ref_file` and `cellassign_ref_file`? You'll see the docs are a bit weird (although I think fairly clear) for the content of these columns. We may wish to alter the cell type process to only accept singler & cell assign reference _names_, and we would create the meta variables for the reference paths themselves in the celltype annotation subworkflow. If we want to take that approach, I'd update docs here accordingly and then in a separate PR we can update how we handle the metadata in the workflow.

PR #565, opened 2023-11-14
Update get_refs.py to download celltype references

### LINKED ISSUE #560: Update get_refs.py to download celltype references
We want to continue to be able to accommodate users who do not have access to S3, so we need to update [get_refs.py](https://github.com/AlexsLemonade/scpca-nf/blob/main/get_refs.py) to download the cell type reference files from S3. I think it is probably fine to keep full paths in the project reference file, but we might want to write out the cell type reference paths somewhere as part of that process. Alternatively, we could make the cell type reference path a parameter and make the actual file paths relative.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/502#pullrequestreview-1728356783_

Further questions before implementation: Should we download only the "default" references that we are using, meaning or all of the celldex & pangalaodb references we are creating?

jashapiro:
Closes #560

This ended up being a bit of a refactor, as the get_refs.py script was getting more than a bit long in the tooth.  So I broke things up to make it a bit easier to chunk out the different tasks as separate functions.

Most of the functionality is unchanged, but the main addition is, as expected, to download the cell type references. 

This requires building up the file paths and filenames including the version numbers. We didn't have an easy way to guess those, so I added a little bit to the reference building workflow to write out version numbers for the reference files. I do this with a file named `<source>_version.txt` which is in the references folder for both singler and cellassign (not the singler models directory). I thought about making this a more structured file, but I think just a little string is enough for now?

The bulk of the added work is in the  `get_celltype_files()` function, so that is where reviewing effort should be focused. I may not have done things the most efficient way there, but it does work. 

I also updated the code that writes out the parameters file to write out all of the relevant local reference paths; before it had only written out the root directory path, but I had some fears that this would not always be properly substituted, so I wrote my own little substitution code, which seems to work. 

I also added a `--dry_run` option to the script for testing. That one does get remote files required for building other paths, but does not download any of the large files.

Note that to test this you will probably want to use `python get_refs.py -r development` to be sure that it looks at the correct repo to find download paths. 

I am sure there will be questions with this large refactor. Please let me know where I should add clarifying comments and/or modify logic for easier flow!


get_refs.py
```diff
@@ -18,259 +19,417 @@
...
+        help="get STAR index (required for genetic demultiplexing)",
+    )
+    parser.add_argument(
+        "--cellranger_index",
+        action="store_true",
+        help="get Cell Ranger index (required for spatial data)",
+    )
+    parser.add_argument(
+        "--docker", action="store_true", help="pull and cache images for docker"
+    )
```
allyhawkins, line 62:
```suggestion
    parser.add_argument(
        "--docker", 
        action="store_true", 
        help="pull and cache images for docker"
    )
```

allyhawkins: Although I'm surprised pre-commit didn't get this?

get_refs.py
```diff
@@ -18,259 +19,417 @@
...
+    ref_paths += get_barcode_files(ref_params)
+
+    # paths to celltype reference files
+    # read in celltype reference metadata
+    celltype_metadata = csv.DictReader(
+        (line.decode() for line in celltype_metadata_file), delimiter="\t"
+    )
+    ref_paths += get_celltype_files(ref_params, celltype_metadata)
+
+    download_files(
```
allyhawkins, line 139: This may be silly, but I would just add a small comment here about downloading all the paths now.

get_refs.py
```diff
@@ -18,259 +19,417 @@
...
+        # if S3, convert bucket path to https:// url
+        bucket_path = root_parts[1].split("/", maxsplit=1)
+        url_root = f"https://{bucket_path[0]}.s3.amazonaws.com"
+        if len(bucket_path) > 1:
+            url_root += f"/{bucket_path[1]}"
+    elif root_parts[0] in ["http", "https", "ftp"]:
+        # otherwise, just get the location
+        url_root = root_uri
+    else:
+        raise ValueError("`ref_rootdir` is not a supported remote location.")
```
allyhawkins, line 176: I think we want this error message to include the `str` since technically it doesn't have to be `ref_rootdir`?

get_refs.py
```diff
@@ -18,259 +19,417 @@
...
+        "genomeParameters.txt",
+        "Log.out",
+        "SA",
+        "SAindex",
+        "sjdbInfo.txt",
+        "sjdbList.fromGTF.out.tab",
+        "sjdbList.out.tab",
+        "transcriptInfo.tab",
+    ]
+    star_dir = Path(param_dict.get("star_index"))
```
allyhawkins, line 245: What happens if this fails and `star_index` is missing from the file? Does this just continue with no error? Would it be worth adding a check that the key actually exists, to make sure that all paths are able to be acquired?

get_refs.py
```diff
@@ -18,259 +19,417 @@
...
+
+
+def get_celltype_files(
+    param_dict: dict[str, str], celltype_metadata: dict[str, str]
+) -> list[Path]:
+    # get celltype reference paths
+    root_url = get_root_url(param_dict.get("ref_rootdir"))
+
+    root_re = re.compile(r"^\$\{?(params.)?ref_rootdir\}?")
+    celltype_dir_re = re.compile(r"^\$\{?(params.)?celltype_ref_dir\}?")
```
allyhawkins, line 300: Can you add a comment here? It looks like you are removing the ${params.}?

get_refs.py
```diff
@@ -18,259 +19,417 @@
...
+
+    # replace celltype ref placeholders
+    singler_ref_dir = celltype_dir_re.sub(celltype_ref_url, singler_ref_dir)
+    singler_model_dir = celltype_dir_re.sub(celltype_ref_url, singler_model_dir)
+    cellassign_ref_dir = celltype_dir_re.sub(celltype_ref_url, cellassign_ref_dir)
+
+    # alternatively, these might be tied to the root url
+    singler_ref_dir = root_re.sub(root_url, singler_ref_dir)
+    singler_model_dir = root_re.sub(root_url, singler_model_dir)
+    cellassign_ref_dir = root_re.sub(root_url, cellassign_ref_dir)
```
allyhawkins, line 319: Why are you grabbing these twice? Is in case we ever switch `{params.celltype_ref_dir}` with `{params.ref_rootdir}` in these paths?

get_refs.py
```diff
@@ -18,259 +19,417 @@
...
+
+    # get versions of celltype reference files
+    celldex_version_url = f"{singler_ref_dir}/celldex_version.txt"
```
allyhawkins, line 322: I had this idea in my review of #502, but I do think writing out a json (or a table for easier viewing for users) would make this code easier and you could avoid the below code of getting the version and building all the file names.

get_refs.py
jashapiro: It all fits on one line, so this was actually the formatter's choice!

get_refs.py
jashapiro:
```suggestion
    # download all the files stored in `ref_paths`
    download_files(
```

get_refs.py
jashapiro: Agree!
```suggestion
        raise ValueError(f"`{root_uri}` is not a supported remote location.")
```

get_refs.py
allyhawkins: Ah that makes sense! I was wondering why it was like that.

get_refs.py
jashapiro: If it is missing, we have a problem! I can add a check, but the data is coming from this repository (and I don't provide an option to change it), so we should be pretty safe.

get_refs.py
jashapiro: Not quite... I'm matching any of `$params.celltype_ref_dir` `$celltype_ref_dir` `${celltype_dir}`, etc. But worth a comment as it is an ugly regex!

get_refs.py
jashapiro: Yeah, that is the main reason. The idea is that we can substitute either kind of base path.

get_refs.py
allyhawkins: Yea I don't think it's super important, I'm just wondering if we were to change the ref config file and then a key didn't exist, how would running this go? I would like to think that it would fail if a key were missing, so we probably don't need the check. But if it doesn't fail and just continues downloading without you knowing that something is missing, that seems bad.

jashapiro:
@allyhawkins In https://github.com/AlexsLemonade/scpca-nf/pull/565/commits/dbdf82b7e2f40e2eb7648139c8ec95ae42b66dfb I added a script and processes for making a cell type table. It isn't done or tested yet, but I thought you might want to have a look to see if it is going in a direction you are happy with.

get_refs.py
jashapiro: I just checked... it will actually fail here as `get` will return `None` and `Path(None)` is an error. We could provide a nicer error, but since this _should never happen_ I am kind of okay with an ugly error when it does.

allyhawkins: That works for me!

allyhawkins:
> @allyhawkins In [dbdf82b](https://github.com/AlexsLemonade/scpca-nf/commit/dbdf82b7e2f40e2eb7648139c8ec95ae42b66dfb) I added a script and processes for making a cell type table. It isn't done or tested yet, but I thought you might want to have a look to see if it is going in a direction you are happy with.

I took a quick look of how it was setup and the approach looks good to me.

jashapiro:
Okay, this is ready for another look. I added a script to make a file table from a list of filenames, and added processes to use it (separate for cellassign and singler). Note that I made the script use just the `tidyverse` packages, so we could use the rocker image without installing anything (the reason is that the image is much smaller, so it seemed worth doing for a little script like this. I suppose I could have written it all in base R for even lighter dependencies, but that seemed a bit too much).

I then updated `get_refs.py` to use this new file, which means no more building file names!

bin/make_celltype_ref_table.R
```diff
@@ -0,0 +1,49 @@
...
+
+# Usage:
+# Rscript make_celltype_ref_table.R <list,of,file,paths> <output_table.tsv>
+
+
+
+# not using optparse as to avoid dependencies
+args <- commandArgs(trailingOnly = TRUE)
+if (length(args) != 2) {
+  stop("Incorrect number of arguments")
```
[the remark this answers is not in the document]
sjspielman: Ehh, actually, nah. If we ever run this script "by hand" we'd want this, but since we're mostly (only?) running it as part of the workflow this probably doesn't help anyone. Plus it's also kind of annoying to run outside the context of the workflow with the `list,of,files` so I don't see running this "by hand" anyways. 

So, nevermind probably!

allyhawkins (approved):
LGTM 👍 I also tested with `--dry_run` and everything seemed to be there. 

The only other thing that is outside the scope of this PR is about how we plan on surfacing the files with the table of reference information to external users. I think part of my idea for creating the file to begin with is to have an easy list of references to view that is not just in S3. I'm fine with outputting it as part of the workflow to S3, but does anyone have thoughts on how we might want to sync the contents of the file in S3 with the contents of a file we keep in the repo? Maybe the answer is we just do it manually as part of running the workflow to build cell type refs.

sjspielman:
> I'm fine with outputting it as part of the workflow to S3, but does anyone have thoughts on how we might want to sync the contents of the file in S3 with the contents of a file we keep in the repo? Maybe the answer is we just do it manually as part of running the workflow to build cell type refs.

I really can't think of anything besides manual, unfortunately. So maybe we handle this by creating an issue template with two steps - 
1. rebuild references
2. copy the file to repo but sent _straight to_ `main` I would think, since S3 is public.

jashapiro:
> I'm fine with outputting it as part of the workflow to S3, but does anyone have thoughts on how we might want to sync the contents of the file in S3 with the contents of a file we keep in the repo? Maybe the answer is we just do it manually as part of running the workflow to build cell type refs.

I think we could just provide the links, which will always be up to date: 
- https://scpca-references.s3.amazonaws.com/celltype/cellassign_references/cellassign_references.tsv
- https://scpca-references.s3.amazonaws.com/celltype/singler_models/singler_models.tsv

bin/make_celltype_ref_table.R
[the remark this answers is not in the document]
jashapiro:
```suggestion
  stop("Exactly two arguments are required: <list,of,files> <output.tsv>")
```

PR #569, opened 2023-11-15
Standardize how we store cell type reference information

allyhawkins:
Mentioned in https://github.com/AlexsLemonade/scpca-docs/pull/193, we had both `ref` and `reference` in the names for items in the metadata of the final SCE object. This just changes everything to `reference`. Let me know if I missed any spots!


sjspielman (approved):
:rocket: 

:saddest face in the world:

:where are my emojis:

sjspielman:
What we have learned from the above comment is that github will indeed RENDER my emojis, but it will not preview it while I write it? But it does preview it in the `Preview` tab? 
:sob:

Issue #570, opened 2023-11-16
Add missing example files


sjspielman:
As submitter cell types are not necessary, we do not currently provide metadata examples in `examples/` for how to use them. In the future, we may want to add example files to show clearly how to provide your own cell type annotations as part of the workflow.

---

**Edit** - 

Noting this context from #571:
https://github.com/AlexsLemonade/scpca-nf/pull/571#issuecomment-1814755230

> >But, do we want to then create example_metadata_files/submitter_celltypes.tsv? I had had something for that in the external instructions PR that just went in - see this commit: [38dba37](https://github.com/AlexsLemonade/scpca-nf/commit/38dba37ba7d0321d4c046dc0d01bb7e7b3b4f9cb) . I can bring something like that file back.

> Seems worth having.

While this issue was originally opened to create `example_metadata_files/submitter_celltypes.tsv`, it seems that in [`example_run_metadata.tsv`](https://github.com/AlexsLemonade/scpca-nf/blob/619e638d10eef6de342639bf28f232d56ac3f5de/examples/example_run_metadata.tsv), we have several analogous files that don't exist. 

- `example_barcode_files/cite_barcodes.tsv`
- `example_barcode_files/cellhash_barcodes.tsv`
- `example_metadata_files/submitter_celltypes.tsv`

We should probably make them all if we're going to make one. This issue should track them all.

sjspielman:
After some discussion in #571, we're headed there shortly!

sjspielman:
Noting this context:
https://github.com/AlexsLemonade/scpca-nf/pull/571#issuecomment-1814755230

> >But, do we want to then create example_metadata_files/submitter_celltypes.tsv? I had had something for that in the external instructions PR that just went in - see this commit: [38dba37](https://github.com/AlexsLemonade/scpca-nf/commit/38dba37ba7d0321d4c046dc0d01bb7e7b3b4f9cb) . I can bring something like that file back.

> Seems worth having.


In [`example_run_metadata.tsv`](https://github.com/AlexsLemonade/scpca-nf/blob/619e638d10eef6de342639bf28f232d56ac3f5de/examples/example_run_metadata.tsv), we have some other analogous files that don't exist. 

We should probably make them all if we're going to make one - 

- `example_barcode_files/cite_barcodes.tsv`
- `example_barcode_files/cellhash_barcodes.tsv`
- `xample_metadata_files/submitter_celltypes.tsv`

jashapiro:
[...]

Issue #469, opened 2023-09-22
Update example output with cell type annotations


allyhawkins:
Once we have finalized how we are going to add in cell type annotations into the workflow we will want to update the example output to reflect those changes. This should be done before we make a new release with the cell type annotation in the workflow.

sjspielman:
While we're here, also update some of the example language to not make it seem like submitter is mandatory - 

https://github.com/AlexsLemonade/scpca-nf/blob/df98da392c7e9e0f4d67c008dcc4208cee9db4d4/examples/README.md?plain=1#L19-L25

allyhawkins:
I believe we could just remove this all together since it's not required to run the workflow?

sjspielman:
@allyhawkins agree, this should suffice for us - https://github.com/AlexsLemonade/ScPCA-admin/issues/634

sjspielman:
Noting that :point_up: comments about removing submitter things got handled as part of #502

sjspielman:
Looks like we actually still have a `submitter_cell_types_file` column in [`examples/example_run_metadata.tsv`](https://github.com/AlexsLemonade/scpca-nf/blob/cf78909b57f0ff4f22bc9b926e43754e2ad94738/examples/example_run_metadata.tsv). It also looks like we have one, filled with `NA` on the `example_run_metadata.tsv` in S3. 

We'll want to remove these columns altogether - The first one we'll need to do a PR for here, and the second update directly in S3 as part of generating the new example output.

sjspielman:
> the second update directly in S3 as part of generating the new example output.

✅

sjspielman:
Womp. Signing off for tonight, but this is next on deck...!

```
[-        ] process > bulk_quant_rna:fastp                           -
[-        ] process > bulk_quant_rna:salmon                          -
[-        ] process > bulk_quant_rna:merge_bulk_quants               -
[-        ] process > map_quant_rna:alevin_rad                       -
[a9/fbf0b6] process > map_quant_rna:fry_quant_rna (run01-rna)        [100%] 1 of 1 ✔
[17/361a8f] process > generate_sce:make_unfiltered_sce (library01)   [100%] 1 of 1 ✔
[7d/423c9b] process > generate_sce:filter_sce (library01)            [100%] 1 of 1 ✔
[-        ] process > map_quant_feature:index_feature                -
[-        ] process > map_quant_feature:alevin_feature               -
[...]

PR #571, opened 2023-11-16
Remove submitter column

sjspielman:
As noted in this comment https://github.com/AlexsLemonade/scpca-nf/issues/469#issuecomment-1814583495, we still had some one submitter thing hanging out in `examples/`. This PR removes it.

A `grep` now shows we have removed all submitter language - 

Screenshot 2023-11-16 at 9 46 55 AM

Note that I've also opened this issue if we want to have examples for submitter cell types at some future point -  #570.


sjspielman:
> If we need to add a little documentation separately, we should do that, but I don't think we should be removing it.

That's definitely fine with me! I'm not sure we actually need any more docs since we have this already:

https://github.com/AlexsLemonade/scpca-nf/blob/74b744b8ba978d1745df17a7834d04e451d09fae/external-instructions.md?plain=1#L418-L430


But, do we want to then _create_ `example_metadata_files/submitter_celltypes.tsv`? I had had something for that in the external instructions PR that just went in - see this commit: https://github.com/AlexsLemonade/scpca-nf/commit/38dba37ba7d0321d4c046dc0d01bb7e7b3b4f9cb . I can bring something like that file back.

jashapiro:
> But, do we want to then _create_ `example_metadata_files/submitter_celltypes.tsv`? I had had something for that in the external instructions PR that just went in - see this commit: [38dba37](https://github.com/AlexsLemonade/scpca-nf/commit/38dba37ba7d0321d4c046dc0d01bb7e7b3b4f9cb) . I can bring something like that file back.

Seems worth having.

sjspielman:
> Seems worth having.

Ok, I'm going to close this PR and start fresh!

PR #574, opened 2023-11-16
Add multiplex warnings to QC report

### LINKED ISSUE #519: Update cell type report to account for multiplexed samples
Once we have made the necessary changes to the cell type report for regular single-cell/single-nuclei samples, we will want to account for displaying cell type annotations in multiplexed samples. 

- We should include a warning that these are multiplexed samples. We have one at the top of the main report, but we might add a note in the cell type section to remind users. We should also include the warning in the supplemental report. 
- We may want to either remove the cluster comparison or re-structure it to account for each sample in the multiplexed library. 

I'm sure there are other things that I am missing, but we should use this issue to track any changes we might want to make when accounting for multiplexed cell type annotations. Then we can break out into additional issues if necessary.

sjspielman:
Closes #519 

As discussed*, I've added heatmap alerts - `warning` for the top of the celltype report, and `info` for the other spots. 

*Slight wording tweak but very happy to backtrack to the original suggestion in Slack for the UMAP text! I went ahead and added "confounded" in there, which is a concept I like to emphasize, but it might be getting too much in the weeds. Please suggest rephrasing wherever you see fit!

I also got changed all the `heatmap_height`s to also be `plot_height`, for simplicity, while I was here. No need for 2 unused variable definitions when we could just have the 1, I figured.

Here are screenshots of the relevant rendered parts!

-----------------------

Screenshot 2023-11-16 at 12 25 26 PM

-----------------------

Screenshot 2023-11-16 at 12 25 33 PM

------------------------

Screenshot 2023-11-16 at 12 25 39 PM

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -191,7 +191,7 @@ umap_df <- lump_celltypes(celltype_df)
 ```{r, eval = has_multiplex, results='asis'}
 glue::glue("
   <div class=\"alert alert-secondary\">
-    This library contains multiple samples that have not been integrated, which may confound clustering assignments.
+    This library contains multiple samples without batch-correction, which may confound clustering assignments.
     Please use caution when interpreting these results.
   </div>
 ")
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -191,7 +191,7 @@ umap_df <- lump_celltypes(celltype_df)
 ```{r, eval = has_multiplex, results='asis'}
 glue::glue("
   <div class=\"alert alert-info\">
-    This library contains multiple samples that have not undergone batch correction, which may confound clustering assignments.
+    This library contains multiple samples that have not undergone batch correction, so clustering assignments may be confounded.
     Please use caution when interpreting these results.
   </div>
 ")
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -191,7 +191,7 @@ umap_df <- lump_celltypes(celltype_df)
 ```{r, eval = has_multiplex, results='asis'}
 glue::glue("
   <div class=\"alert alert-secondary\">
-    This library contains multiple samples without batch-correction, which may confound clustering assignments.
+    This library contains multiple samples without batch correction, which may confound clustering assignments.
     Please use caution when interpreting these results.
   </div>
 ")
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -191,7 +191,7 @@ umap_df <- lump_celltypes(celltype_df)
 ```{r, eval = has_multiplex, results='asis'}
 glue::glue("
   <div class=\"alert alert-info\">
-    This library contains multiple samples that have not undergone batch correction, so clustering assignments may be confounded.
+    This library contains multiple samples that have not been batch-corrected, which may confound clustering assignments.
     Please use caution when interpreting these results.
   </div>
 ")
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -190,7 +190,7 @@ umap_df <- lump_celltypes(celltype_df)
 
 ```{r, eval = has_multiplex, results='asis'}
 glue::glue("
-  <div class=\"alert alert-info\">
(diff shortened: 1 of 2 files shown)


jashapiro (commented):
Couple text edits and a little logic thing I noticed. And maybe a grey alert? Not sure.

sjspielman:
I wasn't able to get `alert-secondary` (or `alert-light`, similar!) working. They just render as plain text in the browser (chrome and firefox checked). Looking at the rendered HTML directly - 

```
<script>/*!
 * Bootstrap v3.3.5 (http://getbootstrap.com)
 * Copyright 2011-2015 Twitter, Inc.
 * Licensed under the MIT license
 */
```

Indeed, at `3.3.5` there are only these alerts - https://bootstrapdocs.com/v3.3.5/docs/components/#alerts-examples. The expanded ones are from bootstrap v4.

I tracked down this issue and it looks like we only get bootstrap v4 alerts with `bslib` (https://github.com/rstudio/rmarkdown/issues/1260#issuecomment-883269556) which actually is in our `scpcaTools` environment. So it looks like we can do something like this to set v4 bootstrap and get nice things -
https://rstudio.github.io/bslib/articles/any-project/index.html#r-markdown

I went ahead and implemented this super barebones so we can see how it looks - with the bump to v4, the over all look does shift but we do get a gray box with `alert-secondary` (`info` ended up being too light). I'll note also it changed the html file size from 3.6 MB -> 3.7 MB, which seems ok. 

Here's a report with this implemented. What do we we think of the overall look of the report? Do we want to move to this more generally? If we don't like this look and the _only reason_ for including `bslib` is for an alert box, probably this is excessive.

[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13381252/celltypes_supplemental_report.html.zip)

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -24,6 +26,7 @@ knitr::opts_chunk$set(
 
 library(SingleCellExperiment)
 library(ggplot2)
+library(bslib)
```
jashapiro, line 29: I don't think this is needed with `bslib: true`
```suggestion
```

jashapiro (approved):
Looks like you went right to v5, which is fine, though you said v4. 

If we change one theme, we have to change them all. And I feel like I can't make the decision alone... too much pressure.

I'm fine with whatever we decide, and the rest looks good.

sjspielman:
> Looks like you went right to v5, which is fine, though you said v4.

😂 this was in the docs example and my brain glossed over that this wasn't the number 4 entirely. Or, I am just really precocious.

sjspielman:
> And I feel like I can't make the decision alone... too much pressure.

@allyhawkins, help us! ;)

allyhawkins:
> > And I feel like I can't make the decision alone... too much pressure.
> 
> @allyhawkins, help us! ;)

Is the question to use the grey or blue boxes? Because I am going to vote blue. Go blue 〽️

jashapiro:
I should have known. Blue it is, and we'll keep bootstrap5 in our back pocket.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -187,6 +187,17 @@ umap_df <- lump_celltypes(celltype_df)
 
 
 <!-- First UMAP: clusters -->
+
+```{r, eval = has_multiplex, results='asis'}
+glue::glue("
+  <div class=\"alert alert-secondary\">
+    This library contains multiple samples that have not been integrated, which may confound clustering assignments.
```
allyhawkins, line 194:
```suggestion
    This library contains multiple samples without batch-correction, which may confound clustering assignments.
```
Just to bring up the idea that people can get confused by the word integration sometimes. I think I would be more specific that there's no batch correction.

allyhawkins (commented):
This looks good to me other than a small wording comment. As far as the blue/grey boxes, I do obviously like blue. But if you guys fell really strongly about grey then I would just make sure all the other boxes in the main report are also grey. Whatever we pick we should be consistent.

sjspielman:
Noting that I ended up landing at 
```
This library contains multiple samples that have not undergone batch correction, which may confound clustering assignments.
```

sjspielman:
In the end, I implemented this suggestion but with this phrasing (starred for emphasis below only) -  https://github.com/AlexsLemonade/scpca-nf/pull/574#discussion_r1396282656

```
 This library contains multiple samples **that have not been batch-corrected**, which may confound clustering assignments.
```

PR #578, opened 2023-11-16
Use words to describe files

jashapiro:
This makes the changes I was suggesting in my slightly late review: https://github.com/AlexsLemonade/scpca-nf/pull/577#pullrequestreview-1735637935

While doing this I also caught some places the headings/links no longer matched.

PR #580, opened 2023-11-17
Add example config

sjspielman:
Part of #469 

Here, I'm adding an config file we can use to process the example data. I made a few more updates while I was here - 

- Update instructions in `internal-instructions.md` for how to use the config
  - This has one relative link that won't work until #578 goes in, fyi
  - I tried to simplify (bulletify) some of the language here too, feedback welcome
- Add a `README` to the config directory. Please weigh in on language there, thanks!


config/profile_example.config
jashapiro, line 1: For this to be active, you will need to update the profiles list in `nextflow.config` as well.

config/README.md
```diff
@@ -0,0 +1,13 @@
+This folder contains additional configuration files for running the `scpca-nf` workflow.
+Some of these files are called from `nextflow.config`, and others can be optionally specified at run time.
```
jashapiro, line 2: _All_ of the configs have to be referenced in `nextflow.config`, otherwise nextflow would have no way of finding them.

config/profile_example.config
```diff
@@ -0,0 +1,23 @@
...
+
+  // the example run_ids
+  run_ids = "run01"
+
+  // celltyping on
+  perform_celltyping = true
+
+  // Private CCDL containers
+  CELLRANGER_CONTAINER = '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:6.1.2'
+  SPACERANGER_CONTAINER = '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-spaceranger:1.3.1'
```
jashapiro, line 22: We don't use these in the example `run01`, I don't think, and they probably don't need to be added here. In theory, anyone _could_ use this profile for a test run with their own infrastructure (if nodes have internet access).

config/README.md
sjspielman: Ahhh I see, thanks!! I think I never looked at that part of the `nextflow.config` file and figured it knew about these because of some automatic naming scheme! I'll delete this sentence and replace with something like they must be included for the main config to know about them.

config/profile_example.config
sjspielman: left over from copy/paste from `ccdl`...will remove!

config/README.md
```diff
@@ -0,0 +1,12 @@
+This folder contains additional configuration files for running the `scpca-nf` workflow.
+To use a configuration file, it must be included in the main configuration file (e.g., [`.nextflow.config`](../nextflow.config)).
+
+| Config file | Description | Included from `nextflow.config`? |
```
jashapiro, line 4: Delete this column, since the only thing not included is the markdown file?  Actually, this probably doesn't need to be a table at all
```suggestion
| Config file | Description |
```

nextflow.config
```diff
@@ -90,4 +90,8 @@ profiles{
   stub {
     includeConfig 'config/profile_stub.config'
   }
+  // re-process example data
```
jashapiro, line 93:
```suggestion
  // settings to process example data
```

sjspielman:
Any _specific_ wording feedback this time?

config/profile_example.config
jashapiro, line 22:
```suggestion
```

config/README.md
```diff
@@ -0,0 +1,12 @@
+This folder contains additional configuration files for running the `scpca-nf` workflow.
+To use a configuration file, it must be included in the main configuration file (e.g., [`nextflow.config`](../nextflow.config)).
```
jashapiro, line 2: A bit more detail, if we are going to say anything.
```suggestion
To use a configuration file, it must be referenced in the main configuration file (i.e., [`nextflow.config`](../nextflow.config)) using `includeConfig`.
```

jashapiro (commented):
Looks good overall, but lets convert the file list back to just a bulleted list. Table seems like overkill here.

config/profile_example.config
sjspielman: Huh I thought I did this.. something must have gotten lost in merging.

config/README.md
```diff
@@ -0,0 +1,11 @@
+This folder contains additional configuration files for running the `scpca-nf` workflow.
+To use a configuration file, it must be referenced in the main configuration file (i.e., [`nextflow.config`](../nextflow.config)) using `includeConfig`.
+
+
+- `containers.config` specifies Docker containers to use when running the `scpca-nf` workflow
+- `process_base.config` specifies resource settings for the `scpca-nf` workflow
```
jashapiro, line 6:
```suggestion
- `process_base.config`: default resource settings for `scpca-nf` workflow processes
```

config/README.md
```diff
@@ -0,0 +1,11 @@
...
+- `profile_awsbatch.config` specifies AWS resources for the `scpca-nf` workflow
```
jashapiro, line 7:
```suggestion
- `profile_awsbatch.config`: legacy AWS Batch configuration for Data Lab users
```

config/README.md
```diff
@@ -0,0 +1,11 @@
...
+- `profile_awsbatch_auto.config` specifies AWS resources for the `scpca-nf` workflow
```
jashapiro, line 8:
```suggestion
- `profile_awsbatch_auto.config`: AWS Batch configuration with auto-scaling disks for Data Lab users
```

config/README.md
```diff
@@ -0,0 +1,11 @@
...
+- `profile_ccdl.config` is used internally by the Data Lab for the `scpca-nf` workflow, as described in [`internal-instructions.md`](../internal-instructions.md#running-scpca-nf-as-a-data-lab-staff-member)
```
jashapiro, line 9:
```suggestion
- `profile_ccdl.config`: file locations used internally by the Data Lab, as described in [`internal-instructions.md`](../internal-instructions.md#running-scpca-nf-as-a-data-lab-staff-member)
```

config/README.md
```diff
@@ -0,0 +1,11 @@
+This folder contains additional configuration files for running the `scpca-nf` workflow.
+To use a configuration file, it must be referenced in the main configuration file (i.e., [`nextflow.config`](../nextflow.config)) using `includeConfig`.
+
+
+- `containers.config` specifies Docker containers to use when running the `scpca-nf` workflow
```
jashapiro, line 5:
```suggestion
- `containers.config` specifies Docker containers used by `scpca-nf` workflow processes
```

jashapiro (commented):
I made some updates to the wording of specific file descriptions. As I was doing it, I realized that I was writing them with colons and sentence fragments rather than sentences. So we should be uniform, but I'll leave it to you which direction to go.

jashapiro (approved):
👍🏼

PR #581, opened 2023-11-17
Compress the RDS files

allyhawkins:
Here I'm making sure that we compress all of our output files. I'm going to make the same change to the cluster script on `main`.


jashapiro (approved):
🗜️

PR #582, opened 2023-11-17
Compress cluster output and update tag

allyhawkins:
This PR includes the update to the clustering script to be sure that the processed RDS files are compressed. 
I also used this opportunity to update the version tag so I can release these changes. 

The last thing that I was forced to do was cherry pick the addition of pre-commit. Without the config file it wouldn't let me commit anything...


jashapiro (approved):
🗜️ 

(in the future, you can temporarily uninstall pre-commit with `pre-commit uninstall` to get it to not complain)

PR #583, opened 2023-11-17
Accommodate running multiple projects

allyhawkins:
This has been bothering me for a while, but right now we don't support running multiple project ids at once. I think for scenarios like having to re-run everything quickly, being able to at least group multiple projects together would be really helpful. 

Here, I made a few small adjustments to account for multiple projects being run at the same time. If you don't like this change, I can remove it.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -73,6 +73,8 @@ workflow {
   if (params.project){
     // projects will use all runs in the project & supersede run_ids
     run_ids = []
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
I was worried that we might not have the code in already to make sure the bulk merging to a table was by project, but we do. So this seems fine to me!

Issue #541, opened 2023-11-06
Improve approach for sizing heatmaps


sjspielman:
We have heatmaps in the supplemental cell type report, but the way we determine sizing still needs to be tweaked. We'll need to process more projects first to help us identify the right approach for general use.

sjspielman:
Noting font size may also come into play: https://github.com/AlexsLemonade/scpca-nf/commit/07bf56a4fe7dcfdda5fef4b2d9ed7b7b2b0a82b2#r131828986

allyhawkins:
I think we can close this since we looked at a bunch of reports and things looked good? @sjspielman

sjspielman:
> I think we can close this since we looked at a bunch of reports and things looked good?

Yup, sounds good. We can always reopen if we encounter edge cases down the line!

PR #584, opened 2023-11-17
Update plot dimensions

sjspielman:
This PR addresses some QC report things identified in #469 we'd like to fix.

- Use `theme(aspect.ratio = 1)` instead of `coord_fixed()` for UMAPs
- Bigger width for diagnostic cell type plots
- Ensure heatmaps are at least 4" tall (related to #541)
- Bonus, fix a bug introduced in https://github.com/AlexsLemonade/scpca-nf/pull/574/files
  - We moved some text into an `as_is` chunk with `glue::glue()`, so need to use `{}` for variables, not \``r stuff`\`

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -223,13 +223,15 @@ find_label_size <- function(input_labels) {
 #' @param col_names Vector of column names
 #' @param n_spacers Number of spacers between stacked heatmaps
 #' @param spacer_size Spacer size in inches
+#' @param min_height Minimize height required. Default 4"
 #'
 #' @return Heatmap height in inches
 calculate_plot_height <- function(
     row_names,
     col_names,
     n_spacers,
-    spacer_size = heatmap_padding) {
+    spacer_size = heatmap_padding, 
+    min_height = 4) {
   # first, based on number of cells in row_names:
   #   6 cell types to an inch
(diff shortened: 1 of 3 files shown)


jashapiro:
This all looks good to me. I don't know if you had any plans for further modifications?

Oh, one minor one: if we add `+ xlim(0,1)` to the CellAssign plot, I think it should get the densities to extend to the full range of the plot, would look nice.

sjspielman:
> This all looks good to me. I don't know if you had any plans for further modifications?

No further plans now, just later as part of #541. I'll go ahead and mark this ready for review, and add `xlim()`

sjspielman:
Full range!
Screenshot 2023-11-17 at 2 17 00 PM

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -740,7 +744,8 @@ ggplot(celltype_df) +
     x = "Probability of annotated cell type",
     y = ""
   ) +
-  coord_cartesian(xlim = c(0, 1)) +
+  #coord_cartesian(xlim = c(0, 1)) +
```
jashapiro, line 747: No need to keep the old one around, I don't think
```suggestion
```

jashapiro (approved):
LGTM!

sjspielman:
After this goes in, I'll regenerate the reports, and move to zipping up example data.

PR #588, opened 2023-11-20
Update example file docs

### LINKED ISSUE #570: Add missing example files

sjspielman:
Closes #570 

Two changes:

- In `external_instructions.md`, I added some bullets to actually describe what goes into the TSV files. But, do we want to link out to third party format, and if so where specifically to link out to? Is all the wording ok as is, too?
- In `examples/README.md`, I tell readers to go look at those file descriptions. Is this info in an ok place?

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -14,6 +14,13 @@ This directory contains the following example files:
 
 These files provide examples of expected formatting and content, but note that the specific values in these files may not be applicable or sufficient for running `scpca-nf` directly on your system.
 
+Note that `example_run_metadata.tsv` additionally refers to several example files, whose descriptions can be found as follows:
+
+- The format for the `example_barcode_files/cite_barcodes.tsv` file is described in the [ADT processing section of `external-instructions.md`](../external-instructions.md#libraries-with-additional-feature-data-adt-or-cellhash).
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
This looks fine to me!

PR #596, opened 2023-11-27
Account for mismatches between old cell typing results and current processed object and handle missing UMAPs

### LINKED ISSUE #591: Figure out what to do when the numbers of cells don't match up between CellAssign results and processed SCE object
I was testing the workflow and ran into an issue with rendering the supplemental cell type report. The problem is in creating the CellAssign density plots. When running a library we already had processed, there were now `NA`'s for the CellAssign prediction score causing an error in: 
https://github.com/AlexsLemonade/scpca-nf/blob/c6930febc1fdf461418fa474e8555bd9f469952d/templates/qc_report/celltypes_supplemental_report.rmd#L710-L715

After some investigation, it looks like the NA's are coming from barcodes that are not present in the CellAssign results, but are present in the processed object. The problem that I think we're going to have moving forward is sometimes the exact cells present in the final object may change slightly when filtering. I double checked that we set seeds in all the scripts where we need it (including when running CellAssign), but we have seen scenarios where the exact number of cells is altered slightly from run to run. 

So we need to decide what to do about repeating CellAssign in that scenario. I think we have a few options: 

1. Remove the ability to skip CellAssign. This option makes me really sad, but it would ensure that any barcodes in the object are in fact getting annotated by CellAssign. 😢 
2. Continue to skip CellAssign, but remove any cells that are not classified before creating plots and print out a warning. The problem with this approach is that it seems misleading to me. It's not that CellAssign couldn't assign a score, it's that we didn't provide that cell to CellAssign to begin with, so we don't really know what it is. 

I don't love option 1, but I think it's the only thing I can think of to be sure that the barcodes are the same. The only other thing would be actually checking that the barcodes in the files are the same before skipping it. 

The other thing to note is that this is based on the assumption that CellAssign returns a score for every single cell that is provided as input. This is my understanding and looking back at some of the analysis I did in `sc-data-integration` this seemed to be the case. So it seems unlikely that CellAssign just didn't assign a score and also didn't return it in the predictions matrix. 

Tagging @jashapiro @sjspielman for any thoughts on this.

allyhawkins:
Closes #591 

Based on discussion in #591, we decided to handle any potential mismatches in barcodes between the processed object and the existing `CellAssign` or `SingleR` results by labeling any cells not present in the results as `Unclassified cells`. This PR makes that adjustment and fixes some other smaller errors I discovered during test runs. 

- I accounted for potential fails in `density()` following the suggestion in https://github.com/AlexsLemonade/scpca-nf/issues/591#issuecomment-1823531553. 
- All of the missing cells have been annotated as `Unclassified cells` in the processed SCE object. This avoids any cells being labeled as `NA` accidentally. This happens as part of `add_celltypes_to_sce.R`. 
- In the QC report, I check for any missing cells and output a warning that some cells may be missing from the cell type results. I also removed any `Unclassified cells` before continuing with plotting. 
- I also had some issues where UMAP was not being calculated, causing a failure when rendering the cell type section of the report. The UMAP issues have been resolved by updating `scpcaTools` (https://github.com/AlexsLemonade/scpcaTools/pull/242). However, I realized we probably want to account for potentially having no UMAP results and still having cell type results. We already have handling for missing UMAPs in the main report, but we don't have that for the cell type report, so I added that here. This involved updating the function for creating `celltype_df` to account for potentially missing UMAP results and then using `has_umap` throughout the report to ensure no UMAPs are printed if UMAP is missing. 

Here's a copy of a rendered main and supplemental report with these changes: 
[SCPCL000495_qc.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13479606/SCPCL000495_qc.html.zip)

[SCPCL000495_celltype-report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13479608/SCPCL000495_celltype-report.html.zip)

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -145,17 +145,29 @@ celltype_df <- create_celltype_df(processed_sce)
 ```
 
 
-```{r, eval = has_cellassign || has_singler, results='asis'}
+```{r, eval = (has_cellassign || has_singler), results='asis'}
 unclassified_methods <- c()
 
 # check for unclassified SingleR cells
-if (any(celltype_df$singler_celltype_annotation == "Unclassified cell")) {
+if (has_singler && any(celltype_df$singler_celltype_annotation == "Unclassified cell")) {
   unclassified_methods <- c(unclassified_methods, "singler")
+
+  # remove all unclassified cells
+  celltype_df <- celltype_df |>
+    dplyr::filter(
+      singler_celltype_annotation != "Unclassified cell"
+    )
 }
 
 # check for unclassified cellassign cells
-if (any(celltype_df$cellassign_celltype_annotation == "Unclassified cell")) {
+if (has_cellassign && any(celltype_df$cellassign_celltype_annotation == "Unclassified cell")) {
   unclassified_methods <- c(unclassified_methods, "cellassign")
+
+  # remove unclassified cells
+  celltype_df <- celltype_df |>
+    dplyr::filter(
+      cellassign_celltype_annotation != "Unclassified cell"
+    )
 }
 
 # print a warning if either singleR or cellAssign have any unclassified cells
@@ -171,13 +183,6 @@ if (length(unclassified_methods) > 0) {
   </div>
 ")
 }
-
-# perform filtering of cell type df
-celltype_df <- celltype_df |>
-  dplyr::filter(
-    singler_celltype_annotation != "Unclassified cell",
-    cellassign_celltype_annotation != "Unclassified cell"
-  )
 ```
 
 
@@ -211,13 +216,17 @@ create_celltype_n_table(celltype_df, cellassign_celltype_annotation) |>
   format_celltype_n_table()
 ```
 
-## UMAPs
+
+```{r, eval = has_umap}
+knitr::asis_output(glue::glue("## UMAPs\n
 
 In this section, we show UMAPs colored by clusters.
-Clusters were calculated using the graph-based `r metadata(processed_sce)$cluster_algorithm` algorithm with `r metadata(processed_sce)$cluster_weighting` weighting.
+Clusters were calculated using the graph-based {metadata(processed_sce)$cluster_algorithm} algorithm with  {metadata(processed_sce)$cluster_weighting} weighting.
+"))
+```
 
 
-```{r}
+```{r, eval = has_umap}
 # Create dataset for plotting UMAPs with lumped cell types
 umap_df <- lump_celltypes(celltype_df)
 ```
@@ -225,7 +234,7 @@ umap_df <- lump_celltypes(celltype_df)
 
 <!-- First UMAP: clusters -->
 
-```{r, eval = has_multiplex, results='asis'}
+```{r, eval = has_umap, eval = has_multiplex, results='asis'}
 glue::glue("
   <div class=\"alert alert-info\">
     This library contains multiple samples that have not been batch-corrected, which may confound clustering assignments.
@@ -235,7 +244,7 @@ glue::glue("
 ```
 
 
-```{r message=FALSE, warning=FALSE}
+```{r eval = has_umap, message=FALSE, warning=FALSE}
 clusters_plot <- plot_umap(
   umap_df,
   cluster,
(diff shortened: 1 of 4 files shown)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -146,6 +145,47 @@ celltype_df <- create_celltype_df(processed_sce)
...
+  # remove unclassified cells
+  celltype_df <- celltype_df |>
+    dplyr::filter(
+      cellassign_celltype_annotation != "Unclassified cell"
+    )
+}
+
+# print a warning if either singleR or cellAssign have any unclassified cells
+if (length(unclassified_methods) > 0) {
+  unclassified_methods <- paste(unclassified_methods, collapse = " and ")
```
jashapiro, line 175: Future-proofing for if there are more than 2 methods (and showing off a new fancy function I learned about that does the right thing with oxford commas... the comma will be left out when there are only two items. It's kind of magic)
```suggestion
  unclassified_methods <- stringr::str_flatten_comma(unclassified_methods, last = ", and ")
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -176,21 +216,25 @@ create_celltype_n_table(celltype_df, cellassign_celltype_annotation) |>
   format_celltype_n_table()
 ```
 
-## UMAPs
+
+```{r, eval = has_umap}
+knitr::asis_output(glue::glue("## UMAPs\n
 
 In this section, we show UMAPs colored by clusters.
```
jashapiro, line 223: I tend to think it is nicer to have all the strings on a separate line for this. (And I am pretty sure the extra `\n` isn't needed. 
```suggestion
knitr::asis_output(glue::glue("
## UMAPs

In this section, we show UMAPs colored by clusters.
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -176,21 +216,25 @@ create_celltype_n_table(celltype_df, cellassign_celltype_annotation) |>
...
+```{r, eval = has_umap}
 # Create dataset for plotting UMAPs with lumped cell types
 umap_df <- lump_celltypes(celltype_df)
 ```
 
 
 <!-- First UMAP: clusters -->
 
-```{r, eval = has_multiplex, results='asis'}
+```{r, eval = has_umap, eval = has_multiplex, results='asis'}
```
jashapiro, line 237: I think just one `eval` statement?
```suggestion
```{r, eval = has_umap && has_multiplex, results='asis'}
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -220,15 +264,20 @@ if (length(levels(umap_df$cluster)) <= 8) {
 ```
 
 
-Next, we show UMAPs colored by cell types.
-A separate UMAP is shown for each cell typing method used.
+```{r, eval = has_umap}
+knitr::asis_output(
+  'Next, we show UMAPs colored by cell types.
+A separate UMAP is shown for each cell typing method used.\n
```
jashapiro, line 270: I don't think this should be needed as long as there is the blank space that you have there?
```suggestion
A separate UMAP is shown for each cell typing method used.
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -220,15 +264,20 @@ if (length(levels(umap_df$cluster)) <= 8) {
...
 For legibility, only the seven most common cell types are shown.
 All other cell types are grouped together and labeled "All remaining cell types" (not to be confused with "Unknown cell type" which represents cells that could not be classified).
+  '
+)
+```
 
 <!-- Now, UMAPs of cell types, where present -->
 
-```{r eval = has_submitter, message=FALSE, warning=FALSE}
+```{r eval = has_submitter & has_umap, message=FALSE, warning=FALSE}
```
jashapiro, line 280: Since we want this atomic, double `&` (some things may complain)
```suggestion
```{r eval = has_submitter && has_umap, message=FALSE, warning=FALSE}
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -240,7 +289,7 @@ plot_umap(
 ```
 
 
-```{r eval=has_singler, message=FALSE, warning=FALSE}
+```{r eval=has_singler & has_umap, message=FALSE, warning=FALSE}
```
jashapiro, line 292:
```suggestion
```{r eval=has_singler && has_umap, message=FALSE, warning=FALSE}
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -252,7 +301,7 @@ plot_umap(
 ```
 
 
-```{r, eval = has_cellassign, message=FALSE, warning=FALSE}
+```{r, eval = has_cellassign & has_umap, message=FALSE, warning=FALSE}
```
jashapiro, line 304:
```suggestion
```{r, eval = has_cellassign && has_umap, message=FALSE, warning=FALSE}
```

jashapiro (approved):
I think this all looks good. I had a number of small suggestions, but I don't think anything that should require another look.

templates/qc_report/celltypes_qc.rmd
allyhawkins: 🤯

PR #597, opened 2023-11-28
Move out integration processes

### LINKED ISSUE #592: Move integration processes to a separate module
We are planning to repurpose the existing workflow for integration to just be for merging objects. This means we no longer need the processes that perform integration via Harmony and fastMNN. Instead of delete those entirely, we plan to move these into a separate Nextflow module that will not currently be used. That way, if we ever do want to integrate, we will be able to use those processes.

sjspielman:
Closes #592 

A little 🍬 to get the ball rolling! This PR does what it says.


integrate.nf
jashapiro, line 1: Do you want to rename this while you are at it, since it isn't integration at all now?

integrate.nf
```diff
@@ -180,4 +129,3 @@ workflow {
     // generate integration report
     integration_report(integrate_harmony.out, file(integration_template))
```
jashapiro, line 130: Probably want to rename this report, and we need to get rid of the integration process calls above here?

jashapiro (commented):
I think there a few more things we need to do here before merging, so we still have a working workflow... Mostly we need to either remove or rename some things so that we are not calling processes that the script does not know about. (The alternative would be to import the module, but I don't think that is the direction we want to go at this point).

sjspielman:
> I think there a few more things we need to do here before merging

👏

sjspielman:
In actual response to your feedback here, I was wary about going too far here and making `integrate` -> `merge` changes, but I can certainly get that ball rolling as part of this PR too! But since you're on board with tackling now, I'll do some renaming and as you say, make sure the workflow still works/flows.

integrate.nf
sjspielman: Maybe just comment this out for now, since we also don't have the `rmd` yet?

merge.nf
sjspielman, line 1: Apparently github felt this update was sufficiently different to call it a whole new file instead of a diff!

sjspielman:
Let me know what you think of this now! I made some changes to get `merge` language in there all around, but left the code mostly as-is to not break the workflow altogether. Do you suggest that I take it farther and/or back off a little bit?

integrate.nf
sjspielman: > and we need to get rid of the integration process calls above here?

Ah, I totally missed this 🙃. I blame my long 🦃 break. These lines are gone now!

PR #600, opened 2023-11-28
Fix merge conflicts and update scpcaTools version

allyhawkins:
It looks like there were a few things that were in `main` that aren't in `development` that caused some conflicts. I can't directly push those changes to `development` because branch protection, so I'm fixing the conflicts here and also updating the container for `scpcaTools`.

### DIFF
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -139,15 +139,26 @@ if (!is.null(opt$feature_name)) {
   # extract altExp
   alt_sce <- altExp(sce, opt$feature_name)
 
-  # add sample metadata from main sce to alt sce metadata
-  metadata(alt_sce)$sample_metadata <- sample_metadata
+  # only convert altExp with > 1 rows
+  if (nrow(alt_sce) > 1) {
+    # add sample metadata from main sce to alt sce metadata
+    metadata(alt_sce)$sample_metadata <- sample_metadata
 
-  # make sce czi compliant
-  alt_sce <- format_czi(alt_sce)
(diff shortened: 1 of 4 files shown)


main.nf
```diff
@@ -73,6 +73,8 @@ workflow {
   if (params.project){
     // projects will use all runs in the project & supersede run_ids
     run_ids = []
+    // allow for processing of multiple projects at once
+    project_ids = params.project?.tokenize(',')
```
jashapiro, line 77: I think this should have come with changes at line 134 to use this list of project_ids?

jashapiro (commented):
This looks good and minor, but I wonder if there is one change missing? Maybe it wasn't a conflict, but it seems like the project filtering changes should be incorporated fully here?

bin/sce_to_anndata.R
```diff
@@ -139,15 +139,26 @@ if (!is.null(opt$feature_name)) {
...
+      alt_sce,
+      anndata_file = opt$output_feature_h5
+    )
+  } else {
+    # warn that the altExp cannot be converted
+    message(
+      glue::glue(
+        "Only 1 row found in altExp named: {opt$feature_name}.
+                       This altExp will not be converted to an AnnData object."
+      )
```
jashapiro, line 161: Formatting here seemed odd.
```suggestion
      glue::glue("
        Only 1 row found in altExp named: {opt$feature_name}.
        This altExp will not be converted to an AnnData object.
      ")
```

jashapiro (approved):
LGTM, with just one (optional) formatting update.

PR #601, opened 2023-11-28
Account for missing project ids

allyhawkins:
Went to do a test run with just a single sample and had an error that we didn't define `project_ids`. I fixed that here. I'll make the same change to `development`.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -74,7 +74,7 @@ workflow {
     // projects will use all runs in the project & supersede run_ids
     run_ids = []
     // allow for processing of multiple projects at once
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
👍🏼

main.nf
```diff
@@ -77,6 +77,7 @@ workflow {
     project_ids = params.project?.tokenize(',')
```
jashapiro, line 77: For safety also
```suggestion
    project_ids = params.project?.tokenize(',') ?: []
```

PR #602, opened 2023-11-28
Project ID fix to `development`

allyhawkins:
Same change as in #601


main.nf
```diff
@@ -77,6 +77,7 @@ workflow {
     project_ids = params.project?.tokenize(',')
```
jashapiro, line 77: For safety also
```suggestion
    project_ids = params.project?.tokenize(',') ?: []
```

jashapiro (approved):
Little additional change

PR #603, opened 2023-12-04
Include example output in release issue template

### LINKED ISSUE #585: Add example data processing to release issue template
...[F]or safety, we should make an issue template for re-generating example data. I think the right place for that is the release checklist issue template. We could add a section "do you need to update the example data? here are the steps for that, and if not check 'em off for free"

_Originally posted by @sjspielman in https://github.com/AlexsLemonade/scpca-nf/issues/469#issuecomment-1816627846_
            
----

We should consider adding a section to the [release checklist](https://github.com/AlexsLemonade/scpca-nf/blob/development/.github/ISSUE_TEMPLATE/release-checklist.md) issue template for re-processing example data. We'll need at least three bullets:

- Re-process the data (link to internal instructions)
- Check that the data looks correct
- Zip up into `scpca-out.zip` on S3

sjspielman:
Closes #585 

Here, I've added a little checklist for the example output into the release template.

Note that instead of what I've done here, we could also have a different issue template for this, if we think it's too much of a burden on whoever is assigned to the release issue?

I also made some other text changes for spell check appeasement. FYI this change (added backticks) can be reverted, but we'd have to add `vX` to the dictionary. Any preference?
```
title: Prepare for scpca-nf release `vX.X.X`
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-checklist.md
+++ b/.github/ISSUE_TEMPLATE/release-checklist.md
@@ -1,7 +1,7 @@
 ---
 name: Release checklist
 about: Prepare for a new release version of scpca-nf
-title: Prepare for scpca-nf release vX.X.X
+title: Prepare for scpca-nf release `vX.X.X`
 labels: release
 assignees: ''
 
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/release-checklist.md
```diff
@@ -13,15 +13,20 @@ assignees: ''
...
-  - [ ] [external-instructions.md](https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md)
+  - [ ] [`nextflow.config`](https://github.com/AlexsLemonade/scpca-nf/blob/main/nextflow.config)
+  - [ ] [`internal-instructions.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/internal-instructions.md)
+  - [ ] [`external-instructions.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md)
 - [ ] Test that the workflow is in good working order with `nextflow run alexslemonade/scpca-nf -latest -r development`
 - [ ] File a PR from the `development` branch to the `main` branch. This should include all of the changes that will be associated with the next release.
+- [ ] (Optional) Generate new example `scpca-nf` output files.
+If updating the example output is not necessary for this release, check these boxes off for free.
+  - [ ] [Re-process the example data](https://github.com/AlexsLemonade/scpca-nf/blob/main/internal-instructions.md#processing-example-data) through the `scpca-nf` workflow and ensure it looks correct.
+  - [ ] Compress the example output in `scpca_out` to create `scpca_out.zip`, as described in the instructions.
```
allyhawkins, line 24: We probably want to also note that the file should be set to public read-only.

allyhawkins (commented):
Mostly looks good, just one small comment about making sure the file is set to public.

PR #605, opened 2023-12-05
Repurpose integration workflow to be specifically for merging

### LINKED ISSUE #593: Repurpose current integration workflow for merging objects only
We would like to edit the existing integration workflow, [`integrate.nf`](https://github.com/AlexsLemonade/scpca-nf/blob/main/integrate.nf) to only be used for merging objects and creating a summary report for those merged objects. Once we remove the integration processes (see #592), we should do the following: 

- [x] Rename the workflow to `merge.nf`
- [x] Remove the parameters for `integration_metafile` and `integration_group`. 
- [x] Group the library ids by `scpca_project_id` for merging. 
- [x] Temporarily remove rendering the integration report, since we will later turn that into a merged report

allyhawkins:
Closes #593 

This PR completes the steps needed to repurpose the integration workflow to be for merging. 

- I removed the parameters for the`merge_metafile` and the `merge_group`. I also removed the part of the workflow that reads in the `merge_metafile` and joins it with the `run_metafile`. 
- Instead, we want to merge all libraries by project. I required the `project` parameter to be provided and filter the `run_metafile` to only contain libraries in the specified project. 
- Before merging, I group by the project ID so that each merged object contains all libraries for a given project. 

These changes were pretty straightforward, but I haven't had a chance to test them yet. I'm using all the resources to run CellAssign, so I'll request a review once I have the chance to test this.

### DIFF
--- a/merge.nf
+++ b/merge.nf
@@ -109,7 +109,7 @@ workflow {
       // make sure we don't have any duplicates of the same library ID hanging around
       // this shouldn't be the case since we removed CITE-seq and cell-hashing
       .unique()
-      // grouped tuple of [merge_group, [library_id1, library_id2, ...], [sce_file1, sce_file2, ...]]
+      // group tuple by project id, [project_id, [library_id1, library_id2, ...], [sce_file1, sce_file2, ...]]
       .groupTuple(by: 0)
 
     merge_sce(grouped_libraries_ch)
--- a/merge.nf
+++ b/merge.nf
@@ -98,10 +98,12 @@ workflow {
       .filter{it.seq_unit in ['cell', 'nucleus']}
       // create tuple of [project id, library_id, processed_sce_file]
       .map{[
-        project_id: it.scpca_project_id,
-        library_id: it.scpca_library_id,
-        scpca_nf_file: "${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
+        it.scpca_project_id,
+        it.scpca_library_id,
+        "${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
       ]}
+      // only include samples that have been processed through scpca-nf
+      .filter{file(it[2]).exists()}
       // make sure we don't have any duplicates of the same library ID hanging around
       // this shouldn't be the case since we removed CITE-seq and cell-hashing
       .unique()
--- a/merge.nf
+++ b/merge.nf
@@ -4,18 +4,18 @@ nextflow.enable.dsl=2
 // Workflow to merge SCE objects into a single object.
 // This workflow does NOT perform integration, i.e. batch correction.
 
-
-// merge-specific parameters
-// TODO: Update approach to define merge groupings.
-params.merge_metafile = 's3://ccdl-scpca-data/sample_info/scpca-integration-metadata.tsv'
-params.merge_group = "All"
-
 // define path to merge template
 // TODO: Establish this merge-report.Rmd file
 //merge_template = "${projectDir}/templates/merge-report.Rmd"
 
 // parameter checks
 param_error = false
+
+// check that at least one project has been provided
+if(!params.project) {
+  log.error("At least one 'project' must be specified for merging.")
+  param_error = true
+}
 
 if (!file(params.run_metafile).exists()) {
   log.error("The 'run_metafile' file '${params.run_metafile}' can not be found.")
@@ -60,7 +60,7 @@ process merge_sce {
 
 }
 
-// create merge report 
+// create merge report
 process merge_report {
   container params.SCPCATOOLS_CONTAINER
   publishDir "${params.results_dir}/merged/${merge_group}"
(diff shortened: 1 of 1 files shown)


allyhawkins:
I had to make a small syntax update for this to work. Also, the way I had it set up implied that every library listed in the library metadata for the specified project would have to exist. I don't know if we want to have that requirement or not, since when working with files in `scpca/processed` not every file for a project will exist. 

On the other hand, when we actually run this for production, we will want to make sure that every file exists in a given project. For now, I added in a filtering step to remove any libraries that don't have a processed file. If we keep this in, we probably want a way to track if any libraries were removed so one thought I had was to print out a list of libraries that aren't included in the object because of missing processed files. I'm curious what others think about this.

sjspielman:
> I'm using all the resources to run CellAssign

😭

sjspielman:
> On the other hand, when we actually run this for production, we will want to make sure that every file exists in a given project. For now, I added in a filtering step to remove any libraries that don't have a processed file. If we keep this in, we probably want a way to track if any libraries were removed so one thought I had was to print out a list of libraries that aren't included in the object because of missing processed files. I'm curious what others think about this.

This is a good question. For _prod_, I think I would like to see a separate check to ensure that all processed files exist before heading into merging in the first place, and then just quit if we don't have them all. But, I agree it makes sense to leave for now. I would say, let's open an issue to remove this filter and replace it with a check that processed SCEs are present for all libraries in the given integration group (& for the time being and foreseeable future, "integration group" = "project"!)?

That said, I'm not sure we need to print out missing libraries during this development stage? Seems like extra code and extra work...? Am I missing something?

merge.nf
```diff
@@ -90,45 +86,31 @@ process merge_report {
...
-        library_id: it.scpca_library_id,
-        scpca_nf_file: "${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
+        it.scpca_project_id,
+        it.scpca_library_id,
+        "${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
       ]}
+      // only include samples that have been processed through scpca-nf
+      .filter{file(it[2]).exists()}
+      // make sure we don't have any duplicates of the same library ID hanging around
+      // this shouldn't be the case since we removed CITE-seq and cell-hashing
```
sjspielman, line 108: noting my previous comment - didn't we want these libraries?

merge.nf
```diff
@@ -90,45 +86,31 @@ process merge_report {
...
-      // pull out library_id from meta and use to join
-      .combine(libraries_ch.map{[it.library_id, it.scpca_nf_file]}, by: 0)
-      // create tuple of merge group, library ID, and output file from scpca_nf
-      .map{[
-        it[1], // merge_group
-        it[0], // library_id
-        file(it[2]) // scpca_nf_file
-        ]}
-      // grouped tuple of [merge_group, [library_id1, library_id2, ...], [sce_file1, sce_file2, ...]]
+      // group tuple by project id, [project_id, [library_id1, library_id2, ...], [sce_file1, sce_file2, ...]]
```
sjspielman, line 110:
```suggestion
      // group tuple by project id: [project_id, [library_id1, library_id2, ...], [sce_file1, sce_file2, ...]]
```

merge.nf
```diff
@@ -90,45 +86,31 @@ process merge_report {
...
       .filter{it.seq_unit in ['cell', 'nucleus']}
+      // create tuple of [project id, library_id, processed_sce_file]
       .map{[
-        library_id: it.scpca_library_id,
-        scpca_nf_file: "${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
+        it.scpca_project_id,
+        it.scpca_library_id,
+        "${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds"
       ]}
+      // only include samples that have been processed through scpca-nf
```
sjspielman, line 105:
```suggestion
      // only include libraries that have been processed through scpca-nf
```

sjspielman (commented):
Seems fine to me, just a couple "comment comments" and a question about altexps.

allyhawkins:
> This is a good question. For prod, I think I would like to see a separate check to ensure that all processed files exist before heading into merging in the first place, and then just quit if we don't have them all. But, I agree it makes sense to leave for now. I would say, let's open an issue to remove this filter and replace it with a check that processed SCEs are present for all libraries in the given integration group (& for the time being and foreseeable future, "integration group" = "project"!)?

Yes, I agree with this! So before we release this, we will want to quit if we don't have them all. 

> That said, I'm not sure we need to print out missing libraries during this development stage? Seems like extra code and extra work...? Am I missing something?

I don't think we need it, I just wanted to see if others thought we would need it. I think if we are planning to make sure that we have all files in the production version, then we don't need to add this in just for development purposes. 

I added in your other suggestions, so this should be ready for another look.

merge.nf
```diff
@@ -90,45 +86,31 @@ process merge_report {
...
-
-    // channel with run metadata, keeping only the columns we need
-    libraries_ch = Channel.fromPath(params.run_metafile)
+    // read in run metafile, filter to projects of interest, and group by project
+    grouped_libraries_ch = Channel.fromPath(params.run_metafile)
       .splitCsv(header: true, sep: '\t')
+      // filter to only include specified project ids
+      .filter{it.scpca_project_id in project_ids}
       // only include single-cell/single-nuclei and make sure no CITE-seq/ hashing libraries
       .filter{it.seq_unit in ['cell', 'nucleus']}
```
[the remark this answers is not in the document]
sjspielman: Aha, makes sense!
So, maybe we don't need to filter at all and we can just keep the `.unique()` at the end?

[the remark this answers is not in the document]
allyhawkins: I think I actually prefer to keep this filter in. This also makes sure we remove any spatial or bulk libraries.

merge.nf
```diff
@@ -90,45 +86,31 @@ process merge_report {
...
-      .filter{merge_groups_all  || (it.merge_group in merge_groups)}
-
-    // channel with run metadata, keeping only the columns we need
-    libraries_ch = Channel.fromPath(params.run_metafile)
+    // read in run metafile, filter to projects of interest, and group by project
+    grouped_libraries_ch = Channel.fromPath(params.run_metafile)
       .splitCsv(header: true, sep: '\t')
+      // filter to only include specified project ids
+      .filter{it.scpca_project_id in project_ids}
       // only include single-cell/single-nuclei and make sure no CITE-seq/ hashing libraries
```
sjspielman, line 97: how about this? feel free to change my language, just wanted to add a bit more about the reasoning!
```suggestion
      // only include single-cell/single-nuclei which already contain processed altexps, and ensure no spatial or bulk
```

sjspielman (approved):
Good to go! Just one comment about expanding the filtering comment to be a little more specific about why we need it!

PR #607, opened 2023-12-07
Bump initial memory and cpu request for CellAssign

allyhawkins:
In attempting to run the first two projects through CellAssign, I am noticing a few things: 

- The first is that most of the samples are failing in the initial two attempts for running CellAssign, and are more successful on the third attempt (although not always). This tells me we should be initially requesting more memory. I think I would also bump up the cpus so that hopefully this takes less time and is less likely to get killed. 
- The second is that, every run has been killed by the external system after ~5-7 hours. In some cases, CellAssign has completed and in others none of the processes have completed. I'm not entirely sure what else we can do other than bump up requests? I wonder if there is a time out that's happening? 

Below is a copy of the error that I'm consistently getting: 
Screenshot 2023-12-07 at 9 38 51 AM

I've also attached the log file so that you can see the complete output. 
[nextflow.log](https://github.com/AlexsLemonade/scpca-nf/files/13602054/nextflow.1.log)

### DIFF
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -46,7 +46,7 @@ process classify_cellassign {
       pattern: "${cellassign_dir}"
     )
   label 'mem_128'
-  label 'cpus_32'
+  label 'cpus_24'
   tag "${meta.library_id}"
   input:
     tuple val(meta), path(processed_rds), path(cellassign_reference_file)
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -36,8 +36,8 @@ process{
   withLabel: cpus_12 {
     cpus = {check_cpus(12)}
   }
-  withLabel: cpus_32 {
-    cpus = {check_cpus(32)}
+  withLabel: cpus_24 {
+    cpus = {check_cpus(24)}
   }
 }
 


modules/classify-celltypes.nf
```diff
@@ -45,8 +45,8 @@ process classify_cellassign {
       mode: 'copy',
       pattern: "${cellassign_dir}"
     )
-  label 'mem_32'
-  label 'cpus_12'
+  label 'mem_128'
+  label 'cpus_32'
```
jashapiro, line 49: I'm a little worried that increasing both threads and memory at the same time will mean that you will still hit memory limits. So I might be more conservative here and only up the cpus to 24 for now? Also multithreading tends to be less efficient with more cores, depending how well it is implemented.

jashapiro (approved):
I suggested being a bit more conservative with CPU requests, but other than that, this looks fine.

The jobs getting killed could be because these are spot instances, and AWS may be killing them randomly. The best way around that is to make them run faster, so upping the CPUs is probably the right move for that as well. But I am still wary of jumping _too_ high on the CPU count if the payoff isn't there.

modules/classify-celltypes.nf
allyhawkins: Okay yea I wasn't sure how much to bump it up to. I'll change it to 24 and then give it a go 🤞

Issue #589, opened 2023-11-21
Update description of tissue type and organs for CellAssign references in supplemental report


allyhawkins:
In #587, we are making some modifications to references that mean multiple tissues/ organs will be found in a single reference for CellAssign. I think it would be useful to be able to include the list of organs found in the reference in the actual supplemental cell type report. This would mean pulling in the organs from the reference metadata file some way, or maybe we could just link to that file in the repo so users can see which organs are found in that reference. I don't think this needs to be done before the release, I just wanted to note that we may want to address this in the future. Right now, we just print out the reference name, e.g., `blood-compartment`.

sjspielman:
As discussed in DSTM, we'll also need to track organs, so we'll need a separate step here of also reading in [the reference file which contains the organs](https://github.com/AlexsLemonade/scpca-nf/blob/main/references/celltype-reference-metadata.tsv). We'll want to read this file in and add organs into the metadata so that we can incorporate them into the report here. 

Noting this file is already in config since it's used for the reference building workflow - https://github.com/AlexsLemonade/scpca-nf/blob/30c7822e205c43aedb5484ab1e40589b68b9bc6e/config/reference_paths.config#L26

sjspielman:
Reminding myself about our cell typing code and thinking about workflow updates to be able to update the report - 

We don't actually pass around the reference names directly in nextflow, only the reference _filenames_. In `add_celltypes_to_sce.R`, we extract the reference name from the reference filename to include in SCE metadata. So, if we were to want to read in the celltype reference file (with organs) into the workflow as a map, we'd have to add in some nextflow string manipulation to join in the organs with our existing channels, and I think we might want to avoid that because it introduces redundancy - whatever string things we would have to add into nextflow are already being done by R.

So, are there any particular objections to just reading in the celltype reference file (with organs) directly into `add_celltypes_to_sce.R` and bypassing all nextflow manipulation? We'd have to add an argument to this script with the file path, but nextflow code won't really be changed except for a file exists check. @allyhawkins @jashapiro

jashapiro:
I don't think I see a problem with that method. You will need to pass the file itself into the workflow process as well to get it copied to the execution environment, but that should be about it.

allyhawkins:
That plan sounds fine to me.

sjspielman:
Closed by #608 & #609

PR #608, opened 2023-12-07
Add cellassign organs to metadata

sjspielman:
Part of #589 

This is the first of two PRs to get cellassign organs/tissues into the cell type report. Here, I just modify the `add_celltypes_to_sce` process and associated R script to get info into the metadata. Since we only need to do this for cellassign references, I set up nextflow to only pass in the file to R if cellassign is present.

Screen shot of resulting `metadata(sce)` - 

Screenshot 2023-12-07 at 12 28 48 PM

While I was there, I also added in some file checks for cell typing files, which seemed like it couldn't hurt! One question is should we call these `organs` or `tissues`? I went with organs since that's what we already call them. Also, my checks for whether we grabbed the organs successfully are maybe overkill, any thoughts?

FYI I opened a related docs issue as well - https://github.com/AlexsLemonade/scpca-docs/issues/211

*Edit - the next PR will get this into the cell type report! ~Probably the main report cell type section so it appears in both htmls.~ *edit 2: nope, we refer folks to the supplemental report for this sort of information!

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -47,7 +47,7 @@ option_list <- list(
     opt_str = c("--celltype_ref_metafile"),
     type = "character",
     help = "Metadata TSV file containing cell type reference metadata.
-            This file is used to obtain a list of organs used for CellAssign annotation."
+            This file is used to obtain a list of organs used for CellAssign annotation and is only required if CellAssign results provided as input."
   )
 )
 


modules/classify-celltypes.nf
```diff
@@ -234,7 +236,10 @@ workflow annotate_celltypes {
 
 
     // incorporate annotations into SCE object
-    add_celltypes_to_sce(assignment_input_ch.add_celltypes)
+    add_celltypes_to_sce(
+      assignment_input_ch.add_celltypes,
+      params.celltype_ref_metadata // file with CellAssign reference organs
```
jashapiro, line 241: You always want to pass in a file object when going to a path. This allows Nextflow to properly stage the files. With S3 files it usually works without it, but it may fail with local paths.
```suggestion
      file(params.celltype_ref_metadata) // file with CellAssign reference organs
```

bin/add_celltypes_to_sce.R
```diff
@@ -216,6 +226,17 @@ if (!is.null(opt$cellassign_predictions)) {
...
   #  come out to just the string "cellassign"
   metadata(sce)$celltype_methods <- c(metadata(sce)$celltype_methods, "cellassign")
+
+  # add cellassign reference organs to metadata
+  cellassign_organs <- opt$celltype_ref_metafile |>
+    readr::read_tsv() |>
+    dplyr::filter(celltype_ref_name == cellassign_ref_info[["ref_name"]]) |>
+    dplyr::pull(organs)
+
+  if (cellassign_organs == "" | is.na(cellassign_organs)) {
```
jashapiro, line 236: Truthiness is fun! (And Groovy treats NA as falsey)
```suggestion
  if (!cellassign_organs) {
```

bin/add_celltypes_to_sce.R
```diff
@@ -42,6 +42,12 @@ option_list <- list(
     type = "character",
     help = "Name of marker by cell type reference file used with CellAssign.
             File name is expected to be in form: `<ref_name>_<source>_<version>.tsv`"
+  ),
+  make_option(
+    opt_str = c("--celltype_ref_metafile"),
+    type = "character",
+    help = "Metadata TSV file containing cell type reference metadata.
+            This file is used to obtain a list of organs used for CellAssign annotation."
```
allyhawkins, line 50:
```suggestion
            This file is used to obtain a list of organs used for CellAssign annotation and is only required if CellAssign results provided as input."
```
Let's just be explicit that it's only required if CellAssign results are provided.

bin/add_celltypes_to_sce.R
```diff
@@ -216,6 +226,17 @@ if (!is.null(opt$cellassign_predictions)) {
   # note that if `metadata(sce)$celltype_methods` doesn't exist yet, this will
   #  come out to just the string "cellassign"
   metadata(sce)$celltype_methods <- c(metadata(sce)$celltype_methods, "cellassign")
+
+  # add cellassign reference organs to metadata
+  cellassign_organs <- opt$celltype_ref_metafile |>
```
allyhawkins, line 231: I say we just keep it at organs to keep it consistent with how PanglaoDB (and we) refer to them.

allyhawkins (approved):
LGTM, just one minor docs update!

PR #610, opened 2023-12-07
Account for feature data when merging objects

### LINKED ISSUE #595: Account for libraries that have feature data in merging workflow
The current plan for merging objects is to make sure that any ADT data is still present in the final merged object. However, with multiplexed data, we do not want to include the HTOs or feature data. This is because we already have sample calls that we have added to the `colData` and having multiple clashing HTOs could be confusing. We are planning on implementing a flag in the function for merging SCE's to tell the function to return a merged object with merged `altExp`s or without (see https://github.com/AlexsLemonade/scpcaTools/issues/150#issuecomment-1828426277). 

In order to pass that information along to the function mentioned ([currently run in `merge_sces.R`](https://github.com/AlexsLemonade/scpca-nf/blob/b865a3adf81757ede5650ff50d0aea81e271827c/bin/merge_sces.R#L107)), we will need to do the following: 

- Grab the technology information for each library when reading in the library metadata
- Create a `has_citeseq` and `is_multiplexed` flag that can be passed to the `merge_sce` process. 
- Add a flag for keep / don't keep altExp data to `merge_sces.R` and pass along to `merge_sce_list`. 

I think in the actual workflow, we really only need to remove the `altExp` if it's multiplexed. 
So then we should be able to create something like `--remove_altExp` flag that is used in `merge_sces.R`. 
We could do the opposite of that and use `has_citeseq ? --keep_altExp : ''`. But I don't actually think we will need both. 

```
merge_sces.R \
      --input_library_ids "${input_library_ids}" \
      --input_sce_files "${input_sces}" \
      --output_sce_file "${merged_sce_file}" \
      --n_hvg ${params.num_hvg} \
     ${is_multiplexed ? --remove_altExp : '' } \
      --threads ${task.cpus}
```

allyhawkins:
Closes #595 

This PR accounts for libraries that have feature data before merging. Specifically, any libraries that have ADT/CITE-seq data will retain the `altExp` in the object, while any libraries with cell hashing will not include the `altExp` in the merged object. I made the following changes: 

- To the script for merging SCE's, I added an option to `--include_alt_exp`. This mirrors the argument being added in https://github.com/AlexsLemonade/scpcaTools/pull/243, but here I am setting the default to be false. So by default the `altExp` will not be included unless the data specifically has CITE-seq/adt data. 
- I create a list of all project ids that have at least one library with CITE-seq data. For each project being processed, I created a `has_adt` value that gets passed to the process for merging. If `has_adt` is true, then the `--include_alt_exp` flag is used. For all other cases, no altExp data will be kept. 
- This means we do not explicitly check for multiplexed data, but I don't think we need that information since they will be treated like regular single-cell/ single-nuclei where no `altExp` will be kept. 
- This approach also implies that if a project has some libraries with CITE/hashing and some libraries without, that will be handled within the function in https://github.com/AlexsLemonade/scpcaTools/pull/243. My understanding of that PR is that is true. 
- The other thing I did was change the `publishDir` for the merge step. Now that we are only producing the merged objects and not integrating, we want to make sure we output this. 

The last thing I want to note is that previously there was a question about why we were including a filter for `seq_unit in ["cell", "nucleus"]` and the `unique` call. On working on this, I realized that CITE/cell hashing data will have two libraries that have `seq_unit` equal to `cell` or `nucleus`.  However the contents of the `tecnology` column will be different. So the first filter is to just remove bulk and spatial data from being merged. The `unique` is to remove duplicates because of CITE/ cell hashing for the same library. I updated the comments here to reflect that.

### DIFF
--- a/merge.nf
+++ b/merge.nf
@@ -31,11 +31,11 @@ if(param_error){
 process merge_sce {
   container params.SCPCATOOLS_CONTAINER
   label 'mem_16'
-  publishDir "${params.checkpoints_dir}/merged"
+  publishDir "${params.results_dir}/merged/${project_id}"
   input:
-    tuple val(project_id), val(library_ids), path(scpca_nf_file)
+    tuple val(project_id), val(has_adt), val(library_ids), path(scpca_nf_file)
   output:
-    tuple val(project_id), path(merged_sce_file)
+    tuple val(project_id), val(has_adt), path(merged_sce_file)
   script:
     input_library_ids = library_ids.join(',')
     input_sces = scpca_nf_file.join(',')
@@ -46,6 +46,7 @@ process merge_sce {
       --input_sce_files "${input_sces}" \
       --output_sce_file "${merged_sce_file}" \
       --n_hvg ${params.num_hvg} \
+      "${has_adt ? "--include_alt_exp" : ''} \
       --threads ${task.cpus}
     """
   stub:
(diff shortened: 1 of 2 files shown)


merge.nf
```diff
@@ -31,11 +31,11 @@ if(param_error){
 process merge_sce {
   container params.SCPCATOOLS_CONTAINER
   label 'mem_16'
-  publishDir "${params.checkpoints_dir}/merged"
+  publishDir "${params.results_dir}/merged/${project_id}"
   input:
-    tuple val(project_id), val(library_ids), path(scpca_nf_file)
+    tuple val(project_id), val(has_adt), val(library_ids), path(scpca_nf_file)
```
sjspielman, line 36: I wonder if we want to change the name of this value `project_id`  _within this process only_ to something like `merge_group_id`, for future us? For now, it will always refer to a project id, but the name `merge_group_id` is slightly more flexible for different groupings down the line. There's a few spots to replace this value in this process (L34, L36 [here], and L38) if you want to make that change.

sjspielman (approved):
Honestly this seems entirely fine to me and ready to go, no notes! 🎉 

Well ok, one note that I can't tell how I feel about, so tell me how you feel about it and that will be the answer! But, I feel fine approving as is! Tag me back for more discussion as needed.

PR #613, opened 2023-12-08
Add process for converting merged objects to AnnData

### LINKED ISSUE #604: Process for converting merged SCE to merged AnnData
Once we create a merged SCE object, we will also want to make it available as an AnnData object. This should just involve converting the SCE to AnnData as we do in the main workflow. For any objects where we keep the `altExp` when merging, this will mean we will have one file for RNA and one for the feature data. 

When we do this, we will need to decide how much reformatting we want to do with the merged SCE prior to conversion to AnnData. Right now, we do some reformatting of the individual SCE objects prior to conversion so that the objects are ready for CELLXGENE. I think generally we do want the format of the AnnData objects to be consistent, so we could probably just use the same `sce_to_anndata.R` script that we use in the main workflow. 

In `sce_to_anndata.R`, the following function is used to re-format. We should decide if there is anything in here that we don't want to included in the merged AnnData objects. 
https://github.com/AlexsLemonade/scpca-nf/blob/30c7822e205c43aedb5484ab1e40589b68b9bc6e/bin/sce_to_anndata.R#L58-L104

allyhawkins:
Closes #604 
⚠️ Stacked on #610 (Because this is slightly dependent on decisions made in #610, I'm going to wait to request review)

This PR adds a process to convert the merged SCE objects to an AnnData object. When converting to AnnData, I chose to use the same scripts that we use in the main workflow to convert the object and then move the counts. The reason for this is I think it's probably best to be consistent in how we format our objects. In particular, with the AnnData objects, we add the sample metadata as columns in the `colData` and do a little bit of renaming. I don't think there's any reason we shouldn't be doing that here. 

This process doesn't differ much from the conversion process in the main workflow, except here, we know that the only time we will have an altExp is if the `has_adt` value is `true`. I use that value to determine if the feature file should be created. 

Also, all of the objects are considered processed objects here, so I run `move_counts_anndata.py` for all HDF5 files. Again, I'm doing this so that how we name things is consistent with the other AnnData objects. 

The last idea I had was maybe we also want to add the sample metadata to the `colData` of the merged objects. That way, things like diagnosis, age, etc., would be easier to work with instead of living in a separate data frame in the `metadata`. What do we think?

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -29,7 +29,7 @@ option_list <- list(
             the default is n_hvg = 2000"
   ),
   make_option(
-    opt_str = c("--include_alt_exp"),
+    opt_str = c("--include_altexp"),
     action = "store_true",
     default = FALSE,
     help = "Keep any altExp present in the merged object."
@@ -115,7 +115,7 @@ merged_sce <- scpcaTools::merge_sce_list(
   batch_column = "library_id",
   preserve_rowdata_cols = "gene_symbol",
   cell_id_column = "cell_id",
-  include_alt_exp = include_alt_exp
+  include_altexp = opt$include_altexp
 )
 
 
--- a/config/containers.config
+++ b/config/containers.config
@@ -1,5 +1,5 @@
 // Docker container images
-SCPCATOOLS_CONTAINER = 'ghcr.io/alexslemonade/scpca-tools:v0.3.1'
+SCPCATOOLS_CONTAINER = 'ghcr.io/alexslemonade/scpca-tools:edge'
 
 ALEVINFRY_CONTAINER = 'quay.io/biocontainers/alevin-fry:0.7.0--h9f5acd7_1'
 BCFTOOLS_CONTAINER = 'quay.io/biocontainers/bcftools:1.14--h88f3f91_0'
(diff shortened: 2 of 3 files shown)


allyhawkins:
I've gone ahead and tested this and this is ready for review!

jashapiro (approved):
LGTM!

sjspielman (approved):
LGTM! The only thing I am thinking is I know I saw some comment go by _somewhere_ (github? slack?) recently from @jashapiro about updating the extension we use for hdf5 files... Should we do this? @jashapiro please weigh in!

allyhawkins:
> LGTM! The only thing I am thinking is I know I saw some comment go by _somewhere_ (github? slack?) recently from @jashapiro about updating the extension we use for hdf5 files... Should we do this? @jashapiro please weigh in!

I think if we did that then we would need to change all the individual AnnData file extensions and then dev would also have to update their code. They can have `.hdf5`, `.h5ad`, or `.h5`, so I don't think it's worth the headache personally.

jashapiro:
It is in #616, which seems like the place to do it (but https://github.com/AlexsLemonade/scpcaTools/issues/244 has to happen first, as we do checks in `scpcaTools`)

PR #614, opened 2023-12-11
Swap in one more merge_project_id

sjspielman:
Related to #610 

What it says!


allyhawkins (approved):
Thanks for catching this!

PR #615, opened 2023-12-11
Create faceted UMAPs by cell type

### LINKED ISSUE #611: Fix incorrect labels for cell type annotation comparison heatmap
For the heatmap comparing assignments from SingleR and CellAssign in the `Automated annotations` section, the x and y-axis labels are incorrect. This should be fixed so that the x-axis is SingleR annotations and then y-axis CellAssign annotations.

### LINKED ISSUE #612: Update colors used for cell type UMAPs
Right now, we are using the `Dark2` palette. However, the two greens that are showing up in this palette are a bit hard to distinguish from each other, which can make it hard to see which cell is assigned to which cell type. We should update the color palette that we use so it's easier to distinguish the cells on the plot. Maybe we could use `Okabe-Ito` instead?

allyhawkins:
Closes #611 
Closes #612 

In evaluating some of the reports from running cell typing, I noticed that it was hard to discern which cell belonged to which cell type because of the colors we were using. I tried some other color palettes, including `Okabe-Ito`, and it improved the colors, but there are definitely still some issues with discerning cells in the plot. After talking through this with Stephanie on Slack, we decided to try a different approach and make a panel of UMAPs. Here, each UMAP shows the cell type of interest in red and all other cells in grey. We are still limiting it to only show the top cell types, but I actually think this is a lot more clear and easier to distinguish cells. To accomplish this I did the following: 

- I created a new function for producing each of the UMAP panels. For each cell type, we have to make a new column designating if the cell belongs to the cell type of interest or other cells. That column is used to color the UMAP. 
- I also created a function to generate the list of top cell types. I tried just pulling the unique labels from the `celltype_annotation_lumped` column, but doing that pulled them in a random order and I want the UMAPs to be in order of cell type frequency. I also want `All remaining cell types` to be the last plot. 
- For each of the top cell types, I generate a UMAP panel and then use `patchwork::wrap_plots` to combine them. 
- I updated the number of cell types to keep before lumping all the rest to be 8. We were previously only showing 7 because we were limited by color palettes. Now by using 8 + the `All remaining cell types`, we have a 3x3 grid for the UMAPs. 

While working on this, I also switched the axes labels for the automated annotations heatmap to address #611. 

A few other notes: 

- I was seeing cases where we had ties, so the `lump_celltypes` was actually returning more than the specified cell type number. I updated that to always take the first match. 
- I did hard code the figure height and width for the UMAP plots to be 9 x 9. I don't imagine very many scenarios where we have fewer cell types, because I haven't really seen that when looking at some of the reports. However, that doesn't mean that can't happen. Do we want to try and adjust the number of cols, rows, and then figure height and width based on the number of cell types?
- I didn't adjust the text yet, because I want to make sure we like these plots. Then I can update the explanatory text for interpreting the plots in another PR.

Here's an example of the updated plots.
Screenshot 2023-12-11 at 1 27 48 PM

And an updated report. 
[main_qc_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13640139/main_qc_report.html.zip)

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -118,7 +118,6 @@ plot_umap <- function(
 #'
 #' @return ggplot object containing a faceted UMAP where each cell type is a facet.
 #'   In each panel, the cell type of interest is colored and all other cells are grey.
-
 faceted_umap <- function(umap_df,
                          annotation_column) {
   # color by the annotation column but only color one cell type at a time
@@ -320,7 +319,7 @@ if (length(levels(umap_df$cluster)) <= 8) {
 knitr::asis_output(
   'Next, we show UMAPs colored by cell types.
 For each cell typing method, we show a separate faceted UMAP.
-In each panel, cells that correspond to a single cell type are colored, while all other cells are in grey.
+In each panel, cells that were assigned the given cell type label are colored, while all other cells are in grey.
 
 For legibility, only the seven most common cell types are shown.
 All other cell types are grouped together and labeled "All remaining cell types" (not to be confused with "Unknown cell type" which represents cells that could not be classified).
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -131,7 +131,7 @@ faceted_umap <- function(umap_df,
       data = select(
         umap_df, -{{ annotation_column }}
       ),
-      color = "grey",
+      color = "gray80",
       alpha = 0.5,
       size = 0.3
     ) +
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -50,12 +50,12 @@ format_celltype_n_table <- function(df) {
 #' @param n_celltypes Number of groups to lump into, with rest put into "Other" group. Default is 7.
 #'
 #' @return Updated df with new column of lumped celltypes for each present method
-lump_celltypes <- function(df, n_celltypes = 7) {
+lump_celltypes <- function(df, n_celltypes = 8) {
   df <- df |>
     dplyr::mutate(
       across(
         ends_with("_celltype_annotation"),
-        \(x) forcats::fct_lump_n(x, n_celltypes, other_level = "All remaining cell types"),
+        \(x) forcats::fct_lump_n(x, n_celltypes, other_level = "All remaining cell types", ties.method = "first"),
         .names = "{.col}_lumped"
       )
     )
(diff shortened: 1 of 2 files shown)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -109,6 +109,74 @@ plot_umap <- function(
...
+#' @param celltype_df Data frame with cell types
+#' @param celltype_column Column containing cell type annotations
+#' @param num_celltypes Number of cell types to keep
+#'
+#' @return A list of the top cell types, with the last element being "All remaining cell types"
+
+list_celltypes <- function(celltype_df,
+                           celltype_column,
+                           num_celltypes = 8) {
+  celltype_list <- create_celltype_n_table(celltype_df, !!sym(celltype_column)) |>
```
sjspielman, line 171: did you try `{{}}` here? if you did and it didn't work, ignore this comment!

```suggestion
  celltype_list <- create_celltype_n_table(celltype_df, {{celltype_column}}) |>
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -109,6 +109,74 @@ plot_umap <- function(
...
+      size = 0.3,
+      alpha = 0.5
+    ) +
+    # remove axis numbers and background grid
+    scale_x_continuous(labels = NULL, breaks = NULL) +
+    scale_y_continuous(labels = NULL, breaks = NULL) +
+    theme(
+      # no legend here
+      legend.position = "None"
+    ) +
```
sjspielman, line 151: I think we'll want a few theme tweaks here - 

- The font (both plot and axis titles) are really large compared to other figures. Let's drop these down.
- We might also want to fiddle with plot margins to reduce the amount of whitespace between panels. That's something to handle on the ggplot, not patchwork, side of things.

jashapiro:
Just a comment that I would rather see this as real facets, to save some redundancy in the axis labels. You can do this with something like this:

```
ggplot(umap_df)+
    geom_point(data = select(umap_df, -annotation_column), color = "grey") +
    geom_point() +
    facet_wrap(vars(annotation_column))
```
This can preserves the multiple colors if you like, which could be nice. Or you can assign the foreground colors as you like in the second geom_point()

sjspielman (commented):
Off to a good start, and I just learned from https://github.com/AlexsLemonade/scpca-nf/pull/615#issuecomment-1850767819 that faceting is possible 😲 

A couple other comments that I couldn't leave directly - 

Should we bump this from 2 --> 3? I'm on the fence...is the TOC bonkers at 3? https://github.com/AlexsLemonade/scpca-nf/blob/733a1f7dc5de0bb9a61098659b73b994ff718c08/templates/qc_report/main_qc_report.rmd#L17

This text in the main QC report needs updating (L341-342) to explain the new plot layouts.
> For legibility, only the seven most common cell types are shown.
All other cell types are grouped together and labeled "All remaining cell types" (not to be confused with "Unknown cell type" which represents cells that could not be classified).

jashapiro:
> I just learned from [#615 (comment)](https://github.com/AlexsLemonade/scpca-nf/pull/615#issuecomment-1850767819) that faceting is possible 😲

Full disclosure, I learned it from Hadley: https://ggplot2-book.org/facet#sec-group-vs-facet (last plot in the section)

allyhawkins:
🤯 This was so much easier! I used ggplot directly to create the faceted plot as Josh suggested and it is so much simpler. Here I am keeping the original color palette and coloring each cell type with a unique color. The only problem there is that we can only keep 7 cell types again because of color restrictions so it sadly doesn't make a perfect grid. What do others think?

Screenshot 2023-12-11 at 2 33 46 PM

I also was able to remove the little helper function for re-ordering, since we don't need it this way. It automatically is ordering them the way I want them, by frequency with `All remaining cell types` last. 

I wasn't planning on updating the text, because I was going to do that separately. Since this PR just got smaller, I went ahead and updated it. Definitely looking for feedback on the wording there!

jashapiro:
Looks good!

> The only problem there is that we can only keep 7 cell types again because of color restrictions so it sadly doesn't make a perfect grid. What do others think?

Let's turn that negative into a positive! We can put the legend in the last spot... I haven't tested this, but something like:
```
theme(
  legend.position = c(1,0), 
  legend.justification(legend.justification = c("right", "bottom"))
)
```

sjspielman:
> Let's turn that negative into a positive! 

I think you meant, "Let's make lemonade out of lemons!" 🍋

sjspielman:
Faceting for the win! I think we might want to go with a lighter gray though to emphasize some of the points that are also light colored. Maybe `gray80`?

templates/qc_report/celltypes_qc.rmd
```diff
@@ -109,6 +109,56 @@ plot_umap <- function(
...
+
+
+#' Create a faceted UMAP panel where each panel has only one cell type colored
+#'
+#' @param umap_df Data frame with UMAP1 and UMAP2 columns
+#' @param annotation_column Column containing cell type annotations
+#'
+#' @return ggplot object containing a faceted UMAP where each cell type is a facet.
+#'   In each panel, the cell type of interest is colored and all other cells are grey.
+
```
sjspielman, line 121:
```suggestion
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -109,6 +109,56 @@ plot_umap <- function(
...
+  faceted_umap <- ggplot(
+    umap_df,
+    aes(x = UMAP1, y = UMAP2, color = {{ annotation_column }})
+  ) +
+    # set points for all "other" points
+    geom_point(
+      data = select(
+        umap_df, -{{ annotation_column }}
+      ),
+      color = "grey",
```
sjspielman, line 134: I should have just left this comment here... can we try out a lighter gray?

allyhawkins:
What do we think? I moved the legend and used lighter grey. I'm 50/50 on the legend placement, but because we are using different colors I do think we should keep it. 

Screenshot 2023-12-11 at 3 11 35 PM

sjspielman:
> What do we think? I moved the legend and used lighter grey. 

The lighter gray is much easier to see!! I still sort of think the legend isn't crucial given the faceting, but I also think it's charming in its little bottom right corner home. Charming wins.

sjspielman (approved):
> I didn't adjust the text yet, because I want to make sure we like these plots. Then I can update the explanatory text for interpreting the plots in another PR.

I totally missed this bullet point in your opening PR text, sorry about that!! I think what you've already done here is enough explanation. 

🚀

PR #617, opened 2023-12-12
Add y axis label for CellAssign probability plot

allyhawkins:
Here I just added a label for the y-axis for the CellAssign plot to mirror what we have in the SingleR delta median plot.


sjspielman (approved):
👍

PR #621, opened 2023-12-14
Compress AnnData output

### LINKED ISSUE #616: Support compressing AnnData objects
Following https://github.com/AlexsLemonade/scpcaTools/issues/244 should support 

We should add compression to both
`bin/sce_to_anndata.R` and 
`bin/move_counts_anndata.py`

The former depends on https://github.com/AlexsLemonade/scpcaTools/issues/244, but the latter does not, and can be implemented with just a `compression` argument to [`AnnData.write`](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.write.html). Unfortunately, we don't always run `move_counts_anndata.py`, so we can't just put compression there. 

We could make the compression in `sce_to_anndata.R` an argument, which might be useful so we don't waste time compressing processed data before we move the counts and recompress. 

**Additional context**
Just noting that we use `.hdf5` everywhere, but maybe we want to switch to `.h5ad`?

allyhawkins:
Closes #616 

This PR adds a new flag, `--compress_output` to the script for converting objects from SCE to AnnData. This means that by default we set compression to `none`, unless the flag is used. This allows us to specify compressing only for files that are not processed. 

For zipping the processed files, I added the argument to the `object.writeH5AD()` function. 

I did not change the extension of the HDF5 files here, because that change doesn't just affect us. I think if we want to do that we should do that separately. 

I tested both the main and merging workflow to be sure that the new additions work.

### DIFF
--- a/bin/move_counts_anndata.py
+++ b/bin/move_counts_anndata.py
@@ -22,14 +22,14 @@ parser.add_argument(
 args = parser.parse_args()
 
 # compile extension regex
-file_ext = re.compile(r"\.hdf5$|.h5$", re.IGNORECASE)
+file_ext = re.compile(r"\.hdf5$|.h5$|.h5ad$", re.IGNORECASE)
 
 # check that input file exists, if it does exist, make sure it's an h5 file
 if not os.path.exists(args.anndata_file):
     raise FileExistsError("`input_anndata` does not exist.")
 elif not file_ext.search(args.anndata_file):
     raise ValueError(
-        "--input_anndata must end in either .hdf5 or .h5 and contain a processed AnnData object."
+        "--input_anndata must end in either .hdf5, .h5, or .h5ad, and contain a processed AnnData object."
     )
 
 # read in anndata
@@ -44,4 +44,4 @@ if "logcounts" in object.layers:
     object.uns["X_name"] = "logcounts"
 
     # export object
-    object.write_h5ad(args.anndata_file)
+    object.write_h5ad(args.anndata_file, compression="gzip")
--- a/bin/move_counts_anndata.py
+++ b/bin/move_counts_anndata.py
@@ -18,6 +18,13 @@ parser.add_argument(
     required=True,
     help="Path to HDF5 file with processed AnnData object",
 )
+parser.add_argument(
+    "-u",
+    "--uncompressed",
+    dest="compress",
+    action="store_false",
+    help="Output an uncompressed HDF5 file",
+)
 
 args = parser.parse_args()
 
@@ -44,4 +51,4 @@ if "logcounts" in object.layers:
     object.uns["X_name"] = "logcounts"
 
     # export object
-    object.write_h5ad(args.anndata_file, compression="gzip")
+    object.write_h5ad(args.anndata_file, compression="gzip" if args.compress else None)
--- a/bin/move_counts_anndata.py
+++ b/bin/move_counts_anndata.py
@@ -22,7 +22,7 @@ parser.add_argument(
 args = parser.parse_args()
 
 # compile extension regex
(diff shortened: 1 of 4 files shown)


bin/move_counts_anndata.py
```diff
@@ -22,14 +22,14 @@
 args = parser.parse_args()
 
 # compile extension regex
-file_ext = re.compile(r"\.hdf5$|.h5$", re.IGNORECASE)
+file_ext = re.compile(r"\.hdf5$|.h5$|.h5ad$", re.IGNORECASE)
```
jashapiro, line 25: Escape all the dots
```suggestion
file_ext = re.compile(r"\.hdf5$|\.h5$|\.h5ad$", re.IGNORECASE)
```

bin/move_counts_anndata.py
```diff
@@ -44,4 +44,4 @@
     object.uns["X_name"] = "logcounts"
 
     # export object
-    object.write_h5ad(args.anndata_file)
+    object.write_h5ad(args.anndata_file, compression="gzip")
```
jashapiro, line 47: I think I would make compression optional as an arg to this function, just in case.

I can't suggest the option, but I can implement it here...

```suggestion
    object.write_h5ad(args.anndata_file, compression="gzip" if args.compress else None)
```

I might do this as a "negative" option, so the default is true, which saves you from having to edit anything else:
```python
parser.add_argument(
    "-u",
    "--uncompressed",
    dest="compress",
    action="store_false",
    help="Output an uncompressed HDF5 file",
)
```

allyhawkins:
I just tested merging with more than 2 libraries and found some things that were causing issues with merging. We don't want to re-add the `library_id` column if it's already there and we want to make sure we are preserving the row names originally found in the `colData(sce)` since that will contain the modified cell ids to have `{library_id}-{barcode}`.  Otherwise, we were accidentally over writing them. 

I also just bumped up the memory a little since using ~ 10 libraries caused retries for both merging and conversion to AnnData.

bin/sce_to_anndata.R
```diff
@@ -114,12 +122,19 @@ sample_metadata <- metadata(sce)$sample_metadata
...
+}
+
 # export sce as anndata object
 # this function will also remove any R-specific object types from the SCE metadata
 #   before converting to AnnData
 scpcaTools::sce_to_anndata(
   sce,
-  anndata_file = opt$output_rna_h5
+  anndata_file = opt$output_rna_h5,
+  compression = compression
```
jashapiro, line 137: I think I would just inline this for directness
```suggestion
  compression = ifelse(opt$compress_output, "gzip", "none")
```

bin/sce_to_anndata.R
```diff
@@ -114,12 +122,19 @@ sample_metadata <- metadata(sce)$sample_metadata
 # make main sce czi compliant
 sce <- format_czi(sce)
 
+if (opt$compress_output) {
+  compression <- "gzip"
+} else {
+  compression <- "none"
+}
+
```
jashapiro, line 130: Delete if you take my inline suggestion
```suggestion
```

merge.nf
```diff
@@ -102,7 +102,8 @@ process export_anndata{
         --input_sce_file ${merged_sce_file} \
         --output_rna_h5 ${rna_hdf5_file} \
         --output_feature_h5 ${feature_hdf5_file} \
-        ${has_adt ? "--feature_name adt" : ''}
+        ${has_adt ? "--feature_name adt" : ''} \
+        --compress_output
```
jashapiro, line 106: I don't think we want to compress here as we will be moving the outputs below
```suggestion
        ${has_adt ? "--feature_name adt" : ''} 
```

jashapiro (approved):
Looks good.  I suggested a few changes, but I don't think I need to see it agian.

PR #628, opened 2023-12-19
Update batch profile to set max aws retries and On Demand queue

### LINKED ISSUE #623: Decrease number of spot attempts and use OnDemand for third retry
Before we run the next projects through cell typing, we want to adjust the workflow process parameters so that AWS only retries 2 times instead of the default 5 times. 
We can use the `aws.batch.maxSpotAttempts` parameter to set this. 

We also want to update our retry strategy so that we change the queue to use OnDemand on the third retry. This cna be done after we actually set up a queue for OnDemand.

allyhawkins:
Closes #623 

Here I updated the `batch` profile to use `maxSpotsAttempts = 2`, instead of the default of 5. 
I also udpated the `queue` so that for the first two tries we use the auto scale queue we have been using, but for the third retry we use an On Demand queue. 

I think we might want to actually test this with a few libraries before we merge this in? 

Also, I'm merging this into main so that we can release this along with any CPU changes that may happen before processing the projects through CellAssign again.

### DIFF
--- a/config/profile_awsbatch_auto.config
+++ b/config/profile_awsbatch_auto.config
@@ -3,10 +3,10 @@ workDir = 's3://nextflow-ccdl-data/work'
 aws{
   batch{
     cliPath = '/home/ec2-user/miniconda/bin/aws'
-    maxTransferAttempts = 2
     volumes = [
       '/ebs-autoscale:/tmp'
     ]
+    maxTransferAttempts = 2
     maxSpotAttempts = 2
   }
   region = 'us-east-1'
--- a/config/profile_awsbatch_auto.config
+++ b/config/profile_awsbatch_auto.config
@@ -15,5 +15,5 @@ aws{
 
 process{
   executor = 'awsbatch'
-  queue = { task.attempt < 2 ? 'nextflow-batch-autoscale-queue' : 'nextflow-batch-priority-queue'}
+  queue = { task.attempt < 3 ? 'nextflow-batch-autoscale-queue' : 'nextflow-batch-priority-queue'}
 }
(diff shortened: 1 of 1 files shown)


config/profile_awsbatch_auto.config
```diff
@@ -7,12 +7,13 @@ aws{
     volumes = [
       '/ebs-autoscale:/tmp'
     ]
+    maxSpotAttempts = 2
```
jashapiro, line 10: Super minor, but let's move `maxTransferAttempts` down here for consistency...

config/profile_awsbatch_auto.config
```diff
@@ -7,12 +7,13 @@ aws{
...
   }
   region = 'us-east-1'
 }
 
 
 process{
   executor = 'awsbatch'
-  queue = 'nextflow-batch-autoscale-queue'
+  queue = { task.attempt < 2 ? 'nextflow-batch-autoscale-queue' : 'nextflow-batch-priority-queue'}
```
jashapiro, line 18: I think we want to do this on the third try?
```suggestion
  queue = { task.attempt < 3 ? 'nextflow-batch-autoscale-queue' : 'nextflow-batch-priority-queue'}
```

jashapiro (commented):
Looks good, but I am pretty sure `task.attempt` counts from 1, so we need to change that.

I agree that we should test a bit, but I think you can merge this in without too much worry. It is a config change only (and you could really run with the current version and just these config updates locally, using `-c nextflow.config`)

jashapiro (approved):
👍🏼

Issue #622, opened 2023-12-18
Test CPU usage for CellAssign


allyhawkins:
We recently increased the number of CPUs and memory requested for running `CellAssign`. We want to do a test to make sure that when we do increase the number of CPUs, we decrease the run time. 

We are using `SCPL000015` to do this test. It's already been run with 24 CPUS and 96 GB at a run time of 4 h 16 min. 
We should also run with 32 CPUs and 128 GB and see if the run time decreases. When we do this remember to increase the `max.cpus` and `max.memory`.

allyhawkins:
Just noting that the run time for 32 CPUs and 128 GB was 4 h 8 min. 

I will do another run with 12 CPUs and 64 GB.

jashapiro:
So it seems that the 12/CPU/64GB version failed with an out of memory error. 

The 12 CPU/96GB version took ~5 h 6 min (second try with one spot instance kill). I would say that means we should probably run cellassign at that config, as the benefits of more CPUs are not likely to be worth the cost.

allyhawkins:
I don't think that run actually finished? Tower doesn't show that CellAssign was completed since this message pops up. 

<img width="835" alt="Screenshot 2023-12-20 at 8 10 41 AM" src="https://github.com/AlexsLemonade/scpca-nf/assets/54039191/3af27cd1-8917-44f8-88db-289f0602746e">

jashapiro:
Hmmm. According to AWS it completed with exit code 0. That error seems like a nextflow something that failed. Hopefully transient?

allyhawkins:
Looking at the time stamps for the results and the checkpoints folder for that library, the last timestamp is 12/18. If it succeeded, there should have been a timestamp for a file on 12/19.

jashapiro:
Yes, nextflow clearly failed to get the results and proceed, but the batch job itself seems to have succeeded. So I expect we can trust the timing. But It makes sense to test again... with 96 GB RAM.

allyhawkins:
> Yes, nextflow clearly failed to get the results and proceed, but the batch job itself seems to have succeeded. So I expect we can trust the timing. But It makes sense to test again... with 96 GB RAM.

Okay I'll start that run now.

allyhawkins:
Confirming that the 12 CPU/ 96 GB ran in 5 h 2 min, so we should update to use this config moving forward.

allyhawkins:
I'm going to close this since we completed the test.

PR #631, opened 2023-12-20
Update CPU and mem requests for CellAssign

allyhawkins:
Based on what we found in #622, we want to use 96 GB and 12 cpus for classifying with CellAssign. This PR adds a new label for 96, removing the old 128 GB label. And then updates the process to use the desired resources.

### DIFF
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -51,6 +51,6 @@ def check_memory(memory){
 
 def check_cpus(cpus, max_cpus){
   cpus = cpus as int
-  def max_cpus = max_cpus as int
+  max_cpus = max_cpus as int
   cpus < max_cpus ? cpus : max_cpus
 }
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -22,7 +22,7 @@ process{
     memory = {check_memory(32.GB * task.attempt)}
   }
   withLabel: mem_96 {
-    memory = {check_memory(96.GB * task.attempt)}
+    memory = {check_memory(48.GB  + 48.GB * task.attempt)}
   }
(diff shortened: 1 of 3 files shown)


config/process_base.config
```diff
@@ -21,8 +21,8 @@ process{
   withLabel: mem_32 {
     memory = {check_memory(32.GB * task.attempt)}
   }
-  withLabel: mem_128 {
-    memory = {check_memory(128.GB * task.attempt)}
+  withLabel: mem_96 {
+    memory = {check_memory(96.GB * task.attempt)}
```
jashapiro, line 25: I kind of feel like we shouldn't just double once we get this large, so here I am suggesting increasing by only 50% at a time. 
```suggestion
    memory = {check_memory(48.GB  + 48.GB * task.attempt)}
```

We should probably update the max_memory along with this... but I think we can do that in `config/profile_ccdl.config`. We can go up to 256GB there. We could also go up to 64 CPUs. (I have a little worry that namespaces might not work quite right with the order things are defined, but I can come back to that.)

allyhawkins: I went ahead and did that. I looked at the main `nextflow.config` file and we pull in the ccdl profile after we pull in the process config, so maybe it will be okay?

allyhawkins:
I went ahead and made those changes here so we don't need to have a separate PR. If you want to take another look @jashapiro.

jashapiro:
> I went ahead and made those changes here so we don't need to have a separate PR.

Looks good, but the memory function & calls need a similar update!

jashapiro (approved):
LGTM!

PR #635, opened 2023-12-22
Move faceted_umap function

sjspielman:
Does what it says, to match our other UMAPs!

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -109,56 +109,6 @@ plot_umap <- function(
       aspect.ratio = 1
     )
 }
-
-
-#' Create a faceted UMAP panel where each panel has only one cell type colored
-#'
-#' @param umap_df Data frame with UMAP1 and UMAP2 columns
-#' @param annotation_column Column containing cell type annotations
-#'
-#' @return ggplot object containing a faceted UMAP where each cell type is a facet.
-#'   In each panel, the cell type of interest is colored and all other cells are grey.
-faceted_umap <- function(umap_df,
-                         annotation_column) {
-  # color by the annotation column but only color one cell type at a time
-  faceted_umap <- ggplot(
-    umap_df,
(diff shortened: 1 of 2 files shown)


allyhawkins:
Do you have an example of what this looks like in a report? I'm pretty sure I removed this and am manually setting the sizes to be squares because adding the aspect ratio made really small UMAPs in a grid that were not really legible.

sjspielman:
> Do you have an example of what this looks like in a report?

I do not... Was just writing my own faceted umaps and copied this code as my starting point, and saw it was missing and added it. But if you explored this and saw it was bad in the faceted UMAP, tha will also be relevant to #636! I will investigate with a cell type report.

sjspielman:
With the library I'm using (`SCPCL000295`), the UMAP doesn't look squished at all to me.
But, we have other problems... I see this both and without `aspect.ratio = 1`. 

## singler plot
> the 0.9 in `legend.position = c(.9, 0),` is not general enough 




## submitter celltypes plot
> we don't lump for this plot, which is the right choice, but then we have legend placement problems.

sjspielman:
I've reverted the aspect ratio commit for now, but I did realize something was funky while I was looking into this. The function for umaps was in the wrong file! You got away with it working before because the celltypes_qc report is called from the supplemental celltype report, but the faceted umaps are only in the supplemental celltype report so that's where the function should live. I went ahead and moved it in this PR!

allyhawkins:
> The function for umaps was in the wrong file! You got away with it working before because the celltypes_qc report is called from the supplemental celltype report, but the faceted umaps are only in the supplemental celltype report so that's where the function should live. I went ahead and moved it in this PR!

This isn't correct... the faceted UMAPs should be shown in the main cell type report.

sjspielman:
> This isn't correct... the faceted UMAPs should be shown in the main cell type report.

Oh man. Of course you are correct. I am so tired.

sjspielman:
Ok, this was a useful journey to discover the legend bug and open #637, but otherwise, it's clearly time for winter break 2024. Closing out this PR as "not actually useful."

PR #632, opened 2023-12-21
Add tables to merge report

allyhawkins:
Closes #625 
Stacked on #629 

Here I'm adding two separate sections that include summary tables. One for library statistics and one for sample statistics. 

- The library statistics section includes a table that summarizes the types of libraries, including tech and seq unit information. This table is shown for every single library type. 
- To account for ADT or multiplexed data, I added two separate tables that show the total number of libraries with/without ADT or multiplexed data. These are only shown if at least one library has adt or cellhash. 
  - To detect if any libraries have ADT/multiplexing, I went back into the `merge_sces.R` script and added in an `additional_modalities` column to each SCE before merging. I use the names of the `altExp` in the original SCE objects to populate this column. If there are no `altExp`, then this is set to `NA`. 
  - Having this column in the merged object then allows me to count the total number of libraries with each type of modality. 
- The sample section includes two tables - one summarizing the types of diagnoses that we have, including subdiagnoses, and the other one summarizing the `disease_timing` column. I did think of combining these into one table, but I personally felt it would get overcrowded. We have some very unique disease timing entries, so I think separating it is the way to go. 
- If any libraries in the project are multiplexed, then we won't have added the sample metadata to the `colData` (this is dependent on any changes in #630). We don't have information about individual multiplexed samples, so I added a pretty blue (go blue 〽️) message box that says this object contains multiplexed samples that have not been demultiplexed so no sample summaries are shown. 

Because this is somewhat dependent on #630, I'm filing this as a draft for now. 

Here is a zip with three different versions of the report - one with no additional modalities, one with adt, and one with multiplex. 
[report-examples.zip](https://github.com/AlexsLemonade/scpca-nf/files/13743508/report-examples.zip)

PR #636, opened 2023-12-22
Add merge report UMAPs

### LINKED ISSUE #626: Merged report: UMAPs
_Broken out of #594_

This issue tracks making the third merged report section: UMAPs

  - Big yellow banner _caution this is not integrated._
  - Sentence explaining how the UMAPs were calculated (they are not the same as UMAPs in the individual-library objects)
  - Show faceted UMAP with gray background points and a single color all around for library of interest
     - We should probably also state that each facet (below) is a _library, not a sample_, if the data is multiplexed
     - We'll also need to dynamically set the chunk plot size based on the number of libraries. 
  - _Optional_: A second banner stating "I map, you map, we all map for map map."

sjspielman:
(probably?) Closes #626

Here we have some merged report UMAPs. I added a facet UMAP and some text that the coordinates differ from individual UMAPs.

There are two items that we will need to circle back to once we start running some libraries through the merge workflow. Because of these, I'm not sure if we'll want to keep #626 open or make new issues (hence the `(probably?)` above). I think we probably want new issues..?

- Sizing! I took a first stab at some kind of sizing for the overall faceted plot, but we'll definitely want to look at all of them to choose the right calculations. If this seems not unreasonable as a first attempt, maybe let's leave it for now and figure it out later.
- Is it multiplexed? I wondered if we'll want an extra warning that each facet is a library, not sample. I've gotten that code started but we'll want to circle back after ongoing `scpcaTools` changes go in to make sure the parsing makes sense.

Here's the UMAP section! 😉 
-----
Screenshot 2023-12-22 at 12 44 17 PM


jashapiro:
Ummm... are those the colors you mean to use?

sjspielman:
> Ummm... are those the colors you mean to use? (also, I'd lose the grid lines)

They have been temporarily chosen for the amusement of the reviewer, but don't tell her!

allyhawkins:
The blue needs to be a little bit lighter, but I say we keep them 😄

sjspielman:
> The blue needs to be a little bit lighter, but I say we keep them 😄

You'll have to take that up them! https://teamcolorcodes.com/michigan-wolverines-color-codes/

allyhawkins:
> > The blue needs to be a little bit lighter, but I say we keep them 😄
> 
> You'll have to take that up them! https://teamcolorcodes.com/michigan-wolverines-color-codes/

Wow thank you for helping me learn about this! Can't wait for all my future plots to incorporate these colors.

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+table(params$merged_sce[[batch_column]])
+```
+
+## UMAPs
+
+```{r, ref.label=c('integration-warning'), eval=TRUE, results='asis'}
+```
+
+In this section, we present UMAPs of all libraries in the merged object, highlighting one library in each facet. 
+In each panel, cells from the given library are colored, while all other cells are in gray.
```
allyhawkins, line 86:
```suggestion
In this section, we present UMAPs showing all cells from all libraries included in the merged object. 
For each library, a separate panel is shown, and cells from that library are colored, while all other cells are gray.
```
Just a little rewording that I think might clarify what is being shown.

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+
+**These merged object UMAP coordinates differ from UMAPs in individual library files.**
```
allyhawkins, line 88: I think if you're going to say this I would add a little more context. Maybe mention that after merging, we re-calculate PCA and UMAP. Because of this, the coordinates found in the merged object will differ from the coordinates found in the individual libraries.

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+
+```{r}
+# Determine appropriate dimensions for UMAP based on number of libraries
+n_libraries <- length(unique(params$merged_sce[[batch_column]]))
+
+# Determine faceting layout based on n_libraries
+facet_ncol <- dplyr::case_when(
+  n_libraries <= 20 ~ 4,
+  n_libraries <= 50 ~ 6,
+  n_libraries <= 80 ~ 8,
```
allyhawkins, line 99: I think this is fine for now, but we might want to revisit when we actually create the merged object for projects this big. I'm worried things will get too small.

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+  .default = 10
+)
+
+# 3 inches per UMAP row?
+umap_height <- ceiling(n_libraries / facet_ncol) * 3
+
+# TODO: until we run the workflow through again after scpcaTools, this will not work. We will need to circle back to this.
+has_multiplex <- FALSE # temporary
```
allyhawkins, line 107: You might be able to remove this, since I'm adding this in https://github.com/AlexsLemonade/scpca-nf/pull/632/files

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+
+# Determine if library is multiplexed, for printing a warning.
+# max_libraries_per_sample <- metadata(merged_sce)$sample_metadata |>
+#  dplyr::count(library_id) |>
+#  dplyr::pull(n) |>
+#  max()
+# has_multiplex <- max_libraries_per_sample > 1
```
allyhawkins, line 114: same applies to this, we probably wont' need it here since we should have `is_multiplexed` set when we create the tables.

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+
+```{r, fig.width = 8, fig.height = umap_height}
+umap_df <- reducedDim(params$merged_sce, "UMAP") |>
+  as.data.frame() |>
+  tibble::rownames_to_column("library") |>
+  tidyr::separate(
+    library,
+    into = c("library", "barcode"),
+    sep = "-"
+  )
```
allyhawkins, line 137:
```suggestion
umap_df <- reducedDim(params$merged_sce, "UMAP") |>
  as.data.frame() 
```
You won't need to do any of this since `library_id` is a column in the merged object.

allyhawkins: Related to this, we will actually want to use the `batch_column` parameter.

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+
+ggplot(
+  umap_df,
+  aes(x = UMAP1, y = UMAP2, color = library)
```
allyhawkins, line 141:
```suggestion
  aes(x = UMAP1, y = UMAP2, color ={{params$batch_column}} )
```

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+) +
+  # set points for all "other" points
+  geom_point(
+    data = dplyr::select(
+      umap_df, -library
```
allyhawkins, line 146:
```suggestion
      umap_df, -{{params$batch_column}}
```
I actually am not sure this formatting will work, but I think it should 🤷‍♀️

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+    ),
+    color = "#ffcb05",
```
allyhawkins, line 148: Just a reminder to sadly change these colors 😢

templates/merge-report.Rmd
```diff
@@ -69,7 +74,91 @@ To conduct any downstream analyses that require batch correction (e.g., UMAP or
...
+    alpha = 0.5,
+    size = 0.3
```
allyhawkins, line 150: I think the points actually look kind of big and all lump together. I'm worried this will get worse as we have more libraries, so maybe going down to 0.1?

templates/merge-report.Rmd
```diff
@@ -26,10 +25,16 @@ knitr::opts_chunk$set(
 library(SingleCellExperiment)
 library(ggplot2)
 
-# Set default ggplot theme
+# Set default ggplot theme, customized for UMAPs
 theme_set(
   theme_bw() +
```
allyhawkins, line 30: I think we don't want to use `theme_bw()` for the UMAPs because we want to remove the grid.

allyhawkins (commented):
Thanks for getting this started, and using the best colors 💙 💛 
I had a few comments, but I also wanted to point out that we should probably wait to add any code regarding if the object contains multiplexed data because I added some things to capture that information in #630 and #632.

templates/merge-report.Rmd
sjspielman: > but we might want to revisit when we actually create the merged object for projects this big

Absolutely... I have no idea what's going to happen...

templates/merge-report.Rmd
sjspielman: I think we should keep it and just override the grid as I've done here. Either way, we'd have to override something - if we go with default ggplot, we'd have to remove the gray background, add in panel borders, and maybe also add facet strip outlines. I thought of maybe `theme_classic()` too, but we'd have to add in panel borders there too.. Either way something will have to get overridden, but `theme_bw()` gives us the panel borders we probably want to keep at least. Thoughts?

https://ggplot2.tidyverse.org/reference/ggtheme.html - 
Screenshot 2024-01-02 at 12 29 38 PM

templates/merge-report.Rmd
sjspielman: For now, maybe I'll keep a _temporary_ `is_multiplexed = FALSE` (with other `has_multiplex` -> `is_multiplexed`)?

templates/merge-report.Rmd
sjspielman: Yep, I realized about `batch_column` too....

But, the `reducedDims` don't have the `library_id` since that's in the `colData`... I might restore this code? Am I missing something?

sjspielman: Yeah, I'm missing something. I'll need to _join_ the coldata to get the batch column

templates/merge-report.Rmd
allyhawkins: I think we want to keep the theme but then add this to when we actually create the UMAP so we remove the grid. 
https://github.com/AlexsLemonade/scpca-nf/blob/55e891648b801da5cb91384aa8fe5a951533e03e/templates/qc_report/celltypes_qc.rmd#L141-L144

templates/merge-report.Rmd
allyhawkins: Sure. I think we just want to wait to approve until #632 goes in then.

templates/merge-report.Rmd
allyhawkins: Oh yes you are right if you are just pulling the reducedDims and not the `colData`. I think we could grab the colData and merge with the reducedDims by the barcode/ colnames and then use the `params$batch_column` to make sure we are grabbing the correct ID's for the cells. Although the library ID's should be in the colnames, I think we want to maintain using the batch column.

allyhawkins: I did not see your previous comment when writing this, but your plan sounds good!

templates/merge-report.Rmd
sjspielman: So, `panel.grid = element_blank(),` is already removing the grid. Here's where we are after https://github.com/AlexsLemonade/scpca-nf/pull/636/commits/f4ff899344a8f9eec33c6c937777bfa34f0691a4 (I used `firebrick3` which is just a color I like, open to discussion!)

Screenshot 2024-01-02 at 1 18 15 PM

sjspielman: noting this probably snuck past review since it was missing from the original image/PR! It was added here: https://github.com/AlexsLemonade/scpca-nf/pull/636/commits/16a46764e99fd680db553ed17b4be136702092a2

allyhawkins: I like this without the grid, and I also am a fan of the color! I mean red is gross generally (for obvious sports reasons),  but it looks good here!

sjspielman: >  I mean red is gross generally (for obvious sports reasons), but it looks good here!

😂 One reason I like this red since it isn't _as_ blinding as regular `red`, but this is also a good reason haha

sjspielman:
This is now ready for another look! I have merged in the tables code and cleaned everything up to work smoothly. I also updated the description of why these UMAPs are different from the individual library UMAPs. I also made two "house keeping" changes - 

- To match other report files, I changed the extension from `Rmd` -> `rmd`.
- I added `merged-report.html` to the `.gitignore`, as we do for the other reports. 

Here's a rendered example: 
[merged-report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13844340/merged-report.html.zip)

Also noting that once this PR gets merged, I'll add a link to the relevant `TODO` lines we'll need to revisit in #641

templates/merge-report.Rmd
sjspielman: Noting #641 has been opened

templates/merge-report.Rmd
sjspielman: Added phrasing in https://github.com/AlexsLemonade/scpca-nf/pull/636/commits/3093c1f01e3e146de52b7943906bd784752d962b

allyhawkins (approved):
LGTM 🚀

PR #644, opened 2024-01-05
Update UMAP faceting code

### LINKED ISSUE #637: Revisit the facet umap function
See original comment: https://github.com/AlexsLemonade/scpca-nf/pull/635#issuecomment-1868031802

The legend placement is not sufficiently general for the plots we'll be using it for, so we'll need to revisit this code.

- We don't lump cell types for submitters, so the legend needs to be placed "normally" outside the margins
- For singler and cellassign, the positioning may also be dependent on text length

sjspielman:
Closes #637 

This PR cleans up legend placement and makes associated changes for the faceted UMAPs where we highlight 1 cell type in each facet. Here's an overview of the changes:

- I modified the lumping function to also wrap labels to 35 characters, which seems to be about right for the legend "slot" and associated spacing changes I made
  - Note that for the wrapping, I wrote a separate `mutate()` altogether since I think that's a bit more legible. 
  - One wrinkle for wrapping is that, when there are two lines of text in the facet strip label, it slightly change the aspect ratio so it's not quite equal to 1. I don't think this a big deal though? 
- I fixed factor ordering - After running `fct_lump()`, we lost our original factor order and levels became alphabetical. Now, we are back to ordering by frequency, with "Unknown cell type" and "All remaining cell types" as the last two levels
- I modified the legend text size and placement, which is now conditional
  - In a shocking twist of events, I was unable to help myself and did a `%%3` anyways for determining legend placement, but there is still only 1 `if` statement so I think it's fine that I did that thing! I also covered the case of 1 cell type, which _really_ should realistically never happen (the only circumstance I can think of is a method totally fails and assigns everything as unknown), but again, there's only 1 `if` :)
- I enforced three columns for `facet_wrap()` which seemed a little safer to code explicitly given the legend placement choices
- Note that `dplyr::` had to be added into the umap function as described in this commit - 3fa66d04f0e8038ea1997870a1936d07e33f3119

Here are two rendered celltype reports:

- Without submitter: [celltypes_supplemental_report-no_submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13845596/celltypes_supplemental_report-no_submitter.html.zip)

- With (fake) submitter (where there are N=6 submitter cell types): [celltypes_supplemental_report-with_submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13845597/celltypes_supplemental_report-with_submitter.html.zip)

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -155,7 +155,10 @@ faceted_umap <- function(umap_df,
     ) +
     # set points for desired cell type
     geom_point(size = 0.3, alpha = 0.5) +
-    facet_wrap(vars({{ annotation_column }})) +
+    facet_wrap(
+      vars({{ annotation_column }}),
+      ncol = 3
+    ) +
     scale_color_brewer(palette = "Dark2") +
     # remove axis numbers and background grid
     scale_x_continuous(labels = NULL, breaks = NULL) +
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -127,18 +127,14 @@ plot_umap <- function(
 #' Create a faceted UMAP panel where each panel has only one cell type colored
 #'
 #' @param umap_df Data frame with UMAP1 and UMAP2 columns
+#' @param n_celltypes The number of cell types (facets) displayed in the plot
 #' @param annotation_column Column containing cell type annotations
 #'
 #' @return ggplot object containing a faceted UMAP where each cell type is a facet.
 #'   In each panel, the cell type of interest is colored and all other cells are grey.
 faceted_umap <- function(umap_df,
+                         n_celltypes,
                          annotation_column) {
-  # Find total number of cell types for determining legend placement
-  total_celltypes <- umap_df |>
-    dplyr::pull({{ annotation_column }}) |>
-    levels() |>
-    length()
-
   # color by the annotation column but only color one cell type at a time
   faceted_umap <- ggplot(
     umap_df,
@@ -175,7 +171,7 @@ faceted_umap <- function(umap_df,
     )
 
   # Determine legend placement based on total_celltypes
-  if (total_celltypes %% 3 != 0 & total_celltypes != 1) {
+  if (n_celltypes %% 3 != 0 & n_celltypes != 1) {
     faceted_umap <- faceted_umap +
       theme(
         legend.position = c(0.68, 0.33),
@@ -368,31 +364,62 @@ All other cell types are grouped together and labeled "All remaining cell types"
 
 <!-- Now, UMAPs of cell types, where present -->
 
-```{r, eval = has_submitter && has_umap, message=FALSE, warning=FALSE, fig.height = 9, fig.width=9}
+```{r}
+# Function to determine plot height, and width should sort itself out?
+determine_umap_height <- function(n_types) {
+  dplyr::case_when(
+    n_types <= 2 ~ 3,
+    n_types <= 3 ~ 4,
+    n_types <= 5 ~ 6,
+    n_types <= 6 ~ 7,
+    .default = 8
+  )
+}
+```
+
+```{r, eval = has_submitter && has_umap}
+submitter_n_celltypes <- length(levels(umap_df$submitter_celltype_annotation_lumped))
+submitter_height <- determine_umap_height(submitter_n_celltypes)
+```
+
+```{r, eval = has_submitter && has_umap, message=FALSE, warning=FALSE, fig.height = submitter_height, fig.width = 8}
 # umap for cell assign annotations
 faceted_umap(
   umap_df,
+  submitter_n_celltypes,
   submitter_celltype_annotation_lumped
 ) +
   ggtitle("UMAP colored by submitter-provided annotations")
 ```
 
 
-```{r, eval=has_singler && has_umap, message=FALSE, warning=FALSE, fig.height=9, fig.width=9}
+```{r, eval = has_singler && has_umap}
+singler_n_celltypes <- length(levels(umap_df$singler_celltype_annotation_lumped))
+singler_height <- determine_umap_height(singler_n_celltypes)
+```
+
+
+```{r, eval=has_singler && has_umap, message=FALSE, warning=FALSE, fig.height = singler_height, fig.width = 8}
 # umap for cell assign annotations
 faceted_umap(
   umap_df,
+  singler_n_celltypes,
   singler_celltype_annotation_lumped
 ) +
   ggtitle("UMAP colored by SingleR annotations")
 ```
 
 
+```{r, eval = has_cellassign && has_umap}
+cellassign_n_celltypes <- length(levels(umap_df$cellassign_celltype_annotation_lumped))
+cellassign_height <- determine_umap_height(cellassign_n_celltypes)
+```
 
-```{r, eval = has_cellassign && has_umap, message=FALSE, warning=FALSE, fig.height = 9, fig.width=9}
+```{r, eval = has_cellassign && has_umap, message=FALSE, warning=FALSE, fig.height = cellassign_height, fig.width = 8}
 # umap for cell assign annotations
 faceted_umap(
   umap_df,
+  cellassign_n_celltypes,
   cellassign_celltype_annotation_lumped
 ) +
   ggtitle("UMAP colored by CellAssign annotations")
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -219,7 +219,7 @@ determine_umap_dimensions <- function(n_celltypes) {
     n_celltypes == 1 ~ c(3.75, 3.5), # 1: no legend
     n_celltypes == 2 ~ c(5.5, 5), # 2: bottom legend
     n_celltypes == 3 ~ c(8, 5.5), # 3: bottom legend
-    n_celltypes <= 5 ~ c(8, 7), # 4 & 5: inset legend
+    n_celltypes <= 5 ~ c(9, 6), # 4 & 5: inset legend
     n_celltypes <= 6 ~ c(8, 7.5), # 6: bottom legend
     .default = c(10, 9.5) # 7 & 8: inset legend
   )
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -178,17 +178,19 @@ faceted_umap <- function(umap_df,
   if (total_celltypes %% 3 != 0 & total_celltypes != 1) {
     faceted_umap <- faceted_umap +
       theme(
-        legend.position = c(0.95, 0),
-        legend.justification = c(1, -0.1),
+        legend.position = c(0.68, 0.33),
+        legend.justification = c("left", "top"),
         legend.title.align = 0.5,
         # use slightly smaller legend text, which helps legend fit and prevents
         #  long wrapped labels from bunching up
-        legend.text = element_text(size = rel(0.85))
+        legend.text = element_text(size = rel(0.85)),
+        aspect.ratio = 1
       )
   } else {
     faceted_umap <- faceted_umap +
       theme(
-        legend.position = "bottom"
+        legend.position = "bottom",
+        aspect.ratio = 1
       )
   }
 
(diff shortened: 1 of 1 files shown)


jashapiro:
> * One wrinkle for wrapping is that, when there are two lines of text in the facet strip label, it slightly change the aspect ratio so it's not quite equal to 1. I don't think this a big deal though?

Can we add back `aspect.ratio = 1`?

jashapiro:
> > * One wrinkle for wrapping is that, when there are two lines of text in the facet strip label, it slightly change the aspect ratio so it's not quite equal to 1. I don't think this a big deal though?
> 
> Can we add back `aspect.ratio = 1`?

Related to this: I think you want a bit of code to determine the height of these figures based on the # of cell types. (I think maybe that is why you got out of square?)

allyhawkins (dismissed):
All these changes look good to me. I'm fine without the aspect ratio, because when I had it I was seeing some really small UMAPs that didn't look good. I think the scenarios where we have long cell type names > 35 characters is minimal too. But feel free to test it out.

sjspielman:
Thanks for reminder about aspect ratio @jashapiro! Let me play around with this and height...

> I think the scenarios where we have long cell type names > 35 characters is minimal too 

One of the attached reports had this scenario - some of the singler cell type names are reaaallly long.

sjspielman:
>  I'm fine without the aspect ratio, because when I had it I was seeing some really small UMAPs that didn't look good.

Yeah, this is what I'm seeing, eg - 
Screenshot 2024-01-05 at 2 52 22 PM

But more importantly, changing the chunk's plot size doesn't make a difference. It seems we'd have to modify the actual legend coordinates based on the plot's coordinates, which seems like more trouble than it's worth?

sjspielman:
>  I think you want a bit of code to determine the height of these figures based on the # of cell types. (I think maybe that is why you got out of square?)

All the UMAP chunks have height/width of 9, and cellassign + singler both had the same number of cell types (n=8). So, I think it's a product of the strip taking up 2 lines?

allyhawkins:
> But more importantly, changing the chunk's plot size doesn't make a difference. It seems we'd have to modify the actual legend coordinates based on the plot's coordinates, which seems like more trouble than it's worth?

I would agree that I don't think it's worth it and I think you could leave it the way it is. The only other thing I would suggest if you really want to change it is just setting the figure height and width based on the number of cell types.

jashapiro:
> But more importantly, changing the chunk's plot size doesn't make a difference. It seems we'd have to modify the actual legend coordinates based on the plot's coordinates, which seems like more trouble than it's worth?

This is because you kept the bottom right positioning though, right? If the legend were positioned by the top left corner, it would fix this, wouldn't it?

jashapiro:
I actually think keeping the aspect ratio fixed is pretty important, otherwise you get skewing between different plots.

jashapiro:
> All the UMAP chunks have height/width of 9, and cellassign + singler both had the same number of cell types (n=8). So, I think it's a product of the strip taking up 2 lines?

I was specifically thinking of submitter plot where you only have 6. It becomes too long, so those aren't square.

sjspielman:
Ok, I think I've gotten there -

```
  # Determine legend placement based on total_celltypes
  if (total_celltypes %% 3 != 0 & total_celltypes != 1) {
    faceted_umap <- faceted_umap +
      theme(
        legend.position = c(0.68, 0.33),
        legend.justification = c("left", "top"),
        legend.title.align = 0.5,
        # use slightly smaller legend text, which helps legend fit and prevents
        #  long wrapped labels from bunching up
        legend.text = element_text(size = rel(0.85)), 
        aspect.ratio = 1
      )
  } else {
    faceted_umap <- faceted_umap +
      theme(
        legend.position = "bottom", 
        aspect.ratio = 1
      )
  }
```

All of them look -

Screenshot 2024-01-05 at 3 18 38 PM

sjspielman:
Next up, going to modify figure size params.

sjspielman:
I think setting size conditionally is a bit of a pain because we have to both contend with whether the legend is inset or on the bottom, as well as the number of cell types. So before I launch into that, what if we just move to 8x8 all around instead of 9x9...? Just make things a little less domineering.

For example - 
[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13846296/celltypes_supplemental_report.html.zip)

jashapiro:
You can do it in one case statement, something like:
```
fig_height <- dplyr::case_when(
  n_types <= 2 ~ 3, 
  n_types <= 3 ~ 4, 
  n_types <= 5 ~ 6, 
  n_types <= 6 ~ 7, 
  .default = 9
)
```

But maybe still not worth it.

sjspielman:
Getting closer (arrived?). Found a bug too for n=2 while exploring - legend has to be on the bottom here, since even though ncol = 3, there won't be a 3rd panel! I also killed the legend for n=1.

Reports for each N<=6 (see submitter UMAPs):

[1-submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13846710/1-submitter.html.zip)
[2-submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13846703/2-submitter.html.zip)
[3-submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13846704/3-submitter.html.zip)
[4-submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13846705/4-submitter.html.zip)
[5-submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13846706/5-submitter.html.zip)
[6-submitter.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13846707/6-submitter.html.zip)

templates/qc_report/celltypes_qc.rmd
```diff
@@ -150,13 +168,36 @@ faceted_umap <- function(umap_df,
...
+      theme(
+        legend.position = c(0.68, 0.33),
+        legend.justification = c("left", "top"),
+        legend.title.align = 0.5,
+        # use slightly smaller legend text, which helps legend fit and prevents
+        #  long wrapped labels from bunching up
+        legend.text = element_text(size = rel(0.85)),
+        aspect.ratio = 1
+      )
+  } else {
```
allyhawkins, line 191: Can you just add a comment here that this is specifically for when n = 2 that because we have < 3 columns, we need the legend on the bottom? 
Maybe another comment for each of the `if` conditions would be helpful to say why each decision was made here.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -329,31 +370,62 @@ All other cell types are grouped together and labeled "All remaining cell types"
...
-```{r, eval = has_submitter && has_umap, message=FALSE, warning=FALSE, fig.height = 9, fig.width=9}
+```{r}
+# Helper function to determine UMAP plot height based on the number of celltypes
+determine_umap_height <- function(n_celltypes) {
+  dplyr::case_when(
+    n_celltypes == 1 ~ 3.5,
+    n_celltypes <= 3 ~ 4,
+    n_celltypes <= 5 ~ 6,
+    n_celltypes <= 6 ~ 7,
+    .default = 8
```
allyhawkins, line 381: How are you setting width? I think we want it to be square and they are not square right now, even if they don't have long cell type names. 
I think I would go back to the default being the 9x9.

allyhawkins (commented):
My biggest comment is that the 8x8 doesn't look square to me. I would go back to 9x9 as the default so we have square UMAPs, at least in the case of cell type labels that are not on the longer end.

templates/qc_report/celltypes_qc.rmd
sjspielman: > How are you setting width?

I just kept it fixed at 8; because of the aspect ratio, the whole space should only be taken up if needed, is what I saw while exploring. But will bump to 9 and explore a bit more!

sjspielman:
I have explored much more! 

- Width and height are now both dynamically set, using test celltypes that have long names to ensure spacing works out ok
- I aligned figures in the center. At first I tried to left-align them all, but this didn't work as expected which was disappointed. So, center was as close as I got.

Here's a set of reports across number of cell types: [reports-across-ncelltypes.zip](https://github.com/AlexsLemonade/scpca-nf/files/13865290/reports-across-ncelltypes.zip)

templates/qc_report/celltypes_qc.rmd
```diff
@@ -150,15 +168,62 @@ faceted_umap <- function(umap_df,
...
+#'
+#' @param n_celltypes Number of cell types (here, facets) displayed in plot
+#'
+#' @return A vector of c(width, height) in inches
+determine_umap_dimensions <- function(n_celltypes) {
+  dplyr::case_when(
+    n_celltypes == 1 ~ c(3.75, 3.5), # 1: no legend
+    n_celltypes == 2 ~ c(5.5, 5), # 2: bottom legend
+    n_celltypes == 3 ~ c(8, 5.5), # 3: bottom legend
+    n_celltypes <= 5 ~ c(8, 7), # 4 & 5: inset legend
```
allyhawkins, line 222:
```suggestion
    n_celltypes <= 5 ~ c(9, 6), # 4 & 5: inset legend
```
How about this? I think we want to keep the dimensions so they are square. If each panel is 3x3, then these should be the dimensions. That should hopefully get rid of some of the non-square ness that we are seeing.

templates/qc_report/celltypes_qc.rmd
```diff
@@ -150,15 +168,62 @@ faceted_umap <- function(umap_df,
...
+    n_celltypes <= 6 ~ c(8, 7.5), # 6: bottom legend
+    .default = c(10, 9.5) # 7 & 8: inset legend
```
allyhawkins, line 224:
```suggestion
    .default = c(9, 9) # 7 & 8: inset legend
```
I think I would still keep this as 9x9 so that it's square. Any time we have the legend inset I think we can keep it square.

allyhawkins (commented):
I'm a little confused as to why you don't like sticking with making each plot 3x3 so the default is 9x9? Is something not looking right? Because the plots used to look much more square then they do now.

sjspielman:
> I'm a little confused as to why you don't like sticking with making each plot 3x3 so the default is 9x9? Is something not looking right? Because the plots used to look much more square then they do now.

I think the problem is me 😬 ! I'm sure I've been focusing more on overall size than overall squareness... I'll do another round shortly!

sjspielman:
Noting that I downloaded a ruler app which actually works really well, and I measured all the panels for different N's. Each panel is indeed square; `aspect.ratio = 1` is working as expected. Any appearance of non-square-ness is a visual artifact maybe from the facet labels. Either way, I'll make those changes to hopefully get the _overall appearance_ looking more square.

sjspielman:
Some images for your perusal @allyhawkins, let me know how you feel about these dimensions:

### 9"x9" for 3x3 grids, with and without wrapped labels. 

9x9 - wrapped labels
9x9 - not wrapped labels


### Three versions of a 5-facet plot:

#### 8"x7" (WxH)

8x7

#### 9"x7" (WxH)

9x7


#### 9"x6" (WxH)

9x6

allyhawkins:
Okay I think I know why they aren't looking square to me. By using `aspect.ratio = 1`, the actual panel is square, but the addition of the label makes it look elongated. I'm attaching a copy of what was in the report until these changes, and to me they look visually more appealing, but if we want to keep `aspect.ratio = 1`, then I think I would choose the 9x9 and the 9x6 but move the legend up a little bit. 
Screenshot 2024-01-09 at 10 44 56 AM

sjspielman:
> Okay I think I know why they aren't looking square to me. By using aspect.ratio = 1, the actual panel is square, but the addition of the label makes it look elongated.

Yup, that's the culprit! It's the labels.


> but if we want to keep aspect.ratio = 1, then I think I would choose the 9x9 and the 9x6 but move the legend up a little bit.

It seems like we do for overall consistency across all report UMAPs. I'll set those width/heights!

But, in terms of moving the legend up, I agree, but I don't think we can do it without getting overly into the weeds. The reason that the legend is so low in the example "loooong" cell type images I have is because the legend position is based around the _center_ of the legend itself. The fact that all the text is wrapped changes the overall legend dimensions and it ends up being lower. There is no way to get that legend higher while ensuring the _other legends_ in the report don't then overlap with the facet above. Does my description make sense and did I understand your comment correctly?

sjspielman:
Oh wait actually, the positioning is actually fixable - I was wrong about the wrapping being the culprit, it's the fact that there aren't three rows. I think it's ok to add another `if` here. Incoming.

sjspielman:
Ok, how does it look??

[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/13876857/celltypes_supplemental_report.html.zip)

allyhawkins (approved):
Sorry for the 🎢, but looks good to me!

sjspielman:
> Sorry for the 🎢

never apologize for making me stay in `ggplot2` land 😉 !

Issue #646, opened 2024-01-08
Add instructions for running the merging workflow


allyhawkins:
I noticed that we don't have any documentation on the `merge.nf` workflow. We might want to include that at minimum, in the internal instructions. 

We should probably also think about if we want to include a section in the external instructions?

sjspielman:
Noting I'll pick up at least the internal instructions part!

sjspielman:
In DSTM Pre-planning meeting on 1/16/24, we decided we would probably like to do this _before the paper_ goes out, but we don't need it for the cell type release.

sjspielman:
External instructions have been written in #688, so all parts of this issue are set

PR #647, opened 2024-01-09
Add merge workflow internal instructions

sjspielman:
Part of #646

This PR ads some internal instructions for the merged workflow. The workflow is pretty cut-and-dry, so I didn't think of much more to write than the small section I added! Can we think of anything else?

Also, I've never actually _run_ the workflow, but I've made what I hope is the right educated guess for how to run it!

### DIFF
--- a/internal-instructions.md
+++ b/internal-instructions.md
@@ -10,13 +10,18 @@
 - [Maintaining references for `scpca-nf`](#maintaining-references-for-scpca-nf)
   - [Adding additional organisms](#adding-additional-organisms)
   - [Adding additional cell type references](#adding-additional-cell-type-references)
+- [Running the merge workflow](#running-the-merge-workflow)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Running `scpca-nf` as a Data Lab staff member
 
 This section provides instructions for running the main workflow, found in [`main.nf`](main.nf).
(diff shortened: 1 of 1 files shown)


internal-instructions.md
```diff
@@ -199,3 +204,27 @@ Corresponding "trained" model files for use in the cell type annotation workflow
   - `<celltype_ref_name>` is a given reference name established by the Data Lab.
   - `<celltype_ref_source>` is `PanglaoDB`
   - `<date>` is the `PanglaoDB` date, which serves as their version, in ISO8601 format.
+
+
+## Running the merge workflow
+
+The workflow in `merge.nf` merges (but **does not integrate**) processed objects produced by the main `scpca-nf` workflow.
+For a given SCPCA project id, the workflow creates a merged `SCE` object, a merged `AnnData` object, and an associated merged object HTML report.
+Note that the workflow does also merge any present alternative experiments (e.g., ADT data from CITE-seq) _except_ for cell hashing alternative experiments.
```
allyhawkins, line 213:
```suggestion
Note that the workflow will also merge any present alternative experiments (e.g., ADT data from CITE-seq) _except_ for cell hashing alternative experiments.
```

allyhawkins (approved):
One small comment, but otherwise looks good. Thanks for doing this!

PR #649, opened 2024-01-10
Fail for missing mapping files, and other missing files

### LINKED ISSUE #639: Mapping workflow should fail if mapping files are missing
In the event that mapping files (fastq) are not present, then we need to make sure that any workflow run that is meant to include mapping should rather than potentially overwriting previous results.

We should first check the current behavior, making espeically sure that running with `--repeat-mapping` when there are no fastq files in the specified directory does not result in a blank output from alevin mapping steps. If this is not currently the case, we should modify the workflow to correct this.

sjspielman:
Closes #639

This PR adds a whole bunch of checks for files existing, including the original scope of #639 and more spots that _seemed_ like they could use it.

#### Part 1: Mapping files

I changed the `has_rad` and `make_rad` order since that made more sense in my brain. Here's how I set it up in `af-rna.nf`:

- We set up to run mapping when (in order) the following are met. Otherwise we do not run mapping.
  - mapping is requested
  - there is no rad directory
  - the assembly has changed.
- If we _are_ mapping:
  - If fastq do not exist, then this will error out on lines 141-142
- If we are _not_ mapping:
  - There needs to be a rad directory that already exists. If it doesn't exist, then this will error out on line 151.

The one item I'm not sure I fully covered is ensuring the rad directory gets created if we are running mapping _and_ the rad directory does not exist (second condition in my initial list of when we run mapping). Is there something more that needs to be done here? I assumed running `alevin-fry` itself creates it, but am I wrong?

Are there any other logic foibles that I missed?

#### Part 2: The rest

I have a bunch of smaller-scope commits that add `checkIfExists: true` to a bunch of `file()` calls throughout, as well as a few checks at the top of `main.nf` for the existence of the report files. Please have a look at these and advise if they are reasonable to keep or should be reverted! I also "duplicated" the `af-rna.nf` changes in `af-features.nf`. I did not touch the spaceranger module, though, since I'm more hesitant to touch that file in general. Let me know if I should!
Edit* A note that I'm not _super sure_ about these checks for the `?: empty_file` places. I think maybe they are wrong, but put them out there for more eyes in case they are right!

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -112,7 +112,7 @@ workflow {
     log.info("Executing workflow for all runs in the run metafile.")
   }
 
-  ref_paths = Utils.readMeta(file(params.ref_json))
+  ref_paths = Utils.readMeta(file(params.ref_json, checkIfExists: true))
 
   unfiltered_runs_ch = Channel.fromPath(params.run_metafile)
     .splitCsv(header: true, sep: '\t')
@@ -223,7 +223,7 @@ workflow {
       single: true
     }
   // apply cellhash demultiplexing
-  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file))
+  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file, checkIfExists: true))
   merged_sce_ch = cellhash_demux_ch.mix(feature_sce_ch.single)
 
   // join SCE outputs and branch by genetic multiplexing
--- a/main.nf
+++ b/main.nf
@@ -51,7 +51,12 @@ if (!file(params.run_metafile).exists()) {
   param_error = true
 }
 
-sample_metafile = file(params.sample_metafile)
+sample_metafile = file(params.sample_metafile) // we make this for passing into later processes
+if (!sample_metafile.exists()) {
+  log.error("The 'sample_metafile' file '${params.sample_metafile}' can not be found.")
+  param_error = true
+}
+
 if (!sample_metafile.exists()) {
   log.error("The 'sample_metafile' file '${params.sample_metafile}' can not be found.")
   param_error = true
@@ -73,6 +78,17 @@ if (params.perform_celltyping) {
     log.error("The 'celltype_ref_metadata' file '${params.celltype_ref_metadata}' can not be found.")
     param_error = true
   }
+
+  if (!file("${projectDir}/templates/qc_report/${celltype_report_template_file}").exists()) {
+    log.error("The 'celltype_report_template_file' file '${celltype_report_template_file}' can not be found.")
+    param_error = true
+  }
+}
+
+// QC report check
+if (!file("${projectDir}/templates/qc_report/${report_template_file}").exists()) {
+  log.error("The 'report_template_file' file '${report_template_file}' can not be found.")
+  param_error = true
 }
 
 if(param_error){
--- a/main.nf
+++ b/main.nf
@@ -118,7 +118,7 @@ workflow {
     log.info("Executing workflow for all runs in the run metafile.")
   }
 
-  ref_paths = Utils.readMeta(file(params.ref_json, checkIfExists: true))
+  ref_paths = Utils.readMeta(file(params.ref_json))
 
   unfiltered_runs_ch = Channel.fromPath(params.run_metafile)
     .splitCsv(header: true, sep: '\t')
--- a/main.nf
+++ b/main.nf
@@ -25,7 +25,7 @@ citeseq_techs = single_cell_techs.findAll{it.startsWith('CITEseq')}
 cellhash_techs = single_cell_techs.findAll{it.startsWith('cellhash')}
 
 // report template paths
-report_template_dir = file("${projectDir}/templates/qc_report", type: 'dir')
+report_template_dir = file("${projectDir}/templates/qc_report", type: 'dir', checkIfExists: true)
 report_template_file = "main_qc_report.rmd"
 celltype_report_template_file = "celltypes_supplemental_report.rmd"
 report_template_tuple = tuple(report_template_dir, report_template_file, celltype_report_template_file)
--- a/main.nf
+++ b/main.nf
@@ -68,6 +68,17 @@ if (!resolution_strategies.contains(params.af_resolution)) {
   param_error = true
 }
 
+if (params.cellhash_pool_file && !file(params.cellhash_pool_file).exists()){
+  log.error("The 'cellhash_pool_file' file ${cellhash_pool_file} can not be found.")
+  param_error = true
+}
+
+// QC report check
+if (!file("${projectDir}/templates/qc_report/${report_template_file}").exists()) {
+  log.error("The 'report_template_file' file '${report_template_file}' can not be found.")
+  param_error = true
+}
+
 // cell type annotation file checks
 if (params.perform_celltyping) {
   if (!file(params.project_celltype_metafile).exists()) {
@@ -85,11 +96,6 @@ if (params.perform_celltyping) {
   }
 }
 
-// QC report check
-if (!file("${projectDir}/templates/qc_report/${report_template_file}").exists()) {
-  log.error("The 'report_template_file' file '${report_template_file}' can not be found.")
-  param_error = true
-}
 
 if(param_error){
   System.exit(1)
@@ -223,7 +229,7 @@ workflow {
       single: true
     }
   // apply cellhash demultiplexing
-  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file, checkIfExists: true))
+  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file))
   merged_sce_ch = cellhash_demux_ch.mix(feature_sce_ch.single)
 
   // join SCE outputs and branch by genetic multiplexing
--- a/modules/af-features.nf
+++ b/modules/af-features.nf
@@ -145,7 +145,7 @@ workflow map_quant_feature{
     feature_barcodes_ch = feature_channel
       .map{meta -> tuple(
         meta.feature_barcode_file,
-        file("${meta.feature_barcode_file}")
+        file("${meta.feature_barcode_file}", checkIfExists: true)
       )}
       .unique()
     index_feature(feature_barcodes_ch)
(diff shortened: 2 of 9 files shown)


modules/af-features.nf
```diff
@@ -160,13 +160,16 @@ workflow map_quant_feature{
...
-          && file(it.feature_rad_dir).exists()
-          && Utils.getMetaVal(file("${it.feature_rad_dir}/scpca-meta.json"), "ref_assembly") == "${it.ref_assembly}"
+        make_rad: (
+          // repeat has been requested
+          params.repeat_mapping
+          // the feature directory does not exist
+          || !file(it.feature_rad_dir).exists()
+          // the assembly has changed; if feature_dir doesn't exist, this line won't get hit
+          || Utils.getMetaVal(file("${it.feature_rad_dir}/scpca-meta.json"), "ref_assembly") != "${it.ref_assembly}"
         )
```
jashapiro, line 170: We don't really need to check the reference version here, as the feature files don't use it (they just use the `feature_barcode_file`). This was probably some overzealous copying.

We do still want to skip if the input files don't exist though. Maybe this works?

```suggestion
        make_rad: (
          // input files exist
          file($it.files_directory, type: "dir").exists() && (
            // and repeat has been requested
            params.repeat_mapping
            // or the feature rad file directory does not exist
            || !file(it.feature_rad_dir).exists()
          )
        )
```

modules/af-rna.nf
```diff
@@ -120,31 +120,35 @@ workflow map_quant_rna {
...
-          && file(it.rad_dir).exists()
-          && Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "ref_assembly") == "${it.ref_assembly}"
+        make_rad: (
+          // repeat has been requested
+          params.repeat_mapping
+          // the rad directory does not exist
+          || !file(it.rad_dir).exists()
+          // the assembly has changed; if rad_dir doesn't exist, this line won't get hit
+          || Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "ref_assembly") != "${it.ref_assembly}"
         )
```
jashapiro, line 132: Here too we want to make sure the input files exist  before doing remapping.
```suggestion
        make_rad: (
          // input files exist
          file($it.files_directory, type: "dir").exists() && (
            // and repeat has been requested
            params.repeat_mapping
            // the rad directory does not exist
            || !file(it.rad_dir).exists()
            // the assembly has changed; if rad_dir doesn't exist, this line won't get hit
            || Utils.getMetaVal(file("${it.rad_dir}/scpca-meta.json"), "ref_assembly") != "${it.ref_assembly}"
          )
        )
```

modules/af-rna.nf
```diff
@@ -120,31 +120,35 @@ workflow map_quant_rna {
...
-        make_rad: true
-       }
+        has_rad: true
```
jashapiro, line 133: I wonder if we want to be more explicit careful here. We can check that the rad dir exists here, and quietly fail if it doesn't by putting the files with no input files or rad dir into a separate branch. 

We could then do something with `view` or some other logging mechanism that I haven't explored to flag anything that ends up in this branch. It _should_ be nothing, but it would be nice to know that!
```suggestion
        has_rad: file(it.rad_dir).exists()
        missing_inputs: true
```

jashapiro (commented):
Looks partway there. I'd guess most of the `checkIfExists` you added here aren't really needed, but they shouldn't hurt anything. 

The main thing I was thinking about was the case where we call `--repeat-mapping` and the input files are not there, so I suggested checking for input files at the top of the `make_rad` branch. But then I had a fear that would fail, dropping us straight to `has_rad`, even if the rad file doesn't exist, so I thought maybe we could make a separate branch for files with neither. What to do with that branch I left open: we might just add a `.view()` with an appropriate message, or we could look for a better way to log those situations. The advantage of the setup I am suggesting as that we should not get full workflow failures, just libraries that don't proceed through the rest of the workflow.

modules/af-features.nf
sjspielman: Noting the overzealous copying wasn't me 😉 

 https://github.com/AlexsLemonade/scpca-nf/blame/fbf8e84bac065d6e7cbbf72f685fc0e680408f8e/modules/af-features.nf#L166

main.nf
```diff
@@ -207,7 +223,7 @@ workflow {
       single: true
     }
   // apply cellhash demultiplexing
-  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file))
+  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file, checkIfExists: true))
```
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/649/commits/21cfebb8127b1f70591452afb148047273637d4e

main.nf
```diff
@@ -73,6 +78,17 @@ if (params.perform_celltyping) {
...
+    log.error("The 'celltype_report_template_file' file '${celltype_report_template_file}' can not be found.")
+    param_error = true
+  }
+}
+
+// QC report check
+if (!file("${projectDir}/templates/qc_report/${report_template_file}").exists()) {
+  log.error("The 'report_template_file' file '${report_template_file}' can not be found.")
+  param_error = true
 }
```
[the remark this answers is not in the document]
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/649/commits/21cfebb8127b1f70591452afb148047273637d4e

modules/af-rna.nf
jashapiro: Followup thought: We can probably use `.subscribe` after this to write to stderr anything that ends up in the missing inputs branch
Something like:
```
rna_channel.missing_inputs.subscribe{
  log.error("The expected input fastq or rad files for run ${it.run_id} are missing.")
}
```

sjspielman: Can super confirm that, surprisingly literally no one, `.view()` won't work to send anything into the log, so I'll try this now!

sjspielman: This was giving 1 line/run id, but this is all in one go:

```
    missing_ch = rna_channel.missing_inputs
      .map{it.run_id}
      .toList()

    missing_ch.subscribe{
      log.error("The expected input fastq or rad files for the following runs are missing: ${it}.")
    }
```

sjspielman: I'd like to find a way to only do this if there _are_ ids though. Otherwise we get an error even if none are missing! This does seem like an error though, so `log.info` might not be the answer.

sjspielman: Also noting that the `STUBR16` is not getting caught like you thought it might?

main.nf
```diff
@@ -96,7 +112,7 @@ workflow {
     log.info("Executing workflow for all runs in the run metafile.")
   }
 
-  ref_paths = Utils.readMeta(file(params.ref_json))
+  ref_paths = Utils.readMeta(file(params.ref_json, checkIfExists: true))
```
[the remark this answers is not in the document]
sjspielman: removed https://github.com/AlexsLemonade/scpca-nf/pull/649/commits/c4260ecdd6a91ea8542468ff5a9d9a7b4af0ffad

modules/af-rna.nf
jashapiro: > I'd like to find a way to only do this if there _are_ ids though. Otherwise we get an error even if none are missing! This does seem like an error though, so `log.info` might not be the answer.

I think I liked one error line per run, and with that mode there will be no error written if there is nothing in the channel. 

But if you _do_ want write only one line, something like this should work (zero length list is false):
```
.subscribe{
  if(it){
    log.error(...)
  }
}
```

(I'll also note that you don't need to save anything to an output channel: no need for `missing_ch = `)

sjspielman: > no need for missing_ch = 

Ah yes thanks, that was left over from a previous attempt at a different approach!

sjspielman: > I think I liked one error line per run, and with that mode there will be no error written if there is nothing in the channel.

I prefer this!

modules/af-rna.nf
```diff
@@ -120,31 +120,47 @@ workflow map_quant_rna {
...
+        has_rad: file(it.rad_dir).exists()
+        missing_inputs: true
+      }
+
+    // send run ids in rna_channel.missing_inputs to log
+    rna_channel.missing_inputs
+      .subscribe{
+        if(it){
+          log.error("The expected input fastq or rad files for ${it.run_id} are missing.")
+        }
```
jashapiro, line 145:
```suggestion
        log.error("The expected input fastq or rad files for ${it.run_id} are missing.")
```

modules/af-features.nf
```diff
@@ -185,8 +198,8 @@ workflow map_quant_feature{
     // create tuple of metdata map (read from output) and rad_directory to be used directly as input to alevin-fry quantification
     feature_rad_ch = feature_ch.has_rad
       .map{meta -> tuple(
-        Utils.readMeta(file("${meta.feature_rad_dir}/scpca-meta.json")),
-        file(meta.feature_rad_dir, type: 'dir')
+        Utils.readMeta(file("${meta.feature_rad_dir}/scpca-meta.json", checkIfExists: true)),
```
jashapiro, line 201: Just noting that in [c4260ec](https://github.com/AlexsLemonade/scpca-nf/pull/649/commits/c4260ecdd6a91ea8542468ff5a9d9a7b4af0ffad) you got rid of one of these `checkIfExists`, and you could presumably get rid of any of them  that are within a `readMeta()` or `readMetaVal()`, but it isn't really much of a concern either way.

modules/af-rna.nf
```diff
@@ -120,31 +120,47 @@ workflow map_quant_rna {
         meta.barcode_file = "${params.barcode_dir}/${params.cell_barcodes[meta.technology]}";
         meta // return modified meta object
       }
-       // split based in whether repeat_mapping is false and a previous dir exists
+       // branch based on whether mapping should be run (make_rad) or skipped (has_rad)
```
jashapiro, line 123:
```suggestion
       // branch based on whether mapping should be run (make_rad) or skipped (has_rad).
       // if neither fastq or rad dir are present, run goes into missing_inputs branch
```

modules/classify-celltypes.nf
```diff
@@ -188,13 +188,13 @@ workflow annotate_celltypes {
     // creates [meta, processed sce, singler model file]
     singler_input_ch = celltype_input_ch
       // add in singler model or empty file
-      .map{it.toList() + [file(it[0].singler_model_file ?: empty_file)]}
+      .map{it.toList() + [file(it[0].singler_model_file ?: empty_file, checkIfExists: true)]}
```
jashapiro, line 191: I just want to note that I expect the `?: empty_file` to work fine here (since you did ask in your original PR comment).

jashapiro (approved):
LGTM! I just had a couple of very minor suggestions, and a confirmation comment. Should not need to see this again.

I see at https://github.com/AlexsLemonade/scpca-nf/actions/runs/7489840073/job/20387333860#step:4:9 that we are getting the output we expect for STUB16, which is great.

modules/af-features.nf
sjspielman: If I did it, I should do it! Cleaning up by finding spots with:

```
grep "readMeta" *nf | grep "checkIfExists" 
grep "getMetaVal" *nf | grep "checkIfExists" 
```

and removing the `checkIfExists`.

PR #651, opened 2024-01-11
Make sure dimensions for UMAPs exist

allyhawkins:
When trying to generate the test data, I came across this error in the cell type report, where the dimensions set in the chunk options don't always exist. For each cell type annotation method, we have a chunk that is evaluated _only_ if that method exists. Here was where we were defining the dimensions to use to create the UMAP in the following chunk. The problem with this approach is that if a method doesn't exist, then the dimensions don't get assigned to a variable. Then in the chunk where we create the UMAP we define the figure height and width with variables that don't exist. So here I'm just setting all the dimensions to a dummy set of dimensions in the event that a cell type method wasn't used. That way the variable exists when going to the UMAP chunk, even if that chunk doesn't get evaluated.

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -401,9 +401,14 @@ All other cell types are grouped together and labeled "All remaining cell types"
 
 <!-- Now, UMAPs of cell types, where present -->
 
-```{r, eval = has_submitter && has_umap}
-submitter_n_celltypes <- length(levels(umap_df$submitter_celltype_annotation_lumped))
-submitter_dims <- determine_umap_dimensions(submitter_n_celltypes)
+```{r}
+if (has_submitter & has_umap) {
+  submitter_n_celltypes <- length(levels(umap_df$submitter_celltype_annotation_lumped))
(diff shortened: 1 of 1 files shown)


sjspielman (approved):
Seems fine to me!

Though, a less verbose way to do this could be something like...

```
# before anything else
umap_dims <- c(1,1)
```

And updating those values with submitter, singler, and cellassign for each subsequent chunk. We'd then change all the chunk params to `fig.width = umap_dims[1], fig.height = umap_dims[2]`. In other words, there would not be `submitter_dims`, `singler_dims`, or `cellassign_dims` variables anymore.

I don't really know which is better. I love the verbosity of variable names to make sure things are in the right spot, but the option I suggested above reduces the amount of code. So...pick your favorite! 

Either way, happy to approve now since this code works too!

allyhawkins:
> I don't really know which is better. I love the verbosity of variable names to make sure things are in the right spot, but the option I suggested above reduces the amount of code. So...pick your favorite!

Hmmm I think I like being explicit since the dims could be different based on the number of cell types. So instead of overwriting them, we should name them separately.

PR #652, opened 2024-01-11
Fix merge script join bug

sjspielman:
While running the merge workflow to help prepare to write docs, it had an error in the merge script. Here, I've found and fixed the bug!

The `dplyr::left_join()` code was using `relationship = "one-to-one"`, but this means _only 1 row_ should match between data frames. Instead, we expect one row in `y` to match _many rows_ in `x`. So, I changed this to `relationship = "many-to-one"` and (edit, typo) the script to works now locally!
Here are `left_join` docs for reference:  https://dplyr.tidyverse.org/reference/mutate-joins.html

I'm re-running the workflow now to double check it works, but wanted to get this filed since I have a quick appointment to head out for (be back in about an hour or so).

Noting this original context for the dplyr code: https://github.com/AlexsLemonade/scpca-nf/pull/630#discussion_r1439704871

Also, I fixed a little purrr code to return a logical vector so R doesn't throw a warning about coercing it from a list to logical vector.


bin/merge_sces.R
```diff
@@ -165,7 +165,7 @@ library_df <- names(input_sce_files) |>
 # join tech and EFO with colData
 colData(merged_sce) <- colData(merged_sce) |>
   as.data.frame() |>
-  dplyr::left_join(library_df, by = c("library_id"), relationship = "one-to-one") |>
+  dplyr::left_join(library_df, by = c("library_id"), relationship = "many-to-one") |>
```
jashapiro, line 168: My bad! many to one does seem like it should be the correct relationship for this join!

sjspielman: I really does! I was surprised at first why it failed!

sjspielman:
Everyone ignore https://github.com/AlexsLemonade/scpca-nf/pull/652/commits/9979c1c276349cbf27f097584aecdaa8c434020d and https://github.com/AlexsLemonade/scpca-nf/pull/652/commits/4deee3761eee9a9c8b5ecff2998978cdb8ffb719 🥴

allyhawkins (dismissed):
Thanks for catching this 🚀

sjspielman:
> Thanks for catching this 🚀

There are more bugs, testing fixes now... you'll be sad... But next commit is nearly incoming!

sjspielman:
Two more fixes in https://github.com/AlexsLemonade/scpca-nf/pull/652/commits/14ed3643a71433d6360e720e8494bf4cda1ef3b6 and https://github.com/AlexsLemonade/scpca-nf/pull/652/commits/1788591a30bf9ed6580250f8fb00347e8f88bf57. In the main QC report, the SCEs we pass in are actual SCEs, but not so here! Have to read in.

sjspielman:
Noting that `SCPCP000001` ran to completion, but with this warning. Looks like the filter code to remove multiplex isn't working since there are two `false`'s?
```
WARN: Input tuple does not match input set cardinality declared by process `export_anndata` -- offending value: [/nextflow-ccdl-data/work/f5/e77aefb0863692784f13d98712bd86/SCPCP000001_merged.rds, SCPCP000001, false, false]
```

sjspielman:
I think we can change...
```
    // export merged objects to AnnData
    anndata_ch = merge_sce.out
      .filter{!it[3]} // remove multiplexed samples before export
      .take(3) // keep everything but multiplexed
``` 

---> 


```
    // export merged objects to AnnData
    anndata_ch = merge_sce.out
      .take(3) // keep everything but multiplexed
```

since multiplexed is the 4th anyways?

`merge_sce` emits: 

```
tuple path(merged_sce_file), val(merge_group_id), val(has_adt), val(multiplexed)
```

sjspielman:
Ok, `take` doesn't seem to be working. I don't know what the deal there is? Going to come back to this tomorrow!


```
merge_sce.out.view()

// export merged objects to AnnData
anndata_ch = merge_sce.out
     .take(3) // keep everything but multiplexed

anndata_ch.view()
```

--->     
```
[/nextflow-ccdl-data/work/eb/4115cff62aee5e5279760282a3a037/SCPCP000001_merged.rds, SCPCP000001, false, false]

[/nextflow-ccdl-data/work/eb/4115cff62aee5e5279760282a3a037/SCPCP000001_merged.rds, SCPCP000001, false, false]
```

jashapiro:
> Ok, `take` doesn't seem to be working. I don't know what the deal there is? Going to come back to this tomorrow!
> 
> ```
> merge_sce.out.view()
> 
> // export merged objects to AnnData
> anndata_ch = merge_sce.out
>      .take(3) // keep everything but multiplexed
> 
> anndata_ch.view()
> ```


I think the issue is that `.take()` in the context here is acting on the channel, so it is taking the first three channel elements, not the first three elements of the list. I think we want `.map{it.take(3)}` 

So: 
```
 // export merged objects to AnnData
    anndata_ch = merge_sce.out
      .filter{!it[3]} // remove multiplexed samples before export
      .map{it.take(3)} // keep everything but multiplexed
```

sjspielman:
Ok, this approach definitely works:
```
 // export merged objects to AnnData
    anndata_ch = merge_sce.out
      .filter{!it[3]} // remove multiplexed samples before export
      .map{it.take(3)} // keep everything but multiplexed
```

But I'm not understanding why we need both `filter` and `take`?

The "before/after" is the same with and without `filter`. My sense was `filter` was left over from before we added `take`, but is there something I'm missing?

```
[/nextflow-ccdl-data/work/86/5d9a4c0a34d1b2d51702a6b64d2b3f/SCPCP000001_merged.rds, SCPCP000001, false, false]
[/nextflow-ccdl-data/work/86/5d9a4c0a34d1b2d51702a6b64d2b3f/SCPCP000001_merged.rds, SCPCP000001, false]
```

jashapiro:
> But I'm not understanding why we need both `filter` and `take`?
> 
> The "before/after" is the same with and without `filter`. My sense was `filter` was left over from before we added `take`, but is there something I'm missing?

`filter` is removing the channel elements that are multiplexed. `map{take}` is removing the 4th element from the non-multiplexed samples for the next step.

sjspielman:
thanks @jashapiro !

sjspielman:
`SCPCP000001` has run to completion, but `SCPCP000003` is still chugging along. It's exported the RDS, but still waiting on anndata because files are BIG. But I'm feeling confident that bugs have all been squashed, so going to re-request review now!

merge.nf
```diff
@@ -171,7 +171,8 @@ workflow {
     // export merged objects to AnnData
     anndata_ch = merge_sce.out
       .filter{!it[3]} // remove multiplexed samples before export
-      .take(3) // keep everything but multiplexed
+      .map{it.take(3)} // keep everything but multiplexed
```
allyhawkins, line 174:
```suggestion
      .map{it.take(3)} // keep everything but multiplexed logical
```
Just something to indicate that here we are moving the logical, not actual multiplexed samples

allyhawkins (approved):
Looks good, just one comment update!

PR #656, opened 2024-01-17
Keep cell type columns when merging

### LINKED ISSUE #655: Keep cell type annotation coldata in the merge workflow
Currently, the merge workflow only retains columns as specified in the default `scpcaTools::merge_sce_list()` argument `retain_coldata_cols`: https://github.com/AlexsLemonade/scpcaTools/blob/2ebdc4f4dfc4233fad97805f9a9a5e3bc6919f1e/R/merge_sce_list.R#L50C5-L60 (as well as any altExp QC stats columns). 

We will need to ensure that cell type annotation columns are also retained. We can do this by adding relevant columns to the `retain_coldata_cols`, _as well as the current default column names_, argument when calling this function from the merge script. We'll need to keep the following columns if each of these is present:

- submitter
  - `submitter_celltype_annotation`
  - note also that, if at least one library has submitter values, we'll want to populate this column in all other libraries with `"Submitter excluded"`
- SingleR
  - `singler_celltype_annotation`  
  - `singler_celltype_ontology`, only if `params.label_type` is ontology
- CellAssign
  - `cellassign_celltype_annotation`
  - `cellassign_max_prediction`

sjspielman:
Closes #655 

Now in the correct repo! This PR updated `merge_sces.R` to handle cell type columns better:

- Find all the cell types that are present in _any_ SCE
- Define a baseline (taken from `scpcaTools` default) set of columns to keep as `retain_coldata_columns`
- For each cell type that is present in at least 1 SCE, add the relevant cell type columns into `retain_coldata_columns`
  - For SingleR, I only add `singler_celltype_ontology` if ontology labels were used for at least one library 
- Finally, tell `merge_sce_list` about the `retain_coldata_columns` vector
 
Moreover, here's how I handle libraries without the given cell type method, but where that method exists in other project libraries:

- Submitter
  - Add the column `submitter_celltype_annotation` with the value `"Submitter-excluded"`
- SingleR
  - Add the column  `singler_celltype_annotation` with the value `"Cell type annotation not performed`"
  - If ontologies were used, then I do the same for `singler_celltype_ontology`
- CellAssign
  - Add the column  `cellassign_celltype_annotation` with the value `"Cell type annotation not performed`"
  - Add the column  `cellassign_max_prediction` with the value `NA_real_`, just to be safe. This is probably overkill since the merge function will add this column with `NA`s anyways, but I like that it's specifically `NA_real_` here

I've tested the code with an SCE list where some, but not all, have cell types and it works as expected! 

Let me know what you think of some of those string choices I made. I'll note the conclusions from this PR over in `scpca-docs` in the issue about merged contents - https://github.com/AlexsLemonade/scpca-docs/issues/219

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -142,6 +142,15 @@ if ("submitter" %in% all_celltypes) {
     retain_coldata_columns,
     "submitter_celltype_annotation"
   )
+
+  # Add `"Submitter-excluded"` value to any libraries without submitter
+  sce_list <- sce_list |>
+    purrr::map(\(sce){
+      if (!"submitter_celltype_annotation" %in% names(sce(colData))) {
+        colData(sce)$submitter_celltype_annotation <- "Submitter-excluded"
+      }
+      return(sce)
+    })
 }
 if ("singler" %in% all_celltypes) {
   # Check if the label used for annotation was ontology in at least 1 SCE
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -146,7 +146,7 @@ if ("submitter" %in% all_celltypes) {
   # Add `"Submitter-excluded"` value to any libraries without submitter
   sce_list <- sce_list |>
     purrr::map(\(sce){
-      if (!"submitter_celltype_annotation" %in% names(sce(colData))) {
+      if (!"submitter" %in% metadata(sce)$celltype_methods) {
         colData(sce)$submitter_celltype_annotation <- "Submitter-excluded"
       }
       return(sce)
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
I think this is fine. At first I wasn't sure that we need the full text of "Cell type annotation not performed" and we should just leave as NA, but maybe it's good to be explicit. This also distinguishes from `SingleR` NA results. The only scenario where I expect this to even be an issue is with a project that has a mix of tissue and cell line samples.

sjspielman:
> This also distinguishes from SingleR NA results.

This is the specific case I had in mind! 

> The only scenario where I expect this to even be an issue is with a project that has a mix of tissue and cell line samples.

Definitely agreed, I think it's unlikely in any other circumstance... unless, like, CellAssign simply refuses to run...

PR #657, opened 2024-01-17
Remove cluster metadata from SCE before merging

sjspielman:
Does what it says!
Since we aren't keeping clusters, we should not keep the cluster parameters. 

This PR was brought to you by me actually writing docs 💪 and realizing it would be super weird to document this.

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -206,7 +206,9 @@ if ("cellassign" %in% all_celltypes) {
 
 # Merge SCEs -------------------------------------------------------------------
 
-# Add a new column with any additional modalities
+# Update some SCE information:
+# - Add a new colData column with any additional modalities
+# - Remove cluster parameters from metadata
 sce_list <- sce_list |>
   purrr::map(\(sce){
     # value will be adt, cellhash, or NA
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
🐈 🐈

sjspielman:
@allyhawkins before I merge this, I might actually want to make a few small other changes I'm noticing while writing docs. What do you think of doing these?

- We use the metadata field `merged_hvgs` but the individual object is `highly_variable_genes`. Should we change this to be `merged_highly_variable_genes` for consistency?
- Looks like we have _named_ vector for `library_id` and `sample_id`. Should we unname?
  ```
  > metadata(sce)$library_id
    SCPCL000001   SCPCL000002   SCPCL000015 
  "SCPCL000001" "SCPCL000002" "SCPCL000015" 
  > metadata(sce)$sample_id
    SCPCL000001   SCPCL000002   SCPCL000015 
  "SCPCS000001" "SCPCS000002" "SCPCS000015" 
  ```

allyhawkins:
> We use the metadata field merged_hvgs but the individual object is highly_variable_genes. Should we change this to be merged_highly_variable_genes for consistency?

I don't have strong feelings about this one, but the longer one is more explicit so that's fine with me. 

> Looks like we have named vector for library_id and sample_id. Should we unname?

This stems from how we merge the metadata using the `scpcaTools` function for merging SCE's. I don't think it's a big deal to keep it and I also don't think specifying named vs. unnamed is important to mention in the docs. So if we change it later we don't have to worry about maintaining that in docs.

sjspielman:
Can I get another 🐈 ? https://github.com/AlexsLemonade/scpca-nf/pull/657/commits/abb2a53ce613fdf622e5d05af1bb9b31fdc20949

allyhawkins (approved):
More 🐈 🐈 (by the way I am very sad github doesn't seem to recognize business cat?!)

PR #658, opened 2024-01-18
Retain filter columns when merging

sjspielman:
This PR is related to this comment https://github.com/AlexsLemonade/scpca-docs/pull/237#discussion_r1457593604:

> And found something else we need to fix! Looks like merged objects are missing `scpca_filter` and `adt_scpca_filter` (if at least 1 library has cite). I feel pretty sure we want to keep those columns around! I'll file a PR in `scpca-nf` to retain those columns.

Here is that PR! Script runs locally ✅

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -165,8 +165,8 @@ if ("singler" %in% all_celltypes) {
     purrr::map(\(sce){
       metadata(sce)$singler_reference_label == "label.ont"
     }) |>
-    # avoid warning with unlist; can't use map_lgl above or reduce(any) below
-    #  the mapping would always need to return length 1
+    # avoid warning with unlist; can't use map_lgl since ^ would always need to
+    #  return length 1
     unlist() |>
     any()
 
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -116,7 +116,7 @@ if (!all(sce_checks)) {
   )
 }
 
-# Determine colData columns to retain  -----------------------------------------
+# Determine cell type annotation columns to retain  -----------------------------------------
 
 
 # Default vector of colData columns to retain from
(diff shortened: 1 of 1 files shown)


bin/merge_sces.R
```diff
@@ -136,8 +130,33 @@ retain_coldata_columns <- c(
...
+adt_present <- sce_list |>
+  purrr::map(\(sce) {
+    "adt_scpca_filter" %in% names(colData(sce))
+  }) |>
+  purrr::reduce(any)
+
+if (adt_present) {
+  retain_coldata_columns <- c(retain_coldata_columns, "adt_scpca_filter")
+}
+
```
allyhawkins, line 150: I think to simplify this you could just move this line `retain_coldata_columns <- c(retain_coldata_columns, "adt_scpca_filter"` to be within the block starting on 231 like so: 

```
sce_list <- sce_list |>
  purrr::map(\(sce){
    # value will be adt, cellhash, or NA
    additional_modalities <- altExpNames(sce)
    if (length(additional_modalities) == 0) {
      additional_modalities <- NA
    } 
    sce$additional_modalities <- additional_modalities

    if("adt" %in% additional_modalities){
       retain_coldata_columns <- c(retain_coldata_columns, "adt_scpca_filter")
    }

    metadata(sce)$cluster_algorithm <- NULL
    metadata(sce)$cluster_weighting <- NULL
    metadata(sce)$cluster_nn <- NULL

    return(sce)
  })
```

allyhawkins: Then you could get rid of all the `adt_present` checks.

allyhawkins (commented):
I suggested just a one line addition that should help simplify what you have.

bin/merge_sces.R
sjspielman: good call!

PR #664, opened 2024-01-22
Skip sample metadata addition to colData if cellhash is present

allyhawkins:
When running the multiplexed samples through the workflow, I was getting the following error: 

```
Command error:
  Warning message:
  replacing previous import ‘S4Arrays::makeNindexFromArrayViewport’ by ‘DelayedArray::makeNindexFromArrayViewport’ when loading ‘SummarizedExperiment’ 
  Error in scpcaTools::metadata_to_coldata(sce, join_columns = "library_id") : 
    The specified `join_columns` are producing multiple matches, but only one match is allowed.
  Calls: format_czi -> <Anonymous>
  Execution halted

```

This is because we are using the `sce_to_anndata.R` script to convert all objects to AnnData prior to running `CellAssign`. In a separate process, we use this same script to do the conversion to produce the final AnnData output. However, in that process, we are skipping any multiplexed samples and when running `CellAssign`, we are keeping multiplexed samples. The issue here is that we have a step in the `format_czi` function that merges the `colData` with the `sample_metadata` using `library_id` as the join column. This won't work when trying to convert an SCE that contains multiplexed data since the `library_id` will match up with multiple samples. 

To get around this, I added a check in the `format_czi` function to make sure that `cellhash` is not an `altExp` found in the SCE object. If so, no sample metadata is added to the `colData`. I also updated the check for converting `altExp` objects to only convert if the feature name was not `cellhash`.

### DIFF
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -165,9 +165,9 @@ if (!is.null(opt$feature_name)) {
       # warn that the altExp cannot be converted
       warning(
         glue::glue("
-        Only 1 row found in altExp named: {opt$feature_name}.
-        This altExp will not be converted to an AnnData object.
-      ")
+          Only 1 row found in altExp named: {opt$feature_name}.
+          This altExp will not be converted to an AnnData object.
+        ")
       )
     }
   }
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -128,40 +128,46 @@ scpcaTools::sce_to_anndata(
 # AltExp to AnnData -----------------------------------------------------------
 
 # if feature data exists, grab it and export to AnnData
-if (!is.null(opt$feature_name) & (opt$feature_name != "cellhash")) {
-  # make sure the feature data is present
-  if (!(opt$feature_name %in% altExpNames(sce))) {
-    stop("feature_name must match name of altExp in provided SCE object.")
-  }
-
-  # check for output file
-  if (!(stringr::str_ends(opt$output_feature_h5, ".hdf5|.h5"))) {
-    stop("output feature file name must end in .hdf5 or .h5")
-  }
-
-  # extract altExp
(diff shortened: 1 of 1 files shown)


jashapiro (commented):
Looks good overall, but I suggested a tiny restructure to reduce nesting and I think you have the test backward?

allyhawkins:
I updated the order of the `if` statement and then used `warning` for both warning messages. This should be ready for another look.

bin/sce_to_anndata.R
```diff
@@ -129,36 +132,43 @@ if (!is.null(opt$feature_name)) {
...
+        alt_sce,
+        anndata_file = opt$output_feature_h5
+      )
+    } else {
+      # warn that the altExp cannot be converted
+      warning(
+        glue::glue("
         Only 1 row found in altExp named: {opt$feature_name}.
         This altExp will not be converted to an AnnData object.
       ")
```
jashapiro, line 170: Spacing
```suggestion
          Only 1 row found in altExp named: {opt$feature_name}.
          This altExp will not be converted to an AnnData object.
        ")
```

jashapiro (approved):
LGTM, with one spacing update

bin/sce_to_anndata.R
```diff
@@ -129,36 +132,43 @@ if (!is.null(opt$feature_name)) {
...
+
+      # make sce czi compliant
+      alt_sce <- format_czi(alt_sce)
+
+      # export altExp sce as anndata object
+      scpcaTools::sce_to_anndata(
+        alt_sce,
+        anndata_file = opt$output_feature_h5
+      )
+    } else {
```
jashapiro, line 164: Oh wait, this else is another else... we need to fix that

jashapiro: nvm, this is fine, it is the nested `if`! (This is what I was saying about failing first... just helps my brain parse it)

PR #665, opened 2024-01-22
Actually remove unclassified cells from heatmap

allyhawkins:
When working on a plot for the paper, I noticed that we don't remove the `Unclassified cells` from the heatmaps, even though we have a nice little blue box that says we do.

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -272,7 +272,7 @@ unclassified_methods <- c()
 
 # check for unclassified SingleR cells
 if (has_singler && any(celltype_df$singler_celltype_annotation == "Unclassified cell")) {
-  unclassified_methods <- c(unclassified_methods, "singler")
+  unclassified_methods <- c(unclassified_methods, "SingleR")
 
   # remove all unclassified cells
   celltype_df <- celltype_df |>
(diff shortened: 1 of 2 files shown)


sjspielman:
This code is fine, but I want to take a step back to figure out how it relates to this code, which appears right after we create `celltype_df`. Do we still need it? I have no memory of what exactly we were talking about with this `unclassified_methods` vector.

https://github.com/AlexsLemonade/scpca-nf/blob/f8905ef8601f017219e0f8e315a22a37fded75b8/templates/qc_report/celltypes_qc.rmd#L270-L308

allyhawkins:
> This code is fine, but I want to take a step back to figure out how it relates to this code, which appears right after we create celltype_df. Do we still need it? I have no memory of what exactly we were talking about with this unclassified_methods vector.

Okay so I realized this when I was making the heatmap separately, but the annotations are being saved as factors in the `create_celltype_df` function. That means even if you filter out those values after then the level is still there, and it was still showing up as a row/column in the heatmap. Let me revisit this again, because we still need to produce the warning, which is what that chunk you posted is doing.

sjspielman:
> That means even if you filter out those values after then the level is still there, and it was still showing up as a row/column in the heatmap. 

uggghhhhh factors

sjspielman:
@allyhawkins is this useful? https://forcats.tidyverse.org/reference/fct_drop.html

allyhawkins:
> https://forcats.tidyverse.org/reference/fct_drop.html

Yes thank you! I updated this to just use that to drop the level after removing the unclassified cells from `celltype_df`.

sjspielman (approved):
Looks good, nice and clean!

Two other quick changes that I can't comment on - 
L275 and L287 where `unclassified_methods` is being defined, can you change "singler" -> "SingleR" and "cellassign" to "CellAssign"? I'm feeling OCD about it.

Don't need to see again though!

PR #666, opened 2024-01-23
Update colData merging based on altExps 😈 

### LINKED ISSUE #661: Retain adt colData columns when merging, and keep cellhash main colData
Once we have https://github.com/AlexsLemonade/scpcaTools/issues/260 implemented, we'll need to revisit `merge.nf` and the `merge_sce()` function to give `scpcaTools::merge_sce_list()` a named list of colData columns to keep. It should look something like:

```
adt_possible <- c("sum.controls", "high.controls", "ambient.scale", "high.ambient", "discard")
adt_present <- adt_list |>
  purrr::map(\(adt) names(colData(adt))) |> 
  purrr::reduce(union)

# this is the argument:
list("adt" = intersect(adt_possible, adt_present))
```

In parallel, we'll want to ensure we keep cellhash colData columns in the _main_ experiment (not altExp!). We can do something similar with `cellhash_possible <- c(hashedDrops_sampleID", "HTODemux_sampleID", "vireo_sampleid")`. Tack this bit into the _main_ vector `retain_coldata_columns`.

sjspielman:
Closes #661 
I am opening this PR as a draft, since we are still working to finalize the `scpcaTools::merge_sce_list()` function updates https://github.com/AlexsLemonade/scpcaTools/pull/261 which this depends on. But, I have locally run the code with the current version of `merge_sce_list()` and it is working. If anyone wants to have a look at the code before I actually mark this ready for review, feel free!

First, I determine all modalities which are present. Then, handle columns which need to be included for each of `"adt"` and `"cellhash"`, if present. It also occurred to me that we should maybe keep the cellhash _stats_ columns, even though we aren't keeping the cellhash altExp itself (any disagreement?). We don't need to be explicit about those columns for `"adt"`, since `merge_sce_list()` will add them as part of altExp preparation. But, those steps won't be taken for `"cellhash"` since altExps will be skipped in the merge function.

FYI I rearranged a little bit too - I realized that by adding the `adt_scpca_filter` column within the existing `purrrr::map()`, we were going to keep readding it...and readding the vector...again and again for each SCE. So, this got moved out too.

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -155,10 +155,7 @@ if (opt$feature_dir != "") {
 
   # if CITE, add `adt_name` column to rowData with rownames
   if (opt$feature_name == "adt") {
-    rowdata_rownames <- altExp(unfiltered_sce, opt$feature_name) |>
-      rowData() |>
-      rownames()
-    rowData(altExp(unfiltered_sce, opt$feature_name))$adt_name <- rowdata_rownames
+    rowData(altExp(unfiltered_sce, "adt"))$adt_name <- rownames(rowData(altExp(unfiltered_sce, "adt")))
   }
 }
 
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -152,6 +152,14 @@ if (opt$feature_dir != "") {
   unfiltered_sce <- merge_altexp(unfiltered_sce, feature_sce, opt$feature_name)
   # add alt experiment features stats
   altExp(unfiltered_sce, opt$feature_name) <- scuttle::addPerFeatureQCMetrics(altExp(unfiltered_sce, opt$feature_name))
+
+  # if CITE, add `adt_name` column to rowData with rownames
+  if (opt$feature_name == "adt") {
+    rowdata_rownames <- altExp(unfiltered_sce, opt$feature_name) |>
+      rowData() |>
+      rownames()
+    rowData(altExp(unfiltered_sce, opt$feature_name))$adt_name <- rowdata_rownames
+  }
 }
 
 
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -153,9 +153,9 @@ if (opt$feature_dir != "") {
   # add alt experiment features stats
   altExp(unfiltered_sce, opt$feature_name) <- scuttle::addPerFeatureQCMetrics(altExp(unfiltered_sce, opt$feature_name))
 
-  # if CITE, add `adt_name` column to rowData with rownames
+  # if CITE, add `adt_id` column to rowData with rownames
   if (opt$feature_name == "adt") {
-    rowData(altExp(unfiltered_sce, "adt"))$adt_name <- rownames(rowData(altExp(unfiltered_sce, "adt")))
+    rowData(altExp(unfiltered_sce, "adt"))$adt_id <- rownames(rowData(altExp(unfiltered_sce, "adt")))
   }
 }
 
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -207,6 +207,7 @@ possible_columns <- c(
   "miQC_pass",
   "prob_compromised",
   "scpca_filter",
+  "additional_modalities",
   # cell type names
   "submitter_celltype_annotation",
   "singler_celltype_annotation",
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -258,7 +258,7 @@ if (is.null(adt_possible_columns) & length(adt_altexps) > 0) {
 }
 
 retain_altexp_coldata_list <- list("adt" = intersect(adt_possible_columns, adt_present_columns))
-preserve_altexp_rowdata_list <- list("adt" = c("adt_name", "target_type"))
+preserve_altexp_rowdata_list <- list("adt" = c("adt_id", "target_type"))
 
 # Merge SCEs -------------------------------------------------------------------
 
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -210,9 +210,6 @@ if ("cellassign" %in% all_celltypes) {
 
 
 
-
-
-
 # Update some SCE information  -----------------------
 # - Add a new colData column with any additional modalities
 # - Remove cluster parameters from metadata
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -100,7 +100,7 @@ if (opt$threads > 1) {
   bp_param <- BiocParallel::SerialParam()
 }
 
-# Read in SCEs -------------------------------------------------------------------
+# Read in SCEs -----------------------------------------------------------------
 
 # get list of sces
 sce_list <- purrr::map(input_sce_files, readr::read_rds)
@@ -116,26 +116,9 @@ if (!all(sce_checks)) {
   )
 }
 
-# Determine cell type annotation columns to retain  -----------------------------------------
+# Add cell type annotation columns where needed  -------------------------------
 
-
-# Default vector of colData columns to retain from
-# https://github.com/AlexsLemonade/scpcaTools/blob/2ebdc4f4dfc4233fad97805f9a9a5e3bc6919f1e/R/merge_sce_list.R
-retain_coldata_columns <- c(
-  "sum",
-  "detected",
-  "total",
-  "subsets_mito_sum",
-  "subsets_mito_detected",
-  "subsets_mito_percent",
-  "miQC_pass",
-  "prob_compromised",
-  "barcodes",
-  # additional column not in scpcaTools:
-  "scpca_filter"
-)
-
-# check if any SCEs have cell types; if so, we want to retain those colData columns
+# check for present cell type annotations
 all_celltypes <- sce_list |>
   purrr::map(\(sce) {
     metadata(sce)$celltype_methods
@@ -143,13 +126,7 @@ all_celltypes <- sce_list |>
   purrr::reduce(union)
 
 
-# add relevant cell type columns to `retain_coldata_columns`
 if ("submitter" %in% all_celltypes) {
-  retain_coldata_columns <- c(
-    retain_coldata_columns,
-    "submitter_celltype_annotation"
-  )
-
   # Add `"Submitter-excluded"` value to any libraries without submitter
   sce_list <- sce_list |>
     purrr::map(\(sce){
@@ -170,13 +147,6 @@ if ("singler" %in% all_celltypes) {
     unlist() |>
     any()
 
-  retain_coldata_columns <- c(
-    retain_coldata_columns,
-    "singler_celltype_annotation",
-    # only use ontology if TRUE
-    ifelse(use_ontology, "singler_celltype_ontology", NULL)
-  )
-
   # Add `"Cell type annotation not performed"` string to libraries without SingleR
   sce_list <- sce_list |>
     purrr::map(\(sce){
@@ -190,12 +160,6 @@ if ("singler" %in% all_celltypes) {
     })
 }
 if ("cellassign" %in% all_celltypes) {
-  retain_coldata_columns <- c(
-    retain_coldata_columns,
-    "cellassign_celltype_annotation",
-    "cellassign_max_prediction"
-  )
-
   # Add `"Cell type annotation not performed"` string to libraries without CellAssign,
   #  and make the max prediction `NA_real_` for extra safety
   sce_list <- sce_list |>
@@ -209,8 +173,7 @@ if ("cellassign" %in% all_celltypes) {
 }
 
 
-
-# Update some SCE information  -----------------------
+# Update some SCE information  -------------------------------------------------
 # - Add a new colData column with any additional modalities
 # - Remove cluster parameters from metadata
 sce_list <- sce_list |>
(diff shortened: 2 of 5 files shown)


merge.nf
```diff
@@ -146,7 +146,7 @@ workflow {
       .map{[
         it.scpca_project_id,
         it.scpca_library_id,
-        file("${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id}/${it.scpca_library_id}_processed.rds")
+        file("${params.results_dir}/${it.scpca_project_id}/${it.scpca_sample_id.split(";").sort().join(",")}/${it.scpca_library_id}_processed.rds")
```
jashapiro, line 149: Why don't we do this on read (line 129) to make it a bit simpler and make sure we are always using the same value (in case `scpca_sample_id` comes up again).

(I also wouldn't be opposed to removing `scpca_` prefixes, as we do in `main.nf`, and in the process removing columns we don't need from the `meta` object)

bin/merge_sces.R
```diff
@@ -224,25 +222,78 @@ sce_list <- sce_list |>
...
+  # determine which cellhash stats columns to keep of the possible columns which may exist
+  cellhash_possible <- c(
+    "hashedDrops_sampleID",
+    "HTODemux_sampleID",
+    "vireo_sampleid",
+    # Also consider cellhash stats columns, which will not get
+    # added in merge_sce_list since include_altexp will be FALSE
+    "altexps_cellhash_sum",
+    "altexps_cellhash_detected",
+    "altexps_cellhash_percent"
```
jashapiro, line 272: I think we want to keep any `altexps` columns for the ADTs too?

In general, is there any reason to do all of these steps multiple times?

I would think we can just set up a `coldata_possible` vector up toward the top that includes all modality values, then intersect it with `coldata_present` one time to get all columns all at once for the final value of `retain_coldata_cols`

bin/merge_sces.R
```diff
@@ -224,25 +222,78 @@ sce_list <- sce_list |>
...
+# Merge SCEs -------------------------------------------------------------------
+
+
 # create combined SCE object
 merged_sce <- scpcaTools::merge_sce_list(
   sce_list,
   batch_column = "library_id",
+  cell_id_column = "cell_id",
   retain_coldata_cols = retain_coldata_columns,
+  include_altexp = opt$include_altexp,
```
jashapiro, line 292: For the future, part of me wonders if we should allow this to be a vector of `altexps` to include. This would be a change in `scpcaTools`, and for backward compatibility we should allow a `TRUE/FALSE` to mean keep all or keep none, with a vector meaning (implicitly) keep some. 

(Turn this comment into an issue there, but do not assign it as something _to do_.)

bin/merge_sces.R
sjspielman: Noting we do already keep for ADTs, because we keep ADT altExps -
https://github.com/AlexsLemonade/scpcaTools/blob/978018e3e78be4ba27c3c54df0e233884755e843/R/merge_sce_list.R#L158-L166

sjspielman: I think if we are more explicit here we'd end up having a non-unique vector in the scpcaTools function, so we'd want to add a `retain_coldata_cols <- unique(retain_coldata_cols)` somewhere in there.

jashapiro: Throwing a `unique()` in there seems like a good idea. 

As did the code you noted in `merge_sce_list()` at the time, but I wonder if we might want to take it out so that the `colData` columns that come out of  always match `retain_coldata_cols`. I'd note that we don't really describe this additional behavior in the function docs, and it is hard to explain!

sjspielman: > but I wonder if we might want to take it out so that the colData columns that come out of always match retain_coldata_cols

Yeah, I'm definitely ok with this move.

merge.nf
sjspielman: ✔️ https://github.com/AlexsLemonade/scpca-nf/pull/666/commits/106be4875a6ce4cdfc567b58637f7ba6974a425c

sjspielman:
I'm hoping this is finally at a place where I can request formal review! Running `merge.nf` through now and if everything looks ok, will de-draft.

jashapiro (commented):
Looks good to start, but I have a few suggestions to simplify, I hope.

The main one is that I think we probably should build our `retain_coldata_columns` only once, and list all possible columns. Or maybe in two parts: `required` & `possible`, if we want to get fancy. 

Basically, I would expect that we can just make a single list of rows we might see across all modalities and then filter it once. This could also include all of the possible cell typing methods: if we have them we will include them, if not, we won't!

If we leave the initial set of columns as "required", we could have a single place where we do something like the following, just once:

```
coldata_possible <- c(
  "submitter_celltype_annotation",
  "singler_celltype_annotation",
{...}
   "altexps_adt_detected",
   "altexps_adt_percent",
   "hashedDrops_sampleID",
    "HTODemux_sampleID",
    "vireo_sampleid",
 {...}
)

coldata_present <- sce_list |>
    purrr::map(\(sce) names(colData(sce))) |>
    purrr::reduce(union)

retain_coldata_columns <- c(retain_coldata_columns, intersect(coldata_possible, coldata_present))
```

Similarly, there should not be a need to define `retain_altexp_coldata_list` and  `preserve_altexp_rowdata_list` within logical blocks. The idea of making those arguments lists was that they could be defined once for all possible structures, and the merge function would handle whether each was needed.

sjspielman:
Massively simplified! I kept some of the code to create `all_modalities`, but only for the purposes of a sanity check on the adt columns. The check is almost certainly not needed, but I love extra checks...

bin/generate_unfiltered_sce.R
```diff
@@ -152,6 +152,11 @@ if (opt$feature_dir != "") {
   unfiltered_sce <- merge_altexp(unfiltered_sce, feature_sce, opt$feature_name)
   # add alt experiment features stats
   altExp(unfiltered_sce, opt$feature_name) <- scuttle::addPerFeatureQCMetrics(altExp(unfiltered_sce, opt$feature_name))
+
+  # if CITE, add `adt_name` column to rowData with rownames
+  if (opt$feature_name == "adt") {
+    rowData(altExp(unfiltered_sce, "adt"))$adt_name <- rownames(rowData(altExp(unfiltered_sce, "adt")))
+  }
 }
```
allyhawkins, line 160: In reviewing https://github.com/AlexsLemonade/scpca-docs/pull/258/files, I was trying to figure out the purpose of this addition? Are you using it somewhere else? 
Also, is there a reason we don't do this for all feature data?

sjspielman: It came from a quick discussion over in `scpcaTools`: https://github.com/AlexsLemonade/scpcaTools/pull/261#discussion_r1465153941

This way, the ADT is not _only_ in the row name; it's also a column. We already have a column reflecting the cellhash rownames called `barcode_id`, so this is just making ADTs more consistent.

bin/merge_sces.R
```diff
@@ -224,25 +185,87 @@ sce_list <- sce_list |>
...
+  intersect(possible_coldata)
+
+# Define altExp columns to retain/preserve, currently only for "adt"
+adt_columns <- sce_list |>
+  # only consider "adt" altExps
+  purrr::keep(\(sce) "adt" %in% altExpNames(sce)) |>
+  purrr::map(\(sce) names(colData(altExp(sce, "adt")))) |>
+  # use .init to handle an empty input; will return the .init value
+  # if input is empty, then there are no "adt" altExps anyways
+  purrr::reduce(union, .init = NULL)
```
jashapiro, line 243: Is there a reason not to predefine the `adt_columns` in the script as we do with `possible_coldata`?

sjspielman: nope!

sjspielman: https://github.com/AlexsLemonade/scpca-nf/pull/666/commits/665c0867b98309b750c399a5a18ba3838a8063f5 

I still have a check in there, but can still remove it too!

bin/generate_unfiltered_sce.R
jashapiro: Do we want to go with `adt_id` to be even more consistent? We are at the mercy of what people name their ADTs, so maybe "name" is the wrong word?

bin/merge_sces.R
```diff
@@ -224,25 +185,94 @@ sce_list <- sce_list |>
...
+# Define altExp columns to retain/preserve, currently only for "adt"
+adt_possible_coldata <- c(
+  "zero.ambient",
+  "high.ambient",
+  "ambient.scale",
+  "sum.controls",
+  "high.controls",
+  "discard"
+)
+adt_possible_columns <- sce_list |>
```
jashapiro, line 245: Since this is now what is there, not just what could be, suggesting a rename.
```suggestion
adt_columns_present <- sce_list |>
```

bin/merge_sces.R
```diff
@@ -224,25 +185,94 @@ sce_list <- sce_list |>
...
+  purrr::reduce(union, .init = NULL)
+
+# ensure that there are indeed no "adt" altExps if adt_possible_columns is empty
+adt_altexps <- sce_list |>
+  purrr::keep(\(sce) "adt" %in% altExpNames(sce))
+if (is.null(adt_possible_columns) & length(adt_altexps) > 0) {
+  stop("Error in determining which adt altExp columns should be retained.")
+}
+
+retain_altexp_coldata_list <- list("adt" = intersect(adt_possible_coldata, adt_possible_columns))
```
jashapiro, line 260:
```suggestion
retain_altexp_coldata_list <- list("adt" = intersect(adt_possible_coldata, adt_columns_present))
```

bin/merge_sces.R
```diff
@@ -224,25 +185,94 @@ sce_list <- sce_list |>
...
+  "hashedDrops_sampleID",
+  "HTODemux_sampleID",
+  "vireo_sampleid"
+)
+
+# Define colData columns to retain based on intersection with present columns
+retain_coldata_columns <- sce_list |>
+  purrr::map(\(sce) names(colData(sce))) |>
+  purrr::reduce(union) |>
+  intersect(possible_coldata)
```
jashapiro, line 234: Because we don't necessarily know the order that will come out of the `union` statement, I think we should put the `possible_coldata` first in the `intersect` statement. This will ensure the order is defined by what we specified above.

```suggestion
columns_present <- sce_list |>
  purrr::map(\(sce) names(colData(sce))) |>
  purrr::reduce(union)
retain_coldata_columns <-  intersect(possible_coldata, columns_present)
```

jashapiro (commented):
Looks good. Just a few little comments to finish this up.

bin/merge_sces.R
sjspielman: this was my bad..!

bin/merge_sces.R
```diff
@@ -224,25 +185,94 @@ sce_list <- sce_list |>
...
+  purrr::keep(\(sce) "adt" %in% altExpNames(sce)) |>
+  purrr::map(\(sce) names(colData(altExp(sce, "adt")))) |>
+  # use .init to handle an empty input; will return the .init value
+  # if input is empty, then there are no "adt" altExps anyways
+  purrr::reduce(union, .init = NULL)
+
+# ensure that there are indeed no "adt" altExps if adt_possible_columns is empty
+adt_altexps <- sce_list |>
+  purrr::keep(\(sce) "adt" %in% altExpNames(sce))
+if (is.null(adt_possible_columns) & length(adt_altexps) > 0) {
```
jashapiro, line 256: Updating variable names (& `&&` for safety and fewer future complaints)
```suggestion
# ensure that there are indeed no "adt" altExps if adt_present_columns is empty
adt_altexps <- sce_list |>
  purrr::keep(\(sce) "adt" %in% altExpNames(sce))
if (is.null(adt_present_columns) && length(adt_altexps) > 0) {
```

jashapiro (approved):
LGTM, with one little fix.

sjspielman:
> LGTM, with one little fix.

One more fix incoming too - `additional_modalities` needs to be one of the retained columns!

bin/merge_sces.R
sjspielman: ugh, thank you.

jashapiro:
> also retain additional_modalities column

Ugh... we actually shouldn't retain that column in its current form. Really, we need to update https://github.com/AlexsLemonade/scpca-nf/blob/e253f2050117dfaa490abea2498f20878161be9a/bin/merge_sces.R#L186
so that we don't save a vector into that space. 

If we every have more than one modality, that code will break, as it will recycle the values.

So we need to transform additional_modalities there into a string with something like:
```
sce$additional_modalities <- paste(additional_modalities, collapse=";")
```
Then anywhere we use it from that column we need to split on the separator.

sjspielman:
> If we every have more than one modality, that code will break, as it will recycle the values.

Future Us thanks you! Any qualms about me opening a separate issue for that and coming back to it later this sprint?

jashapiro:
> > If we every have more than one modality, that code will break, as it will recycle the values.
> 
> 
> 
> Future Us thanks you! Any qualms about me opening a separate issue for that and coming back to it later this sprint? 

I think we can add the `paste` here and a separate issue for re-splitting if there are places we use the column?

sjspielman:
@jashapiro got it all here!

jashapiro (approved):
LGTM, with one little change/option.

PR #671, opened 2024-01-25
Fix missing files directory bug

### LINKED ISSUE #670: Fix bug in files_directory checking
We have uncovered a bug related to changes that were previously made in 

Specifically, these lines fail altogether if the fastq's are gone (and they are now gone!)

https://github.com/AlexsLemonade/scpca-nf/blob/development/modules/af-rna.nf#L128
https://github.com/AlexsLemonade/scpca-nf/blob/development/modules/af-features.nf#L167

We'll want to first check that `it.files.directory` is `true` before this check happens.

sjspielman:
Closes #670 

This PR fixes a bug to ensure that `it.file_directory` is not `false` in the first place before we check if it exists. Workflow now runs without the associated error.

### DIFF
--- a/modules/af-features.nf
+++ b/modules/af-features.nf
@@ -164,7 +164,7 @@ workflow map_quant_feature{
       .branch{
         make_rad: (
           // input files exist
-          file(it.files_directory, type: "dir").exists() && (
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
This is good. I opened #672 to handle my suggestion separately.

PR #673, opened 2024-01-25
Account for missing inputs everywhere

### LINKED ISSUE #672: Add missing_inputs branches to non-alevin mapping
Related, and we should probably fix at the same time:

In `bulk-salmon.nf`, `spaceranger.nf` and `genetic-demux.nf` we should update branching with the same pattern as `af-rna.nf` to prevent errors when raw data files do not exist.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/670#issuecomment-1910462148_

Adding detail here that this means changing branches for skipping mapping or quantification to check for the existence of the files directory variable and contents, and allowing for a missing_inputs branch. These updates should include sending `missing_inputs` branches to the log, as is done in `af-rna.nf`

jashapiro:
closes #672

Stacked on #671

Here I updated the spaceranger, genetic demux and salmon workflows to make sure the input directories exist. 

I mostly emulated the changes we had made to the alevin-based mapping steps, including adding a logging step for missing inputs.

The one additional change was to make sure the function that gets cell ranger sample ids for spaceranger will not fail if given an empty input.

### DIFF
--- a/modules/bulk-salmon.nf
+++ b/modules/bulk-salmon.nf
@@ -131,7 +131,7 @@ workflow bulk_quant_rna {
             params.repeat_mapping
             // the results directory does not exist
             || !file(it.salmon_results_dir).exists()
-            // the assembly has changed; if rad_dir doesn't exist, these lines won't get hit
+            // the assembly has changed; if salmon_results_dir doesn't exist, these lines won't get hit
             || Utils.getMetaVal(file("${it.salmon_results_dir}/scpca-meta.json"), "ref_assembly") != "${it.ref_assembly}"
             || Utils.getMetaVal(file("${it.salmon_results_dir}/scpca-meta.json"), "t2g_bulk_path") != "${it.t2g_bulk_path}"
           )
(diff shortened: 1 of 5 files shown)


jashapiro:
There still seems to be a bug here in genetic demux when not all bulk samples are present: it should fail in that case but does not. I will look back into this.

modules/bulk-salmon.nf
```diff
@@ -121,16 +121,29 @@ workflow bulk_quant_rna {
...
-          && Utils.getMetaVal(file("${it.salmon_results_dir}/scpca-meta.json"), "ref_assembly") == "${it.ref_assembly}"
-          && Utils.getMetaVal(file("${it.salmon_results_dir}/scpca-meta.json"), "t2g_bulk_path") == "${it.t2g_bulk_path}"
+        make_quants: (
+          // input files exist
+          it.files_directory && file(it.files_directory, type: "dir").exists() && (
+            // and repeat has been requested
+            params.repeat_mapping
+            // the results directory does not exist
+            || !file(it.salmon_results_dir).exists()
+            // the assembly has changed; if rad_dir doesn't exist, these lines won't get hit
```
sjspielman, line 134:
```suggestion
            // the assembly has changed; if salmon_results_dir doesn't exist, these lines won't get hit
```

sjspielman: or you can delete this comment (copied from my code, presumably) altogether - up to you!

sjspielman (approved):
LGTM! just one comment on a comment.

PR #683, opened 2024-02-01
Update merged object for anndata export

### LINKED ISSUE #681: Missing information in merged AnnData objects
While working towards https://github.com/AlexsLemonade/scpca-docs/issues/241, I have identified some discrepancies in the exported `AnnData` objects. We might want to split this issue up, but getting all my thoughts down here to start.

###  Discrepancy 1
The `.obs` slot is missing two expected columns: `assay_ontology_term_id` and `suspension_type`. However, the column `seq_unit` does exist in the object; it just didn't make it into the `suspension_type` column too. 

Here is where we add this information to `AnnData`:
https://github.com/AlexsLemonade/scpca-nf/blob/c0e257b2d6c5eb438e1677737393f5d4ca28e258/bin/sce_to_anndata.R#L82-L91

But, this won't work for merged objects, because we have placed all these metadata columns into library-specific data frames stored as `metadata(merged_sce)$library_metadata$SCPCL000000`. As a consequence, this information is missing from `.obs`. 

#### Solution to discrepancy 1

For _merged objects_, we already have added this information to `colData` already before getting to `AnnData` conversion:
https://github.com/AlexsLemonade/scpca-nf/blob/c0e257b2d6c5eb438e1677737393f5d4ca28e258/bin/merge_sces.R#L278-L287

Therefore, during `AnnData` conversion, we don't actually need to bring `assay_ontology_term_id` or `seq_unit` into the `colData`. We just need to add `is_primary_data = FALSE`. 

I suggest we add a check for if we're working with a merged object, and if so, only run: `sce$is_primary_data = FALSE` without messing with any other `colData` columns. We could add a check in a few ways, e.g.:
- a flag to the script itself
- check if, for example, `metadata(sce)$library_metadata`  is not `NULL`

###  Discrepancy 2

It turns out we actually have no `library_metadata` represented anywhere in the `AnnData` object! This information is totally lost in the conversion, since we have to remove all list & df objects from the sce metadata in `scpcaTools`:
https://github.com/AlexsLemonade/scpcaTools/blob/cec26960ed3fbfc852146e801f7c82c42c50d1b2/R/sce_to_anndata.R#L64-L67

I do think it's worth finding a way to retain this information, which you can see laid out in this table: https://github.com/AlexsLemonade/scpca-docs/blob/3afe4ae3fb4ab356d96caa8cf5298a7058065bb2/docs/merged_objects.md?plain=1#L127-L174

#### Solution to discrepancy 2

What if we just port these columns into the `colData`? Again, we'd want to leverage a check in `sce_to_anndata.R` for whether we're working with a merged object. Other than stashing these in `colData`, I don't have any particular ideas that will appease an SCE -> AnnData conversion.

Tagging @allyhawkins @jashapiro for discussion/thoughts

sjspielman:
Closes #681

This PR addresses issues raised about handling merged object during AnnData export:

1) In `merge_sces.R`, I ensured that `colData` column is called `suspension_type`, not `seq_unit`, and I updated `merge-report.rmd` accordingly.
2) I added a flag to `sce_to_anndata.R` to turn on in `merge.nf`.
3) I added a function `format_merged_sce()` to `sce_to_anndata.R` to format merged objects differently than single objects. This function only updates the reduced dim names, but even so I think the modularity is useful in case we ever want/need to change something else. Also, I use this function for altExps too. We don't expect reducedDims in the altExp so it really has no effect, but again one day we might need to add something else so I thought mirroring the code between main/altexp processing made sense. Any thoughts there?

I tested `merge.nf` with `SCPCP000003` (aka with an altExp) and confirmed it runs successfully.

### DIFF
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -170,8 +170,15 @@ if (!is.null(opt$feature_name)) {
     # add sample metadata from main sce to alt sce metadata
     metadata(alt_sce)$sample_metadata <- sample_metadata
 
-    # make sce czi compliant
-    alt_sce <- format_czi(alt_sce)
+    # make altExp sce czi compliant, for single SCE objects
+    if (!is_merged) {
+      sce <- format_czi(sce)
+    } else {
+      # paste X to reduced dim names if present
+      if (!is.null(reducedDimNames(sce))) {
+        reducedDimNames(sce) <- glue::glue("X_{reducedDimNames(sce)}")
+      }
+    }
 
     # export altExp sce as anndata object
     scpcaTools::sce_to_anndata(
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -139,7 +139,7 @@ sce <- readr::read_rds(opt$input_sce_file)
 sample_metadata <- metadata(sce)$sample_metadata
 
 # make main sce czi compliant for single objects, or format merged objects
-if (!is_merged) {
+if (!opt$is_merged) {
   sce <- format_czi(sce)
 } else {
   sce <- format_merged_sce(sce)
@@ -179,7 +179,7 @@ if (!is.null(opt$feature_name)) {
     metadata(alt_sce)$sample_metadata <- sample_metadata
 
     # make altExp sce czi compliant for single objects, or format merged objects
-    if (!is_merged) {
+    if (!opt$is_merged) {
       sce <- format_czi(sce)
     } else {
       sce <- format_merged_sce(sce)
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -39,6 +39,12 @@ option_list <- list(
     action = "store_true",
     default = FALSE,
     help = "Compress the HDF5 file containing the AnnData object"
+  ),
+  make_option(
+    opt_str = c("--is_merged"),
+    action = "store_true",
+    default = FALSE,
+    help = "Whether the input SCE file contains a merged object"
   )
 )
 
--- a/bin/sce_to_anndata.R
+++ b/bin/sce_to_anndata.R
@@ -127,8 +127,17 @@ sce <- readr::read_rds(opt$input_sce_file)
 # we need this if we have any feature data that we need to add it o
 sample_metadata <- metadata(sce)$sample_metadata
 
-# make main sce czi compliant
-sce <- format_czi(sce)
+# make main sce czi compliant, for single SCE objects
+if (!is_merged) {
+  sce <- format_czi(sce)
+} else {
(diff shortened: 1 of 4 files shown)


bin/sce_to_anndata.R
```diff
@@ -121,8 +138,14 @@ sce <- readr::read_rds(opt$input_sce_file)
...
 sample_metadata <- metadata(sce)$sample_metadata
 
-# make main sce czi compliant
-sce <- format_czi(sce)
+# make main sce czi compliant for single objects, or format merged objects
+if (!opt$is_merged) {
+  sce <- format_czi(sce)
+} else {
+  sce <- format_merged_sce(sce)
+}
```
jashapiro, line 146: remove gratuitous `!`
```suggestion
if (opt$is_merged) {
  sce <- format_merged_sce(sce)
} else {
  sce <- format_czi(sce)
}
```

bin/sce_to_anndata.R
```diff
@@ -56,6 +62,17 @@ if (!(stringr::str_ends(opt$output_rna_h5, ".hdf5|.h5|.h5ad"))) {
...
 }
 
+# Merged object function  ------------------------------------------------------
+
+# this function updates merged object formatting for anndata export
+format_merged_sce <- function(sce) {
+  # paste X to reduced dim names if present
+  if (!is.null(reducedDimNames(sce))) {
+    reducedDimNames(sce) <- glue::glue("X_{reducedDimNames(sce)}")
+  }
```
jashapiro, line 72: By my quick test, this is not `NULL` if there are no reducedDims; it is an empty vector. Since `glue` with an empty vector will return an empty vector, we don't need the `if` 
```suggestion
  reducedDimNames(sce) <- glue::glue("X_{reducedDimNames(sce)}")
```

I realize that this is in the `format_czi` code as well, but we should update that too!

bin/sce_to_anndata.R
```diff
@@ -155,8 +178,12 @@ if (!is.null(opt$feature_name)) {
...
     metadata(alt_sce)$sample_metadata <- sample_metadata
 
-    # make sce czi compliant
-    alt_sce <- format_czi(alt_sce)
+    # make altExp sce czi compliant for single objects, or format merged objects
+    if (!opt$is_merged) {
+      sce <- format_czi(sce)
+    } else {
+      sce <- format_merged_sce(sce)
+    }
```
jashapiro, line 186:
```suggestion
    if (opt$is_merged) {
      sce <- format_merged_sce(sce)
    } else {
      sce <- format_czi(sce)
    }
```

jashapiro (approved):
Some small changes suggested, but I should not need to see it again!

PR #685, opened 2024-02-05
Remove duplicated code in cell type module

sjspielman:
Hot fix into `main` to remove duplicated code that snuck in there during a merge.


allyhawkins (approved):
Thanks for catching this!

PR #687, opened 2024-02-05
Add cellassign note to QC if not run

### LINKED ISSUE #663: Maybe add a note to QC report about when skipping CellAssign
> Do we want to include a note in the report that CellAssign failed due to a low number of cells? Right now, it just won't show any plots that require CellAssign as it won't be present as a cell type method in the metadata.

I saw this question, but I'm not really sure! I feel like it is probably worth a note, but I would not want to dwell on it. Skipping the plots definitely seems correct, though.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/662#issuecomment-1904399967_

Just to note that if CellAssign is skipped because there aren't enough cells (n=30), the cell type annotation column for CellAssign will have the value "Not run" for all cells. But no other CellAssign metadata or columns will be present.

sjspielman:
Closes #663 

This PR adds a quick sentence to the top of the QC report if CellAssign was attempted but not run, where relevant
While I was here, I also added a `warning=FALSE` for an area in the UMAPs that was printing a warning (but code is all fine) that users don't need to see.

Here's how the sentence I added looks:

---

Screenshot 2024-02-05 at 11 41 08 AM

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -371,7 +371,9 @@ glue::glue("{methods_text}")
 if (cellassign_not_run) {
   # this needs two \n\n to appear on a separate line
   glue::glue(
-    "\n\nCell type annotation with [`CellAssign`](https://github.com/Irrationone/cellassign), a marker-gene-based approach ([Zhang _et al._ 2019](https://doi.org/10.1038/s41592-019-0529-1)), was attempted but not run due to an insufficient number of cells.
+   <div class=\"alert alert-info\">
+    \n\nCell type annotation with [`CellAssign`](https://github.com/Irrationone/cellassign), a marker-gene-based approach ([Zhang _et al._ 2019](https://doi.org/10.1038/s41592-019-0529-1)), was attempted but not run due to an insufficient number of cells.
+    </div>
     "
   )
 }
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -369,8 +369,9 @@ if (has_cellassign) {
 glue::glue("{methods_text}")
 
 if (cellassign_not_run) {
+  # this needs two \n\n to appear on a separate line
   glue::glue(
-    "\n\nCell type annotation with [`CellAssign`](https://github.com/Irrationone/cellassign), a marker-gene-based approach ([Zhang _et al._ 2019](https://doi.org/10.1038/s41592-019-0529-1), was attempted but did not run due to an insufficient number of cells.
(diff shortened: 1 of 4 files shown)


templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -356,6 +367,14 @@ if (has_cellassign) {
 }
 
 glue::glue("{methods_text}")
+
+if (cellassign_not_run) {
+  # this needs two \n\n to appear on a separate line
+  glue::glue(
+    "\n\nCell type annotation with [`CellAssign`](https://github.com/Irrationone/cellassign), a marker-gene-based approach ([Zhang _et al._ 2019](https://doi.org/10.1038/s41592-019-0529-1)), was attempted but not run due to an insufficient number of cells.
```
allyhawkins, line 374:
```suggestion
  glue::glue(
   <div class=\"alert alert-info\">
    \n\nCell type annotation with [`CellAssign`](https://github.com/Irrationone/cellassign), a marker-gene-based approach ([Zhang _et al._ 2019](https://doi.org/10.1038/s41592-019-0529-1)), was attempted but not run due to an insufficient number of cells.
    </div>
```
Can we make this a blue box?

sjspielman:
🟦 

---

Screenshot 2024-02-06 at 9 37 46 AM

allyhawkins (commented):
One thought is that this will only show up in the supplemental report. I feel like this should be at the beginning of the main cell type report so that it's both in the main report and in the supplemental report? Similar to how we have the check for unclassified cells.

sjspielman:
> I feel like this should be at the beginning of the main cell type report so that it's both in the main report and in the supplemental report? Similar to how we have the check for unclassified cells.

Yeah, I think you're right

sjspielman:
I rearranged some code a bit to get the blue box in both spots. I wanted the blue box to still be in the top intro section of the supplemental report, so I added a util function to determine if cellassign is present which can be called in both reports. This is necessary to do the printing in separate places.

Looking at how it turned out, I'm not sure if I like the blue box in the celltype section of the main report. I still think we should communicate the information, but just maybe as a regular sentence. What do you think?

----

# main report screenshot:
Screenshot 2024-02-06 at 12 36 14 PM


---

# supplemental report screenshot:
Screenshot 2024-02-06 at 12 36 25 PM

allyhawkins:
> Looking at how it turned out, I'm not sure if I like the blue box in the celltype section of the main report. I still think we should communicate the information, but just maybe as a regular sentence. What do you think?

You might be right 😢 Probably better to just do a sentence.

sjspielman:
> You might be right 

You may be right, or I may be crazy 🎶 

I do think I'm right, but I saw a chance to quote Billy Joel and I took it.

sjspielman:
Here's where I landed for the celltype section of the main report. I also removed the CellAssign details, mirroring the level of info we give for Singler. The supplemental report still has the full citation information and description in the blue box.

---

Screenshot 2024-02-06 at 1 19 32 PM

allyhawkins (approved):
LGTM

Issue #619, opened 2023-12-13
Update main overview figure


allyhawkins:
Right now we have an overview figure that shows that the workflow outputs SCE objects and that the workflow stops at normalization and dimensionality reduction. We should update this figure to include conversion to AnnData, clustering, and cell type annotation.

allyhawkins:
Closed by #721

PR #688, opened 2024-02-05
Add merge.nf external instructions

### LINKED ISSUE #646: Add instructions for running the merging workflow

sjspielman:
Closes #646 

This PR adds external instructions for the `merge.nf` workflow. Let me know about the level of detail and the order in which I provided information. Also, am I missing anything..?

Also, do we want to indicate in the repo's `README` that this workflow exists? Since other `README` updates are coming down the line (https://github.com/AlexsLemonade/scpca-nf/issues/619), I wondered if we should wait until those additional changes happen. Or, since we already link the external instructions in the `README`, it may not be necessary at all.  

Primarily requesting Ally for review here, but @jashapiro please feel free to chime in if you have any wording/organizational suggestions!

### DIFF
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -641,17 +641,15 @@ If you specified a different parameter value from the default `scpca-out` for th
 Results from running the `merge.nf` workflow will also be added to this `params.outdir` directory in a sub-directory called `merged`.
 The `merge.nf` workflow will create merged objects for all processed library objects present for the specified project id.
 
-In order to run this workflow, you will first need to clone the `scpca-nf` repository and navigate into it:
 
-```sh
-git clone git@github.com:AlexsLemonade/scpca-nf.git
-cd scpca-nf
-```
-
-Then, you can run the workflow as follows:
+The workflow can be run as shown below.
+To be sure that you are using a consistent version, you can specify use of a release tagged version of the workflow, set below with the `-r` flag.
+The command below will pull the `scpca-nf` workflow directly from Github using the `v0.7.2` version.
+Released versions can be found on the [`scpca-nf` repository releases page](https://github.com/AlexsLemonade/scpca-nf/releases).
 
 ```sh
 nextflow run AlexsLemonade/scpca-nf/merge.nf \
+  -r v0.7.2 \
   -config <path to config file>  \
   -profile <name of profile> \
   --project <project id whose libraries should be merged>
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -33,7 +33,10 @@
   - [Overview](#overview-1)
   - [Steps for running the `merge.nf` workflow](#steps-for-running-the-mergenf-workflow)
     - [Running the `merge.nf` workflow](#running-the-mergenf-workflow)
+  - [Output files](#output-files-1)
   - [Special considerations for specific data types when running `merge.nf`](#special-considerations-for-specific-data-types-when-running-mergenf)
+    - [Merging libraries with CITE-seq data](#merging-libraries-with-cite-seq-data)
+    - [Merging libraries with cellhash data](#merging-libraries-with-cellhash-data)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -638,11 +641,9 @@ The `merge.nf` workflow requires two parameters to run:
   In this case, a separate merged object will be created for each id.
 - `run_metafile`, the run metadata file which was previously prepared when running the main workflow
 
-<!-- I'm not sure this stuff about outdir is needed actually??-->
 The `merge.nf` workflow runs by first finding all libraries present, for each project, in the specified `params.outdir`, which represents the output directory where `scpca-nf` will have stored results from a prior run.
 If you specified a different parameter value from the default `scpca-out` for the `outdir` parameter in your `scpca-nf` configuration file, you will need to ensure that same value is provided to `merge.nf`.
 Results from running the `merge.nf` workflow will also be added to this `params.outdir` directory in a sub-directory called `merged`.
-The `merge.nf` workflow will create merged objects for all processed library objects present for the specified project id.
 
 The workflow can be run as shown:
 
@@ -665,14 +666,42 @@ nextflow run AlexsLemonade/scpca-nf/merge.nf \
   --project <project id whose libraries should be merged>
 ```
 
+### Output files
+
+The `merge.nf` workflow will output, for each specified project id, an `.rds` file containing a merged `SingleCellExperiment` object, an `.hdf5` file containing a merged `AnnData` object, and a report which provides a brief summary of the types of libraries and their samples' diagnoses included in the merged object, as well as UMAP visualizations highlighting each library.
+
+These output files will follow this structure:
+
+```
+merged
+└── <project id>
+    ├── <project_id>_merged.rds
+    ├── <project_id>_merged_rna.hdf5
+    └── <project_id>_merged-summary-report.html
+```
+
 
 ### Special considerations for specific data types when running `merge.nf`
 
 There are some additional considerations to be aware of for libraries which contain additional modalities, such as ADT counts from CITE-seq or HTO counts from multiplexing.
 
+#### Merging libraries with CITE-seq data
+
 If any libraries in a merge group have ADT counts, these counts will also be merged and included in the final merged object.
 In the case of `SingleCellExperiment` objects, ADT counts will be provided as an alternative experiment called `"adt"` in same object.
 In the case of `AnnData`, a separate file will be exported with the extension `_adt.hdf5` that contains the merged ADT counts.
+The output files will follow this structure if CITE-seq data is present:
+
+```
+merged
+└── <project id>
+    ├── <project_id>_merged.rds
+    ├── <project_id>_merged_rna.hdf5
+    ├── <project_id>_merged_adt.hdf5
+    └── <project_id>_merged-summary-report.html
+```
+
+#### Merging libraries with cellhash data
 
 The `merge.nf` workflow currently does not support merging HTO counts from multiplexed libraries.
 If any libraries contain HTO counts, the RNA counts will still be merged and exported, but the HTO counts will not be included.
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -67,12 +67,24 @@ Here we provide an overview of the steps you will need to complete:
 The standard configuration the `scpca-nf` workflow expects that compute nodes will have direct access to the internet, and will download reference files and container images with any required software as required.
 If your HPC system does not allow internet access from compute nodes, you will need to download the required reference files and software before running, [following the instructions we have provided](#using-scpca-nf-on-nodes-without-direct-internet-access).
 
-Once you have set up your environment and created the metadata and configuration files, you will be able to start your run as follows, adding any additional optional parameters that you may choose.
+Once you have set up your environment and created the metadata and configuration files, you will be able to start your run as follows, adding any additional optional parameters that you may choose:
 
-You will also need to specify a release tagged version of the workflow with the `-r` flag.
+```sh
+nextflow run AlexsLemonade/scpca-nf \
+  -config <path to config file>  \
+  -profile <name of profile>
+```
+
+Where `<path to config file>` is the **relative** path to the [configuration file](#configuration-files) that you have setup and `<name of profile>` is the name of the profile that you chose when [creating a profile](#setting-up-a-profile-in-the-configuration-file).
+This command will pull the `scpca-nf` workflow directly from Github, and run it based on the settings in the configuration file that you have defined.
+
+**Note:** `scpca-nf` is under active development.
+Using the above command will run the workflow from the `main` branch of the workflow repository.
+To update to the latest released version you can run `nextflow pull AlexsLemonade/scpca-nf` before the `nextflow run` command.
+
+To be sure that you are using a consistent version, you can specify use of a release tagged version of the workflow, set below with the `-r` flag.
 The command below will pull the `scpca-nf` workflow directly from Github using the `v0.7.2` version.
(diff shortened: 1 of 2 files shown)


external-instructions.md
```diff
@@ -606,3 +607,52 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
 nextflow run AlexsLemonade/scpca-nf \
   --repeat_mapping
 ```
+
+
+## Running the `merge.nf` workflow
+
+In addition to the main `scpca-nf` workflow, this repository contains a separate workflow called `merge.nf` that will merge a set of processed ScPCA `SingleCellExperiment` objects into a single merged `SingleCellExperiment` object containing all counts from the specified libraries.
+Specifically, for a given SCPCA project id, this workflow creates a merged `SingleCellExperiment` object, a merged `AnnData` object, and an associated merged object HTML report.
+This workflow is specifically designed to run on processed `SingleCellExperiment` object files created by the `scpca-nf` workflow.
```
allyhawkins, line 616:
```suggestion
This workflow is specifically designed to run on processed `SingleCellExperiment` object files output by the `scpca-nf` workflow.
```

external-instructions.md
```diff
@@ -606,3 +607,52 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
...
+
+There are several important items to be aware of when running this workflow:
+
+* **This workflow does not integrate libraries or perform any batch-correction**.
+It only merges objects into a single object to facilitate downstream analyses such as, but not limited to, integration.
+* Running this workflow may require a significant amount of memory and up to 250 GB of storage.
```
allyhawkins, line 622: I don't think we actually know how much storage is required? It's going to be dependent on what users are running and how many libraries are being merged. Maybe just a note that the merged object will total the storage of all individual libraries (or be less). 
Also, do we know what the memory is? Is it any different then the max memory we require to run the main workflow.

external-instructions.md
```diff
@@ -606,3 +607,52 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
 nextflow run AlexsLemonade/scpca-nf \
   --repeat_mapping
 ```
+
+
+## Running the `merge.nf` workflow
+
+In addition to the main `scpca-nf` workflow, this repository contains a separate workflow called `merge.nf` that will merge a set of processed ScPCA `SingleCellExperiment` objects into a single merged `SingleCellExperiment` object containing all counts from the specified libraries.
+Specifically, for a given SCPCA project id, this workflow creates a merged `SingleCellExperiment` object, a merged `AnnData` object, and an associated merged object HTML report.
```
allyhawkins, line 615:
```suggestion
Specifically this workflow creates a merged `SingleCellExperiment` object, a merged `AnnData` object, and an associated merged object HTML report encompassing all libraries with the same `scpca_project_id`. 
```
Generally, since these are external docs, we require the `scpca_project_id` column, but it doesn't have to actually correspond to a project ID for the portal. Since we refer to it in the section on preparing a metadata file, we may want to link to that section. But we also want to convey that that column is used for grouping.

external-instructions.md
```diff
@@ -606,3 +607,52 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
...
+
+There are several important items to be aware of when running this workflow:
+
+* **This workflow does not integrate libraries or perform any batch-correction**.
+It only merges objects into a single object to facilitate downstream analyses such as, but not limited to, integration.
+* Running this workflow may require a significant amount of memory and up to 250 GB of storage.
+* If ADT counts from a CITE-seq experiment are included in at least one library being merged, these results will be included in the final merged object.
+* Any HTO counts from multiplexed libraries will not be included in the final merged object.
+* At this time, the workflow only supports merging libraries from the specified ScPCA project id(s).
+It is not currently possible to specify different groupings.
```
allyhawkins, line 626: Same comment as above, we should rephrase this to just say that the `scpca_project_id` column is how to designate groupings.

external-instructions.md
```diff
@@ -606,3 +607,52 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
...
+
+```sh
+git clone git@github.com:AlexsLemonade/scpca-nf.git
+cd scpca-nf
+```
+
+Then, you can run the workflow as follows:
+
+```sh
+nextflow run merge.nf \
```
allyhawkins, line 654:
```suggestion
nextflow run AlexsLemonade/scpca-nf/merge.nf \
```
I think this should work? If so, we can remove the cloning instructions.

allyhawkins (commented):
Thanks for doing this and getting this started. Generally, I think we want to do some reorganization from what you have. I would approach it similarly to how we approach the main workflow and present a series of steps you need to accomplish to run the workflow. 

- Intro to what the merged workflow is (currently found in lines 614-616). Here add a reminder that the workflow does not integrate and talk about the purpose briefly (line 620-621). 
- Steps to run the workflow (list the steps and then expand each of these as their own section)
  - Prepare the metadata file
  - Set up environment
  - Run the main workflow to produce scpca-nf output
  - Run the merged workflow (this is where you would mention the parameters needed for this workflow). 
- Special considerations for libraries with CITE-seq or cell hashing. This should be pretty short, but just a space to mention what is/is not included for these libraries.

sjspielman:
Thanks for the feedback, I'll re-organize as suggested! Once clarification about this comment - 

>  (list the steps and then expand each of these as their own section) 

Do you think I should reiterate certain aspects (e.g. preparing the metadata file) that are already present for scpca-nf, or just link the sections? To me, "expand" suggests to re-iterate, but wanted to make sure we're on the same page before I dive in.

external-instructions.md
sjspielman: Since all objects are being loaded into memory at once, I imagine quite a bit, but really we don't know. Maybe we come back to this sizing section once we actually have the information after running the projects through.

external-instructions.md
sjspielman: > But we also want to convey that that column is used for grouping.

You know, I guess then technically they could specify their own grouping by leveraging this column, but they'd also have to do a bunch of file re-organization. Do you think I should go into this?

allyhawkins:
> Do you think I should reiterate certain aspects (e.g. preparing the metadata file) that are already present for scpca-nf, or just link the sections? To me, "expand" suggests to re-iterate, but wanted to make sure we're on the same page before I dive in.

This is a good question. I was thinking about creating something similar to what we included in the `Overview` where we have a list of steps and links to more information about them (https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md#overview). So maybe I would just list the steps and number them and for the configuration, metadata file, and running scpca-nf link to the appropriate sections in the main workflow. Then for the run the merged workflow link to a section below that expands on how to set up the run/ specify params. 

## Merged workflow overview 
Intro 

1. Prep metadata file 
2. Configure environment
3. Run scpca-nf 
4. Run merged workflow 

### Running the merged workflow
params information + command to run 

### Special considerations

external-instructions.md
sjspielman: Yeah, seems to work! But looks like users need to provide a tag with `-r`, too, so I'll add that also.

sjspielman:
> So maybe I would just list the steps and number them and for the configuration, metadata file, and running scpca-nf link to the appropriate sections in the main workflow. Then for the run the merged workflow link to a section below that expands on how to set up the run/ specify params.

Sounds good to me!

external-instructions.md
allyhawkins: They shouldn't have to do any file reorganization. Whatever unique ID is used in `scpca_project_id` column will be used to name the output folders when running `scpca-nf`. And then also used for grouping. I think we just want to emphasize that the grouping is based on the contents of that column which is set by the user.

sjspielman: I guess I was thinking of a scenario where they wanted to specify a grouping that was _different_ from the id they used to run the main workflow. This would involve reorg, which I don't really want to dive into...unless someone actually asks us about it; _then_ we can add it.

sjspielman:
Next round! This is _exactly_ what we talked about, since while writing it, it felt weird to say "prepare the run metadata file, and then follow every other instruction to prepare for the whole workflow." So, I followed the overall structure we discussed, but I deviated a bit in the details. How does it look to you?

Also, it seems that the workflow _always wants_ the `-r` version flag. If you try to run without one, it fails with the error:
```
Project `AlexsLemonade/scpca-nf` is currently stickied on revision: v0.7.2 -- you need to explicitly specify a revision with the option `-r` in order to use it
```
So, I did a bit of cleanup across docs to convey this as a mandatory argument. In addition, I removed "extra" CLI args from where we show `--publish_fry_outs`, since we follow a convention where we don't repeat all flags when showing a new one like that.

jashapiro:
> Also, it seems that the workflow _always wants_ the `-r` version flag. If you try to run without one, it fails with the error:
> 
> ```
> Project `AlexsLemonade/scpca-nf` is currently stickied on revision: v0.7.2 -- you need to explicitly specify a revision with the option `-r` in order to use it
> ```

This error is not an _always_ error. It has to do with what the last pulled version was. If you run `nextflow pull` without `-r` you should get the latest version and then be able to run it just fine. Which is to say that `-r` is not always required, but it is a often good idea.

jashapiro:
> > Also, it seems that the workflow _always wants_ the `-r` version flag. If you try to run without one, it fails with the error:
> > ```
> > Project `AlexsLemonade/scpca-nf` is currently stickied on revision: v0.7.2 -- you need to explicitly specify a revision with the option `-r` in order to use it
> > ```
> 
> This error is not an _always_ error. It has to do with what the last pulled version was. If you run `nextflow pull` without `-r` you should get the latest version and then be able to run it just fine. Which is to say that `-r` is not always required, but it is a often good idea.

Rather than adding `-r` everywhere, maybe we add a "troubleshooting" section? I would say if you see that error, try `nextflow pull AlexsLemonade/scpca-nf` to get the latest release, then try again. In general, I would rather have external users use the latest version unless they have a reason to stick with an old one.

external-instructions.md
```diff
@@ -491,8 +496,6 @@ This will tell the workflow to save the `alevin-fry` outputs to a folder labeled
 
 ```sh
 nextflow run AlexsLemonade/scpca-nf \
-  -config <path to config file>  \
-  -profile <name of profile> \
```
allyhawkins, line 495: Was there a reason you deleted these?

external-instructions.md
```diff
@@ -606,3 +609,70 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
...
+#### Running the `merge.nf` workflow
+
+The `merge.nf` workflow requires two parameters to run:
+
+- `project`, the `scpca_project_id` whose objects should be merged
+  * A comma-separated list of `scpca_project_id` values can also be provided.
+  In this case, a separate merged object will be created for each id.
+- `run_metafile`, the run metadata file which was previously prepared when running the main workflow
+
+<!-- I'm not sure this stuff about outdir is needed actually??-->
```
allyhawkins, line 641: I think you can keep it, it probably doesn't hurt?

external-instructions.md
```diff
@@ -606,3 +609,70 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
...
+The `merge.nf` workflow runs by first finding all libraries present, for each project, in the specified `params.outdir`, which represents the output directory where `scpca-nf` will have stored results from a prior run.
+If you specified a different parameter value from the default `scpca-out` for the `outdir` parameter in your `scpca-nf` configuration file, you will need to ensure that same value is provided to `merge.nf`.
+Results from running the `merge.nf` workflow will also be added to this `params.outdir` directory in a sub-directory called `merged`.
+The `merge.nf` workflow will create merged objects for all processed library objects present for the specified project id.
```
allyhawkins, line 645: I don't know if you need this sentence though? Maybe we need a small section on output files? like we have in the main instructions? It would be very brief though and just state that you get a merged object and merged report and maybe a tree showing the expected output structure.

external-instructions.md
```diff
@@ -606,3 +609,70 @@ To force repeating the mapping process, use the `--repeat_mapping` flag at the c
...
+  -profile <name of profile> \
+  --project <project id whose libraries should be merged>
+```
+
+
+### Special considerations for specific data types when running `merge.nf`
+
+There are some additional considerations to be aware of for libraries which contain additional modalities, such as ADT counts from CITE-seq or HTO counts from multiplexing.
+
+If any libraries in a merge group having ADT counts, these counts will also be merged and included in the final merged object.
```
allyhawkins, line 673:
```suggestion
If any libraries in a merge group have ADT counts, these counts will also be merged and included in the final merged object.
```

allyhawkins (commented):
I like this structure! I think the only thing that's missing is an explicit explanation of the output files? But I think it can be really short.

external-instructions.md
sjspielman: Yeah, there seems to be a convention where we don't include these lines when introducing additional flags to use for the `scpca-nf` workflow. Personally, I am very fine keeping these, but I thought that maybe it wasn't at the right "style" we were going for? So I can definitely put these lines back, but I want to be on the record that I am also in favor of including them in the other places we introduce flags (repeat mapping/celltyping, etc).

sjspielman:
I've added a small output section s requested! 
Let me know what your preference is for this - https://github.com/AlexsLemonade/scpca-nf/pull/688/files#r1481908533

external-instructions.md
allyhawkins: I'm 50/50 on this. I think you're fine to delete this.

allyhawkins (approved):
Looks good!

PR #689, opened 2024-02-06
QC report aesthetic updates

### LINKED ISSUE #618: Adjust UMAP point sizes based on number of cells
In reviewing some of the cell type reports, I noticed that the points on the UMAPs look kind of small for samples that have < ~1000 cells. We might want to consider dynamically changing the point size based on having a certain number of cells so that libraries with fewer cells have larger points.

### LINKED ISSUE #620: Aesthetic updates to miQC plots
Idle thought: do we want to apply these aesthetic improvements to `scpca-nf` too?

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/CART_Monje/pull/173#discussion_r1426925647_

In the Monje workflow, we made some small adjustments to the lines printed on the miQC plots. We might consider adding those changes to the miQC plots in this repo.

sjspielman:
Closes #618
Closes #620 

This PR makes two viz changes to the QC report:

- For #620, I updated the miQC plot line thickness as described in the isssue
- For #618, I updated the UMAP point size. I explored a variety of different numbers of cells and tweaked numbers accordingly. In the attached zip here, I include a bunch of reports where the file name contains the number of cells. All these reports include the update miQC line sizing, too.
  - Once we agree on the point sizes, I will port the code defining the umap point sizes into the cell type supplemental report. I feel like we might want to put that in a function for Future Us, which might entail making a new `utils/report_functions.rmd` file? Thoughts?

Let me know what you think of sizing choices!

[qc_reports_cell_counts.zip](https://github.com/AlexsLemonade/scpca-nf/files/14184266/qc_reports_cell_counts.zip)

### DIFF
--- a/templates/qc_report/main_qc_report.rmd
+++ b/templates/qc_report/main_qc_report.rmd
@@ -53,6 +53,12 @@ reformat_nulls <- function(df) {
 set.seed(params$seed)
 ```
 
+
+<!-- source functions --> 
+```{r, child='utils/report_functions.rmd'}
+```
+
+
 ```{r sce_setup}
 # save some typing later
 library_id <- params$library
@@ -135,19 +141,10 @@ if ((has_singler | has_cellassign) & is.null(params$celltype_report)) {
 has_multiplex <- length(sample_id) > 1
 sample_types <- metadata(unfiltered_sce)$sample_type
 
-# determine UMAP point size based on number of processed cells
-n_processed_cells <- ncol(processed_sce)
-umap_point_size <- dplyr::case_when(
-  n_processed_cells <= 500 ~ 0.8,
-  n_processed_cells <= 1500 ~ 0.6,
-  .default = 0.3
-)
-umap_facet_point_size <- dplyr::case_when(
-  n_processed_cells <= 500 ~ 0.5,
-  n_processed_cells <= 1500 ~ 0.25,
-  n_processed_cells <= 5000 ~ 0.1,
-  .default = 0.01
-)
+# determine UMAP point sizing
+umap_points_sizes <- determine_umap_point_size(ncol(processed_sce))
+umap_point_size <- umap_points_sizes[1]
+umap_facet_point_size <- umap_points_sizes[2]
 ```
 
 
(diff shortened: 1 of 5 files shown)


templates/qc_report/main_qc_report.rmd
```diff
@@ -134,6 +134,20 @@ if ((has_singler | has_cellassign) & is.null(params$celltype_report)) {
 # check if we have multiplex
 has_multiplex <- length(sample_id) > 1
 sample_types <- metadata(unfiltered_sce)$sample_type
+
+# determine UMAP point size based on number of processed cells
+n_processed_cells <- ncol(processed_sce)
+umap_point_size <- dplyr::case_when(
+  n_processed_cells <= 500 ~ 0.8,
+  n_processed_cells <= 1500 ~ 0.6,
+  .default = 0.3
```
allyhawkins, line 143: The report that has 1570 looks to have a larger point size than the 5245. Shouldn't they be the same since its > 1500? 
It's hard to comment without seeing this, but I think 1500 might be too low of a cut-off and maybe we want to do 2000? I like the way the reports currently look so if it's true that the point size is larger with 1570 then we should keep that.

allyhawkins (commented):
I actually think this approach looks pretty decent. My only question is about the 1500/2000 for the cut-off of when to use the default size. It honestly probably doesn't matter too much, but I'm just curious since it looked like what's in the code here didn't match up with the report with 1507 cells.

templates/qc_report/main_qc_report.rmd
sjspielman: > The report that has 1570 looks to have a larger point size than the 5245. Shouldn't they be the same since its > 1500?

Yeah, I see this... I wonder if I put a wrong version of that file in there. But I agree with you on the size looking good, so I'll bump. Let me explore around 2000-3000 to see if maybe 2000 or 2500 is the right spot!

sjspielman:
I played around with a couple different SCEs looking at a 2000 vs 2500 cutoff, and honestly it all looked fairly similar to me so i went with a 2250 cutoff 😂 . I wrote this up into a function in a new utils file which is now "sourced" in both the main and supplemental cell type report.

PR #691, opened 2024-02-07
Standardize celltype report filenames

arkid15r:
This find/replace change makes the celltype file report names look consistent with other files: `_celltype-report.html` -> `_celltype_report.html`


jashapiro (commented):
This was actually an intentional change to make "semantic chunks" of file names be separated by `_` with word spacing within those chunks be `-`.

We might want to update the instructions docs to show that format a bit better by changing `library_id` to `library-id`, but I think we are going to leave this particular format as is.

arkid15r:
Closing this based on the Josh's comment.

PR #693, opened 2024-02-12
add some pre-commit.ci options

jashapiro:
I recently added [pre-commit.ci](http://pre-commit.ci) to watch this repository, so I wanted to see how it works.

I added two options here, one of which is to _not_ fix anything it finds automatically, but for this repo we might want to leave that on.

The other is the target for the pre-commit hook update prs (which pre-commit.ci does automatically), which we want to be `development`, so I have added it there. 

The main question here is whether these actions will run for all PRs (including this one) or only those that target `main`. Hoping for the former!


.pre-commit-config.yaml
```diff
@@ -16,7 +16,14 @@ repos:
...
         language: fail
         files: '(?i)\.(Rhistory|RData|rds)$'
-        # `exclude: <regex>` to allow committing specific files
+        # `exclude` files here are _allowed_ data files.
+        # Here they are specified by a full path from the repository root in a multiline regex.
+        exclude: |
+          (?x)^(
+            test/references/celltype/singler_models/singler_model_file.rds|
+            test/output/results/STUBP01/STUBS01/STUBL01_processed.rds|
+            test/output/results/STUBP01/STUBS16/STUBL16_processed.rds
```
sjspielman, line 25: why only these?

jashapiro: Those are the only ones tracked in the repo.

sjspielman: ah well, there's that!

sjspielman (approved):

PR #694, opened 2024-02-13
add autofix setting & update

jashapiro:
Here I am adding a line to show how to disable the autofixes by pre-commit.ci

I also updated the versions of a few hooks.


sjspielman:
This is fine, but I wonder if we also want to say something about `precommit-ci`, as either a separate or nested topic for `pre-commit hooks` in general, in the contributing file?

sjspielman:
> I would be fine with that, but I do not have time to do it today.

That's fine, let's get this in now and circle back!

sjspielman (approved):

Issue #682, opened 2024-01-31
Deal with libraries that don't have any cells after removing empty droplets


allyhawkins:
We have come across a case where a library doesn't have any remaining cells after removing empty droplets. When this happens, the workflow will fail during the `filter_sce.R` script since we try to add columns with values to an object with no rows. We should consider handling this without the whole workflow failing, but still providing a message that the library doesn't have any cells. 

One thought is to write out an empty filtered file or no filtered file and check in a branch for either the file existing or the file size. Similar to how we handle missing input files, one branch would proceed with the workflow and the other branch would write out a message to the log. 

Another option is to create a process for any objects that have no filtered files and write out just the unfiltered file.

allyhawkins:
Closed by #738

PR #696, opened 2024-02-13
Skip any additional processing to processed objects with 0 cells

allyhawkins:
In processing the next set of projects, I came across an error in the process for running CellAssign. Prior to any actual cell type annotation we do some modifications and convert the object to an `AnnData` object. However, this fails for any objects with 0 cells. We have already accounted for this in the later step for outputting the `AnnData` objects in the workflow, but we are currently still attempting to perform downstream steps on an object with 0 cells. We probably don't want to do this and can avoid any issues with future additions by skipping any downstream processing if the object only has 0 cells. 

Here I modified the `post_process_sce.R` script to create an empty file if the object created in the script has 0 cells. Then the post processing output is branched to include libraries with > 0 cells and libraries with only 0 cells. The libraries with empty objects will get logged in the log file and not included in the remaining processing steps. Finally, they are all combined together before creating the QC report and publishing the files. For the QC report, I check if the files are empty when setting `has_filtered` and `has_processed`. I tested this and everything ran successfully for the library with no cells.


templates/qc_report/main_qc_report.rmd
```diff
@@ -60,8 +60,8 @@ unfiltered_sce <- params$unfiltered_sce
 filtered_sce <- params$filtered_sce
 processed_sce <- params$processed_sce
 
-has_filtered <- !is.null(filtered_sce)
-has_processed <- !is.null(processed_sce)
+has_filtered <- !is.null(filtered_sce) & (file.size(filtered_sce) != 0)
+has_processed <- !is.null(processed_sce) & (file.size(processed_sce) != 0)
```
jashapiro, line 64: Double & for safety/style and >0 just in case we could get a negative number somehow.

```suggestion
has_filtered <- !is.null(filtered_sce) && (file.size(filtered_sce) > 0)
has_processed <- !is.null(processed_sce) && (file.size(processed_sce) > 0)
```

templates/qc_report/main_qc_report.rmd
```diff
@@ -60,8 +60,8 @@ unfiltered_sce <- params$unfiltered_sce
 filtered_sce <- params$filtered_sce
 processed_sce <- params$processed_sce
 
-has_filtered <- !is.null(filtered_sce)
-has_processed <- !is.null(processed_sce)
+has_filtered <- !is.null(filtered_sce) & (file.size(filtered_sce) != 0)
```
jashapiro, line 63: How would a filtered file end up size 0? Is there code for that? It is fine to keep this as a general thing, but just curious.

templates/qc_report/main_qc_report.rmd
jashapiro: I'm wondering if this is the right place for this code? I would expect to check the file sizes in  
https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/sce_qc_report.R

Somewhere before: 
https://github.com/AlexsLemonade/scpca-nf/blob/b734559d8fa0f1ebb7359656cf52d36a555c787b/bin/sce_qc_report.R#L150-L153

jashapiro (commented):
Looks like the right idea for the nextflow code, but I think we want to do the file check in the outer script, not in the report template. Then we can just pass in NULL to the report template.

templates/qc_report/main_qc_report.rmd
allyhawkins: I was originally doing this in anticipation of addressing #682, but thinking about it more we haven't quite decided the approach we want there, so I'll remove it.

allyhawkins:
I moved the check for the file size to be in the Rscript for rendering rather than the actual notebook itself. I also only did it for the processed SCE object.

jashapiro:
You will need to copy the updated precommit config file from development to get this to pass. We need to skip the spelling check, which takes too long to build for precommit CI (and we cover separately)

jashapiro (approved):
I added some file exclusions to the file filter that was failing in precommit, so everything should be good to go now.

PR #697, opened 2024-02-18
Fix merge report bugs related to multiplexing

sjspielman:
This PR addresses a few small merge report bugs that came up when working on updating the ScPCA test data.

- This is the first time running through a project with _both_ multiplex and non-multiplex samples as well as _multiple_ suspension types (we now have cell & nucleus in there). So I fixed a few bugs related to that:
  - Some places where we had `%in%` should have been `==`; when determining if a row is multiplexed in `coldata_df`, we should only compare to that same row's value for `additional_modalities`. 
  - When determining facet UMAP size, we have to count the number of libraries (`tech_table$`Number of libraries`); in the data we've explored so far this data frame had only 1 row. Here, it has multiple because of multiple suspensions. So I now `sum()` this column.

Also, `RMarkdown` apparently was not liking `eval= !<thing>` in R chunks. Apparently `eval` just wants a T/F without having to actually evaluate a statement (ironically, perhaps). So I made a variable `not_multiplexed` we can use to handle some chunk logic.

There was also one chunk I switched from `if (multiplexed)` in the chunk as the only code being run, to `eval=multiplexed` in the R chunk settings instead, as a treat 🍬 

This version runs with the test data! - [merge-report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/14323524/merge-report.html.zip)

### DIFF
--- a/templates/merge-report.rmd
+++ b/templates/merge-report.rmd
@@ -69,13 +69,14 @@ batch_column <- params$batch_column
 
 # set adt/multiplex values
 # this tells us if at least one library contains either of these modalities
-additional_modalities <- merged_sce$additional_modalities |>
+additional_sce_modalities <- merged_sce$additional_modalities |>
   stringr::str_split(pattern = ";") |>
   purrr::reduce(union) |>
   sort()
 
-has_adt <- "adt" %in% additional_modalities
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
Thanks for catching these, looks good 👍

PR #699, opened 2024-02-19
Fix multiplex anndata export

### LINKED ISSUE #698: Ensure merge workflow exports AnnData for multiplexed samples
While looking into https://github.com/AlexsLemonade/ScPCA-admin/pull/823#issuecomment-1952781648, it has become clear that `merge.nf` has actually never properly exported AnnData files at all for multiplexed projects.

This can be seen here: 

```
// mimics merge_sce.out
merge_sce_out = Channel.from(["/nextflow-ccdl-data/work/c0/f8f45d984a4e92a6d3083f7f25ef92/SCPCP999990_merged.rds", "SCPCP999990", "false", "true"])


// next step in workflow:
//  anndata_ch = merge_sce.out
//      .filter{!it[3]} // remove multiplexed samples before export
//      .map{it.take(3)} // keep everything but multiplexed logical


merge_sce_out
  .filter{!it[3]}
  .view() // it's empty.
```

Since `merge_sce.out` doesn't contain _samples_, it just contains project ids, we're losing projects that have any multiplexed samples altogether. This means the `anndata_ch` ends up being empty and we don't actually run the anndata export step at all. I recall being confused by this a while back (see discussion in https://github.com/AlexsLemonade/scpca-nf/pull/652)... The thing is, we don't want to remove SCEs that are multiplexed - we just don't want to export their altexp too!

## Proposed (likely) solution 

We should do what I originally thought we should do - just delete `.filter{!it[3]}` and keep only ` .map{it.take(3)}`: 

```
 anndata_ch = merge_sce.out
  .map{it.take(3)} // keep everything but multiplexed logical
```

Then, I'm fairly sure the anndata process itself should work as expected. The `had_adt` value will be false, so the altExp will not get exported to AnnData, only the main SCE rna counts. I'm running this through now to confirm!

sjspielman:
Closes #698

This PR deletes the `filter` line as described in #698.
I confirmed it works as expected with the test data - you can see the output in `test-data/scpca_out` on `S3`.

### DIFF
--- a/merge.nf
+++ b/merge.nf
@@ -171,6 +171,7 @@ workflow {
         sce_file_list
       )}
 
+    // merge SCEs
     merge_sce(grouped_libraries_ch)
 
     // generate merge report
@@ -178,7 +179,6 @@ workflow {
 
     // export merged objects to AnnData
     anndata_ch = merge_sce.out
-      .filter{!it[3]} // remove multiplexed samples before export
       .map{it.take(3)} // keep everything but multiplexed logical
 
 


sjspielman:
Closing because as discussed in #698, this whole thing was a misunderstanding in the first place!

PR #702, opened 2024-02-21
Restore heatmap legend

### LINKED ISSUE #701: Add missing legend to celltype report
It appears we may be missing the legend for the celltype report's SingleR/CellAssign comparison heatmap. We should look into this!

sjspielman:
edit- closes #701 

While looking over the reports shared here https://github.com/AlexsLemonade/ScPCA-admin/issues/716#issuecomment-1953118312, I noticed that the CellAssign/SingleR heatmaps were missing a legend. This PR restores that legend by adding a quick argument to the heatmap function that says to always retain the legend.

Rendered report -  
[celltypes_supplemental_report.html.zip](https://github.com/AlexsLemonade/scpca-nf/files/14360330/celltypes_supplemental_report.html.zip)

Also, I reverted a change I made in https://github.com/AlexsLemonade/scpca-nf/pull/697; We definitely didn't need that `not_multiplexed` variable I made. I don't really recall/understand what I was experiencing that I thought we needed it, but we really don't need it. `eval = !multiplexed` works just fine.

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -552,7 +552,7 @@ create_single_heatmap(
   row_title = "CellAssign annotations",
   column_title = "SingleR annotations",
   labels_font_size = find_label_size(all_celltypes),
-  keep_legend_name = "SingleR annotations",
+  keep_legend_name = "CellAssign annotations",
   col_fun = heatmap_col_fun,
   # additional arguments
   column_names_rot = 90
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -116,7 +116,6 @@ ComplexHeatmap::ht_opt(TITLE_PADDING = grid::unit(heatmap_padding, "in"))
 #' @param labels_font_size Font size to use for rows and column labels.
 #' @param keep_legend_name The name to use in the legend
 #' @param col_fun Color function for the heatmap palette. Default is `heatmap_col_fun`.
-#' @param override_keep_legend Boolean to always keep the legend regardless of other logic
 #' @param ... Additional arguments to pass to `ComplexHeatmap::Heatmap()`
 #'
 #' @return A ComplexHeatmap object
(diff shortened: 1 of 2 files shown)


templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -152,8 +154,9 @@ create_single_heatmap <- function(
       direction = "horizontal",
       legend_width = unit(1.5, "in")
     ),
-    # only keep legends that match `keep_legend_name`
-    show_heatmap_legend = row_title == keep_legend_name,
+    # only keep legends if override_keep_legend is TRUE, or
+    # if the row title matches `keep_legend_name`
+    show_heatmap_legend = override_keep_legend || row_title == keep_legend_name,
```
jashapiro, line 159: I don't love adding an extra argument here. Could we do this instead by using a vector for `keep_legend_name`? So in the case in question  we would pass in `keep_legend_name = c("SingleR annotations", "CellAssign annotations")`.
```suggestion
    show_heatmap_legend = row_title %in% keep_legend_name,
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -555,7 +558,8 @@ create_single_heatmap(
   keep_legend_name = "SingleR annotations",
```
jashapiro, line 558: Or is the problem here that this is just wrong because it refers to the column not the row? 
```suggestion
  keep_legend_name = "CellAssign annotations",
```

jashapiro (commented):
I'm generally opposed to adding arguments when the existing ones work, and I am somewhat confused by what the problem here really is. So I think a better solution may be one of the two things I suggested below? Either changing the argument in the place where there was a problem, or allowing a vector of "keep" names?

templates/qc_report/celltypes_supplemental_report.rmd
sjspielman: hm. yeah.

sjspielman:
more simple, more better

jashapiro (approved):
LGTM

PR #706, opened 2024-02-27
Update merge publishDir

### LINKED ISSUE #704: Discussion: Should we change the merged publish directory to be the given project directory?
This came up when discussing the scpca test data with @arkid15r. The question was, why aren't we sending merged projects into the project directories, and instead sending into a separate merged directory?
https://github.com/AlexsLemonade/scpca-nf/blob/2e6520087a687d9b7e0983815ba283b5490fb19e/merge.nf#L33

My sense is we made this decision because, in the future, `merge_group_id` values might not be project id's, but some other indicator variable (from a future metadata file) to group, for example, all samples of diagnosis X, which has the potential to include libraries from multiple projects.

This issue is meant to facilitate some discussion about this point. Is this still our goal to keep this flexibility? If not, we might consider changing the publish directory to the project directory itself, not within `merged/`, which may make life a little easier on the dev side.

Also worth noting these external instructions, where we tell users that their merge group is expected to be the same as their project id:
https://github.com/AlexsLemonade/scpca-nf/blob/development/external-instructions.md#the-mergenf-workflow

sjspielman:
Closes #704 

Short and sweet, flips order in the `merge.nf` publishDir so, for merge ids that are projects, their merged objects will be in the project directories.

### DIFF
--- a/merge.nf
+++ b/merge.nf
@@ -30,7 +30,7 @@ if(param_error){
 process merge_sce {
   container params.SCPCATOOLS_CONTAINER
   label 'mem_32'
-  publishDir "${params.results_dir}/merged/${merge_group_id}"
+  publishDir "${params.results_dir}/${merge_group_id}/merged"
   input:
     tuple val(merge_group_id), val(has_adt), val(multiplexed), val(library_ids), path(scpca_nf_file)
   output:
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
🩴

sjspielman:
> 🩴

This took me a second! I get it. I like it.

PR #708, opened 2024-02-29
Specify merging only specific runs from a project

### LINKED ISSUE #705: Specify run ids to use when merging projects
As mentioned in https://github.com/AlexsLemonade/ScPCA-admin/issues/820, we have some scenarios where we will want to create a merged object without including all libraries in that project. To accommodate this we should add the params for run IDs to the merge workflow. We should accommodate run IDs, library IDs, and sample IDs. Then when we filter to the project IDs specified we should also filter to the run IDs (similar to how we select runs in `main.nf`).

allyhawkins:
Closes #705 

This PR adds the capability to specify runs, libraries, and/or samples to include when creating a merged object for a ScPCA project. First, we filter to any libraries that are in a specific project and then I filter to either include all runs or specify based on a parameter. 

I did choose to add a `merge_run_ids` parameter here rather than use the `run_ids` parameter set for the main workflow. The reason is because I noticed we have a default list of test runs in the ccdl profile, rather than using the `All` setting. So if you don't use the flag for `run_ids` then it will only include the test runs we specify in the profile. We would need to remember to always include `--run_ids All` when running the merge workflow, and I don't think we want to have to do that every single time. I can see scenarios of us forgetting this. So now we have a separate `merge_run_ids` param that by default is set to `All` both in the ccdl profile and the main nextflow config.

Within this PR I also added how to run with specific run ids to the internal instructions, but is this something we also want to include in external instructions? It seems like a very specific need and is not a required parameter, but I can add it if we think it's necessary.

### DIFF
--- a/config/profile_ccdl.config
+++ b/config/profile_ccdl.config
@@ -10,9 +10,9 @@ params{
   checkpoints_dir = "${params.outdir}/checkpoints"
   results_dir = "${params.outdir}/results"
 
-  // a set of run_ids for testing
+  // a set of run_ids for testing. used only by the main workflow
   run_ids = "SCPCR000001,SCPCS000101"
-  // include all runs in a merged project
+  // include all runs in a merged project. used only by the merged workflow
   merge_run_ids = "All"
 
   // cell type references
--- a/config/profile_ccdl.config
+++ b/config/profile_ccdl.config
@@ -12,6 +12,8 @@ params{
 
   // a set of run_ids for testing
   run_ids = "SCPCR000001,SCPCS000101"
+  // include all runs in a merged project
+  merge_run_ids = "All"
 
   // cell type references
   project_celltype_metafile = "s3://ccdl-scpca-data/sample_info/scpca-project-celltype-metadata.tsv"
--- a/internal-instructions.md
+++ b/internal-instructions.md
@@ -215,9 +215,9 @@ Note that the workflow will also merge any present alternative experiments (e.g.
 The merge workflow requires two parameters:
 
 - `project`, the SCPCA project id whose objects should be merged
+- `merge_run_ids`, the set of run ids, library ids, or sample ids to include in the merged object, by default all runs for each project will be included
 - `run_metafile`, the metadata file (`scpca-library-metadata.tsv`) which contains information about libraries to merge
   - This is specified in the `ccdl` profile configuration file
-
 
 Data Lab members with access to AWS can run the workflow with the following command(s):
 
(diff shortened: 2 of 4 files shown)


config/profile_ccdl.config
```diff
@@ -12,6 +12,8 @@ params{
 
   // a set of run_ids for testing
   run_ids = "SCPCR000001,SCPCS000101"
+  // include all runs in a merged project
+  merge_run_ids = "All"
```
sjspielman, line 16: I would expand comments a bit, something like - 

```suggestion
  // a set of run_ids for testing. used only by the main workflow
  run_ids = "SCPCR000001,SCPCS000101"
  // include all runs in a merged project. used only by the merged workflow
  merge_run_ids = "All"
```

internal-instructions.md
```diff
@@ -215,10 +215,10 @@ Note that the workflow will also merge any present alternative experiments (e.g.
 The merge workflow requires two parameters:
 
 - `project`, the SCPCA project id whose objects should be merged
+- `merge_run_ids`, the set of run ids, library ids, or sample ids to include in the merged object, by default all runs for each project will be included
```
sjspielman, line 218:
```suggestion
- `merge_run_ids`, the set of run ids, library ids, or sample ids to include in the merged object.
By default all runs ids for the given project will be included
```

internal-instructions.md
```diff
@@ -215,10 +215,10 @@ Note that the workflow will also merge any present alternative experiments (e.g.
 The merge workflow requires two parameters:
 
 - `project`, the SCPCA project id whose objects should be merged
```
sjspielman, line 217: Let's also add here that multiple projects can be run, but make sure the wording indicates that a merged object is only one project at a time - did i achieve this wording?

```suggestion
- `project`, the SCPCA project id whose objects should be merged.
A comma-separated list of project ids can also be provided to merge multiple projects separately
```

sjspielman (approved):
>  I can see scenarios of us forgetting this.

Not this time, Past Us. 

> Within this PR I also added how to run with specific run ids to the internal instructions, but is this something we also want to include in external instructions? It seems like a very specific need and is not a required parameter, but I can add it if we think it's necessary.

I don't think this is a necessary. We could open a "nice to have" issue for it, but either way it's not urgent and I don't think worth picking up unless it's a use case that comes up for someone.

---

LGTM! Left some comments about comments, but don't need to see again. Also I'm approving this under the assumption this code has been _tested and run_ with both the default `"All"` and a set of merge run ids. If this was not a good assumption, I can also run through to help test, lmk!

Issue #709, opened 2024-03-01
[BUG] CellAssign process getting killed with `OutOfMemoryError`


allyhawkins:
**Describe the bug**
If you have an error message or report that you can include, please do!

In running SCPCP000016 through `scpca-nf` with cell typing, I'm consistently getting the same error stating that the task was killed due to being out of memory. This project does use the all pediatric cancer reference, which is not used by any other projects that have been processed, so I do wonder if that's part of it. 

I've tried bumping up the memory to 256 and get the same error everytime. I will note that one library did complete successfully but it has ~500 cells, whereas the other libraries have > 7,000 cells. 

**To reproduce**
If possible, please provide example code to reproduce the problem, including any information about the dataset where the problem occurred, if possible.

The following command but specifying the prod `outdir` and the project: 

`nextflow run AlexsLemonade/scpca-nf -r v0.7.2 -profile ccdl,batch --perform_celltyping`

**Additional context**

This is the main message that Nextflow has for the error: 
```
The exit status of the task that caused the workflow execution to fail was: 137

Error executing process > 'annotate_celltypes:classify_cellassign (SCPCL000835)'

Caused by:
  Essential container in task exited - OutOfMemoryError: Container killed due to memory usage
```

Then here's the output error message from running `CellAssign`: 

```
Command error:
  Warning message:
  replacing previous import ‘S4Arrays::makeNindexFromArrayViewport’ by ‘DelayedArray::makeNindexFromArrayViewport’ when loading ‘SummarizedExperiment’ 
  Registered S3 methods overwritten by 'zellkonverter':
    method                                             from      
    py_to_r.numpy.ndarray                              reticulate
    py_to_r.pandas.core.arrays.categorical.Categorical reticulate
  miQC_model cannot be converted between SCE and AnnData.
  /usr/local/lib/python3.10/dist-packages/scvi/_settings.py:63: UserWarning: Since v1.0.0, scvi-tools no longer uses a random seed by default. Run `scvi.settings.seed = 0` to reproduce results from previous versions.
    self.seed = seed
  /usr/local/lib/python3.10/dist-packages/scvi/_settings.py:70: UserWarning: Setting `dl_pin_memory_gpu_training` is deprecated in v1.0 and will be removed in v1.1. Please pass in `pin_memory` to the data loaders instead.
    self.dl_pin_memory_gpu_training = (
  Global seed set to 2021
[...]

PR #711, opened 2024-03-04
Bump memory for CCDL only

jashapiro:
I think the issue with https://github.com/AlexsLemonade/scpca-nf/issues/709#issuecomment-1977033313 was that we set the max memory in two places: in the base profile for a default, but also in the ccdl profile. So I dropped the base back down and upped the ccdl profile to 1TB, which will hopefully do the trick.

I also changed the label to  `mem_512`, because that better reflects the content.

### DIFF
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -1,6 +1,6 @@
 // Resource maximum settings
 params.max_cpus = 24
-params.max_memory = 488.GB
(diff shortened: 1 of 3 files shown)


allyhawkins (approved):
I knew there was something I was missing. I forgot we set it in two places!

PR #712, opened 2024-03-05
Reduce SCE file sizes

### LINKED ISSUE #710: Ideas for reducing file sizes

Our files are pretty big, especially RDS files. We could save storage and download space with some smallish changes.

A simple and transparent to users change would be to change our RDS compression from gzip to bzip2. This can be implemented with a change in `write_rds()` from `compress='gz'` to `compress='bz2'`, and will be completely transparent to users as far as code goes, as the result is still an `rds` file

In my testing, this takes about 50% longer (e.g. 35 seconds vs 21 seconds, though so not much really), but results in files ~50% smaller. However, the read time increases substantially, from ~1.4 to 9.4 seconds in the example I tested. I tend to think the tradeoff is likely worth it for our use case, but others may disagree, so we should discuss this!

One other idea I had was to remove the `miQC_model` object from the "processed" files. At that point, the data in the object can't really be used as the rejected cells have already been removed. This object is also fairly large (100MB or more) and does not appear to compress well, so removing it would save a significant amount of space, and having it in both the "filtered" and "processed" data seems redundant. This change would require a docs update as well.
 
I also tested `xz` compression, but this was _much_ slower, on the order of 4 minutes for the same file as above, with similar compression to `bz2` (a bit smaller). It did read about 2x as fast as bz2 though. Something about all this doesn't seem quite right, and may be to do with the implementation of xz in R, but it is what it is.

jashapiro:
CLoses #710

Here I changed all compression of SCE RDS files from gzip to bz2, with the exception of merged files, which I left alone with their current faster but less efficient compression algorithm. (Noting also that the gzip there is already set for low compression, high speed)

I also removed the miQC model from the metadata of the processed data only, and from merged objects. I don't think it will be there in normal processing, but it seemed reasonable to remove in case we want to run the merge script on filtered data at some point.

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -175,7 +175,7 @@ if ("cellassign" %in% all_celltypes) {
 
 # Update some SCE information  -------------------------------------------------
 # - Add a new colData column with any additional modalities
-# - Remove cluster parameters from metadata
+# - Remove cluster parameters and miQC model from metadata
 sce_list <- sce_list |>
   purrr::map(\(sce){
     additional_modalities <- altExpNames(sce)
@@ -187,6 +187,7 @@ sce_list <- sce_list |>
     metadata(sce)$cluster_algorithm <- NULL
     metadata(sce)$cluster_weighting <- NULL
     metadata(sce)$cluster_nn <- NULL
+    metadata(sce)$miQC_model <- NULL
 
     return(sce)
   })
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -9,12 +9,12 @@ option_list <- list(
   make_option(
     opt_str = c("--input_library_ids"),
     type = "character",
-    help = "Comma separated list of library IDs corresponding to the libraries being integrated."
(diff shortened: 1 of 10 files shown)


bin/merge_sces.R
```diff
@@ -9,12 +9,12 @@ option_list <- list(
   make_option(
     opt_str = c("--input_library_ids"),
     type = "character",
-    help = "Comma separated list of library IDs corresponding to the libraries being integrated."
```
sjspielman, line 12: 😱

bin/merge_sces.R
```diff
@@ -9,12 +9,12 @@ option_list <- list(
...
+    help = "Comma separated list of library IDs corresponding to the libraries being merged."
   ),
   make_option(
     opt_str = c("--input_sce_files"),
     type = "character",
-    help = "Comma separated list of input sce file paths corresponding to the sces being integrated."
```
sjspielman, line 17: 🙀

allyhawkins (approved):
LGTM

Issue #718, opened 2024-03-06
Future idea: Create merged objects for projects with multiplexed libraries containing all non-multiplexed single-cell libraries


allyhawkins:
**Is your feature request related to a problem? Please describe.**

In #716, we are planning to skip creation of merged objects for any projects with multiplexed libraries. However, we have scenarios where a project could contain some multiplexed libraries and some non-multiplexed libraries. We should still produce a merged object that contains all non-multiplexed single-cell/ single-nuclei libraries. 

**Describe the solution you'd like**

Update the workflow to remove any multiplexed libraries from projects prior to merging. This would create a merged object with only regular single-cell/single-nuclei libraries present.

PR #719, opened 2024-03-06
Skip merging projects with multiplexed libraries

### LINKED ISSUE #716: Skip creation of merged objects

We have decided against creation of merged objects if projects contain multiplexed libraries (for now). 

We should update the `merge.nf` workflow to skip projects containing multiplexed samples. For these projects, no output files would be created.

### LINKED ISSUE #717: Fix column name typos
Should be all lowercase `id`:
https://github.com/AlexsLemonade/scpca-nf/blob/fcf766cc6656f9f90d3b96fa0ae2a6ad8c05b46e/bin/merge_sces.R#L226-L227

allyhawkins:
Closes #716 
Closes #717 

Here I added a line to filter out any projects that have multiplexed libraries from the channel that is used as input to create merged objects. 

Because we have decided against including multiplexed libraries being included in merged objects, I removed any support we had for merging multiplexed libraries in the `merge_sces.R` script. I was a little torn on if we should keep this, but if the plan is to eventually create merged objects for all non-multiplexed libraries in a project, then we don't need this anymore. I can add it back if we want the flexibility to use it in the future, but I think a better change to that script would be to spit out an error if any libraries being merged contained multiplexed data when we address #718. 

As part of that, I also removed the columns to keep referring to cell hash stats and demuxing results, resolving the issue in #717.

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -223,8 +223,8 @@ possible_columns <- c(
   "altexps_cellhash_sum",
   "altexps_cellhash_detected",
   "altexps_cellhash_percent",
-  "hashedDrops_sampleID",
-  "HTODemux_sampleID",
+  "hashedDrops_sampleid",
+  "HTODemux_sampleid",
   "vireo_sampleid"
 )
 
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -35,6 +35,13 @@ option_list <- list(
     help = "Keep any altExp present in the merged object."
   ),
   make_option(
+    opt_str = c("--multiplexed"),
+    action = "store_true",
+    default = FALSE,
+    help = "Indicates if the provided SCE's contain multiplexed data.
+      If so, the sample metadata will not be added to the colData."
+  ),
+  make_option(
     opt_str = c("-t", "--threads"),
     type = "integer",
     default = 1,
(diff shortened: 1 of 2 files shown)


bin/merge_sces.R
jashapiro, line 1: I would probably leave this script alone? Having options that we are not currently using in production does not seem a problem to me, and it might be useful as a base for later testing.

merge.nf
```diff
@@ -162,6 +161,9 @@ workflow {
     grouped_libraries_ch = libraries_ch
       // only include single-cell/single-nuclei which ensures we don't try to merge libraries from spatial or bulk data
       .filter{it.seq_unit in ['cell', 'nucleus']}
+      // remove any multiplexed projects
+      // future todo: only filter library ids that are multiplexed, but keep all other non-multiplexed libraries
+      .filter{!(it.project_id in multiplex_projects.getVal())}
```
jashapiro, line 166: To facilitate the later implementation, why don't we use `.branch` here and then we can log a message about skipping multiplexed libraries in this report.

merge.nf
```diff
@@ -162,6 +161,9 @@ workflow {
...
       // create tuple of [project id, library_id, processed_sce_file]
       .map{[
```
jashapiro, line 168: To facilitate the later implementation, why don't we use .branch here and then we can log a message about skipping multiplexed libraries in this report.

(I used `log.warn`, but maybe just `log.info`?)

```suggestion
    filtered_libraries_ch = libraries_ch
      // only include single-cell/single-nuclei which ensures we don't try to merge libraries from spatial or bulk data
      .filter{it.seq_unit in ['cell', 'nucleus']}
      // remove any multiplexed projects
      // future todo: only filter library ids that are multiplexed, but keep all other non-multiplexed libraries
      .branch{
        multiplexed: it.project_id in multiplex_projects.getVal()
        single_sample: true
      }
    
    filtered_libraries_ch.multiplexed
      .unique{ it.project_id }
      .subscribe{
        log.warn("Not merging ${it.project_id} because it contains multiplexed libraries.")
      }
      
    grouped_libraries_ch = filtered_libraries.single_sample 
      // create tuple of [project id, library_id, processed_sce_file]
      .map{[
```

jashapiro (commented):
Looks good. I made a couple suggestions to make the future easier, one of which was just to leave the script alone.

allyhawkins:
I updated this to use the branching you suggested and then also removed the changes to the script, so that should be the same now, other than addressing the capitalization of `ID` for the demuxing columns. 

I am testing this with running one multiplexed and one non-multiplexed libraries, and I think I like the `log.warn`? It comes up as a message pretty immediately when running so you can see why not as many jobs started as you might expect if a multiplexed library is there: 
Screenshot 2024-03-07 at 9 01 30 AM

jashapiro (approved):
LGTM!

PR #722, opened 2024-03-07
Add max_mem label for Cell Assign

### LINKED ISSUE #709: [BUG] CellAssign process getting killed with `OutOfMemoryError`

allyhawkins:
Closes #709 

Here I added a new label, `max_mem` that uses the 96 gb on the first try and then increases to use the max memory on the second and third try. I'm testing this now, but also waiting for the other test run to actually complete to make sure this solves the problem. I just wanted to get this filed in case we are good to make these changes tomorrow.

### DIFF
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -1,6 +1,6 @@
 // Resource maximum settings
 params.max_cpus = 24
-params.max_memory = 128.GB
+params.max_memory = 488.GB
 
 process{
   memory = {check_memory(4.GB * task.attempt, params.max_memory)}
@@ -24,8 +24,8 @@ process{
   withLabel: mem_96 {
     memory = {check_memory(48.GB  + 48.GB * task.attempt, params.max_memory)}
   }
-  withLabel: mem_256 {
-    memory = {check_memory(128.GB + 128.GB * task.attempt, params.max_memory)}
+  withLabel: mem_488 {
+    memory = {check_memory(256.GB + 256.GB * task.attempt, params.max_memory)}
   }
   withLabel: cpus_2  {
     cpus = {check_cpus(2, params.max_cpus)}
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -1,6 +1,6 @@
 // Resource maximum settings
 params.max_cpus = 24
-params.max_memory = 128.GB
+params.max_memory = 96.GB
 
 process{
   memory = {check_memory(4.GB * task.attempt, params.max_memory)}
@@ -24,8 +24,8 @@ process{
   withLabel: mem_96 {
     memory = {check_memory(48.GB  + 48.GB * task.attempt, params.max_memory)}
   }
-  withLabel: mem_512 {
-    memory = {check_memory(256.GB + 256.GB * task.attempt, params.max_memory)}
+  withLabel: max_mem {
+    memory = {task.attempt > 1 ? check_memory(96.GB, params.max_memory) : params.max_memory}
   }
   withLabel: cpus_2  {
     cpus = {check_cpus(2, params.max_cpus)}
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -1,6 +1,6 @@
 // Resource maximum settings
 params.max_cpus = 24
-params.max_memory = 96.GB
+params.max_memory = 128.GB
 
 process{
   memory = {check_memory(4.GB * task.attempt, params.max_memory)}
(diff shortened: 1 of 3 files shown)


config/process_base.config
```diff
@@ -24,6 +24,9 @@ process{
   withLabel: mem_96 {
     memory = {check_memory(48.GB  + 48.GB * task.attempt, params.max_memory)}
   }
+  withLabel: max_mem {
```
jashapiro, line 27: For parallelism in the tags
```suggestion
  withLabel: mem_max {
```

modules/classify-celltypes.nf
```diff
@@ -49,7 +49,7 @@ process classify_cellassign {
       mode: 'copy',
       pattern: "${cellassign_dir}"
     )
-  label 'mem_96'
+  label 'max_mem'
```
jashapiro, line 52:
```suggestion
  label 'mem_max'
```

jashapiro (approved):
LGTM, with a little name order fix

PR #725, opened 2024-03-11
Handle empty sample type columns and add sentence about references not having tumor cells

### LINKED ISSUE #707: Add sentence to cell type report indicating references come from normal samples
> I think adding a sentence to the section on cell typing in the processing docs would be sufficient, or if we want to make sure users see it we could include it in the report itself.

both? Just a quick, "Please be aware that all cell type references include only normal (not tumor) tissue." would suffice, I think.

_Originally posted by @sjspielman in https://github.com/AlexsLemonade/ScPCA-admin/issues/716#issuecomment-1966685634_

### LINKED ISSUE #713: Add is_xenograft & is_cell_line to example metadata
We need `is_xenograft` and `is_cell_line` columns here: https://github.com/AlexsLemonade/scpca-nf/blob/development/examples/example_sample_metadata.tsv

They are expected here:
https://github.com/AlexsLemonade/scpca-nf/blob/1569842882eadc5a378f5eba51759cde9241ad3d/bin/generate_unfiltered_sce.R#L182-L190

allyhawkins:
Closes #713 
Closes #707 

Here I'm making a few fixes to the workflow: 

- I added a sentence about not having tumor cells in the supplemental cell type report. This is the only place we explicitly talk about references, so I chose to include it only in the supplemental and not in the main report too. 
- I updated `generate_unfiltered_sce.R` to handle missing columns for `is_cell_line` and `is_xenograft`. If both are missing, then `Not provided` is used as the `sample_type`. 
- I will note that we use the `is_cell_line` column specifically when determining if cell typing should be skipped, so I think we do want to note that those columns can be included, even if they aren't required. I accounted for them in the example metadata and readme and then also added a note about it to the external instructions for creating the sample metadata.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -42,12 +42,14 @@ Be sure to enter the **full** path to the directory containing the fastq files i
 You will also need to create a tab-separated values **sample** metadata file.
 At a minimum, the sample metadata file must contain a column with `scpca_sample_id` as the header.
 The contents of this column should contain all unique sample ids that are present in the `scpca_sample_id` column of the run metadata file.
+Additionally, you may include columns indicating the sample type, such as `is_cell_line` and `is_xenograft`.
+If the sample is from a cell line, `is_cell_line` should be `TRUE`, and if the sample is from a patient-derived xenograft, `is_xenograft` should be `TRUE`.
 
 Below is an example of a sample metadata file:
 
-| scpca_sample_id | diagnosis    | age |
-| --------------- | ------------ | --- |
-| sample01        | glioblastoma | 71  |
+| scpca_sample_id |is_cell_line | is_xenograft | diagnosis    | age |
+| --------------- |-------------| -------------|------------ | --- |
+| sample01        | FALSE       | FALSE        | glioblastoma | 71  |
 
 **Note that the `diagnosis` and `age` columns are shown as example sample metadata one might include in the sample metadata file.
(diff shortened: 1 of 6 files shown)


examples/README.md
```diff
@@ -42,12 +42,14 @@ Be sure to enter the **full** path to the directory containing the fastq files i
 You will also need to create a tab-separated values **sample** metadata file.
 At a minimum, the sample metadata file must contain a column with `scpca_sample_id` as the header.
 The contents of this column should contain all unique sample ids that are present in the `scpca_sample_id` column of the run metadata file.
+Additionally, you may include columns indicating the sample type, such as `is_cell_line` and `is_xenograft`.
+If the sample is from a cell line, `is_cell_line` should be `TRUE`, and if the sample is from a patient-derived xenograft, `is_xenograft` should be `TRUE`.
```
sjspielman, line 46: UGH github deleted my comment here! I suggested a rephrase (please check my bullet spacing) to be more explicit - 


```
Additionally, you may include columns `is_cell_line` and `is_xenograft` to indicate the sample type:

- `is_cell_line`: Use `TRUE` if the sample is from a cell line and `FALSE` otherwise.
Cell type annotation will not be performed for samples that are `TRUE`.
- `is_xenograft`: Use `TRUE` is the sample is from a patient-derived xenograft and `FALSE` otherwise.

This information will be reflected in the summary QC report.
```

allyhawkins:
I updated the bullets and also added them to the external instructions. This should be ready for another look.

PR #726, opened 2024-03-11
Use `mem_max` for all merging and sync changes from `main`

allyhawkins:
In preparation for #654, I'm adjusting the processes in the merging workflow to use `mem_max`. I think we want to be able to adjust memory for all processes since after the merged object is created, we may need a lot of memory to read them back into R for the report and exporting AnnData objects. 

While I was here, I also prepped for #654 by resolving any conflicts between `main` and `development`. So there are a few changes that are getting added to keep things in `sync`.


jashapiro (approved):
LGTM

PR #730, opened 2024-03-15
Fix bugs in v0.8.0

allyhawkins:
In processing projects through v0.8.0, I found two bugs that needed to be addressed. 

- The first was we are trying to adjust the linewidth on the miQC plot, independent of if the miQC plot just plots the metrics or the models. If the metrics are plotted without the model, then no lines are present to adjust. I just moved that to be within the `if` statement that actually creates the miQC plot with the model. 
- The second was an extra `&` that seemed to be causing an error in evaluating the `sample_type` for multiplex samples. When I removed this then things worked as expected. The double & only looks at the first element and with multiplex we have a vector. 
- While I was here, I bumped up the memory usage for the fry quant process and qc report, since we were seeing a reasonable number of failures. 

I tested this with the two libraries that were causing issues, so this at least fixes the errors I've come across so far.

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -200,7 +200,7 @@ sample_type <- sample_metadata_df |>
       is_xenograft ~ "patient-derived xenograft",
       is_cell_line ~ "cell line",
       # if neither column was provided, note that
-      is.na(is_xenograft) && is.na(is_cell_line) ~ "Not provided",
+      is.na(is_xenograft) & is.na(is_cell_line) ~ "Not provided",
       .default = "patient tissue"
     )
   ) |>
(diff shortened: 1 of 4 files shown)


bin/generate_unfiltered_sce.R
```diff
@@ -200,7 +200,7 @@ sample_type <- sample_metadata_df |>
       is_xenograft ~ "patient-derived xenograft",
       is_cell_line ~ "cell line",
       # if neither column was provided, note that
-      is.na(is_xenograft) && is.na(is_cell_line) ~ "Not provided",
+      is.na(is_xenograft) & is.na(is_cell_line) ~ "Not provided",
```
jashapiro, line 203: Ah good catch.

jashapiro (approved):
LGTM!

PR #734, opened 2024-04-05
Reduce merge memory a touch

jashapiro:
Now that one of the merges has completed successfully, it was time to submit the changes I included to reduce memory usage during the merge. 

- I moved some of the SCE trimming to when we read in the files (though I realize this won't matter now that we don't have the miQC object in processed data anyway) .
- I also removed a place where we duplicate the list unnecessarily.
- In probably the biggest effect, I remove the `sce_list` once the merged object is created.
- Threw in some `gc()` calls.

I also added an explicit `long_running` tag to shift to the priority queue more quickly, and bumped up the `mem_max` base memory.

### DIFF
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -25,7 +25,7 @@ process {
     memory = {check_memory(48.GB  + 48.GB * task.attempt, params.max_memory)}
   }
   withLabel: mem_max {
-    memory = {task.attempt > 1 ? params.max_memory : check_memory(192.GB, params.max_memory)}
+    memory = {(task.attempt > 2  && task.exitStatus in 137..140) ? params.max_memory : check_memory(192.GB * task.attempt, params.max_memory)}
   }
   withLabel: cpus_2  {
     cpus = {check_cpus(2, params.max_cpus)}
--- a/config/process_base.config
+++ b/config/process_base.config
@@ -25,7 +25,7 @@ process {
     memory = {check_memory(48.GB  + 48.GB * task.attempt, params.max_memory)}
   }
   withLabel: mem_max {
-    memory = {task.attempt > 1 ? params.max_memory : check_memory(96.GB, params.max_memory)}
+    memory = {task.attempt > 1 ? params.max_memory : check_memory(192.GB, params.max_memory)}
   }
   withLabel: cpus_2  {
     cpus = {check_cpus(2, params.max_cpus)}
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -100,21 +100,37 @@ if (opt$threads > 1) {
   bp_param <- BiocParallel::SerialParam()
 }
 
+# Functions
+
+read_trim_sce <- function(sce_file) {
+  # Read in SCE and update some information to reduce size/memory usage
+
+  sce <- readr::read_rds(sce_file)
+  if (!is(sce, "SingleCellExperiment")) {
+    stop("Input file must contain a `SingleCellExperiment` object.")
+  }
+  # Update some SCE information  -------------------------------------------------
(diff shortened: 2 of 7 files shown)


jashapiro:
According to Tower, SCPCP00003 actually only used about 160GB memory, so I changed the error strategy a bit to only go to the actual max if there was an OOM error.

config/process_base.config
```diff
@@ -25,7 +25,7 @@ process {
     memory = {check_memory(48.GB  + 48.GB * task.attempt, params.max_memory)}
   }
   withLabel: mem_max {
-    memory = {task.attempt > 1 ? params.max_memory : check_memory(96.GB, params.max_memory)}
+    memory = {(task.attempt > 2  && task.exitStatus in 137..140) ? params.max_memory : check_memory(192.GB * task.attempt, params.max_memory)}
```
allyhawkins, line 28:
```suggestion
    memory = {(task.attempt > 1  && task.exitStatus in 137..140) ? params.max_memory : check_memory(192.GB * task.attempt, params.max_memory)}
```
I think if it's a memory failure on the first try then we want to increase the memory rather than run it again?

allyhawkins (commented):
Just one question about why you increased the task attempts to be 2 before increasing the memory if there's a memory failure.

config/process_base.config
allyhawkins: Ah okay, I didn't put together that by adding the `192.GB * task.attempt` you were still increasing the memory. This makes sense.

allyhawkins (approved):
This looks good so I'm going to go ahead and approve. One thing that we will need to decide is if we want to still include this merged object or not. We will need to release these changes and re-generate the merged object since it wasn't run with an scpca-nf version though. Ultimately I think we want to do that, but I think there's a question of when we actually do that and how that times with merged objects being released on the Portal. 

Maybe let's wait and see what happens to the other project before making a decision?

allyhawkins:
> I think that was my view. The SCPCP000003 job "only" took ~26 hours, but export to AnnData didn't work on the first try (OOM), so that is still running. I'm not sure that the version matters too much for these specific jobs though... if we run with nextflow run merge.nf does that record the version the same way as running from the repo? Or maybe I invoked it wrong...

You can specify the version with that script by using the `--workflow_version` argument, which by default will use `development`. But you are right that I guess we don't actually record any version information for merged objects. maybe we should create something similar to `scpca-meta.json` that includes that information.

allyhawkins:
> In https://github.com/AlexsLemonade/scpca-nf/commit/9b0379b1fe14d26f19700a272894422e3b289bdc I added some debug messages for logging, and they seemed to confirm that the error is occuring during the actual export step. We get past reading and formatting, but fail before getting to ADT.

Does this mean you are producing a `merged_rna.hdf5`, but not `merged_adt.hdf5`? 

> We could potentially merge this version in its current state, and come back with a bug fix, but I think we probably want to at least test it with a couple smaller merges before doing so.

We can test this with the data that's in the `scpca/processed` folder and do SCPCP000001 and SCPCP000003. Smaller versions of those projects are there for testing. 

In general though, I think I would agree about merging this in and then debugging for SCPCP000008 separately. We can release all the other merged objects in the meantime.

jashapiro:
> Does this mean you are producing a `merged_rna.hdf5`, but not `merged_adt.hdf5`?

No, we are not producing the hdf5 file at all, as far as I can tell. I can't see the files themselves on batch, but I have a message printing immediately before and immediately after the hdf5 file is created. The one before prints but there is a failure before the next one.

allyhawkins (approved):
Going to approve with the note that we will need to revisit the workflow to deal with larger projects like SCPCP000008.

Issue #736, opened 2024-04-09
Discussion: Rename AnnData objects with .h5ad extension


jashapiro:
We have been using `.hdf5` as the extension for AnnData output, but the preferred file extension for these files according to [Anndata docs](https://anndata.readthedocs.io/en/latest/fileformat-prose.html) is the more-specific `.h5ad`, which is also used extensively in other databases.

Making this change will also involve numerous changes to https://github.com/AlexsLemonade/scpca-docs/, https://github.com/AlexsLemonade/scpca-portal/, and https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/357. Most of these should be pretty straightforward, but will include changes to import scripts, website copy, docs copy and docs images. 

Despite the effort involved, I do think this is a change worth making (coupled with some future major release, and before we decide to designate anything as 1.0) to make our output a bit easier to integrate into other workflows.

Tagging @davidsmejia & @dvenprasad for discussion, as I doubt they are watching this repo... science team members will also want to weigh in, I am sure!

dvenprasad:
Seconding David that we don't mention the file extensions on the front-end. The doc illustration changes shouldn't be too.

We'll have to update the AnnData READMEs too.

allyhawkins:
I'm on board with updating the extension. We talked at AACR about making the update when packaging up the files to detect either `.hdf5` or `.h5ad` and then making sure the files included in the zip files are all `.h5ad`. That means we can do this without the science team having to re-process the data. We would then make the changes to the docs and README files and update the workflow to output `.h5ad` for any future processing. 
@davidsmejia is this something that would be feasible?

sjspielman:
@jashapiro Did you mean to link https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/357 here? Or was that meant to be something else?

jashapiro:
> @jashapiro Did you mean to link [AlexsLemonade/OpenScPCA-analysis#357](https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/357) here? Or was that meant to be something else?

I must have meant something else.

jashapiro:
Completed!

PR #737, opened 2024-04-11
Change Anndata extension to h5ad

jashapiro:
This PR should accomplish the main code changes required for #736.

I basically just searched for `hdf5` and updated most of those to `h5ad` as appropriate. The other "major" change was to the arguments for the `predict_cellassign.py` script, which I changed to be more consistent with other scripts and just use `anndata_file` rather than referring to the extension itself.

I updated the docs in this repo that referred to hdf5 files here too. But of course there are other docs changes that will be required.


allyhawkins (approved):
Thanks for doing this!

PR #738, opened 2024-04-15
Skip processing if no cells remain after empty droplet filtering

### LINKED ISSUE #682: Deal with libraries that don't have any cells after removing empty droplets

### LINKED ISSUE #735: [BUG] Age in sample_metadata is inconsistently typed
**Describe the bug**

In the `metadata(sce)$sample_metadata` table, we sometimes have the `age` column as a `double` and sometimes a `character`. 

**To reproduce**
Using the processed output files for two libraries: 

```
> type(metadata(SCPCL000822)$sample_metadata$age)
[1] "double"
> type(metadata(SCPCL000608)$sample_metadata$age)
[1] "character"
```

**Expected behavior**

Always one type!

**Version information**

I believe these outputs are from the latest scpca-nf output, but I am not sure... we should check with 0.8.0 outputs

**Additional context**

The error could easily be in `scpcaTools`; I have not yet gone to look at where the data is read in.

allyhawkins:
Closes #682 
Closes #735 

This PR makes some adjustments to the workflow to account for any libraries that have no cells after removing empty droplets. I took a very similar approach to how we handle objects that have 0 cells after removing low-quality cells in the processed object. 

- In the `filter_sce.R` script, I add a check for the number of columns after removing empty droplets. If that is equal to 0, then an empty `filtered.rds` is created, otherwise processing proceeds as normal. 
- If the `filtered.rds` file is empty, then any additional steps are skipped. The log will include a note about any libraries that have no cells. 
- Finally, the filtered and processed rds are written out as empty files.

While I was here I also accounted for the sample metadata issue and made sure we read in all columns as characters before adding them to the object. 

**Questions for reviewers:**

- First, do we like this approach of creating an empty file? I do wonder if we want to change to logging an error and stopping the workflow? I think it's important to note that there is an error, but I also don't want to kill a run of multiple libraries because one of them is poor quality...
- Right now, we publish an empty processed file and we don't account for that in documentation. I actually think if the file is empty, maybe we just don't want to include it? This would have other implications on the Portal itself since we would have to make sure the script can accommodate missing rds files when zipping everything up. Alternatively, if we keep this route of providing empty files we should just make a note in the documentation.


allyhawkins:
The stub workflow was failing because everything was getting flagged as having no cells since they will all be empty files. I updated this to not consider stub files when checking file size after filtering and processing.

bin/filter_sce.R
```diff
@@ -95,111 +95,117 @@ filtered_sce <- scpcaTools::filter_counts(unfiltered_sce)
...
-
-# grab names of altExp, if any
-alt_names <- altExpNames(filtered_sce)
+# if no cells left, write out empty file, otherwise continue processing
+if (ncol(filtered_sce) == 0) {
+  # make an empty filtered file
+  file.create(opt$filtered_file)
+} else {
+  # need to remove old gene-level rowData statistics first
+  drop_cols <- colnames(rowData(filtered_sce)) %in% c("mean", "detected")
```
jashapiro, line 104: Minor suggestion, just to reduce nesting: You can add an exit in this block, then leave the rest unchanged.

```suggestion
# if no cells left, write out empty file and quit, otherwise continue processing
if (ncol(filtered_sce) == 0) {
  # make an empty filtered file
  file.create(opt$filtered_file)
  quit(save = "no")
}
 
# need to remove old gene-level rowData statistics first
drop_cols <- colnames(rowData(filtered_sce)) %in% c("mean", "detected")
```

main.nf
```diff
@@ -209,7 +209,17 @@ workflow {
     .filter{it[0]["library_id"] in rna_only_libs.getVal()}
   // make rds for rna only
   rna_sce_ch = generate_sce(rna_quant_ch, sample_metafile)
+    // only continue processing any samples with > 0 cells left after filtering
+    .branch{
+      continue_processing: it[2].size() > 0 || it[2].name.contains("STUBL")
```
jashapiro, line 214: Will this work? It is a bit more specific. (though I might reduce the string down to just `"STUB"`)
```suggestion
      continue_processing: it[2].size() > 0 || it[2].name.startsWith("STUBL")
```

jashapiro (approved):
This looks good to me, but I made a suggestion that would reduce the number of lines changed in the script by just exiting rather than having an `else` block. 

I don't think I need to see this again though.

PR #742, opened 2024-04-16
Account for demultiplexing without genetic demultiplexing

### LINKED ISSUE #740: [BUG] Account for grabbing estimated demux cell counts for libraries with no genetic demultiplexing
**Describe the bug**

When looking into our test data more carefully, I noticed that we have an empty array for `sample_cell_estimates` in the final `metadata.json` for all test multiplexed libraries. Instead there should be an array with a sample count for each sample in the multiplexed library. 

These counts are estimated in `sce_qc_report.R` by grabbing the cell counts for one of the demux methods. 
https://github.com/AlexsLemonade/scpca-nf/blob/edc3c2680235cb92d15ebd751abe021e8fbe5574/bin/sce_qc_report.R#L235-L250

However, the default `demux_method` is set to `vireo` in this script and we don't actually provide that argument in the module that runs this script. So for every library, regardless of if genetic demultiplexing was run or not, it's going to grab the demux counts from the column output by vireo. This isn't an issue with any of our current samples on the Portal, but will be an issue if we ever have samples that are multiplexed and don't have matching bulk or if a user is running without genetic demultiplexing. This also affects the test data, since we don't run genetic demultiplexing on those samples, so we probably want to fix it. 

**Proposed solution**

I think we should add a key to the `meta` object indicating if genetic demux was used or not. If it's not used then we want to provide a different `demux_method` (either HashedDrops or HTODemux) to the script. Otherwise, vireo is always used.

allyhawkins:
Closes #740 

Here, I am making a small adjustment to include sample cell estimates for multiplexed libraries when no genetic demultiplexing is run. We already had an argument, `demux_method,` set up in the script to generate the metadata and QC report, but we were not modifying the default of `vireo` regardless of whether genetic demultiplexing was run or not. 

I added another value to the input channel for the QC report indicating if that sample was demultiplexed using genetic demultiplexing. Then, that gets used to determine the value used for `demux_method` when running `sce_qc_report.nf`. If no genetic demultiplexing was used, I'm using `hashedDrops`.  I think we could add a parameter to specify using hashedDrops or HTODemux, but that's probably overkill. 

I did originally think we should add a key to the meta object, but if we want to make sure every library has that key in their meta object we need to do it before any quantification happens. But then if mapping is skipped the meta object that is in `scpca-meta.json` won't have that value. I don't think this is worth making an update to all the `scpca-meta.json` files, so I thought adding it right before we actually need it made the most sense.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -187,7 +187,13 @@ if (!is.null(opt$cellassign_predictions)) {
     stop("Cell type reference metadata filename must be provided")
   }
 
-  predictions <- readr::read_tsv(opt$cellassign_predictions)
+  # if the predictions file isn't emtpy read it in
+  if (file.size(opt$cellassign_predictions) > 0) {
+    predictions <- readr::read_tsv(opt$cellassign_predictions)
+  } else {
+    # if it's empty, then sce could not be converted to anndata and cell assign was not run
+    sce$cellassign_celltype_annotation <- "Not run"
+  }
 
   # if the only column is the barcode column then CellAssign didn't complete successfully
   # otherwise add in cell type annotations and metadata to SCE
--- a/bin/add_demux_sce.R
+++ b/bin/add_demux_sce.R
@@ -107,21 +107,27 @@ cellhash_ids <- rownames(altExp(sce))
 
 # only perform demultiplexing if more than one HTO is detected
 if (length(cellhash_ids) > 1) {
+  demux_methods <- c()
   # add HashedDrops results
   if (opt$hash_demux) {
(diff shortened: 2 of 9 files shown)


jashapiro (commented):
This looks fine, but I think I would rather we got the demux method from the SCE rather than the command line, which would mean we could modify only `sce_qc_report.R`. I feel like that will be more robust to future changes. I think I might remove the  `--demux-method` argument and just get that value from what is present in the files?

jashapiro (approved):
LGTM

I can see a world where we might want to make the preferred demux method an option again, but I think this is fine for the forseeable future

PR #743, opened 2024-04-17
Skip CellAssign if only 1 cell

### LINKED ISSUE #731: Make sure CellAssign is skipped for any objects with just 1 cell
In running through the projects again with cell typing, I noticed a few of the newer low quality samples that we got that aren't yet on the Portal failing because they only had 1 cell when trying to run the process for CellAssign. 

We currently account for the case where there are 0 cells left in the processed object by writing an empty file and skipping any downstream processing. However, we didn't account for if there was only 1 cell. `zellkonverter` doesn't like this, so we want to avoid passing any objects with just 1 cell to `classify_cellassign` and trying to convert to HDF5. 

I haven't quite figured out the best way to do this yet, but we will want to get this fixed before processing the next round of samples for release.

allyhawkins:
Closes #731 

Here I'm making sure we skip trying to run CellAssign if a processed object only has 1 cell. The problem is that we can't convert an object with 1 cell to an AnnData object, so then there's no AnnData object to use as input to CellAssign. 

To handle this I added a check for the number of cells in the SCE object in `sce_to_anndata.R`. If < 2 cells, then the script just quits and doesn't do any conversions. This means no `.h5ad` file gets created. If that's the case then we skip running the `predict_cellassign.py` script and just make an empty predictions file to output from the process. Then when annotations get added to the SCE object, if the predictions file is empty, all cells get annotated with `Not run`. This is the same value we use when CellAssign isn't run because of too few cells. 

While testing, I ran into small errors in the QC report that I also fixed here. Moving the miQC lines was something we had done on `main`, but seemed to be missing from `development`.

### DIFF
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -69,16 +69,25 @@ process classify_cellassign {
         --input_sce_file "${processed_rds}" \
         --output_rna_h5 "processed.h5ad"
 
-    # Run CellAssign
-    predict_cellassign.py \
-      --anndata_file "processed.h5ad" \
-      --output_predictions "${cellassign_dir}/cellassign_predictions.tsv" \
-      --reference "${cellassign_reference_file}" \
-      --seed ${params.seed} \
-      --threads ${task.cpus}
+    # only run cell assign if a h5ad file was able to be created
+    # otherwise create an empty predictions file
+    if [[ -e "processed.h5ad" ]]; then
 
-    # write out meta file
-    echo '${meta_json}' > "${cellassign_dir}/scpca-meta.json"
+      # Run CellAssign
+      predict_cellassign.py \
+        --anndata_file "processed.h5ad" \
+        --output_predictions "${cellassign_dir}/cellassign_predictions.tsv" \
+        --reference "${cellassign_reference_file}" \
+        --seed ${params.seed} \
+        --threads ${task.cpus}
+
+      # write out meta file
+      echo '${meta_json}' > "${cellassign_dir}/scpca-meta.json"
+
+    else
+      touch "${cellassign_dir}/cellassign_predictions.tsv"
+    fi
+
     """
   stub:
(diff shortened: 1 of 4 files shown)


modules/classify-celltypes.nf
```diff
@@ -69,16 +69,25 @@ process classify_cellassign {
...
+      # Run CellAssign
+      predict_cellassign.py \
+        --anndata_file "processed.h5ad" \
+        --output_predictions "${cellassign_dir}/cellassign_predictions.tsv" \
+        --reference "${cellassign_reference_file}" \
+        --seed ${params.seed} \
+        --threads ${task.cpus}
+
+      # write out meta file
+      echo '${meta_json}' > "${cellassign_dir}/scpca-meta.json"
```
jashapiro, line 85: I assume we want to do this either way? So move this out of the `if` block?

templates/qc_report/main_qc_report.rmd
```diff
@@ -130,6 +130,9 @@ if (has_processed) {
 } else {
   has_umap <- FALSE
   has_celltypes <- FALSE
+  has_singler <- FALSE
+  has_cellassign <- FALSE
+  has_submitter <- FALSE
```
jashapiro, line 135: Just to keep order consistent
```suggestion
  has_singler <- FALSE
  has_cellassign <- FALSE
  has_submitter <- FALSE
  has_celltypes <- FALSE
```

modules/classify-celltypes.nf
```diff
@@ -69,16 +69,25 @@ process classify_cellassign {
...
+
+    else
+      touch "${cellassign_dir}/cellassign_predictions.tsv"
```
jashapiro, line 88: We should probably `touch` this file in the stub too?

jashapiro (approved):
Looks good. I had a few minor suggestions, only one of which has a formal suggestion block, but I should not need to see this again.

PR #745, opened 2024-04-19
Account for no cells being assigned to a sample when demultiplexing

allyhawkins:
Apparently, with our test data using `hashedDrops` doesn't assign cells to any of our samples. This prints out an empty list again, which we want to avoid. I checked, and using `HTODemux` only assigns cells to one sample but not all of them. This leads to a list with just one value while the list of sample IDs is longer. This would mean there is no way to match which cell count goes to which sample. 

If no cells are assigned, we should set to 0 rather than having an empty list. This PR makes that adjustment by looking for any missing sample IDs and then assigning those to 0.

### DIFF
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -239,12 +239,9 @@ if (multiplexed) {
   # get cell count estimates
   demux_column <- paste0(demux_method, "_sampleid")
   demux_counts <- colData(filtered_sce)[[demux_column]] |>
+    factor(levels = sample_ids) |> # use a factor get any zero counts
     table() |>
     as.list() # save as a list for json output
-
-  # for any ids that have no cells, set count to 0
-  missing_sample_ids <- setdiff(sample_ids, names(demux_counts))
-  demux_counts[missing_sample_ids] <- 0
 
   # add demux info to the metadata list
   metadata_list <- append(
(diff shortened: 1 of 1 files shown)


bin/sce_qc_report.R
```diff
@@ -242,12 +242,16 @@ if (multiplexed) {
     table() |>
     as.list() # save as a list for json output
 
+  # for any ids that have no cells, set count to 0
+  missing_sample_ids <- setdiff(sample_ids, names(demux_counts))
+  demux_counts[missing_sample_ids] <- 0
+
```
jashapiro, line 248: Your solution works, but a more direct way may be to have the `table()` work on a factor, with the ids already set. Then it will include zeros and preserve the order
```suggestion
    factor(levels = sample_ids) |> # use a factor to get zero counts
    table() |>
    as.list() # save as a list for json output
```

bin/sce_qc_report.R
```diff
@@ -242,12 +242,16 @@ if (multiplexed) {
...
   # add demux info to the metadata list
   metadata_list <- append(
     metadata_list,
     list(
       demux_method = demux_method,
-      demux_samples = sample_ids,
+      demux_samples = names(demux_counts), # make sure order of sample ids matches counts order
```
jashapiro, line 254: You could leave this either way, since the order should match with my suggestion

jashapiro (approved):
I proposed an alternative method which is a bit more direct, but it should be fine either way, really.

PR #753, opened 2024-04-24
Fix bugs with generating example data

allyhawkins:
When working on generating the updated example data for #748, I found a bug in `main.nf` where we were missing a `params.` call before a variable. I also realized we were missing the cellhash pools files in the example config, so it was looking for a file that doesn't exist and giving an error. I added the cellhash pool file that we have in examples to S3 and then to the profile for example. Alternatively, I could have just left it empty in the example config, but since we have the file we mine as well use it. This means that all example metadata files should now also be on S3.


sjspielman (approved):
LGTM!

PR #755, opened 2024-04-25
Fix bugs for dealing with processed objects with only 1 cell

allyhawkins:
While processing data for the Portal, I found an object that only has 1 cell in the processed object so it tested out some of the code we added to handle that scenario. There were a few issues that I had to fix to get this to work and then I also made a design change that we will need to decide on if it's the right move or not. 

- If any object only has 1 cell then it doesn't have PCA or clusters. We have a check for PCA/UMAP when generating the reports, but not for clusters. Here I added a `has_clusters` variable to make sure no plots that use clusters are created. 
- There was also a small error in the `add_celltypes_to_sce.R` where we were still attempting to add in cell type annotations if the predictions file was empty. So I adjusted the logic there to account for the missing file properly. 

The other thing I did was choose to not create the supplemental report if there's only 1 cell in the object. The report was just some text with one table showing the assignment of that single cell in SingleR, but no other plots. This doesn't seem super informative to me, and we would still have the table in the main QC report. What do others think of this choice? This should not affect very many samples, but we would want to make sure we check how this affects zipping up the files for the Portal. 

@davidsmejia Would you guys have to make changes when zipping up the files for the Portal if a sample was missing the `celltype-report.html` file? I assume you have checks to make sure the files are all there, so would we be able to make the existence of this file optional?

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -191,13 +191,13 @@ if (!is.null(opt$cellassign_predictions)) {
   if (file.size(opt$cellassign_predictions) > 0) {
     predictions <- readr::read_tsv(opt$cellassign_predictions)
   } else {
-    # if it's empty, then sce could not be converted to anndata and cell assign was not run
-    sce$cellassign_celltype_annotation <- "Not run"
+    predictions <- NULL
   }
 
-  # if the only column is the barcode column then CellAssign didn't complete successfully
+  # if the only column is the barcode column or if the predictions file was empty
+  # then CellAssign didn't complete successfully
   # otherwise add in cell type annotations and metadata to SCE
-  if (all(colnames(predictions) == "barcode")) {
+  if (is.null(predictions) | all(colnames(predictions) == "barcode")) {
     # if failed then note that in the cell type column
     sce$cellassign_celltype_annotation <- "Not run"
   } else {
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -197,7 +197,7 @@ if (!is.null(opt$cellassign_predictions)) {
   # if the only column is the barcode column or if the predictions file was empty
   # then CellAssign didn't complete successfully
   # otherwise add in cell type annotations and metadata to SCE
-  if (is.null(predictions) | all(colnames(predictions) == "barcode")) {
+  if (is.null(predictions) || all(colnames(predictions) == "barcode")) {
     # if failed then note that in the cell type column
     sce$cellassign_celltype_annotation <- "Not run"
   } else {
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -370,7 +370,7 @@ create_celltype_n_table(celltype_df, cellassign_celltype_annotation) |>
 ```
 
 
-```{r, eval = has_umap & has_clusters}
+```{r, eval = has_umap && has_clusters}
 knitr::asis_output(glue::glue("
 ## UMAPs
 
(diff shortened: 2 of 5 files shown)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -370,7 +370,7 @@ create_celltype_n_table(celltype_df, cellassign_celltype_annotation) |>
 ```
 
 
-```{r, eval = has_umap}
+```{r, eval = has_umap & has_clusters}
```
jashapiro, line 373:
```suggestion
```{r, eval = has_umap && has_clusters}
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -398,7 +398,7 @@ glue::glue("
 ```
 
 
-```{r eval = has_umap, message=FALSE, warning=FALSE}
+```{r eval = has_umap & has_clusters, message=FALSE, warning=FALSE}
```
jashapiro, line 401:
```suggestion
```{r eval = has_umap && has_clusters, message=FALSE, warning=FALSE}
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -418,7 +418,7 @@ if (length(levels(umap_df$cluster)) <= 8) {
 ```
 
 
-```{r, eval = has_umap}
+```{r, eval = has_umap & has_celltypes}
```
jashapiro, line 421:
```suggestion
```{r, eval = has_umap && has_celltypes}
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -416,7 +420,7 @@ glue::glue("
 ```
 
 <!-- If not multiplexed, show the header, text, and heatmap --> 
-```{r, eval = !has_multiplex, results='asis'}
+```{r, eval = !has_multiplex & has_clusters, results='asis'}
```
jashapiro, line 423:
```suggestion
```{r, eval = !has_multiplex && has_clusters, results='asis'}
```

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -448,7 +452,7 @@ plot_height <- calculate_plot_height(
 ```
 
 
-```{r, eval = !has_multiplex, fig.height = plot_height, fig.width = 8.5, warning = FALSE}
+```{r, eval = !has_multiplex & has_clusters, fig.height = plot_height, fig.width = 8.5, warning = FALSE}
```
jashapiro, line 455:
```suggestion
```{r, eval = !has_multiplex && has_clusters, fig.height = plot_height, fig.width = 8.5, warning = FALSE}
```

bin/add_celltypes_to_sce.R
```diff
@@ -191,13 +191,13 @@ if (!is.null(opt$cellassign_predictions)) {
...
-    sce$cellassign_celltype_annotation <- "Not run"
+    predictions <- NULL
   }
 
-  # if the only column is the barcode column then CellAssign didn't complete successfully
+  # if the only column is the barcode column or if the predictions file was empty
+  # then CellAssign didn't complete successfully
   # otherwise add in cell type annotations and metadata to SCE
-  if (all(colnames(predictions) == "barcode")) {
+  if (is.null(predictions) | all(colnames(predictions) == "barcode")) {
```
jashapiro, line 200:
```suggestion
  if (is.null(predictions) || all(colnames(predictions) == "barcode")) {
```

jashapiro (commented):
LGTM. I just had a number of linting places where I changed `&` -> `&&` and `|` -> `||`. None of them matter in this case, but good practice.

allyhawkins:
@jashapiro are we okay with the decision to not create a supplemental report? I just want to be sure before I merge this and then update the current release.

jashapiro:
> @jashapiro are we okay with the decision to not create a supplemental report? I just want to be sure before I merge this and then update the current release.

Yes, I am fine with that decision!

allyhawkins:
> Yes, I am fine with that decision!

Can you approve please? 😄

jashapiro (approved):
Sorry, thought I did!

PR #757, opened 2024-05-01
Fix some merge bugs

allyhawkins:
In merging the processed data for a project for the Portal, I came across a small error in the `merge_sces.R` script. R was complaining about calculating the `sum()` of a list so I just added an `unlist()` and that fixed that problem. 

The other precautionary measure I took here is to ensure that if a processed file exists but its size is 0, it's not actually included in the merging process. I can imagine that causing problems down the line if we try to read in an empty file. I also added a warning message that should print out any libraries that are not going to be included in the object.

### DIFF
--- a/merge.nf
+++ b/merge.nf
@@ -178,6 +178,15 @@ workflow {
         log.warn("Not merging ${it.project_id} because it contains multiplexed libraries.")
       }
 
+    // print out warning message for any libraries not included in merging
+    merge_libaries = filtered_libraries_ch
+      .collect{it.library_id}
+    libraries_ch
+    .filter{!(it.library_id in merge_libaries.getVal())}
+    .subscribe{
+      log.warn("Processed files do not exist for ${it.library_id}. This library will not be included in the merged object.")
+    }
+
     grouped_libraries_ch = filtered_libraries_ch.single_sample
       // create tuple of [project id, library_id, processed_sce_file]
       .map{[
--- a/merge.nf
+++ b/merge.nf
@@ -179,7 +179,7 @@ workflow {
       }
 
     // print out warning message for any libraries not included in merging
-    merge_libaries = filtered_libraries_ch
+    merge_libaries = filtered_libraries_ch.single_sample
       .collect{it.library_id}
     libraries_ch
     .filter{!(it.library_id in merge_libaries.getVal())}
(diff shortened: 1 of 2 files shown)


merge.nf
```diff
@@ -178,15 +178,24 @@ workflow {
...
       }
 
+    // print out warning message for any libraries not included in merging
+    merge_libaries = filtered_libraries_ch.single_sample
+      .collect{it.library_id}
+    libraries_ch
+    .filter{!(it.library_id in merge_libaries.getVal())}
+    .subscribe{
+      log.warn("Processed files do not exist for ${it.library_id}. This library will not be included in the merged object.")
+    }
```
jashapiro, line 188: Fixing the indentation, but I think the message here is a bit confusing, as there _are_ processed files for multiplexed data, which will be listed again here, along with all of the spatial an bulk libraries. 

```suggestion
    libraries_ch
      .filter{!(it.library_id in merge_libaries.getVal())}
      .subscribe{
        log.warn("Processed files do not exist for ${it.library_id}. This library will not be included in the merged object.")
      }
```

If you want to look for files without processed samples, I think you would have to do that more directly, maybe with a duplicate filter to the one below looking at the files themselves? Something like:
```
filtered_libraries_ch.single_sample
  .map{
        it.library_id,
        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds")
      ]}
  .filter{!(it[1].exists() && it[1].size() > 0)}
  .subscribe{
      log.warn("Processed files do not exist for ${it[0]}. This library will not be included in the merged object.")
    }
```

jashapiro (commented):
These seem like good things to catch! I had a different option for the first, up to you.  But the second I think we want something a bit different?

allyhawkins:
Thanks for the suggestions @jashapiro. I went ahead and incorporated both of them, I think we do want to avoid printing out multiplexed samples in the error message.

jashapiro (approved):
LGTM

merge.nf
```diff
@@ -178,15 +178,26 @@ workflow {
...
       // create tuple of [project id, library_id, processed_sce_file]
       .map{[
         it.project_id,
         it.library_id,
         file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds")
       ]}
-      // only include libraries that have been processed through scpca-nf
-      .filter{file(it[2]).exists()}
+      // only include libraries that have been processed through scpca-nf and aren't empty
+      .filter{file(it[2]).exists() && file(it[2]).size() > 0}
```
jashapiro, line 200: Just a note that because `it[2]` is already a file object, we should be able to drop the wrappers
```suggestion
      .filter{it[2].exists() && it[2].size() > 0}
```

PR #759, opened 2024-05-09
Delete duplicate layer in AnnData objects

jashapiro:
When working on OpenScPCA-nf, I found that when we move raw data in the processed data, we are leaving in place the original "logcounts" layer, making it meaning we have two copies of it around. 

This PR removes that duplication (and merges in changes from `main` to `development`)

Unless we were supposed to leave it for some reason, I think it should save us some space! By my metrics (and logical expectation, nice when they match) this saves ~1/3 of the space for the AnnData processed objects.

I couldn't find any reference to the `layers` in our docs, so this change should not require any updates there.


allyhawkins (approved):
This is fine with me. We don't use the AnnData objects at all after running this script in the workflow so should be good to go. Thanks for merging in those other changes from `main` too!

Issue #761, opened 2024-05-20
Test use of smaller Docker images in workflow


allyhawkins:
Following #760, we will want to test both `scpca-nf` and `merge.nf` with the new Docker images. To do this we should create a testing profile that sets the default run IDs to be one of each type of library. We should also document the addition of this profile and note that it can be used for testing in the future.

allyhawkins:
Closed by #763

PR #762, opened 2024-05-23
Use new, slimmer sometimes, containers 

### LINKED ISSUE #760: Use new smaller images in processes
Before testing as part of #746, we should substitute in the smaller docker images for the full scpcatools image for each process as appropriate.

This means looking at the script(s) for each step and assessing which of the slimmed images will be most appropriate for that process. 

Following this, we will do some additional testing to be sure we made the right choices!

jashapiro:
Closes #760

This PR adds the new scpcaTools containers which upgrade everything to R 4.4.0 and Bioconductor 3.19. I tried to use the slimmest container I could find for each task, and I think I got them all correct, but a bit more testing will be required to be sure that all code paths are tests (which is what #761 is for!)

Before merging these into `main` we will need to cut a release of scpcaTools to fix the docker versions, which are currently set to `edge`.

### DIFF
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -44,7 +44,7 @@ process classify_singler {
 
 
 process classify_cellassign {
-  container params.SCPCATOOLS_CONTAINER
+  container params.SCPCATOOLS_CELLASSIGN_CONTAINER
     publishDir (
       path: "${meta.celltype_checkpoints_dir}",
       mode: 'copy',
@@ -103,7 +103,7 @@ process classify_cellassign {
 }
 
 process add_celltypes_to_sce {
-  container params.SCPCATOOLS_CONTAINER
+  container params.SCPCATOOLS_SLIM_CONTAINER
   label 'mem_4'
   label 'cpus_2'
   tag "${meta.library_id}"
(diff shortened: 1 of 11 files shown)


modules/classify-celltypes.nf
```diff
@@ -44,7 +44,7 @@ process classify_singler {
 
 
 process classify_cellassign {
-  container params.SCPCATOOLS_CONTAINER
+  container params.SCPCATOOLS_CELLASSIGN_CONTAINER
```
allyhawkins, line 47:
```suggestion
  container params.SCPCATOOLS_SCVI_CONTAINER
```
I think this is what you meant?

modules/sce-processing.nf
```diff
@@ -138,7 +138,7 @@ process filter_sce {
 }
 
 process genetic_demux_sce {
-  container params.SCPCATOOLS_CONTAINER
+  container params.SCPCATOOLS_SLIM_CONTAINER
```
allyhawkins, line 141: I can't comment on it, but I think for `process cellhash_demux_sce` you need the Seurat container?

allyhawkins (approved):
Just two minor comments, but when we do the testing we might find some other issues so I'll go ahead and approve.

modules/sce-processing.nf
jashapiro: Yeah, I think I wasn't sure if that had everything we need, but we might as well give it a shot!

PR #763, opened 2024-05-24
Test new docker image and add test samples to profile config

### LINKED ISSUE #761: Test use of smaller Docker images in workflow

allyhawkins:
Closes #761 

I did some testing of the workflow using the changes from #762 and everything ran successfully 🎉 I did not however run anything through `CellAssign`... I can do that though if we would like before closing the testing out. 

To do the testing I was going to create a new test profile, but we already have a `ccdl` profile that specifies a subset of run IDs to run. So rather than create a new profile I'm using the existing `ccdl` profile but have updated the `run_ids` param to include one sample or library from each of the possible modalities - one single-cell with bulk, one CITE, one multiplexed, and one spatial. I think I got everything?

### DIFF
--- a/config/profile_ccdl.config
+++ b/config/profile_ccdl.config
@@ -11,7 +11,9 @@ params {
   results_dir = "${params.outdir}/results"
 
   // a set of run_ids for testing. used only by the main workflow
-  run_ids = "SCPCR000001,SCPCS000101"
+  // one single-cell with bulk, one CITE, one spatial, one multiplexed
+  run_ids = "SCPCS000001,SCPCS000050,SCPCS000203,SCPCL000537"
+
   // include all runs in a merged project. used only by the merged workflow
   merge_run_ids = "All"
 


allyhawkins:
> a test with cellassign (One small sample)

I completed this and results look mostly the same with a few differences in the assigned cell types for CellAssign, but I don't think anything to be concerned about. For reference here's both the new and old cell type report for the same sample: 
[cell-type-reports.zip](https://github.com/AlexsLemonade/scpca-nf/files/15485188/cell-type-reports.zip)

> a merge test (small, again, and I expect this to work based on openscpca-nf)

This is complete without any issues. 

> finally, I do want to make sure we look through these results and don't see any major changes. Looking at the QC reports between what is in prod and the recent runs should be sufficient, I would think?

I looked at the old QC reports currently present in `scpca-prod` for each of the test samples and compared to the new ones. Overall, there is a difference in the number of cells filtered by 1-2 cells which is expected. There was also one library that had a few HVGs switched but the overall HVGs present didn't change. This is probably due to which cells were left after filtering and also not out of the ordinary. 

The biggest differences were seen in the multiplexed library report for the numbers of cells corresponding to each sample as reported by `Seurat::HTODemux`. I feel like this is probably due to some slight changes in Seurat and probably okay to not worry about, but I did want to let you know in case you have other concerns. 
[multiplex-reports.zip](https://github.com/AlexsLemonade/scpca-nf/files/15485304/multiplex-reports.zip)

jashapiro:
I do think it is okay not to worry about, but we should probably make a note in the next release that cell assignments from HTODemux are likely to change. I can't find anything in changelogs about _what_ changed, but I am not at all surprised to see that something might have.

Otherwise, everything looks fine here.

jashapiro (approved):

PR #767, opened 2024-06-26
Move bulk output to separate bulk directory

### LINKED ISSUE #764: Re-order bulk metadata to match order for overall sample metadata
In reviewing the sample metadata, we noticed that the bulk metadata and sample metadata is not in the same order. We have some overlapping columns that are present in both, but in a different order. Although this isn't entirely necessary, it might be a good idea to adjust the order to match what's in https://github.com/AlexsLemonade/scpca-portal/issues/699#issuecomment-2140625806. 

This should be pretty straight forward with some minor edits to [generate_bulk_metadata.R](https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/generate_bulk_metadata.R) to make sure the output TSV has columns in the correct order. 

This should be: 
```
project_id
sample_id
library_id 
seq_unit
technology
total_reads
mapped_reads
genome_assembly
mapping_index
salmon_version
total_reads
date_processed
workflow
workflow_version
workflow_commit
```

### LINKED ISSUE #766: Output bulk data to a `bulk` folder rather than individual files within project directory
Rather than output the bulk data and metadata as individual files within the project directory, we want to output these files to a `bulk` folder. 
This means updating the `publishDir` statement to be `"${params.results_dir}/${meta.project_id}/bulk"` in the line below: 
https://github.com/AlexsLemonade/scpca-nf/blob/a9dc826b8576c48ca38c6ca137f9eeced29c3acc/modules/bulk-salmon.nf#L71.

allyhawkins:
Closes #766 
Closes #764 

Here I just added a `bulk` directory to the existing `publishDir` for the bulk output. This will mean both the quant and metadata files will be saved to `project_id/bulk`. 

While I was here I decided to update the order of the bulk metadata to match the project metadata and `single_cell_metadata.tsv` files that are on the Portal.

### DIFF
--- a/bin/generate_bulk_metadata.R
+++ b/bin/generate_bulk_metadata.R
@@ -89,21 +89,14 @@ bulk_metadata_df <- library_metadata |>
       scpca_project_id %in% opt$project_id
   ) |>
   dplyr::select(
-    scpca_sample_id, scpca_library_id, scpca_project_id,
-    technology, seq_unit
+    scpca_project_id, scpca_sample_id, scpca_library_id,
+    seq_unit, technology
   ) |>
   # rename column names to match format of metadata files from other modalities
   dplyr::rename(
+    project_id = scpca_project_id,
     sample_id = scpca_sample_id,
-    library_id = scpca_library_id,
-    project_id = scpca_project_id
-  ) |>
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
LGTM

PR #769, opened 2024-07-03
Add sample metadata tables to reports 

### LINKED ISSUE #765: Add sample metadata table to QC and cell type reports

See https://github.com/AlexsLemonade/scpca-portal/issues/759

We would like to include a sample metadata table in the main QC report. This should be included along with the processing and library tables that are currently present and should include things like: 

- sample ID
- diagnosis
- subdiagnosis 
- tissue location 
- age 
- sex
- disease timing

We want to include any sample metadata that is consistently collected across samples. 

**Additional context**

We might consider adding in the ontology IDs here as well. We will also have to make sure we account for things being NA as some samples do have missing information.

allyhawkins:
Closes #765 

This PR adds a function to create a sample metadata table to `report_functions.Rmd` and then uses that function in both the main QC report and cell type report to create the sample metadata table. The table is shown for any non-multiplexed samples. Otherwise, we have a nice little blue box that says the metadata will not be displayed since the sample is not demultiplexed. I took the wording from the merge report where we also don't show sample metadata. 

The table is formatted in the same way we have our other tables and I was able to successfully add links for the ontology IDs 🎉. I also accounted for missing ontology IDs and only provide a link if an ID was provided. 

In the main report I adjusted a few headers to account for adding in this table. 

Questions for reviewers: 

- Are we okay with the values that I chose to include? Things I am not displaying are project ID, library ID, submitter ID, submitter, participant ID, or self reported ethnicity term id. 
- How do we feel about the location of the table? 
- I added some explanatory text above the table, but I'm not totally convinced we need it. 

Also, I've been seeing the following error for this table, but I'm not sure how to fix it. The table is still printed, but if we can remove this message that's probably best? 

> Warning message:
In yaml::yaml.load(yaml, handlers = knit_params_handlers(evaluate = evaluate),  :
  an error occurred when handling type 'r'; using default handler

Here's a zip with copies of the updated QC and cell type reports for a multiplexed and non-multiplexed sample: 
[updated_reports.zip](https://github.com/user-attachments/files/16090245/updated_reports.zip)

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -381,6 +381,32 @@ glue::glue("
 ```
 
 
+# Sample metadata 
+
+The below table summarizes clinical metadata for the sample associated with this library. 
+Blue hyperlinks are present for any terms with an ontology term identifier associated with the displayed human readable value. 
+These links will direct you to the [`Ontobee`](https://ontobee.org/) page for that ontology term identifier.  
+
+```{r}
+# extract sce metadata containing processing information as table
+processed_meta <- metadata(processed_sce)
+
+# if data is not multiplexed, print out sample metadata
+if (!has_multiplex) {
+  print_sample_metadata(processed_meta)
+} else {
+  # otherwise print out an info box that no sample metadata will be displayed
+  knitr::asis_output(glue::glue("
+  <div class=\"alert alert-info\">
+
+  This library is multiplexed and contains data from more than one sample.
+  Demultiplexing has not been performed, so sample metadata will not be displayed.
+  </div>
+"))
+}
+```
+
+
 <!------- Call the celltypes_qc report section from the main report ----------->
 ```{r, child='celltypes_qc.rmd'}
 ```
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -385,7 +385,7 @@ glue::glue("
 
 The below table summarizes clinical metadata for the sample associated with this library. 
 Blue hyperlinks are present for any terms with an ontology term identifier associated with the displayed human readable value. 
-These links will direct you to the [`Ontobee`](https://ontobee.org/) page for that ontology term identifier.  
+These links will direct you to a web page with information about that ontology term identifier.   
 
 ```{r}
 # extract sce metadata containing processing information as table
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -396,13 +396,15 @@ if (!has_multiplex) {
   print_sample_metadata(processed_meta)
 } else {
   # otherwise print out an info box that no sample metadata will be displayed
-  knitr::asis_output(glue::glue("
-  <div class=\"alert alert-info\">
+  knitr::asis_output(
+    glue::glue("
+      <div class=\"alert alert-info\">
 
-  This library is multiplexed and contains data from more than one sample.
-  Demultiplexing has not been performed, so sample metadata will not be displayed.
-  </div>
-"))
+      This library is multiplexed and contains data from more than one sample.
+      Demultiplexing has not been performed, so sample metadata will not be displayed.
+      </div>
+    ")
+  )
 }
 ```
 
--- a/templates/qc_report/main_qc_report.rmd
+++ b/templates/qc_report/main_qc_report.rmd
@@ -242,12 +242,12 @@ if (!has_multiplex) {
   # otherwise print out an info box that no sample metadata will be displayed
   knitr::asis_output(
     glue::glue("
-  <div class=\"alert alert-info\">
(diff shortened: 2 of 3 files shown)


templates/qc_report/utils/report_functions.rmd
```diff
@@ -26,3 +26,80 @@ determine_umap_point_size <- function(n_processed_cells) {
...
+      "tissue_location", sample_metadata$tissue_location, sample_metadata$tissue_ontology_term_id
+    ) |>
+      dplyr::mutate(
+        # replace ID with _ to construct url
+        ontology_term_id = stringr::str_replace(ontology_term_id, ":", "_"),
+        # build url
+        url = glue::glue("http://purl.obolibrary.org/obo/{ontology_term_id}"),
+        # format linked url with hr value for table
+        # if NA term ID, don't link
+        url = dplyr::if_else(!is.na(ontology_term_id), kableExtra::text_spec(hr_value, "html", link = url), hr_value)
```
jashapiro, line 61: reversing logic to simplify a bit, then unsimplifying by giving all values a `text_spec
```suggestion
        url = dplyr::if_else(
          is.na(ontology_term_id), 
          kableExtra::text_spec(hr_value, "html"), 
          kableExtra::text_spec(hr_value, "html", link = url)
        )
```

jashapiro:
> Also, I've been seeing the following error for this table, but I'm not sure how to fix it. The table is still printed, but if we can remove this message that's probably best?
> 
> > Warning message:
> > In yaml::yaml.load(yaml, handlers = knit_params_handlers(evaluate = evaluate),  :
> > an error occurred when handling type 'r'; using default handler

Can you narrow down when you are seeing this warning? I don't see it in any of the reports you sent, so is it only with some samples?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -381,6 +381,32 @@ glue::glue("
 ```
 
 
+# Sample metadata 
+
+The below table summarizes clinical metadata for the sample associated with this library. 
+Blue hyperlinks are present for any terms with an ontology term identifier associated with the displayed human readable value. 
+These links will direct you to the [`Ontobee`](https://ontobee.org/) page for that ontology term identifier.
```
jashapiro, line 388: This link isn't always to Ontobee, so you probably want something a bit more general.

allyhawkins:
> Can you narrow down when you are seeing this warning? I don't see it in any of the reports you sent, so is it only with some samples?

It doesn't print out in the reports, but only when I run the chunk that has the new metadata table locally or when I render the report it prints out in R.

jashapiro (commented):
This looks fine overall. I don't have great ideas about the warning you are seeing, but you might be able to narrow it down by turning on debugging with `debug(print_sample_metadata)` and running the chunk. That should allow you to find exactly where the warning is coming from.

I have some suspicion it may because some values are encoded by `text_spec` and some are not, but that is really just a guess.

allyhawkins:
>  I don't have great ideas about the warning you are seeing, but you might be able to narrow it down by turning on debugging with debug(print_sample_metadata) and running the chunk. That should allow you to find exactly where the warning is coming from.

I think it might be something with my Rstudio? I get the warning for any chunk that I run, but the chunk runs successfully. I found a similar issue [here](https://github.com/rstudio/rstudio/issues/7545). I could be way off, but I also ran the code in the console and don't see any warnings. So it's definitely something with the notebook itself. I also don't think it's a huge deal since rendering it works as expected?  

I also incorporated your edits and removed the Ontobee link so this should be ready for another look.

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -381,6 +381,32 @@ glue::glue("
...
+  print_sample_metadata(processed_meta)
+} else {
+  # otherwise print out an info box that no sample metadata will be displayed
+  knitr::asis_output(glue::glue("
+  <div class=\"alert alert-info\">
+
+  This library is multiplexed and contains data from more than one sample.
+  Demultiplexing has not been performed, so sample metadata will not be displayed.
+  </div>
+"))
```
jashapiro, line 405: Indentation
```suggestion
    <div class=\"alert alert-info\">

    This library is multiplexed and contains data from more than one sample.
    Demultiplexing has not been performed, so sample metadata will not be displayed.
    </div>
  "))
```

templates/qc_report/main_qc_report.rmd
```diff
@@ -226,13 +225,40 @@ glue::glue("
...
+  # otherwise print out an info box that no sample metadata will be displayed
+  knitr::asis_output(
+    glue::glue("
+  <div class=\"alert alert-info\">
+
+  This library is multiplexed and contains data from more than one sample.
+  Demultiplexing has not been performed, so sample metadata will not be displayed.
+  </div>
+")
+  )
```
jashapiro, line 251: Formatting here too... Maybe give this a run through `styler`?

jashapiro (approved):
LGTM, but probably worth a code styling run before merging.

PR #777, opened 2024-08-01
Add parameter for maximum number of libraries to allow when merging

### LINKED ISSUE #776: Skip creating merged objects if there are too many samples
We have not been successful in completing the merged workflow when there are > 100 libraries being merged into a single object. Because of this we probably want to add a check for the total number of libraries being merged and if it's > 100 we skip creating the object and emit a warning message about how that object won't be created, similar to what we do for multiplexed samples. 

Something similar to this but for total number of libraries: 
https://github.com/AlexsLemonade/scpca-nf/blob/f215046b3a9d9ddc50fac1a198a57e1bb813aeae/merge.nf#L175-L179

allyhawkins:
Closes #776 

In processing data for the Portal, we have seen that projects with > 100 libraries do not successfully run through the merge workflow due to what we think is limitations of the packages we use. We don't actually know the number of libraries that cause the problem and it's probably based on number of cells and how big the actual matrices are, but as of right now we do not include any [projects with > 100 libraries as merged objects on the Portal](https://scpca.readthedocs.io/en/stable/faq.html#which-projects-can-i-download-as-merged-objects). 

Because of this, I'm adding a parameter to specify the maximum number of libraries allowed in a merged object and setting the default to 100. If any projects have more than this, then we will print a message to the log files and skip processing that sample.

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -61,7 +61,7 @@ params {
 
   // Merge workflow-specfic options
   params.reuse_merge = false // if later steps fail, you can use `--reuse_merge` reuse the merged RDS object during a rerun
-  params.max_merge_libraries = 100 // maximum number of libraries to merge
+  params.max_merge_libraries = 100 // maximum number of libraries that can be merged
 
   // Docker container images
   includeConfig 'config/containers.config'
--- a/nextflow.config
+++ b/nextflow.config
@@ -61,7 +61,7 @@ params {
 
   // Merge workflow-specfic options
   params.reuse_merge = false // if later steps fail, you can use `--reuse_merge` reuse the merged RDS object during a rerun
-
+  params.max_merge_libraries = 100 // maximum number of libraries to merge
 
   // Docker container images
   includeConfig 'config/containers.config'
--- a/merge.nf
+++ b/merge.nf
@@ -144,11 +144,11 @@ workflow {
         || (it.scpca_sample_id in run_ids)
       }
       .map{[
-        seq_unit: it.seq_unit,
-        technology: it.technology,
         project_id: it.scpca_project_id,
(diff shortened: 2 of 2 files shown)


nextflow.config
```diff
@@ -61,7 +61,7 @@ params {
 
   // Merge workflow-specfic options
   params.reuse_merge = false // if later steps fail, you can use `--reuse_merge` reuse the merged RDS object during a rerun
-
+  params.max_merge_libraries = 100 // maximum number of libraries to merge
```
sjspielman, line 64: eh?
```suggestion
  params.max_merge_libraries = 100 // maximum number of libraries that can be merged
```

sjspielman (approved):
This all looks good to me! I'll note that I also tested this locally (I thought about whether we might want to add this to stub testing, but I don't think that's necessary) - 

- I added a couple lines to the stub run metadata file to give project `STUBP08` 5 libraries
- I ran `nextflow merge.nf -stub -profile stub --max_merge_libraries 4 --project STUBP08`
- Indeed, I received the message: `WARN: Not merging STUBP08 because it contains too many libraries.`

allyhawkins:
> This all looks good to me! I'll note that I also tested this locally (I thought about whether we might want to add this to stub testing, but I don't think that's necessary) -
> 
> * I added a couple lines to the stub run metadata file to give project `STUBP08` 5 libraries
> * I ran `nextflow merge.nf -stub -profile stub --max_merge_libraries 4 --project STUBP08`
> * Indeed, I received the message: `WARN: Not merging STUBP08 because it contains too many libraries.`

I also don't think this is necessary. We also skip multiplexed libraries in the same way and don't have a stub run for that so I think it's probably good as is.

PR #778, opened 2024-08-02
Update versions and bug fix for v0.8.3

allyhawkins:
Here I'm updating the release version to be v0.8.3. 

I also did a test run and found a small bug where we were trying to access hvg's in the ADT AnnData objects, which don't exist. So I fixed that by using the `--hvg_name "none"` option.

### DIFF
--- a/bin/reformat_anndata.py
+++ b/bin/reformat_anndata.py
@@ -40,7 +40,7 @@ parser.add_argument(
     default="highly_variable_genes",
     help=(
         "Indicate the name used to store highly variable genes associated with the PCA in the exported AnnData object."
-        " Use the value 'none' if no highly variable genes were not used."
+        " Use the value 'none' if no highly variable genes were used."
     ),
 )
 parser.add_argument(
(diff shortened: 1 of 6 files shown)


bin/reformat_anndata.py
```diff
@@ -40,7 +40,7 @@
     default="highly_variable_genes",
     help=(
         "Indicate the name used to store highly variable genes associated with the PCA in the exported AnnData object."
-        " Use the value 'none' if no highly variable genes were not used."
+        " Use the value 'none' if no highly variable genes were used."
```
sjspielman, line 43:
```suggestion
        "Use the value 'none' if no highly variable genes were used."
```

bin/sce_to_anndata.R
```diff
@@ -232,5 +232,6 @@ if (!is.null(opt$feature_name)) {
         ")
       )
     }
+    message("Exported alt")
```
sjspielman, line 235: just confirming if you meant to keep this or if it was part of testing? ok either way as long as it's what you meant!

sjspielman (approved):
🦏

bin/sce_to_anndata.R
allyhawkins: Yes I did mean to leave it in. We have an `exported RNA` message but no message here and I think it was helpful to see where things fail.

PR #783, opened 2024-08-05
Account for different HVG name in merged object

allyhawkins:
One more fix for the merged object workflow. They have a different name for HVGs that need to be specified using the new reformatting script so I updated that here. Note that I tested this on an object that was failing and it worked as expected.


sjspielman (approved):
🎢

Issue #792, opened 2024-10-30
Use scpcaTools 0.4.1 images


jashapiro:
scpcaTools 0.4.1 will soon be released. This release brings back rounding of counts by default (what we thought we were doing), and some security updates in Python packages that use the `lightning` package. 

Tracking incorporating those images here.

jashapiro:
closed by #795

PR #794, opened 2024-10-30
convert addParams to take

### LINKED ISSUE #791: remove `addParams`
Apparently the `addParams` feature when importing modules is deprecated, so we should modify the places where we use it to pass the parameter values as workflow inputs.

jashapiro:
When doing some testing related to #792 with newer versions of Nextflow, I got warnings that the `addParams()` function was deprecated, so I made some changes to remove that syntax, instead using `take:` to pass the values. Turns out I like this syntax better anyway!

closes #791


sjspielman (approved):
LGTM!

PR #799, opened 2024-11-04
Increase minimum cells to 3

jashapiro:
closes #798

Does what the title says, and also fixes our use of a removed Nextflow option.

I have tested this and it seems to work with at least the cases I have found; I also tested that when there are 3 cells, variance is calculated, as is a PCA (with a complaint, of course).

PR #801, opened 2024-11-05
Add minimum number of cells (3) for merge

jashapiro:
Following #799, here I am also filtering the merged data to not include samples with 2 or fewer cells. This turns out to be necessary for the calculation of tehe grouped PCA (at least that is what I think is going on).

I took a bit of a belt and suspenders approach here, checking both in the metadata during the workflow itself and in the R script where I check the SCEs before merging.

There is also a little cleanup in one of the `Util` functions so it doesn't throw an error if a file doesn't exist.

I did a small test with this and it seemed to work, but a larger test is still in progress, which is why I am filing this as a draft.

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -132,6 +132,18 @@ read_trim_sce <- function(sce_file) {
 # get list of sces
 sce_list <- purrr::map(input_sce_files, read_trim_sce)
 
+# filter out libraries with fewer than 3 cells (causes errors with PCA)
+n_cells <- sce_list |> purrr::map_int(ncol)
+included_libs <- names(sce_list)[which(n_cells >= 3)]
+if (length(included_libs) < length(sce_list)) {
+  message(
+    "The following libraries have fewer than 3 cells and will be excluded from the merged object: ",
+    paste(setdiff(names(sce_list), included_libs), collapse = ", ")
+  )
+}
+sce_list <- sce_list[included_libs]
+
+
 # Add cell type annotation columns where needed  -------------------------------
 
 # check for present cell type annotations
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -135,10 +135,11 @@ sce_list <- purrr::map(input_sce_files, read_trim_sce)
 # filter out libraries with fewer than 3 cells (causes errors with PCA)
 n_cells <- sce_list |> purrr::map_int(ncol)
 included_libs <- names(sce_list)[which(n_cells >= 3)]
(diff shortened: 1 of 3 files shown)


jashapiro:
Tests passed, ready for review!

bin/merge_sces.R
```diff
@@ -132,6 +132,18 @@ read_trim_sce <- function(sce_file) {
...
 
+# filter out libraries with fewer than 3 cells (causes errors with PCA)
+n_cells <- sce_list |> purrr::map_int(ncol)
+included_libs <- names(sce_list)[which(n_cells >= 3)]
+if (length(included_libs) < length(sce_list)) {
+  message(
+    "The following libraries have fewer than 3 cells and will be excluded from the merged object: ",
+    paste(setdiff(names(sce_list), included_libs), collapse = ", ")
+  )
+}
```
sjspielman, line 143: might do this comparison first and use it for check? up to you

```suggestion
lib_diff <- setdiff(names(sce_list), included_libs)
if (length(lib_diff) > 0) {
  message(
    "The following libraries have fewer than 3 cells and will be excluded from the merged object: ",
    paste(lib_diff, collapse = ", ")
  )
}
```

merge.nf
```diff
@@ -199,22 +199,32 @@ workflow {
...
         it.library_id,
-        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds")
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds"),
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
       ]}
-      // only include libraries that have been processed through scpca-nf and aren't empty
-      .filter{it[2].exists() && it[2].size() > 0}
+      // only include libraries that have been processed through scpca-nf and have at least 3 cells
+      .filter{it[2].exists() && it[2].size() > 0 && Utils.getMetaVal(it[3], "processed_cells") >= 3}
+      .map{it[0..2]} // remove metadata file from tuple
```
sjspielman, line 227: I gather `0..2` syntax is the groovy equivalent of R's `0:2` (...ok, `1:3` what with indexing from 1 😄 )

sjspielman (approved):
LGTM, just one suggestion for R code if you want it!

merge.nf
jashapiro: That is correct. I could have done a number of things to get rid of the last element. Probably the best is `it.dropRight(1)` to remove the last element, but this seemed fine here.

PR #806, opened 2025-01-14
Update GHA artifacts

### LINKED ISSUE #805: Bump GHA artifacts to v4
The GHA artifacts v3 is being phased out: https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/

We should replace these occurrences with v4:

- https://github.com/AlexsLemonade/scpca-nf/blob/b4fc58c25e7d583841a3a89e30e5d11325a061bc/.github/workflows/nextflow-stub-check.yaml#L42

- For this one, we might as well switch to our `spellcheck` repo
https://github.com/AlexsLemonade/scpca-nf/blob/b4fc58c25e7d583841a3a89e30e5d11325a061bc/.github/workflows/spell-check.yml#L34

sjspielman:
Closes #805 

This PR updates two workflows:
- The stub workflow gets bumped from artifact v3 -> v4
- The spellcheck workflow gets replaced with a new workflow using our spellcheck repo
  - I did not delete the spell check script since it's used by pre-commit. If we prefer to also use the spellcheck repo for precommit, then we'd need to add a `.pre-commit-hooks.yaml` file to it, which maybe we should just go ahead and do? 

I'm opening this as a draft to confirm first that the GHA works and detects a typo I added in.

### DIFF
--- a/.github/workflows/nextflow-stub-check.yaml
+++ b/.github/workflows/nextflow-stub-check.yaml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Upload nextflow log
         if: ${{ !cancelled() }}
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: nextflow-log
           path: nextflow-runs.log
--- a/.github/workflows/spell-check.yml
+++ b/.github/workflows/spell-check.yml
@@ -8,35 +8,39 @@ on:
       - development
       - main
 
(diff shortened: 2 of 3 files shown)


jashapiro (approved):
LGTM. We could probably remove the spellcheck in precommit and replace it with `typos`, as that will run much faster, and we have the full spellcheck in the GHA, but this is fine for now.

PR #809, opened 2025-01-24
Account for samples being both PDX and cell line

### LINKED ISSUE #808: Account for samples that are both cell lines and PDX
We now have samples that are classified as both a cell line and a PDX as part of the Portal. Currently when we assign `sample_type` in the metadata of the SCE objects we only allow for one of these. Additionally, the code is set up so that `sample_type` is a single value of either `patient-derived xenograft`, `cell line`, `not provided`, or `patient tissue`. I think we want to modify things so that a sample type can include both and will be set as `cell line, patient-derived xenograft`. 

We will need to make the necessary updates to account for a new value in this column of the sample metadata. I believe these are the only two parts of the code that are affected. 

https://github.com/AlexsLemonade/scpca-nf/blob/b4fc58c25e7d583841a3a89e30e5d11325a061bc/bin/generate_unfiltered_sce.R#L196-L215

https://github.com/AlexsLemonade/scpca-nf/blob/b4fc58c25e7d583841a3a89e30e5d11325a061bc/templates/qc_report/main_qc_report.rmd#L151-L208

allyhawkins:
Closes #808 

This PR makes some minor adjustments to how we handle `sample_type` to account for samples that are both a PDX and a cell line. I modified the existing code so that the `sample_type` option can now be a vector with `cell line` and `patient-derived xenograft`. 

I had to make a small change to the table in the report to combine into a single string, otherwise it printed two columns for the sample table. I tested this with the sample this applies to and did a test run using one of each sample type and everything looks as expected. 

Here's an example of the report for the affected sample. 
[SCPCL001201_qc.html.zip](https://github.com/user-attachments/files/18540965/SCPCL001201_qc.html.zip)

### DIFF
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -213,7 +213,8 @@ sample_type <- sample_metadata_df |>
 
 # unname if length is 1, and add to sce metadata
 if (length(sample_type) == 1) {
-  sample_type <- unname(sample_type)
+  sample_type <- unlist(sample_type) |>
+    unname()
 }
 metadata(unfiltered_sce)$sample_type <- sample_type
 
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -197,6 +197,8 @@ sample_type <- sample_metadata_df |>
   dplyr::filter(sample_id %in% sample_ids) |>
   dplyr::mutate(
     sample_type = dplyr::case_when(
+      # account for sample being both cell line and PDX
+      is_xenograft & is_cell_line ~ c("cell line", "patient-derived xenograft"),
       is_xenograft ~ "patient-derived xenograft",
       is_cell_line ~ "cell line",
       # if neither column was provided, note that
--- a/bin/generate_unfiltered_sce.R
+++ b/bin/generate_unfiltered_sce.R
@@ -198,7 +198,7 @@ sample_type <- sample_metadata_df |>
   dplyr::mutate(
     sample_type = dplyr::case_when(
       # account for sample being both cell line and PDX
-      is_xenograft & is_cell_line ~ c("cell line", "patient-derived xenograft"),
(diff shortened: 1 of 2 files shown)


bin/generate_unfiltered_sce.R
```diff
@@ -205,12 +207,14 @@ sample_type <- sample_metadata_df |>
...
-  tibble::deframe()
+  # convert into named list
+  tibble::deframe() |>
+  strsplit(split = ", ")
 
 # unname if length is 1, and add to sce metadata
 if (length(sample_type) == 1) {
-  sample_type <- unname(sample_type)
+  sample_type <- unlist(sample_type) |>
+    unname()
```
jashapiro, line 217: I was surpised you needed both `unlist` and `unname` here, but apparently you do!

bin/generate_unfiltered_sce.R
```diff
@@ -197,6 +197,8 @@ sample_type <- sample_metadata_df |>
   dplyr::filter(sample_id %in% sample_ids) |>
   dplyr::mutate(
     sample_type = dplyr::case_when(
+      # account for sample being both cell line and PDX
+      is_xenograft & is_cell_line ~ "cell line, patient-derived xenograft",
```
jashapiro, line 201: I am wary of two character splits, even though I know this works.
```suggestion
      is_xenograft & is_cell_line ~ "cell line;patient-derived xenograft",
```

bin/generate_unfiltered_sce.R
```diff
@@ -205,12 +207,14 @@ sample_type <- sample_metadata_df |>
     )
   ) |>
   dplyr::select(sample_id, sample_type) |>
-  # convert into named vector
-  tibble::deframe()
+  # convert into named list
+  tibble::deframe() |>
+  strsplit(split = ", ")
```
jashapiro, line 212: Making a single character split.
```suggestion
  strsplit(split = ";")
```

jashapiro (approved):
LGTM. I made one suggestion, but I really don't feel strongly about it.

PR #811, opened 2025-01-27
Use resource limits and update manifest and schema

jashapiro:
I had a branch from a bit ago where I updated a few things to new Nextflow standards, and I thought it might make sense to get this in as part of #810.

The changes here implement using the "new" resource limits in Nextflow. 

While I was at it, I also updated the schema to match the current parameters, which included removing what I think is an unused `CONDA_CONTAINER` parameter.

I also updated some of the manifest to use the updated format, with additional information including the doi for the paper!

### DIFF
--- a/.github/workflows/nextflow-config-check.yaml
+++ b/.github/workflows/nextflow-config-check.yaml
@@ -8,10 +8,10 @@ on:
 jobs:
   nf-config-check:
     runs-on: ubuntu-22.04
-    container: nfcore/tools:2.13.1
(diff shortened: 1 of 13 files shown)


jashapiro:
I realized that this got a bit bigger than I was expecting when I had to update the GitHub actions, so if you think this should wait until after the 0.8.6 release, that is fine with me!

allyhawkins (approved):
Assuming that you did a test run of the workflow, LGTM

PR #816, opened 2025-01-29
Add bulk TPM output to scpca-nf

### LINKED ISSUE #815: Add TPM matrix output for bulk data
We currently output only count data for bulk sequencing, but for some applications, TPM data will be preferred. We can modify the [`merge_bulk_quants` process](https://github.com/AlexsLemonade/scpca-nf/blob/16fb37f8c5b17e41e557db852c9521a90705da2e/modules/bulk-salmon.nf#L68) and the [`merge_counts_tximport.R` script](https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/merge_counts_tximport.R)  to output both TPM and count matrices using the existing salmon output files.

jashapiro:
Closes #815

I was looking at `scpca-nf`, and realized that the changes to add TPM output were quite minimal, so I went ahead and did that. I left the count file name unchanged, despite its ambiguity, and simply added one more output file, in the same directory. 

Since the script name seemed a bit out of date, I modified that as well.

### DIFF
--- a/bin/tximport_bulk.R
+++ b/bin/tximport_bulk.R
@@ -2,7 +2,7 @@
 
 # This script grabs the quant.sf output files from Salmon for all samples within a project
 # The files are then imported use tximport to generate a gene x count matrix
-# Output includes a counts matrix as a tab separated tsv file and a .Rds object containing the tximport object
+# Outputs a counts matrix and TPM matrix as separate TSV files.
 
 
 # load needed packages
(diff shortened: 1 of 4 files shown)


sjspielman (approved):
Fun surprise, looks good!

I approved, but we should update this sentence too: https://github.com/AlexsLemonade/scpca-nf/blob/16fb37f8c5b17e41e557db852c9521a90705da2e/external-instructions.md?plain=1#L488
I guess we'll need some `scpca-docs` issues as well, but this won't make it to the portal for a little bit I imagine.

jashapiro:
> Fun surprise, looks good!
> 
> I approved, but we should update this sentence too:
> 
> https://github.com/AlexsLemonade/scpca-nf/blob/16fb37f8c5b17e41e557db852c9521a90705da2e/external-instructions.md?plain=1#L488
> 
> 
> I guess we'll need some `scpca-docs` issues as well, but this won't make it to the portal for a little bit I imagine.

Turns out that document is a bit out of date... we probably want an issue to update it as well.

PR #817, opened 2025-01-30
Fix contributors formatting

jashapiro:
Apparently the contributors field requires specific formatting, and I had it wrong. What I had works fine for local runs, but causes trouble when using Seqera Tower. This should now be the correct format.


allyhawkins (approved):
🚀

PR #819, opened 2025-02-05
Remove SPATIAL_RNA_COUNTER_CS

### LINKED ISSUE #818: Remove SPATIAL_RNA_COUNTER_CS
From https://github.com/AlexsLemonade/ScPCA-admin/pull/1024#issue-2830969239: 

> The main thought is that there is a very large amount of data that is stored in the `SPATIAL_RNA_COUNTER_CS` directory, which seems to basically be the Space Ranger intermediate files. We don't use anything from this directory in downstream analysis, and I don't expect that we ever will. In fact, we only use files from the `outs` directory, so we could potentially the checkpoint to just that, but removing just the `SPATIAL_RNA_COUNTER_CS` directory is probably going to be more than sufficient.

> For a bit of context on the sizes, the `SPATIAL_RNA_COUNTER_CS` directory in the sample I looked at first was ~2.5 GB, while the outs directory was ~800 MB. So this is not a small savings!

For minimal disruption, I think the best course of action is to update `modules/spaceranger.nf` to add the following line:

 ```
# remove intermediate files
rm -rf ${out_id}/SPATIAL_RNA_COUNTER_CS
```

following https://github.com/AlexsLemonade/scpca-nf/blob/16fb37f8c5b17e41e557db852c9521a90705da2e/modules/spaceranger.nf#L34-L35

jashapiro:
Closes #818, if we want to. 

This is a very small PR, but as noted in #818, it may actually have some substantial savings both in storage costs and maybe even a bit of run time (since it prevents copying files to and from the work directory). 

I can't see a downside at the moment, but I am well aware I could be missing something.

### DIFF
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -31,6 +31,9 @@ process spaceranger {
     # write metadata
     echo '${meta_json}' > ${out_id}/scpca-meta.json
 
+    # remove Space Ranger intermediates directory
+    rm -rf ${out_id}/SPATIAL_RNA_COUNTER_CS
+
     # remove bam and bai files
     rm ${out_id}/outs/*.bam*
     """


allyhawkins (approved):
I am totally okay with this since we don't use it for anything so going to approve.

Issue #821, opened 2025-02-11
Create separate staging and production profiles


jashapiro:
To better ensure that the results from `scpca-nf` are stable and allow easier comparisons between the production results and any changes in progress, we should create separate staging and production profiles, with separate output locations for checkpoints and results.

The current `ccdl` profile uses an output location that is effectively a testing ground; there are no particular controls on the data that are found there. When we are performing a production run, we currently set the output location to the `scpca-prod` directory manually with the `--outdir` parameter, but this is a bit brittle and, well, manual.

Instead, we should define `ccdl_staging` and `ccdl_prod` profiles that set appropriate output paths.

For extra protection, we may want to see if we can restrict the production profile so that will only run with a versioned tag (so we do not accidentally overwrite results from a development version of the workflow).

This issue should also include:

- Creation of an appropriate staging output location (`scpca-staging`), which will begin as a duplicate of the `scpca-prod` directory
- Changes to documentation to reflect usage of the new profiles, including at https://github.com/AlexsLemonade/ScPCA-admin/tree/main/data-processing-and-handoff

jashapiro:
I filed a [separate issue](https://github.com/AlexsLemonade/ScPCA-admin/issues/1028) for the documentation updates in ScPCA-admin, and the copying of files is complete, so I am closing this issue as complete.

PR #827, opened 2025-02-25
Create separate staging and production profiles

jashapiro:
Part of #821

Here I added two additional profiles, `ccdl_staging` and `ccdl_prod` that are largely the same as the current `ccdl` profile, but specify the correct output directories, and process all files by default. 

These profiles also set `perform_celltyping = true`, as well as using the CCDL Seqera Workspace. This last change I was a bit uncertain about, as it does mean that the launch will fail unless a Seqera token is properly set, but I thought it was probably worthwhile. But please let me know if you think that is a step too far. 

I also set the production run to use resume by default, with the initial thought that this would mean that it could use the results from a previous successful staging run to reduce runtimes, as we do in `OpenScPCA-nf`. I'm not sure that this will actually work though, as the checkpoint files that are used as input between staging and production will be different (if only in location). 

Because all of the profiles are related, I decided to move all of the CCDL-specific profiles to be managed by a single file, which also handles the import of the `batch` profiles. While I have not tested this with actual runs, I did test it with commands such as the following:

```
nextflow config . -profile ccdl_staging,batch
```

This allowed me to see that the settings seem to be applied as expected.

I also updated the instruction file to reflect the changes and updated run instructions. My feeling at this point is that we can continue to use the `ccdl` profile for the majority of development and initial testing, and then use the `ccdl_staging` to iron out all the edge case problems before proceeding to `ccdl_prod`. I tried to reflect that in the readme, but I might need to be more explicit. 

I am currently copying all files from `scpca-prod` to `scpca-staging` (with the exception of `SPATIAL_RNA_COUNTER_CS/` directories); when that is done, we should be ready to start using these profiles.


config/profiles_ccdl.config
```diff
@@ -0,0 +1,80 @@
...
+    // paths to output folders, redefined to use `outdir` above
+    checkpoints_dir = "${params.outdir}/checkpoints"
+    results_dir = "${params.outdir}/results"
+
+    // a set of run_ids for testing. used only by the main workflow
+    // one single-cell with bulk, one CITE, one spatial, one multiplexed
+    run_ids = "SCPCS000001,SCPCS000050,SCPCS000203,SCPCL000537"
+
+    // include a sample project for merging. used only by the merged workflow
+    merge_run_ids = "SCPCP000016"
```
allyhawkins, line 32:
```suggestion
    // include all runs in a project when creating merged object
    merge_run_ids = "All"
```
This parameter can only be used to specify run, library, or sample IDs to include in a merged project. To actually run the merge workflow you need a `project`, which we don't set by default here. The purpose of this parameter is to specify if you only want to include a certain set of libraries from a project in the merged object.

We could set SCPCP000016 as the default project to use if we wanted to have one pre-set for testing? It just will also get used when testing the main workflow, so I don't think it's entirely necessary.

config/profiles_ccdl.config
```diff
@@ -0,0 +1,80 @@
...
+    checkpoints_dir = "${params.outdir}/checkpoints"
+    results_dir = "${params.outdir}/results"
+
+    // staging runs always include celltyping
+    perform_celltyping = true
+  }
+}
+
+ccdl_prod {
+  resume = true // prod runs should always use resume to avoid reprocessing
```
allyhawkins, line 60: I'm iffy about including this. When I made the update to OpenScPCA-nf to use a different consensus reference file I ran it on staging and then ran it on prod with resume and when I checked all the processes had cached, but the output files had not changed and did not match what was on staging. I also feel like resume doesn't always work as expected...

config/profiles_ccdl.config
```diff
@@ -0,0 +1,80 @@
...
+    // one single-cell with bulk, one CITE, one spatial, one multiplexed
+    run_ids = "SCPCS000001,SCPCS000050,SCPCS000203,SCPCL000537"
+
+    // include a sample project for merging. used only by the merged workflow
+    merge_run_ids = "SCPCP000016"
+
+  }
+}
+
+ccdl_staging {
```
allyhawkins, line 37: A thought I had that might simplify us having to repeat a lot of the same parameters is that we might want to create a separate `ccdl_metadata` profile that defines all metadata files and then we use `includeConfig` within these profiles to grab those params. Then all you are defining in these profiles are the `outdir` which should be different for each one and then any extra params like `perform_celltyping` and the run Ids.

internal-instructions.md
```diff
@@ -84,12 +90,26 @@ Please refer to our [`CONTRIBUTING.md`](CONTRIBUTING.md#stub-workflows) for more
...
 ```
-nextflow run AlexsLemonade/scpca-nf -r v0.8.6 -profile ccdl,batch --project SCPCP000000
+nextflow run AlexsLemonade/scpca-nf -r v0.8.6 -profile ccdl_staging,batch --project SCPCP000000
 ```
 
+When that run has completed successfully, check that the outputs are as expected.
+You can then run the workflow using the `ccdl_prod` profile:
+
+```
+nextflow run AlexsLemonade/scpca-nf -r v0.8.6 -profile ccdl_staging,batch --project SCPCP000000
```
allyhawkins, line 106:
```suggestion
nextflow run AlexsLemonade/scpca-nf -r v0.8.6 -profile ccdl_prod,batch --project SCPCP000000
```

internal-instructions.md
```diff
@@ -84,12 +90,26 @@ Please refer to our [`CONTRIBUTING.md`](CONTRIBUTING.md#stub-workflows) for more
...
+```
+
+Both of these profiles have `-with-tower` set by default, and will use the [ScPCA workspace](https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch) for monitoring (allowing all team members to see progress).
+*Be sure to have set the `TOWER_ACCESS_TOKEN` environment variable to your Seqera platform token when triggering a run, or these runs will fail.*
```
allyhawkins, line 110: What does this failure look like? Is it obvious that this is the reason it failed?

allyhawkins (commented):
Overall this look good, I just had a few questions and maybe a suggestion to remove the duplication of setting all the metadata files that are the same. 

I think my biggest concern is about using the `resume` option since I just don't trust it to always work in the way we expect. That being said I always use it for running production runs myself so I could be convinced...

config/profiles_ccdl.config
jashapiro: Oh, that is a good idea! I didn't want to repeat myself so much, but I couldn't think of a way to handle this easily.

config/profiles_ccdl.config
jashapiro: I'll take it out. There is also a thought I had that we could update the procedure to not actually run to create a prod release at all, but instead use `aws s3 sync` to copy from staging to prod... The point is to have two copies.

internal-instructions.md
jashapiro: Failure will be either:
```
Missing personal access token -- Make sure there's a variable TOWER_ACCESS_TOKEN in your environment
```
or If there is an invalid token (this happens if you forgot `op run --` and it tries to pass the `op://` string):
```
Unauthorized Seqera Platform API access -- Make sure you have specified the correct access token
```

So pretty good error messages, all told!

allyhawkins: Okay great, I'm good with keeping this in then! I think it will be helpful for monitoring all the runs.

config/profiles_ccdl.config
allyhawkins: > I'll take it out. There is also a thought I had that we could update the procedure to not actually run to create a prod release at all, but instead use `aws s3 sync` to copy from staging to prod... The point is to have two copies.

I'm not opposed to that idea...

jashapiro: > There is also a thought I had that we could update the procedure to not actually run to create a prod release at all, but instead use `aws s3 sync` to copy from staging to prod... The point is to have two copies.
> 
> I'm not opposed to that idea...

Let's keep this in mind for https://github.com/AlexsLemonade/ScPCA-admin/issues/1028. I think we can leave the profile in here for now, as it won't hurt anything for it to exist. 

I think I addressed your comments and this should be ready for another look.

allyhawkins (approved):
LGTM

PR #831, opened 2025-03-19
Add initial table with output metrics for testing

### LINKED ISSUE #822: Compile a list of metrics that will be calculated for each library processed by scpca-nf
While comparing full output files from `scpca-nf` might be ideal, the fact that the output files are large suggests that we might prefer to pre-compute a set of metrics as part of the output. Aa first step of this, we will need a list of specific metrics that we will calculate on each workflow run to identify changes in the workflow outputs.

Some of these, such as number of cells, may already exist in metadata files that we produce, but others will require new calculations.

The expected output here will be a document in this repository that describes the key metrics. We will use that document to guide the next steps of creating an output file with these metrics. 

A partial list of metrics and more information can be found at https://data-lab-knowledge.slab.com/posts/discussion-testing-scpca-nf-outputs-3d4pra87

jashapiro:
Closes #822
Here I am adding a document with a table of metrics that will be calculated as part of testing the workflow for changes. I also included a section for formatting checks, but I have not filled that in yet. I thought we could start with just the metrics.

Most of the metrics here are derived from https://data-lab-knowledge.slab.com/posts/discussion-testing-scpca-nf-outputs-3d4pra87, with a few additions/updates. Note that some of the metrics are actually arrays or dictionaries (i.e. hvgs and  cell types). I think we do want to  include those in the output metric; they will make the file larger but still not _large_.

I wasn't sure whether we wanted to just make these part of the `_metadata.json` file or a separate `_metrics.json` file. Either way seems reasonable enough to me, with the caveat that things like the version numbers could become a bit harder to find in the metadata file if we add too many large fields to it.

Please let me know if there are any metrics that you think I missed, or if you think we should remove ones that I included.


config/workflow-testing.md
```diff
@@ -0,0 +1,34 @@
+# Testing scpca-nf outputs
+
+As `scpca-nf` is updated, we want to be sure that we are aware of any changes in the output from the pipeline before we deploy updated results to the portal.
+This documents outlines the tests that we will run and the key metrics that we will monitor to ensure that the pipeline is functioning as expected.
```
allyhawkins, line 4:
```suggestion
This document outlines the tests that we will run and the key metrics that we will monitor to ensure that the pipeline is functioning as expected.
```

config/workflow-testing.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Data format checks
+
```
allyhawkins, line 7: Maybe make a note that this is coming soon or TBD?

config/workflow-testing.md
```diff
@@ -0,0 +1,34 @@
...
+## Output metrics
+
+The table below describes the key metrics that we will track for the workflow outputs, and which files are required to calculate those metrics.
+These metrics should be calculated, if required, at the end of the workflow just before the publication step, and stored in a JSON file. At the moment, I am not sure if we just add these to `_metadata.json` or create a separate file.
```
allyhawkins, line 11: I think I favor a separate file so as not to crowd the current `_metadata.json` file? Also when we hand off the files the `_metadata.json` has information needed to deploy the samples to the Portal, but the metrics here are for internal use so I think we can keep it separate.

config/workflow-testing.md
```diff
@@ -0,0 +1,34 @@
...
+
+| Metric             | Description                                    | Required files              |
+| ------------------ | ---------------------------------------------- | --------------------------- |
+| `unfiltered_cells` | total number of cells in unfiltered SCE object | `{library_id}_metadata.json` or `{library_id}_unfiltered.rds` |
+| `filtered_cells`   | total number of cells in filtered SCE object   | `{library_id}_metadata.json` or `{library_id}_filtered.rds` |
+| `processed_cells`  | total number of cells in processed SCE object  | `{library_id}_metadata.json` or `{library_id}_processed.rds` |
+| `total_reads`      | total number of reads from FASTQ files | `{library_id}_metadata.json` |
+| `mapped_reads`      | total number of reads mapped by alevin-fry | `{library_id}_metadata.json` |
+| `cell_filtering_method` | method used to filter cells | `{library_id}_metadata.json` or `{library_id}_processed.rds` |
```
allyhawkins, line 20: We should probably add in `droplet_filtering_method` and `normalization_method`.

config/workflow-testing.md
```diff
@@ -0,0 +1,34 @@
...
+| `unfiltered_expressed_genes` | count of genes with total expression > 0 in unfiltered SCE object | `{library_id}_unfiltered.rds` |
+| `filtered_expressed_genes` | count of genes with total expression > 0 in filtered SCE object | `{library_id}_filtered.rds` |
+| `processed_expressed_genes` | count of genes with total expression > 0 in processed SCE object | `{library_id}_processed.rds` |
+| `hv_genes` | the array of the highly variable genes  | `{library_id}_processed.rds` |
+| `cluster_sizes` | an array with the number of cells in each cluster | `{library_id}_processed.rds` |
+| `singler_celltypes` | a dictionary with the number of cells in each SingleR-based cell type  | `{library_id}_processed.rds` |
+| `cellassign_celltypes` | a dictionary with the number of cells in each CellAssign-based cell type  | `{library_id}_processed.rds` |
+| `unfiltered_altexp_total` | total number of reads from altExp (ADT or cellhash) | `{library_id}_unfiltered.rds` |
+| `filtered_altexp_total` | total number of reads from altExp (ADT or cellhash) | `{library_id}_filtered.rds` |
+| `processed_altexp_total` | total number of reads from altExp (ADT or cellhash) | `{library_id}_processed.rds` |
```
allyhawkins, line 34: Do we also want to add in `adt_scpca_filter_count`, similar to the `miqc_pass_count`, but for ADT filtering?

allyhawkins (commented):
This mostly looks good, I had a few suggestions of things we should add. 
Also, I don't know that the location of this file makes the most sense? I think it probably belongs as a new header in the `internal-instructions.md` rather than in the `config` directory?

jashapiro:
> Also, I don't know that the location of this file makes the most sense? I think it probably belongs as a new header in the `internal-instructions.md` rather than in the `config` directory?

That's a mistake! It was meant to be top level; not sure how I ended up saving it in `config`. I didn't want it to be part of internal instructions partly because external users might want to monitor changes as well, and there will be additional files created as part of this, but I could consider making it part of that doc.

allyhawkins:
> > Also, I don't know that the location of this file makes the most sense? I think it probably belongs as a new header in the `internal-instructions.md` rather than in the `config` directory?
> 
> That's a mistake! It was meant to be top level; not sure how I ended up saving it in `config`. I didn't want it to be part of internal instructions partly because external users might want to monitor changes as well, and there will be additional files created as part of this, but I could consider making it part of that doc.

I think making it a separate document in the top level works for me!

jashapiro:
I updated the text with a bit more detail and also added a few more metrics to track. The main additions are the ones you suggestions, but also the `logcounts` matrix sum to check that for changes, and the references for SingleR and CellAssign. The thought with the last of these is that we may well want to skip checking the  assignment counts if the reference has changed, or at least to highlight that as a likely cause of any changes we might see. 

Should be ready for another look.

allyhawkins (approved):
LGTM

PR #834, opened 2025-03-24
Add metrics calculation script

### LINKED ISSUE #832: Add script to compile _metrics.json file
Following the list of metrics defined in #822, we need to write a script that compiles all of the metrics for each library into a `_metrics.json` file. This script will be run as part of the publishing process, so it will take as input the three versions of the each SCE object, as well as the `_metadata.json` file.

jashapiro:
Closes #832 

Here I added a script for calculating the tracking metrics described in the `workflow-testing.md` document.

I added a script to gather the statistics and perform calculations as needed and integrated it into the QC report process of the workflow, which I renamed to make the fact that is the "publishing" part of the workflow.  I added a couple more stats that I thought were worth including, notably the spliced counts. I also made the adt counts a dictionary so that if there were more than one altExp, we will be ready (at least with this part; lots of other changes would still be needed).

As long as I was renaming processes, I went a bit further, renaming some of the confusingly named "merged" processes with names that better reflect that they are for experiments with feature data.

I also made a small change to the output `_metadata.json` file in the QC script to add the `project_id` value there as well. 

If in your review you think of any other stats we should add now, let me know. 

One limitation of the current implementation is that it does not deal with spatial or bulk data at all; since bulk data metrics would probably be at the project level, and spatial is... spatial, I think we probably are okay for now.

### DIFF
--- a/bin/sce_metrics.R
+++ b/bin/sce_metrics.R
@@ -1,0 +1,96 @@
+#!/usr/bin/env Rscript
+
+# This script generates a QC report using scpcaTools from a pair of filtered and unfiltered SCE objects
+
+# import libraries
+suppressPackageStartupMessages({
+  library(optparse)
+  library(SingleCellExperiment)
+})
+
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = "--metadata_json",
+    default = "metadata.json",
+    type = "character",
+    help = "path to metadata json  file",
+    default = ""
+  ),
+  make_option(
+    opt_str = c("-u", "--unfiltered_sce"),
+    type = "character",
+    help = "path to rds file with unfiltered sce object",
+    default = ""
+  ),
+  make_option(
+    opt_str = c("-f", "--filtered_sce"),
+    type = "character",
+    help = "path to rds file with filtered sce object",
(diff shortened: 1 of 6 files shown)


bin/sce_metrics.R
```diff
@@ -0,0 +1,143 @@
...
+    })
+  return(exp_totals)
+}
+
+# set up arguments
+option_list <- list(
+  make_option(
+    opt_str = "--metadata_json",
+    type = "character",
+    help = "path to metadata json  file",
```
allyhawkins, line 39:
```suggestion
    help = "path to metadata json file",
```

allyhawkins (approved):
This looks good to me, and I don't really have any changes other than the one typo. Assuming you tested this locally and it ran successfully, then I think we are good to go.

jashapiro:
> This looks good to me, and I don't really have any changes other than the one typo. Assuming you tested this locally and it ran successfully, then I think we are good to go.

I ran it on batch, actually! I had meant to include example output, but here it is:
[SCPCL000001_metadata.json](https://github.com/user-attachments/files/19457333/SCPCL000001_metadata.json)

PR #835, opened 2025-03-26
Add consensus cell type assignments to processed objects

### LINKED ISSUE #824: Add consensus cell type assignment to scpca-nf
We would like to add a process for adding consensus cell types to the `colData` of the processed SCE objects. Consensus cell type assignments are determined based on the combination of cell types assigned by `SingleR` and `CellAssign` and will use the [reference table in `OpenScPCA-analysis`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv).  

I think the best way to do this would be to modify the existing `add_celltypes_to_sce.R` script, which currently adds a column for each of the cell type annotations to the processed SCE. We should modify that script to add the consensus reference as an argument and then assign the consensus label based on the `SingleR` and `CellAssign` assignments. 

Logic in the script should only assign consensus cell types if both `SingleR` and `CellAssign` results are present. If that is the case, then two new columns will be added for `consensus_annotation` and `consensus_ontology`. 

Outside of changing the script we will also need to update the config file to have a parameter for the consensus reference and pass that into the [`add_celltypes_to_sce` process](https://github.com/AlexsLemonade/scpca-nf/blob/16fb37f8c5b17e41e557db852c9521a90705da2e/modules/classify-celltypes.nf#L105).

allyhawkins:
Closes #824 

Here I am adding the consensus cell types to the processed objects within the `add_celltypes_to_sce.R` script that gets used to add the `SingleR` and `CellAssign` cell types to the `colData` of the object. 

- I added two new parameters, one for the panglao ontology ID reference and one for the consensus cell type reference. The parameters for both of these files point to their location in `OpenScPCA-analysis`.
- Since part of creating the consensus labels included assigning ontology ID terms to the cell types in `PanglaoDB`, I first am adding in a new column, `cellassign_celltype_ontology` which includes the assigned ontology terms for the `CellAssign` cell types. 
- Then if both `CellAssign` and `SingleR` results exist, I add in the consensus annotation and ontology. 
  - If `CellAssign` had an empty predictions file and therefore the values are `Not run`, I do not do any consensus cell type assignments. 
  - Right now, the `consensus_celltype_annotation` and `consensus_celltype_ontology` columns only exist if both `SingleR` and `CellAssign` were successful, otherwise these columns do not exist. Are we okay with that approach or should we fill in with `NA` if one of the items is not run? 
  - In testing, I had one sample with cells in the `CellAssign` predictions file that were no longer present in the filtered object (it was only a handful). To account for this scenario, I added a filtering step to only include barcodes that are in the object when obtaining the annotations. 
- While working in this script I also made some updates to how we handle errors to use `stopifnot` and set some defaults to "".

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -180,13 +180,11 @@ if (has_singler) {
 
   # add note about cell type method to metadata
   metadata(sce)$celltype_methods <- c(metadata(sce)$celltype_methods, "singler")
-} else {
-  has_singler <- FALSE # use to decide if we should assign consensus labels
 }
 
 # CellAssign results -----------------------------------------------------------
 
-has_cellassign <- file.exists(opt$cellassign_predictions) && (file.size(opt$cellassign_predictions) > 0)
+has_cellassign <- file.exists(opt$cellassign_predictions)
 if (has_cellassign) {
   # check that cellassign reference info is provided
   stopifnot(
@@ -198,8 +196,14 @@ if (has_cellassign) {
   # read in panglao ontology reference
   panglao_ref_df <- readr::read_tsv(opt$panglao_ontology_ref)
 
-  # read in predictions file
-  predictions <- readr::read_tsv(opt$cellassign_predictions)
+  # if cell assign predictions file exists but is empty then cell assign failed and we want to account for that with Not Run
+  if (file.size(opt$cellassign_predictions) > 0) {
+    # read in predictions file
+    predictions <- readr::read_tsv(opt$cellassign_predictions)
+  } else {
+    predictions <- NULL
+    has_cellassign <- FALSE # reset to false so that we don't add in consensus cell types
+  }
 
   # if the only column is the barcode column or if the predictions file was empty
   # then CellAssign didn't complete successfully
@@ -262,8 +266,6 @@ if (has_cellassign) {
     }
     metadata(sce)$cellassign_reference_organs <- cellassign_organs
   }
-} else {
-  has_cellassign <- FALSE # use to decide if we should assign consensus labels
 }
 
 # assign consensus cell type labels
@@ -296,9 +298,7 @@ if (has_singler && has_cellassign) {
       relationship = "many-to-many" # account for multiple of the same cell type
     ) |>
     # use unknown for NA annotation but keep ontology ID as NA
-    # if the sample type is cell line, keep as NA
-    dplyr::mutate(consensus_annotation = dplyr::if_else(is.na(consensus_annotation), "Unknown", consensus_annotation))
-
+    tidyr::replace_na(list(consensus_annotation = "Unknown"))
 
   # add consensus cell type and ontology to sce
   sce$consensus_celltype_annotation <- celltype_df$consensus_annotation
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -298,7 +298,7 @@ if (has_singler & has_cellassign) {
         "singler_celltype_ontology" = "blueprint_ontology",
         "cellassign_celltype_ontology" = "panglao_ontology"
       ),
-      relationship = "many-to-many"
+      relationship = "many-to-many" # account for multiple of the same cell type
     ) |>
     # use unknown for NA annotation but keep ontology ID as NA
     # if the sample type is cell line, keep as NA
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -215,6 +215,8 @@ if (file.exists(opt$cellassign_predictions)) {
   } else {
     # get cell type with maximum prediction value for each cell
     celltype_assignments <- predictions |>
+      # account for the fact that we could have lost some cells we had previously when re-processing through filtering steps
+      dplyr::filter(barcode %in% colnames(sce)) |>
       tidyr::pivot_longer(
         !barcode,
         names_to = "celltype",
@@ -277,7 +279,10 @@ if (has_singler & has_cellassign) {
   )
 
   # read in consensus table
-  consensus_ref_df <- readr::read_tsv(opt$consensus_celltype_ref)
+  consensus_ref_df <- readr::read_tsv(opt$consensus_celltype_ref) |>
+    # select unique combinations of consensus refs based on ontology columns
+    dplyr::select(blueprint_ontology, panglao_ontology, consensus_ontology, consensus_annotation) |>
+    unique()
 
   # create df with consensus assignments
   celltype_df <- colData(sce) |>
@@ -292,7 +297,8 @@ if (has_singler & has_cellassign) {
       by = c(
         "singler_celltype_ontology" = "blueprint_ontology",
         "cellassign_celltype_ontology" = "panglao_ontology"
-      )
+      ),
+      relationship = "many-to-many"
     ) |>
     # use unknown for NA annotation but keep ontology ID as NA
     # if the sample type is cell line, keep as NA
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -14,7 +14,8 @@ option_list <- list(
   make_option(
     opt_str = c("-i", "--input_sce_file"),
     type = "character",
-    help = "path to rds file with input sce object"
+    help = "path to rds file with input sce object",
+    default = ""
   ),
   make_option(
     opt_str = c("-o", "--output_sce_file"),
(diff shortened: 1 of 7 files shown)


bin/add_celltypes_to_sce.R
```diff
@@ -169,29 +180,30 @@ if (!is.null(opt$singler_results)) {
...
-if (!is.null(opt$cellassign_predictions)) {
-  # check that cellassign predictions file was provided
-  if (!file.exists(opt$cellassign_predictions)) {
-    stop("Missing CellAssign predictions file")
-  }
-
-  if (is.null(opt$cellassign_ref_file)) {
-    stop("CellAssign reference filename must be provided")
-  }
+if (file.exists(opt$cellassign_predictions)) {
```
jashapiro, line 189: Can we simplify some things by checking the file size here?

Then we don't need the `else` block at the end or the internal check where we set predictions as NULL?

```suggestion
has_cellassign <- file.exists(opt$cellassign_predictions) && (file.size(opt$cellassign_predictions) > 0)
if (has_cellassign) {
```

bin/add_celltypes_to_sce.R
```diff
@@ -74,30 +93,22 @@ get_ref_info <- function(ref_filename, extension) {
...
 
-if (!is.null(opt$singler_results)) {
-  if (!file.exists(opt$singler_results)) {
-    stop("Missing SingleR results file")
-  }
-
-  if (is.null(opt$singler_model_file)) {
-    stop("SingleR model filename must be provided")
-  }
+if (file.exists(opt$singler_results)) {
```
jashapiro, line 107: I think it makes sense to just set this directly and then use it in both places?
```suggestion
has_singler <- file.exists(opt$singler_results)
if (has_singler) {
```

bin/add_celltypes_to_sce.R
```diff
@@ -74,30 +93,22 @@ get_ref_info <- function(ref_filename, extension) {
...
+  # check singler model has been provided
+  stopifnot("Singler model filename must be provided" = opt$singler_model_file != "")
 
+  has_singler <- TRUE # use to decide if we should assign consensus labels
```
jashapiro, line 111: If you take my suggestion above, you can remove this.
```suggestion
```

bin/add_celltypes_to_sce.R
```diff
@@ -169,29 +180,30 @@ if (!is.null(opt$singler_results)) {
 
   # add note about cell type method to metadata
   metadata(sce)$celltype_methods <- c(metadata(sce)$celltype_methods, "singler")
+} else {
+  has_singler <- FALSE # use to decide if we should assign consensus labels
```
jashapiro, line 184: And remove this!
```suggestion
} else {
  has_singler <- FALSE # use to decide if we should assign consensus labels
```

bin/add_celltypes_to_sce.R
```diff
@@ -169,29 +180,30 @@ if (!is.null(opt$singler_results)) {
...
+  panglao_ref_df <- readr::read_tsv(opt$panglao_ontology_ref)
 
   # if the predictions file isn't emtpy read it in
   if (file.size(opt$cellassign_predictions) > 0) {
     predictions <- readr::read_tsv(opt$cellassign_predictions)
+    has_cellassign <- TRUE # use to decide if we should assign consensus labels
   } else {
     predictions <- NULL
+    has_cellassign <- FALSE # if cell assign wasn't run correctly, don't assign consensus
   }
```
jashapiro, line 207: If we check the file size earlier, then we should be good here?

```suggestion
  # read in predictions file
  predictions <- readr::read_tsv(opt$cellassign_predictions)
```

bin/add_celltypes_to_sce.R
```diff
@@ -250,6 +267,47 @@ if (!is.null(opt$cellassign_predictions)) {
     }
     metadata(sce)$cellassign_reference_organs <- cellassign_organs
   }
+} else {
+  has_cellassign <- FALSE # use to decide if we should assign consensus labels
+}
```
jashapiro, line 272: With `has_cellassign` already set, we shouldn't need this.
```suggestion
} 
```

bin/add_celltypes_to_sce.R
```diff
@@ -250,6 +267,47 @@ if (!is.null(opt$cellassign_predictions)) {
...
+
+# assign consensus cell type labels
+if (has_singler & has_cellassign) {
```
jashapiro, line 275: For clarity that there will only be one value.
```suggestion
if (has_singler && has_cellassign) {
```

bin/add_celltypes_to_sce.R
```diff
@@ -250,6 +267,47 @@ if (!is.null(opt$cellassign_predictions)) {
...
+  # now make sure that reference file exists
+  stopifnot(
+    "Consensus cell type reference file does not exist" = file.exists(opt$consensus_celltype_ref)
+  )
+
+  # read in consensus table
+  consensus_ref_df <- readr::read_tsv(opt$consensus_celltype_ref) |>
+    # select unique combinations of consensus refs based on ontology columns
+    dplyr::select(blueprint_ontology, panglao_ontology, consensus_ontology, consensus_annotation) |>
+    unique()
```
jashapiro, line 285: For uniformity (and a touch faster?)
```suggestion
    dplyr::distinct()
```

bin/add_celltypes_to_sce.R
```diff
@@ -250,6 +267,47 @@ if (!is.null(opt$cellassign_predictions)) {
...
+      consensus_ref_df,
+      by = c(
+        "singler_celltype_ontology" = "blueprint_ontology",
+        "cellassign_celltype_ontology" = "panglao_ontology"
+      ),
+      relationship = "many-to-many" # account for multiple of the same cell type
+    ) |>
+    # use unknown for NA annotation but keep ontology ID as NA
+    # if the sample type is cell line, keep as NA
+    dplyr::mutate(consensus_annotation = dplyr::if_else(is.na(consensus_annotation), "Unknown", consensus_annotation))
```
jashapiro, line 305: I like `tidyr::replace_na()` for this to be more explicit about what is being done, and much more concise!
I'm confused by the comment about cell lines though, as it doesn't seem that is implemented here?
```suggestion
    tidyr::replace_na(list(consensus_annotation = "Unknown"))
```

jashapiro (commented):
This looks good overall, but I had some suggestions for simplification. 

The other big thing is that I think we probably do want the ontology table and consensus table to live in this repo; particularly for the ontologies it seems like that should live alongside the panglaodb reference that is already here. The consensus file too would need to be updated if/when the references were, so I feel like it belongs in this repo as well.

config/reference_paths.config
```diff
@@ -22,6 +22,11 @@ singler_references_dir = "${params.celltype_ref_dir}/singler_references"
 singler_models_dir = "${params.celltype_ref_dir}/singler_models"
 cellassign_ref_dir = "${params.celltype_ref_dir}/cellassign_references"
 
+// Reference file used for assigning ontology IDs for CellAssign results and consensus cell types
+// Originally created in the `cell-type-consensus` module of `OpenScPCA-analysis`
+panglao_ref_file = 'https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-consensus/references/panglao-cell-type-ontologies.tsv'
+consensus_ref_file = 'https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv'
```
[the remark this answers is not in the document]
allyhawkins: > I think it probably makes sense for these two files to either live in this repository or be stored on S3. While I am comfortable with the `OpenScPCA-nf` scripts referencing the `OpenScPCA-analysis` repo, I think It may make sense for this one to remain more self-contained.

I'm fine with this approach. The reason I have it like this is because when we talked about it during pre-planning we had decided to just use the file within the analysis repo to minimize where we maintain the file, but I can move it into this repo. I think given the size we can just store it in the repo rather than having to put it on S3.

allyhawkins:
@jashapiro I updated the statements around `has_singler` and `has_cellassign` like you suggested, with one modification. If the `CellAssign` predictions file exists but has size 0 then `CellAssign` failed or didn't complete for some reason. We account for that by putting "Not run" in the assignment columns and I wanted to keep that. 

The other thing I did was add the reference files to the repo and included some supporting documentation. I tested with the new changes and things worked as expected so this should be ready to go.

PR #841, opened 2025-04-09
Account for consensus columns in merged objects

### LINKED ISSUE #840: Account for consensus cell type columns in merged objects
When working on the docs for consensus cell types, I noticed that we have a section of the `merge_sces.R` script where we account for cell types in some but not all libraries. We will need to do the same when we have some libraries with consensus cell types and others without it. We need to update this section of code https://github.com/AlexsLemonade/scpca-nf/blob/16fb37f8c5b17e41e557db852c9521a90705da2e/bin/merge_sces.R#L148 and make sure we include the new cellassign ontology and consensus cell type columns in the `possible_columns` list on L205.

allyhawkins:
Closes #840 

Here I am updating the script for merging SCEs to include the new columns added in #824. We should have one new column for the `cellassign_celltype_ontology` and then two new columns for consensus annotations. I added those columns to the list of expected columns and then updated some of the code we use for handling situations where not all libraries in an object have the same set of cell types. 

In doing this I was able to remove the list of celltype methods and just detect if cell types were annotated based on the names in the `colData`. This just meant deleting a few extra lines of code and moving some of the code up. 

I tested this with a few libraries where only 1 had consensus cell types and the expected columns were present and filled in appropriately. I also did a run with the latest changes to the matrix format in `scpcaTools` and checked that the format of the counts is now correct.

### DIFF
--- a/bin/merge_sces.R
+++ b/bin/merge_sces.R
@@ -152,62 +152,49 @@ present_columns <- sce_list |>
   purrr::map(\(sce) names(colData(sce))) |>
   purrr::reduce(union)
 
-if ("submitter_celltype_annotation" %in% present_columns) {
-  # Add `"Submitter-excluded"` value to any libraries without submitter
-  sce_list <- sce_list |>
-    purrr::map(\(sce) {
-      if (!"submitter" %in% metadata(sce)$celltype_methods) {
-        colData(sce)$submitter_celltype_annotation <- "Submitter-excluded"
-      }
-      return(sce)
-    })
-}
-if ("singler_celltype_annotation" %in% present_columns) {
-  # Check if the label used for annotation was ontology in at least 1 SCE
-  use_ontology <- sce_list |>
-    purrr::map_lgl(\(sce) {
-      any(metadata(sce)$singler_reference_label == "label.ont")
-    }) |>
-    any()
 
-  # Add `"Cell type annotation not performed"` string to libraries without SingleR
-  sce_list <- sce_list |>
-    purrr::map(\(sce) {
-      if (!"singler" %in% metadata(sce)$celltype_methods) {
-        colData(sce)$singler_celltype_annotation <- "Cell type annotation not performed"
-        if (use_ontology) {
-          colData(sce)$singler_celltype_ontology <- "Cell type annotation not performed"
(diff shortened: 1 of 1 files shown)


bin/merge_sces.R
```diff
@@ -236,9 +249,6 @@ possible_columns <- c(
 )
 
 # Define colData columns to retain based on intersection with present columns
-present_columns <- sce_list |>
-  purrr::map(\(sce) names(colData(sce))) |>
-  purrr::reduce(union)
```
sjspielman, line 241: Did you mean to delete this?

sjspielman (commented):
Great catch here! 

In addition to the comment I left, I do feel like this code could be cleaned up for future us. If/when we continue to add cell type annotation columns (e.g., from OpenScPCA-analysis modules), we'll keep needing these kinds of if blocks and that doesn't seem very robust. Potentially making this more modular would be a separate issue/PR though, and not necessarily one to do in the near future. But, maybe we open an issue for whenever the next time we* add a cell type label comes around to remind us to have a fresh look here? What do you think?

bin/merge_sces.R
allyhawkins: Yes I moved it up to line 150 so I could use the `present_columns` variable earlier.

sjspielman: Ah, yes!

sjspielman (approved):
lgtm!

allyhawkins:
> In addition to the comment I left, I do feel like this code could be cleaned up for future us. If/when we continue to add cell type annotation columns (e.g., from OpenScPCA-analysis modules), we'll keep needing these kinds of if blocks and that doesn't seem very robust. Potentially making this more modular would be a separate issue/PR though, and not necessarily one to do in the near future. But, maybe we open an issue for whenever the next time we* add a cell type label comes around to remind us to have a fresh look here? What do you think?

At first I thought this might be too much effort for right now, but then I looked at it a little more and I thought there must be a way to make this easier to follow and add to in the future as you said, so I went ahead and did it. @sjspielman can you take another look?

sjspielman (approved):
this looks right to me, thanks for doing this!! 🎉

PR #843, opened 2025-04-10
Use new scpcaTools container and bump version to v0.8.7

### LINKED ISSUE #838: Update scpcaTools container to use v0.4.2
Once we have a new release of `scpcaTools` we should update the container to use the tag `v0.4.2` (see https://github.com/AlexsLemonade/scpcaTools/issues/290). 
We should also do a test of the merged workflow and ensure that the output SCE object is the correct format once making this change and before doing #837.

allyhawkins:
Closes #838 
Towards #837 

Here I'm bumping the `scpcaTools` container to the latest release, `v0.4.2`. I did a test run with the container using `edge` right before I did the release and things work as expected for both the main and merge workflow. 

I also updated the version tag throughout the config and docs in preparation for the release.


sjspielman (approved):
🐑

PR #844, opened 2025-04-11
Sync `development` to `main` for v0.8.7

allyhawkins:
Towards #837 

Here I'm syncing the changes from `development` to include in `main` for the latest release. This includes the addition of consensus cell types, but also a few more items. I summarized the changes below, is there anything we shouldn't include? 

- [Add bulk TPM file](https://github.com/AlexsLemonade/scpca-nf/pull/816/files)
- [Resource limits and nextflow schema](https://github.com/AlexsLemonade/scpca-nf/pull/811) and [contributors formatting](https://github.com/AlexsLemonade/scpca-nf/pull/817)
- [Remove spatial counter](https://github.com/AlexsLemonade/scpca-nf/pull/819)
- [Separate staging and prod profiles](https://github.com/AlexsLemonade/scpca-nf/pull/827)
- [Metrics table](https://github.com/AlexsLemonade/scpca-nf/pull/834)
- [Consensus cell types](https://github.com/AlexsLemonade/scpca-nf/pull/835) and [merged consensus cell types](https://github.com/AlexsLemonade/scpca-nf/pull/841)

Note that if we keep the bulk TPM file export in this release we should update the hand off script to make sure that only the `bulk_quant.tsv` file gets handed off. Right now, the script will hand off both files: https://github.com/AlexsLemonade/ScPCA-admin/blob/a2b2daf0caac336f391bac974bc1b808f29d5853/data-processing-and-handoff/handoff-scpca-data.py#L110


jashapiro (approved):
I think including all of these changes makes sense to me. I am wondering why the development branch seems to be out of sync with main though? Do we need to make a separate branch to merge from to get things into sync? i.e., branch off main, sync dev into that branch, merge that branch into both main and dev?  I'm approving in case you don't want to do that, but I think we maybe shoud?

Other than that, I agree that we don't want to sync the TPM to the portal at this stage; it is very cheap to output, so I like having it around, but we should file an issue to update the handoff script.

allyhawkins:
> I think including all of these changes makes sense to me. I am wondering why the development branch seems to be out of sync with main though? Do we need to make a separate branch to merge from to get things into sync? i.e., branch off main, sync dev into that branch, merge that branch into both main and dev? I'm approving in case you don't want to do that, but I think we maybe shoud?

I'm not sure I understand this comment. All of these changes were made in `development` first so they should not be in `main`, but this PR will add them to `main`. After which the two branches will be in sync? We did a release in January that cherry picked some changes from `development`, which may be what you are seeing but this should solve that. Unless I'm missing something...

PR #845, opened 2025-04-11
Change to use H5AD in main overview diagram

### LINKED ISSUE #826: Update main diagram to have `H5AD` rather than `HDF5` in icons
I noticed that the [overview diagram](https://github.com/AlexsLemonade/scpca-nf/blob/main/diagrams/scpca-nf-overview.png) we show has the incorrect file extensions in the icons for `AnnData`. We should update this to have `H5AD` in the icons. Note that this image also exists in [`scpca-paper-figures`](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/figures/svgs/overview-scpca-fig2.svg) so we can just grab it from there.

allyhawkins:
Closes #826 

I meant to do this before doing the release but forgot, but still probably worth doing. I just changed the overview diagram to use `H5AD` in the icons for the AnnData output. I'm going to file this change to both `development` and `main`.


sjspielman (approved):
👍

PR #847, opened 2025-04-14
Make sure `has_cellassign` is set to `FALSE` if CellAssign doesn't complete successfully 

allyhawkins:
In re-processing all the data for the Portal, I came across a library that was throwing a join error when assigning consensus cell types. It looks like this library had a failure in `CellAssign` that leads to us labeling the cells as `Not run`. In this case we probably should not attempt to assign consensus cell types, so here I'm adding a `has_cellassign <- FALSE` inside the if statement that checks for if `CellAssign` completed.


jashapiro (approved):
LGTM

PR #849, opened 2025-04-15
Initial outline of change report

### LINKED ISSUE #842: Create outline document of change report
As a first step in #833, we should create an Rmd report file that outlines the format of the final report.

To do this, we would include the headers and notes on the overall structure, including what kinds of tables or figures we might want in the report. This step would not include any code, but would serve as a scaffolding that we can use for discussion and comment, and then as a scaffolding into which we can slot code chunks to implement.

jashapiro:
Closes #842

Here I am adding the start of a template for reporting changes in cell metrics. 

The main design is that this notebook will read all metrics files (and possibly metadata files) from S3 given two sets of paths to results, then generate various summary and detail tables of changes between versions.

I included some of the basic code I plan to use to get and read the files, but it is not set as an executable block at the moment; I mostly just wanted it there for reference.

The main portions of the outline are in the remainder of the document. These are mostly my initial thoughts, and I welcome any suggestions for updates or changes! I expect some of these ideas will need modification as code is added. 

For many of the comparisons, I implemented the report to include both a "Summary" and "Detail" view, implemented with `{.tabset}` so that the detail will be generally hidden but easy to reveal. I was pretty happy when I (re)found this feature, and I think this is a great use for it, but let me know if you think I should take a different approach.

Some questions, like cutoffs for warning displays, etc. may have to wait for implementation and/or real data. I expect to implement the code for this is a series of PRs rather than all at once.

One other question is how to handle package dependencies. I think for now my thought is that I will create a directory within the `scripts` directory that would hold a wrapper script to run this notebook (and maybe move the notebook there?) and also define the renv that it requires (`s3fs`, `jsonlite`, `dplyr`, etc.). I don't know if we plan to create a Nextflow workflow to run this metric specifically, or if we want to trigger it another way, but it seemed like creating a Docker image at this stage was probably overkill, and we can of course convert an renv file to a Docker image without much difficulty. 

We will also want to decide where we want these reports to be stored/shared, but I think that can also wait until we have more of an implementation.


templates/compare-metrics/compare-metrics.rmd
```diff
@@ -0,0 +1,182 @@
...
+### Summary
+For these metrics, we can create table of changes, with the counts of the number of samples with each change.
+Something like the below:
+
+| Processing field | Reference value | Comparison value | Number of samples changed |
+| ----- | --------- | --------- | ---------------- |
+| `droplet_filtering_method` | `emptyDrops` | `emptyDropsCellRanger` | 10 |
+
+### Detail
+If desired, we could also include a hidden table with individual sample changes.
```
allyhawkins, line 94: Yes, I think it would be helpful to list out the samples where these things have changed.

templates/compare-metrics/compare-metrics.rmd
```diff
@@ -0,0 +1,182 @@
...
+
+## Processing changes {.tabset}
+
+Describe any changes to the software versions or parameters used to process the data.
+
+Fields to compare:
+
+- `droplet_filtering_method`
+- `normalization_method`
+- `cell_filtering_method`
```
allyhawkins, line 71: Generally, I think these three items are going to be where we see the most potential changes, since others are parameters we set in the workflow itself but it can't hurt to check all of them just in case.

templates/compare-metrics/compare-metrics.rmd
```diff
@@ -0,0 +1,182 @@
...
+If there are changes in any of those, we will report those, with prominent warnings for any large changes.
+What defines a large change?
+Perhaps start with a change > 0.5% of the total?
+Alternatively, if we see a large fraction of samples changing, we would want to note that.
+
+#### Summary
+The main display item would be a summary table, in the first tab:
+
+| Metric | Number of samples changed | % of samples changed | Mean % change |
+| ----- | --------- | --------- | ---------------- |
```
allyhawkins, line 118: I don't know that this table is going to be the most informative since we expect most of the samples to have some small changes for each of these metrics. So I would expect the number of samples and % of samples to be really high with a low mean percentage. We will need a way to find samples that are outliers in the mean % change. 

On the other hand, if a lot of samples have a large % change then we want to be able to catch that, so then that makes this table useful. I think this particular section will need to be flushed out once you start exploring the data.

allyhawkins: Another though might be plotting the differences in cell numbers for each library. We should expect most to be similar, but then you could find libraries where the difference is much higher?

templates/compare-metrics/compare-metrics.rmd
```diff
@@ -0,0 +1,182 @@
...
+### Clustering
+
+Report changes in the number of clusters, as well as changes in the sizes of the assigned clusters.
+We will ignore changes in the ordering of clusters by sorting by cluster size before comparisons.
+As a summary statistic, we might use adjusted Rand index (ARI) or adjusted mutual information (AMI) to compare the two sets of clusters, reporting when we are well below 1.
+
+### Cell type assignments
+
+Report changes to the set of cell types assigned, and to the number of cells for each cell type.
+This will include SingleR, CellAssign and consensus cell types, if present.
```
allyhawkins, line 180: Note that we will need to add the dictionary for consensus cell types to the metrics file, since currently it is not included.

templates/compare-metrics/compare-metrics.rmd
jashapiro: I like the idea of a plot: A histogram of % changes would be a good summary. 

I was planning to sort the detail table by % change, but I was also considering the [`DT` library](https://rstudio.github.io/DT/) for output to allow user sorting & search.

jashapiro:
Thanks for the comments @allyhawkins! I made a few updates to allow specifying the project if required. I also added some notes about making plots. Do you have any more thoughts on the outline? 

I expect implementation of this to happen in stages, with the next step getting the main infrastructure done and adding the first section of output.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -0,0 +1,200 @@
...
+| `SCPCS999999` | `unfiltered_cells` | 1000 | 1005 | 5 | 0.5 |
+
+
+### Read count changes {.tabset}
+
+If there are changes in the cell counts, we expect changes in the read counts as well.
+So we may only want to proceed with read count changes for samples where cell counts are unchanged.
+Here again, I think we should warn/highlight any changes > 0.5% of the total.
+
+- `unfiltered_total_counts`
```
allyhawkins, line 154: I actually don't think we expect there to be a change in the unfiltered total counts or genes expressed unless there was a change in salmon/alevin-fry. We might want to pull that out into it's own section and only include it if there is a version change in the tools used for mapping? 

Alternatively, we should just check that these numbers are the same and print out any samples where it's different, since that to me is a red flag.

allyhawkins: Hmm thinking about this more, they _could_ change if the version of the mapping tools are the same, but the workflow was run with `--repeat_mapping`. Maybe we want to track that somewhere since generally we are re-running without mapping, so I think having a way to confirm that the unfiltered counts/reads are the same seems like a good check.

allyhawkins (approved):
This looks good to me. I just had one comment about checking unfiltered reads/counts which I think we want to account for and will need to be flushed out when we add those sections in. I don't need to see this again, but maybe at least add a comment about that in the outline so we don't lose track of it.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: Having the check seems like it is low cost, and I was only planning to show changed values anyway, so it should not show up. I want the red flag in case we make some other mistake!

jashapiro: I added a note for both the cell count and the unfiltered totals to make sure we highlight those changes in particular.

PR #856, opened 2025-04-22
Add infrastructure and processing info to change report 

### LINKED ISSUE #851: Implement report steps for reading metrics json files and reporting on processing changes
This PR will set up the renv required to run the change report notebook as well as a wrapper script to perform the rendering.

The content added to the notebook will include reading the metrics json files into a list, and then parsing that list to compile a table of how processing parameters have changed between runs.

jashapiro:
This PR starts the main tasks of creating the report notebook.
Closes #851 

The first parts was initiating the `renv`, which should all be pretty standard.

The main entrypoint for rendering is now the `compare-metrics.R` script. This has a number of options for setting up paths and project options. There is also an option to skip the download of the json files from S3. This is mostly for testing, as the slowest part of the script is listing all the files on S3 and then downloading the `_metrics.json` files (listing is actually slower, because the API does not allow filtering of the files at the AWS end!). As this gets closer to being done, I expect to remove saving and reading the `metrics.rds` file or otherwise cleaning that up.

The notebook itself now includes the initial processing of the `_metrics.json` files, as well as the first set of summaries. Running this at the moment results in no changes, which is to be expected. But I did some quick shuffling to make examples of changed libraries and processing. The result can be found in the attached file:
[compare-metrics.html.zip](https://github.com/user-attachments/files/19854381/compare-metrics.html.zip)

I did go ahead and use `DT` for all of the tables: The formatting is pretty much default at the moment, but it should give a fair amount of flexibility if we want to do things like highlighting cells by value (or even making the cell rows bar plots! see https://rstudio.github.io/DT/functions.html)

I kept the text pretty minimal at this point; I mostly wanted to test out making the tables and such. I don't know that we need too much text, but probably a bit more than what I currently have may make sense.


scripts/compare-metrics/compare-metrics.R
```diff
@@ -0,0 +1,65 @@
...
+    help = "Project ID(s) to filter the metrics files. Default is all projects"
+  ),
+  make_option(
+    c("-o", "--output_file"),
+    type = "character",
+    default = "compare-metrics.html",
+    help = "Output file name for the rendered report"
+  ),
+  make_option(
+    c("--skip_download"),
```
allyhawkins, line 35:
```suggestion
    c("--use_cache"),
```
Maybe we can be more explicit about what this option is doing? I might also mirror this name change in the template.

scripts/compare-metrics/compare-metrics.R
```diff
@@ -0,0 +1,65 @@
...
+
+rmarkdown::render(
+  file.path(script_dir, "compare-metrics-template.rmd"),
+  output_file = basename(opt$output_file),
+  output_dir = dirname(opt$output_file),
+  params = list(
+    reference_s3 = opt$ref_s3,
+    comparison_s3 = opt$comp_s3,
+    project_id = project_ids,
+    download_s3 = !opt$skip_download
```
allyhawkins, line 62:
``` suggestion
    use_cache = opt$use_cache
```
If you take my previous suggestion, I would do this since I think it's a little clearer. Trying to figure out if it was correct to use `!` here was tripping me up since the arguments were the inverse of each other.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -1,8 +1,9 @@
 ---
 params:
-  reference_url: s3://nextflow-ccdl-results/scpca-prod
-  comparison_url: s3://nextflow-ccdl-results/scpca-staging
+  reference_s3: s3://nextflow-ccdl-results/scpca-prod
+  comparison_s3: s3://nextflow-ccdl-results/scpca-staging
   project_id: "all"
+  download_s3: false
```
allyhawkins, line 6:
```suggestion
  use_cache: false
```
If you take my previous suggestions. You'll also need to update in the chunk directives.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,183 @@ output:
...
+  suffix = c(".ref", ".comp")
+)
+# temporary write for testing
+readr::write_rds(metrics_df, "metrics.rds")
+```
+
+```{r, eval = !params$download_s3}
+#| include: false
+# temporary chunk for testing
+metrics_df <- readr::read_rds("metrics.rds")
```
allyhawkins, line 83: I know this is temporary, but we should probably check that this exists earlier, otherwise you will still need to download.

scripts/compare-metrics/compare-metrics.R
allyhawkins: Also since you noted saving the rds file is temporary in the template notebook, does that mean this argument is temporary? If so, let's note that so we remember to remove it. 
Alternatively, we could keep it but ensure that when it runs in a workflow we do a fresh download of the metrics files.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,183 @@ output:
...
+    sample_id,
+    library_id,
+    added_removed
+  )
+```
 
-Output table:
+Below is a table of added or removed samples and libraries.
+There are **`r sum(lib_changes$added_removed == "Added")`** added libraries and **`r sum(lib_changes$added_removed == "Added")`** removed libraries.
+These come from **`r dplyr::n_distinct(lib_changes$project_id)`** project(s).
```
allyhawkins, line 106: I might make printing this text a little more dynamic. You can include this in the `else` statement below and only print the summary of how many libraries are added/removed if there were any. Otherwise just print a statement that no libraries/ samples were added or removed.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,183 @@ output:
...
+    ref != comp
+  )
+```
 
 ### Summary
 
-For these metrics, we can create table of changes, with the counts of the number of samples with each change.
-Something like the below:
+The table below (if present) shows a summary of processing changes between the reference and comparison workflow runs.
+There are a total of **`r dplyr::n_distinct(processing_changes$library_id)`** libraries with processing changes.
```
allyhawkins, line 156: Can we move these summary statements to be above the Summary/Detail tab? Otherwise they only print out in the summary tab and when you click on detail it's gone. I also might be dynamic about what to print here. If there are libraries that change I like including the bolded totals, but if there are no changes we should have a statement that says that and notes that no tables will be shown.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,183 @@ output:
...
-| `SCPCS999999` | `droplet_filtering_method` | `emptyDrops` | `emptyDropsCellRanger` |
-
-
+```{r}
+if (nrow(processing_changes) == 0) {
+  knitr::asis_output(
+    "<p>No processing changes</p>"
+  )
+} else {
+  DT::datatable(
```
allyhawkins, line 189: I think this detailed view might get very overwhelming if we have a lot of little changes across multiple projects. I think we should think about how to break up this table to make it easier to look at. I think having the ability to sort in `DT` is huge, but maybe we should break these detailed tables up so that we have one table per project?

allyhawkins (commented):
This looks mostly good, I just had a few comments about being more dynamic with what we say in the text. I also think we should break up the detailed tables a bit more otherwise if we have a lot of small changes it could be overwhelming.

scripts/compare-metrics/compare-metrics-template.rmd
allyhawkins: > I think the filtering tools in DT will make this pretty easy to handle: If you use the search tab with a project ID, it filters down to that project.

Wow, I totally missed the search bar. I agree that if we have that then no need to break the tables up!

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,247 @@ output:
...
+if (nrow(processing_summary) == 0) {
+  knitr::asis_output(
+    "<p>No processing changes</p>"
+  )
+} else {
+  report_table(
+    processing_summary,
+    colnames = c("Processing field", "Reference value", "Comparison value", "Number of libraries changed")
+  )
+}
```
allyhawkins, line 242:
```suggestion
```{r, eval=has_processing_changes}
processing_summary <- processing_changes |>
  dplyr::count(
    field,
    ref,
    comp,
    name = "n_changed"
  )

  report_table(
    processing_summary,
    colnames = c("Processing field", "Reference value", "Comparison value", "Number of libraries changed")
  )
```
I think since you have the statement that there are no processing changes above, then you could just not print anything? If you do this, then I think you should make the summary/detail tabs dynamic and not have them unless you have processing changes to look at.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,247 @@ output:
...
+    The tables below show the processing changes between the reference and comparison workflow runs.<br>
+    There {ifelse(n_added == 1, 'is', 'are a total of')} **{n_libraries}** {ifelse(n_libraries == 1, 'library', 'libraries')} with processing changes
+    from **{n_projects}** {ifelse(n_projects == 0, 'project', 'projects')}.
+  ") |>
+    knitr::asis_output()
+}
+```
 
-### Detail
+### Summary
```
allyhawkins, line 221:
```suggestion
```{r, eval=has_processing_changes}
knitr::asis_output("
### Summary
")
```
Since we're being dynamic here we could remove these tabs if there are no changes.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,247 @@ output:
...
+    ref != comp
+  )
+```
 
-| Processing field | Reference value | Comparison value | Number of samples changed |
-| ----- | --------- | --------- | ---------------- |
-| `droplet_filtering_method` | `emptyDrops` | `emptyDropsCellRanger` | 10 |
+```{r}
+# Text summary of processing changes
+if (nrow(processing_changes) == 0) {
```
allyhawkins, line 204:
```suggestion
# Text summary of processing changes
has_processing_changes <- nrow(processing_changes) == 0
if (has_processing_changes) {
```
Because you use this in multiple places, it might be good to define it separately first.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,247 @@ output:
...
+} else {
+  report_table(
+    processing_summary,
+    colnames = c("Processing field", "Reference value", "Comparison value", "Number of libraries changed")
+  )
+}
+```
 
+
+### Detail
```
allyhawkins, line 246:
```suggestion
```{r, eval=has_processing_changes}
knitr::asis_output("
### Detail
")
```

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -15,99 +16,247 @@ output:
...
+if (nrow(processing_changes) == 0) {
+  knitr::asis_output(
+    "<p>No processing changes</p>"
+  )
+} else {
+  report_table(
+    processing_changes,
+    colnames = c("Project", "Sample", "Library", "Field", "Reference value", "Comparison value")
+  )
+}
```
allyhawkins, line 258:
```suggestion
```{r, eval=has_processing_changes}
  report_table(
    processing_changes,
    colnames = c("Project", "Sample", "Library", "Field", "Reference value", "Comparison value")
  )
```

allyhawkins (commented):
This looks good and I like the text changes you made! I had one more suggestion to add since we're being dynamic with text then we mine as well remove the summary/detail tabs if we don't need them.

jashapiro:
> This looks good and I like the text changes you made! I had one more suggestion to add since we're being dynamic with text then we mine as well remove the summary/detail tabs if we don't need them.

Done! For posterity, the current version of the report: 
[compare-metrics.html.zip](https://github.com/user-attachments/files/19871787/compare-metrics.html.zip)

allyhawkins (approved):
LGTM

PR #861, opened 2025-04-23
Make test files default (for now) and updates to metrics report

### LINKED ISSUE #852: Create test `metrics.json` files with altered parameters and results
For testing the report, it will be convenient to have a set of metrics files with altered values that we can use to generate reports reflecting changes after processing.

We should be able to do this by copying the just the json files for one project to a new S3 location, and then editing a few of them. One of the simplest edits may be to rename a file or two, swapping identifiers so that the results disagree.

jashapiro:
closes #852 

I created a set of test files at `s3://nextflow-ccdl-results/metrics-test` which includes projects 1-10. I made this the default for the wrapper script for now. 

I swapped around a few samples, which results in one processing change in the current report, and the removal of lots of projects means there are a lot a changed files.

While I was testing this, I found a few little errors/optimizations that I cleaned up as well.
- I list only the `results` subdirectory, which saves a fair amount of time. It still takes a while to download and parse all the json files, but the listing is _much_ faster.
- I also made the regex more specific, as it turns out there are some `*_metrics.json` files that are produced by Space Ranger, which we want to be sure to skip (these are really only in the checkpoints, but better safe than sorry)
- NULL values turned out to cause some trouble (specifically showing up in libraries where there has not been cell typing done), so I added `default=NA` to the list transpose, which handled the job.
  - I wasn't sure if these projects (`SCPCP000020` and `SCPCP000024`) were _supposed_ to not have cell typing? Do we need to reprocess them?

### DIFF
--- a/scripts/compare-metrics/compare-metrics.R
+++ b/scripts/compare-metrics/compare-metrics.R
@@ -17,7 +17,8 @@ option_list <- list(
   make_option(
     c("-c", "--comp_s3"),
     type = "character",
-    default = "s3://nextflow-ccdl-results/metrics-test",
+    default = "s3://nextflow-ccdl-results/metrics-test", # temporary default for testing
+    # default = "s3://nextflow-ccdl-results/scpca-staging/results", # normal default
     help = "S3 URI for the bucket and prefix for the comparison S3 files"
   ),
   make_option(c("-p", "--project_id"),
--- a/scripts/compare-metrics/compare-metrics.R
+++ b/scripts/compare-metrics/compare-metrics.R
@@ -11,13 +11,13 @@ option_list <- list(
   make_option(
     c("-r", "--ref_s3"),
     type = "character",
-    default = "s3://nextflow-ccdl-results/scpca-prod",
+    default = "s3://nextflow-ccdl-results/scpca-prod/results",
     help = "S3 URI for the bucket and prefix for the reference files"
   ),
   make_option(
     c("-c", "--comp_s3"),
     type = "character",
-    default = "s3://nextflow-ccdl-results/scpca-staging",
+    default = "s3://nextflow-ccdl-results/metrics-test",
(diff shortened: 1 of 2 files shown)


scripts/compare-metrics/compare-metrics.R
```diff
@@ -11,13 +11,13 @@ option_list <- list(
...
     type = "character",
-    default = "s3://nextflow-ccdl-results/scpca-prod",
+    default = "s3://nextflow-ccdl-results/scpca-prod/results",
     help = "S3 URI for the bucket and prefix for the reference files"
   ),
   make_option(
     c("-c", "--comp_s3"),
     type = "character",
-    default = "s3://nextflow-ccdl-results/scpca-staging",
+    default = "s3://nextflow-ccdl-results/metrics-test",
```
allyhawkins, line 20: Maybe we should just add a comment here that this is temporary?

allyhawkins (commented):
> I wasn't sure if these projects (SCPCP000020 and SCPCP000024) were supposed to not have cell typing? Do we need to reprocess them?

These are both cell line only projects so they should not have cell typing. 

Everything else here looks good, but you probably want to regenerate the test files once #862 goes in?

scripts/compare-metrics/compare-metrics.R
jashapiro: Sounds reasonable: 
```suggestion
     default = "s3://nextflow-ccdl-results/metrics-test", # temporary default for testing
    # default = "s3://nextflow-ccdl-results/scpca-staging/results", # normal default
```

jashapiro:
> Everything else here looks good, but you probably want to regenerate the test files once #862 goes in?

I don't think we want to regenerate the test files now; we will have to regenerate all of the metrics files at some point after #862, but I don't think we need to do that yet.

allyhawkins (approved):
👍

PR #863, opened 2025-04-24
Add pixi environments for Cavatica (and general development)

### LINKED ISSUE #857: Explore sbpack installation and compatibility
To begin the job of testing `scpca-nf` on Cavatica, we will first need to create an environment with the tools required. This will include `nf-core` and `sbpack` at a minimum, but we need to make sure that we have compatible versions of each.

We can do this by creating a conda environment, but maybe I will pilot [`pixi`](http://pixi.sh) for this? One nice feature of `pixi` (aside from speed) is that it does a nice job of separating dependency description from the lock file.

The extent of this issue is to create the environment and perform an initial run of `sbpack_nf` to test compatibility, generally instructions at https://docs.cavatica.org/docs/bring-nextflow-apps-to-cavatica, using the `--dump-sb-app` to avoid pushing to Cavatica. I will also create a readme file to describe the dependency installation procedure as well as other setup components. The`sb_nextflow_schema.yaml` file that is created will not be included.

jashapiro:
Closes #857

To manage the dependencies for porting to Cavatica, I here set up some `pixi` environments, and added documentation about their use. 

Since we did not already have a `conda` environment defined in this repo, this seemed like a good place to pilot using `pixi`, and I will say that I found the experience pretty enjoyable. It is _much_ faster than `conda` or `mamba` and I like the fact that the environments are defined by the directory rather than given separate names, more like `renv`. I also really like that it automatically updates the environment to match the lockfile when launched, so you don't have to worry about using an old version of the environment if it was updated by another user.

It also makes it relatively easy to define multiple environments, which I have taken advantage of here to allow us to continue to use the latest version of `nf-core` for creating the manifest file, while also allowing the older version that `sbpack` depends on only when using the `cavatica` environment. I also use an earlier version of python there: the latest version that allowed sbpack to run without warnings.

I added documentation to the `CONTRIBUTING.md` file to describe `pixi` installation and use of the default environment. Separately, I made a directory and readme for Cavatica documentation, which includes describing invoking the `cavatica` environment. I'm not sure whether we really need a separate directory for Cavatica work, but it seemed reasonable to start with.

I'm very curious to hear how others find the use of `pixi` and whether there are additional instructions that should be added at this stage (before running any specific `sbpack` steps). 

I recommend testing by installing `pixi` on your system and then running  some of the following commands: 

```
pixi run nextflow -version # should be 24.10.6
pixi run nf-core --version # should be the latest version
pixi run -e cavatica sbpack_nf -h
pixi run -e cavatica nf-core --version # should be version 2.1
```

I'm still not quite sure that having the separate `nf-core` versions for the two steps will be compatible, but they seem to work for now, and should be easy to update if needed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,7 +123,7 @@ As an alternative to `black`, you can also use [`ruff`](https://docs.astral.sh/r
 
 To ease installation of dependencies and help with reproducibility, we have included a [pixi](https://pixi.sh) configuration file (`pixi.toml`) and lockfile (`pixi.lock`) in the root of the repository.
 `pixi` is a package manager that takes advantage of the conda ecosystem, but faster and with some nice extra features like automatically keeping the environment in sync with the lockfile and the ability to create multiple environments for a project.
-The `pixi.toml` file contains a list the high-level dependencies needed to run the workflow, and the `pixi.lock` file contains the exact versions of those dependencies that were used when the lockfile was created.
+The `pixi.toml` file contains a list of the high-level dependencies needed to run the workflow, and the `pixi.lock` file contains the exact versions of those dependencies that were used when the lockfile was created.
 
 The included default environment includes `nextflow`, `nf-core`, `pre-commit`, and their dependencies.
 
@@ -141,7 +141,7 @@ Either method will automatically update the environment to match the package ver
 
 For convenience, we have included a set of [pre-commit hooks](https://pre-commit.com/) that can be used to automatically format code according to the above specifications, as well as to spellcheck and check for other common errors.
 
-To use these hooks if you are using `pixi`, you can run `pixi run pre-commit install` from the `scpca-nf` directory.
+Once you have [installed `pixi`](#environment-management-with-pixi), you can run `pixi run pre-commit install` from the `scpca-nf` directory to use these hooks. 
 Alternatively, install the `pre-commit` package according to your favorite method (`pip`, `conda`, etc.) and then run `pre-commit install` in the `scpca-nf` directory.
 This will install the hooks in the `.git/hooks` directory, and they will be run automatically when you commit changes.
 If any of the hooks fail, the commit will be aborted, and you will need to fix the errors and re-commit.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,25 @@
 
 We welcome contributions to the `scpca-nf` workflow, including usage reports and suggestions from users.
 
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Issues](#issues)
+- [Pull requests and branch structure](#pull-requests-and-branch-structure)
+  - [Updating `nextflow_schema.json`](#updating-nextflow_schemajson)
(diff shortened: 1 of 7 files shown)


pixi.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[tasks]
+
+[dependencies]
+nextflow = ">=24.10.6,<25"
+pre-commit = ">=4.2.0,<5"
+
+
+[feature.nfcore.dependencies]
+nf-core = "3.*"
```
allyhawkins, line 16: It looks like this is getting installed in the main `scpca-nf` environment, but the other dependencies are getting installed in `scpca-nf:cavatica`. I would have naively expected this to be in `scpca-nf:nfcore` just based on how its accounted for in this file?

CONTRIBUTING.md
```diff
@@ -97,11 +118,31 @@ If you use Visual Studio Code, you can install the [`black` extension](https://m
...
+To use the default environment, you can then either prefix your command with `pixi run` (e.g. `pixi run nextflow`), or you can launch a shell with the environment activated by running `pixi shell`.
+Either method will automatically update the environment to match the package versions defined the `pixi.toml` and/or `pixi.lock` files.
+
+
 ## Pre-commit hooks
 
 For convenience, we have included a set of [pre-commit hooks](https://pre-commit.com/) that can be used to automatically format code according to the above specifications, as well as to spellcheck and check for other common errors.
 
-To use these hooks, install the `pre-commit` package according to your favorite method (`pip install pre-commit` or `conda install pre-commit`), then run `pre-commit install` in the `scpca-nf` directory.
+To use these hooks if you are using `pixi`, you can run `pixi run pre-commit install` from the `scpca-nf` directory.
```
allyhawkins, line 144:
```suggestion
Once you have [installed `pixi`](#environment-management-with-pixi), you can run `pixi run pre-commit install` from the `scpca-nf` directory to use these hooks. 
```
Just some rewording and adding the link to installing pixi.

allyhawkins (commented):
This looks pretty good and I like `pixi`! I love that it automatically updates and that you can run directly from the command line without activating the environment first. 
I was able to install it and run the commands you mentioned and just generally play around with it. 

I did have one clarifying question and then two small suggestions for the docs.

pixi.toml
jashapiro: Yeah, I was a bit confused by this at first as well. The tutorial on multi environments was helpful (though I mostly wrote the toml file by hand): https://pixi.sh/latest/tutorials/multi_environment/

`pixi` has separate concepts of "features" and "environments". You specify the packages associated with each feature, and then you can compose sets of features into environments. The main `[dependencies]` block (and `[pypi-dependencies]`, if it exists) defines the "default" feature, and is included in all of the environments. So in this case I always want nextflow and pre-commit to be present, so those are in `[dependencies]` (default). But I had to put `nf-core` in a separate feature (named `nfcore`), as I want to also  have an incompatible version in the `cavatica` environment.

Then at the bottom I define the environments in the [environments] block. The default environment would usually just contain whatever is in `[dependencies]`, but I also want it to include the `nfcore` feature, so I had to define it. I could have made it `["default", "nfcore"]` to be explicit, but the default environment is always included. I also defined the "cavatica" environment there to include the features I had defined for that. 

I'm going to modify it a bit to rename the feature as `sbpack`; that might make it a bit more clear?

allyhawkins: Makes sense, thank you for the explanation!

allyhawkins (approved):
LGTM

PR #864, opened 2025-05-02
Add cell and read count reporting to change tracking notebook

### LINKED ISSUE #853: Add cell and read count reporting to change tracking notebook
Following #851, we will want to add the section to report on changes in the numbers of cells and reads.

This will include both tables and plots, with both summary and detail views, as described in the notebook template.

jashapiro:
Here I am adding the cell count change summaries and detail table to the change report. This is the first half of #853 (but does not fully close it)

There was a bit of error correction I had to do because we are using unboxed JSON for the metrics file, which seems to be causing a bit of a challenge with NA values. But maybe only sometimes? The issue is that we have a few NA values for miQC pass values, which are getting read in as strings. In my testing if we were to export without `auto_unbox`, this would go away. So I added that change for some simplification of reading the file in the future, but for now I had to deal with what we have so there is some added code to be sure the NA values are treated appropriately and the columns turned back into simple vectors where needed. I did pull that out to a separate function at least, which makes other parts a bit simpler anyway.

The remaining code is pretty straightforward as far as building tables goes, I think. The most challenging part was dealing with NA values: we do want to report if a value changed to/from NA, so there is a bit of code I had to add to every `filter()`step to make sure those are retained.  I also added some code in the DT template (adapted from https://github.com/rstudio/DT/issues/322) to make sure they show up as `NA` rather than a blank field in the table. 

I followed the same general plan of attack as before with dynamic text.

I don't know for sure that the plots I made are exactly what we will want to end up with, but it is a bit hard to tell with the number of changes that we have at the moment what those will really look like with real data. So I did not spend too much time on the formatting, but if you have suggestions there, do feel free to make them!

### DIFF
--- a/scripts/compare-metrics/compare-metrics-template.rmd
+++ b/scripts/compare-metrics/compare-metrics-template.rmd
@@ -319,7 +319,7 @@ if (!has_cell_changes) {
   n_libraries <- dplyr::n_distinct(changed_cell_counts$library_id)
 
   glue::glue("
-    The tables below show changes in cell counts changes between the reference and comparison workflow runs at various stages of processing.<br>
+    The tables below show changes in cell counts between the reference and comparison workflow runs at various stages of processing.<br>
     There {ifelse(n_added == 1, 'is', 'are a total of')} **{n_libraries}** {ifelse(n_libraries == 1, 'library', 'libraries')} with cell count changes
     from **{n_projects}** {ifelse(n_projects == 1, 'project', 'projects')}.
   ") |>
--- a/scripts/compare-metrics/compare-metrics-template.rmd
+++ b/scripts/compare-metrics/compare-metrics-template.rmd
@@ -261,30 +261,86 @@ report_table(
 
 ### Cell count changes {.tabset}
 
-For each sample, we will want to first compare:
+```{r}
+# calculate the difference between reference and comparison for selected fields
 
-- `unfiltered_cells`
-- `filtered_cells`
-- `processed_cells`
-- `miqc_pass_count`
-- `scpca_filter_count`
-- `adt_scpca_filter_count`
+cell_fields <- c(
+  "unfiltered_cells",
+  "filtered_cells",
+  "processed_cells",
+  "miqc_pass_count",
+  "scpca_filter_count",
+  "adt_scpca_filter_count"
+)
 
-If there are changes in any of those, we will report those, with prominent warnings for any large changes.
-What defines a large change?
-Perhaps start with a change > 0.5% of the total?
-Alternatively, if we see a large fraction of samples changing, we would want to note that.
-If there are any changes to the number of unfiltered cells, that should be highlighted, as we do not expect this to change between runs.
+cell_changes <- metrics_df |>
+  tidyr::drop_na(starts_with("mapped_reads")) |> # remove added/removed libraries
+  dplyr::select(
+    project_id,
+    sample_id,
+    library_id,
+    starts_with(cell_fields)
+  ) |>
+  tidyr::pivot_longer(
+    cols = starts_with(cell_fields),
+    names_pattern = "(.+)\\.(ref|comp)$",
+    names_to = c("field", ".value") # .value separates into `ref` and `comp` fields
+  ) |>
+  dplyr::filter(is.finite(comp) | is.finite(ref)) |> # keep where at least one is finite
+  dplyr::mutate(
+    change = comp - ref,
+    change_frac = change / ref
+  )
+```
+
+
+The tables below show any changes in cell counts at different stages of filtering.
 
 #### Summary
-The main display item would be a summary table, in the first tab:
 
-| Metric | Number of samples changed | % of samples changed | Mean % change |
-| ----- | --------- | --------- | ---------------- |
-| `scpca_filter_count` | 100 | 20 | -10 |
+```{r}
+# Text summary of cell count changes
+cell_summary <- cell_changes |>
+  dplyr::summarise(
+    .by = c("field"),
+    # count where at least one version has a value
+    n = dplyr::n(),
+    n_changed = sum(is.na(change) | change != 0),
+    sample_change_frac = n_changed / n,
+    mean_change_frac = mean(change_frac, na.rm = TRUE)
+  )
+
+report_table(
+  cell_summary,
+  colnames = c("Metric", "Samples", "Number of samples changed", "% of samples changed", "Mean % change")
+) |>
+  DT::formatPercentage(c(4, 5), 2)
+```
+
 
 We will also include a plot with a histogram of the % changes for each metric where at least one sample changed.
 
+```{r}
+change_fields <- cell_summary |>
+  dplyr::filter(n_changed > 0) |>
+  dplyr::pull(field)
+
+cell_plot_df <- cell_changes |>
+  dplyr::filter(field %in% change_fields, change != 0) |>
+  dplyr::mutate(
+    field = factor(field, levels = cell_fields),
+  )
+
+ggplot(cell_plot_df, aes(x = change_frac)) +
+  geom_histogram(bins = 20) +
+  facet_wrap(vars(field)) +
+  labs(
+    title = "Histogram of changes in cell counts",
+    x = "Change proportion",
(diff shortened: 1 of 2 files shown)


allyhawkins:
@jashapiro do you by any chance have a rendered example report? If so, could you share it to make review easier?

jashapiro:
> @jashapiro do you by any chance have a rendered example report? If so, could you share it to make review easier?

Yes, sorry I had meant to include that! [compare-metrics.html.zip](https://github.com/user-attachments/files/20039617/compare-metrics.html.zip)

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -254,38 +271,152 @@ report_table(
...
+    # count where at least one version has a value
+    n = dplyr::n(),
+    n_changed = sum(is.na(change) | change != 0),
+    sample_change_frac = n_changed / n,
+    mean_change_frac = mean(change_frac, na.rm = TRUE)
+  )
+
+report_table(
+  cell_summary,
+  colnames = c("Metric", "Samples", "Number of samples changed", "% of samples changed", "Mean % change")
```
allyhawkins, line 366: Do we want to count libraries or samples here? In the processing changes table we count libraries, and I think that's what we want to do here.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -254,38 +271,152 @@ report_table(
...
+    field,
+    ref,
+    comp,
+    change,
+    change_frac
+  )
+
+report_table(
+  cell_detail,
+  colnames = c("Project", "Sample", "Library", "Metric", "Reference", "Comparison", "Change", "% change")
```
allyhawkins, line 415: I noticed that I have a weird `c` in the upper right hand corner of the `Detail` table when I view it? Scrolling left/right doesn't move it and I can click on it and the table will sort. I think it's an error with printing the `% change` column? 
Screenshot 2025-05-05 at 11 03 25 AM

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -254,38 +271,152 @@ report_table(
...
+  dplyr::mutate(
+    field = factor(field, levels = cell_fields),
+  )
+
+ggplot(cell_plot_df, aes(x = change_frac)) +
+  geom_histogram(bins = 20) +
+  facet_wrap(vars(field), ncol = 3) +
+  labs(
+    title = "Histogram of changes in cell counts",
+    x = "Change proportion",
```
allyhawkins, line 388: I feel like this should be `Change percentage` and not proportion?

allyhawkins (commented):
Thanks for attaching the rendered report! I had some questions and suggestions about the table that I think should be addressed. 

I think the histograms are good for now and we can always revisit them if we don't feel like they are actually helpful when looking at the real reports. This is what I had envisioned though when we talked about plots in the outline.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -254,38 +271,152 @@ report_table(
...
+}
+```
+
+```{r, eval=has_cell_changes}
+knitr::asis_output("#### Summary")
+```
 
-#### Detail
+```{r}
+# Summary table of cell count changes
```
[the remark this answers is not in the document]
jashapiro: Thank you for catching the 0% error! I'll make this and average of absolute changes

> In this table, the Samples column is the total number of samples where that metric was present? While Number of samples changed is the total number of samples where there is a change in that metric?

That is correct. I want to have both for the cases where a metric might not exist (e.g., miQC pass)

I think the changes we are seeing are correct: I swapped 2 sets of samples; one from project 01 and one from project 03, but the 01 sample I also removed one of the pair. I should probably make a few more select changes in the test data, but I think it is working as expected.

jashapiro (commented):
Thanks for your comments Ally!  I think I got them all, but let me know if there was anything else I missed. 

The one extra change was to make the `adt_scpca_filter_count` `NA` if there is not an ADT table. (Apparently `sum(NULL == "foo")` yields `0`, which is not what I guess I expected). Seeing the effects of this will require a rerun of the workflow, but it should mean that the library counts in the summary will not all be the same (and yes, they should be library counts).

I think I fixed the scrolling bug too: the updated file is attached.

[compare-metrics.html.zip](https://github.com/user-attachments/files/20049201/compare-metrics.html.zip)

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -254,38 +272,152 @@ report_table(
...
+
+cell_summary <- cell_changes |>
+  dplyr::summarise(
+    .by = c("field"),
+    # count where at least one version has a value
+    n = dplyr::n(),
+    n_changed = sum(is.na(change) | change != 0),
+    sample_change_frac = n_changed / n,
```
allyhawkins, line 361:
```suggestion
    library_change_frac = n_changed / n,
```
Since we are talking libraries and not samples

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -254,38 +272,152 @@ report_table(
...
+    mean_change_frac = mean(abs(change_frac), na.rm = TRUE)
```
allyhawkins, line 362: Something still seems off to me... Now the miQC pass count mean percentage is 0.01%. If we're removing NA and we have two libraries with 1.58% and -1.56% change then this should be 1.57% right?

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -254,38 +272,152 @@ report_table(
...
+  )
+
+report_table(
+  cell_summary,
+  colnames = c("Metric", "Libraries", "Number of libraries changed", "% of libraries changed", "Mean % change")
+) |>
+  DT::formatPercentage(c(4, 5), 2)
```
allyhawkins, line 369: I think this is not working and that might be why you are off on the mean %? 
What if you do this: 
```suggestion
  DT::formatPercentage(c("% of libraries changed", "Mean % change"), 2)
```
I don't know why using the indices wouldn't work, but the examples show using the column names 🤷‍♀️ 
https://rstudio.github.io/DT/functions.html

It looks like there's another example that uses two columns with: 
`DT::formatPercentage(4:5, 2)`

allyhawkins (commented):
This looks good, but I still think there's something off with the mean % change. I think it might be how you're using `DT::formatPercentage`.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: I think the issue here is the calculation includes all of the 0 changes. So the mean percent change is low. I could make it the mean given not 0 change, which probably makes more sense with expectations.

allyhawkins: > I think the issue here is the calculation includes all of the 0 changes. So the mean percent change is low. I could make it the mean given not 0 change, which probably makes more sense with expectations.

Ah! That definitely makes sense why the number is so low. But yea I think it would make sense to report the mean change of the samples that actually have a change. Otherwise I think it's misleading to show that the change is really small.

jashapiro: Updated in https://github.com/AlexsLemonade/scpca-nf/pull/864/commits/a614d0d997f4b44d2647c2afca456281c363bd71. I didn't make a new version of the file, but the percentages jumped up as expected.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: This is working as expected, but either way works, and the names are a bit more clear; I just didn't want to type them out... I'll make this change throughout for consistency.

jashapiro: turns out it has to be the original column names, not the display names. Done in  [0516e21](https://github.com/AlexsLemonade/scpca-nf/pull/864/commits/0516e210f858cd10da4c03993c5940215be2e31d)

allyhawkins (approved):
LGTM!

Issue #858, opened 2025-04-23
Create and "optimize"  `sb_nextflow_schema.yaml`


jashapiro:
Following the instructions at https://docs.cavatica.org/docs/bring-nextflow-apps-to-cavatica and https://github.com/rabix/sbpack/blob/master/sbpack/noncwl/Readme.md#sbpack_nf, we will create the initial `sb_nextflow_schema.yaml` file required for creating an app on Cavatica. We will then follow the instructions to update input and output settings.

Updates to instructions should cover the specific sbpack options that were invoked, as well as which changes to yaml files were made as optimizations.

jashapiro:
Closed by #865

PR #865, opened 2025-05-07
Create `sb_nextflow_schema.yaml`

jashapiro:
This is an initial version of the `sb_nextflow_schema.yaml` file, along with some additional tooling and documentation. 
Part of #858

I will start with the tooling: 

As the commands for generating and updating the schema file were getting tedious, I wrote a `cavatica-build.sh` script to apply the settings we want (rather than just listing them in the docs). Of particular note there is the `sb_exclude.txt` file that I am using to pass files and directories that should be ignored when packing up the workflow. Unfortunately, the default ignore list does not include `pixi` directories and files, which I noticed when the compress step was taking forever. Since the excluded files have to be passed in at the command line, but there are a lot of them (you can't just add to the default excludes), I went with the `$(< $exclude_file)` syntax, which seemed to work well.

Using the script I was also able to avoid the annoying behavior of making numbered yaml files rather than replacing the current one. I also keep the `sb_nextflow_schema.yaml` file in the `cavatica` folder rather than at the root of the repo.

I also made an abbreviated docs file so we don't just show the . I imagine we may want to update that in the future, but the current version seems fine for now.

I updated the  `nextflow_schema.json` file a fair amount as part of this PR: I mostly moved `description` fields to `help_text` and shortened the description. I did this to improve the translation of the files to the seven bridges format, where the long descriptions were making the site hard to read.

After generating the initial `sb_nextflow_schema.yaml` file, I removed all of the CONTAINER params from that file; I had hoped that `hidden` flags would work, as they did get copied over to the schema, but they still appear in the params, and for simplicity we don't need to show that in the interface. I can imagine removing other parameters as well to keep the interface simple, but I did not implement any of that yet.

Filing this as a draft, partly because there seems to be something that I am still missing as far as the config goes: the first run attempt got through a single process but failed after that, so I imagine there is a bit more to figure out. But it might make sense to get this part in as a start even so. 

@allyhawkins: Did you have a successful run last year when we were trying this? If so, I might ask you to do a couple things to see if I can compare the settings files.


cavatica/cavatica-build.sh
```diff
@@ -0,0 +1,46 @@
...
+      --sb-doc ${doc_file} \
+      ${mode_arg} \
+      --dump-sb-app
+    mv ../sb_nextflow_schema.yaml .
+elif [ "$ACTION" == "push" ]; then
+    # Push the Docker image to the registry
+    pixi run -e cavatica sbpack_nf \
+      --profile . \
+      --workflow-path .. \
+      --sb-schema sb_nextflow_schema.yaml \
```
allyhawkins, line 40: Should you add a check that this file exists before running it?

cavatica/cavatica-build.sh
```diff
@@ -0,0 +1,46 @@
...
+    fi
+
+    # Build sb_nextflow_schema.yaml
+    pixi run -e cavatica sbpack_nf \
+      --workflow-path .. \
+      --appid ${app_id} \
+      --sb-doc ${doc_file} \
+      ${mode_arg} \
+      --dump-sb-app
+    mv ../sb_nextflow_schema.yaml .
```
allyhawkins, line 34: I think we want to remove the `sb_nextflow_schema.yaml.bak` after running this if it exists?

cavatica/sb_nextflow_schema.yaml
allyhawkins, line 1: Just a note that when I checked out this branch and ran the cavatica-build.sh script, it modifies this file with some minor formatting changes. Mostly indentations, spacing, and using single-quotes instead of double-quotes.

nextflow_schema.json
```diff
@@ -87,150 +76,193 @@
...
           "type": "string",
           "default": "All",
-          "description": "Specify a comma-separated list of run IDs to run. Use \"All\" to run all samples in the run_metafile."
+          "description": "Run IDs",
+          "help_text": "A comma-separated list of run IDs to run. Use \"All\" to run all samples in the run_metafile."
         },
         "project": {
           "type": "string",
-          "description": "Run samples only from a given project ID"
+          "description": "Project ID",
```
allyhawkins, line 84:
```suggestion
          "description": "Project ID(s)",
```
Maybe? Just to show that it can be a single or multiple IDs?

nextflow_schema.json
```diff
@@ -87,150 +76,193 @@
...
+          "help_text": "Run samples only from a given project ID"
```
allyhawkins, line 85:
```suggestion
          "help_text": "A comma-separated list of project IDs. All samples with the provided project IDs will be run."
```
This should also be updated to reflect that you can provide multiple IDs.

nextflow_schema.json
```diff
@@ -87,150 +76,193 @@
...
-          "description": "alevin-fry outputs are not published by default. Use this option to publish them to the `checkpoints` folder."
+          "description": "Publish alevin-fry outputs",
+          "help_text": "alevin-fry outputs are not published by default. Use this option to publish them to the `checkpoints` folder."
         },
         "merge_run_ids": {
           "type": "string",
           "default": "All",
-          "description": "Which runs to merge when using the merge.nf workflow.",
+          "description": "Run IDs to merge",
+          "help_text": "A comma-separated list of run IDs to merge. Use \"All\" to merge all runs in the run_metafile.",
```
allyhawkins, line 121:
```suggestion
          "help_text": "A comma-separated list of run IDs to merge. Use \"All\" to merge all runs in the run_metafile. Only used in the merge.nf workflow",
```
I think we want to note that this is only needed for merge.nf, otherwise it's ignored.

nextflow_schema.json
```diff
@@ -87,150 +76,193 @@
...
           "hidden": true
         },
         "reuse_merge": {
           "type": "boolean",
-          "description": "If merge results already exist, recreating them can be skipped if desired"
+          "description": "Reuse merge results",
+          "help_text": "If merge results already exist, recreating them can be skipped if desired.",
```
allyhawkins, line 127: I might also note that this is only for the merge.nf workflow.

cavatica/cavatica-build.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -euo pipefail
+# This script is used to build and/or push the Cavatica app
+
+# Run from the script file location
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+# cavatica app id: <username>/<project>/<app>
+app_id="jashapiro/scpca-nf-test/scpca-nf"
```
allyhawkins, line 9: I think this is fine for us to use internally, but I don't think we want people pushing to our app. We probably want to make this an argument at some point.

allyhawkins (commented):
I tested the script you had and have a few minor comments, but I was able to build the schema and then push it successfully. Now running the workflow is a whole other story... 
When testing previously I had not had a true successful run that completed the whole workflow...

cavatica/cavatica-build.sh
jashapiro: I want to keep that around for people to be able to compare changes. I certainly used that a lot. I know git can handle it, but I was doing a number of tests between git commits.

cavatica/sb_nextflow_schema.yaml
jashapiro: Yes, this is a function of me doing manual editing afterward and automatic formatting by my editor. I could add running the yaml file through a code formatter to the script. Might do that if I can find an easy one.

allyhawkins (approved):
LGTM, just one typo fix. Note that I did not actually try to run anything on Cavatica though, so let me know if you would like me to do that.

PR #866, opened 2025-05-07
Add read counts

### LINKED ISSUE #853: Add cell and read count reporting to change tracking notebook

jashapiro:
closes #853 

Here I am adding the read count tables and plots to the report. These follow the same general form as the cell counts, but there were a few extra wrinkles for altexp reporting. The main thing was adding a line to break the list column into a set of columns based on which altExp was being used. `tidyr:unnest_wider` was exactly what I was looking for!

I simplified a few other things along the way: the effort to sort the plots non-alphabetically seemed not worth it, so I ended up removing the code that was dealing with factors there.

One important decision I want feedback on was whether or not to include samples that had cell count changes in the read count tables and plots. It seemed obvious to me that any changes in cell counts would result in changes in read counts, so it seemed like reporting those changes was redundant, and potentially confusing. So I only report the libraries that did not have corresponding cell count changes for a given processing step. I had a thought that we could make an additional table with _all_ changes (in a separate tab), but it still seemed like overkill. But please let me know if you have other thoughts!

Attached is the latest output: [compare-metrics.html.zip](https://github.com/user-attachments/files/20089177/compare-metrics.html.zip)

### DIFF
--- a/scripts/compare-metrics/compare-metrics-template.rmd
+++ b/scripts/compare-metrics/compare-metrics-template.rmd
@@ -447,6 +447,65 @@ We will also include changes in altExps using the following fields:
 These will require a bit of extra processing, in that we will probably want to extract the altExp names and add those to the field names.
 For example, we would want to use a field like `unfiltered_altexp_total_adt` in the comparison table.
 
+```{r}
+# calculate the read count differences between reference and comparison for selected fields
+read_fields <- c(
+  "unfiltered_total_counts",
+  "unfiltered_total_spliced",
+  "unfiltered_expressed_genes",
+  "filtered_total_counts",
+  "filtered_total_spliced",
+  "filtered_expressed_genes",
+  "processed_total_counts",
+  "processed_total_spliced",
+  "processed_total_logcounts",
+  "processed_expressed_genes",
+  "unfiltered_altexp_total",
+  "filtered_altexp_total",
+  "processed_altexp_total"
+)
+
+read_changes <- metrics_df |>
+  tidyr::drop_na(starts_with("mapped_reads")) |> # remove added/removed libraries
+  dplyr::select(
+    project_id,
+    sample_id,
+    library_id,
+    starts_with(read_fields)
+  ) |>
+  tidyr::unnest_wider( # create columns for separated altexps
+    contains("altexp"),
+    names_sep = "_"
+  ) |>
+  dplyr::rename_with( # rename to keep ref/comp at end
+    \(name) stringr::str_replace(name, "\\.(ref|comp)(_.+)$", "\\2.\\1"),
+    contains("altexp")
+  ) |>
+  tidyr::pivot_longer(
+    cols = starts_with(read_fields),
+    names_pattern = "(.+)\\.(ref|comp)$",
+    names_to = c("field", ".value") # .value separates into `ref` and `comp` fields
+  ) |>
+  dplyr::filter(is.finite(comp) | is.finite(ref)) |> # keep where at least one is finite
+  dplyr::mutate(
+    change = comp - ref,
+    change_frac = change / ref
+  )
+
+changed_read_counts <- read_changes |>
+  dplyr::filter(is.na(change) | change != 0) |>
+  # filter out fields that correspond to known cell count changes
+  dplyr::mutate(
+    # make corresponding field names for cell counts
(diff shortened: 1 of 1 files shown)


scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -422,39 +427,182 @@ report_table(
...
+} else if (!has_read_changes) {
+  knitr::asis_output(
+    "There may be changes in read counts, but only those expected due to changes in cell counts as described above."
+  )
+} else {
+  n_projects <- dplyr::n_distinct(changed_read_counts$project_id)
+  n_libraries <- dplyr::n_distinct(changed_read_counts$library_id)
+  glue::glue("
+    The tables below show changes in read counts between the reference and comparison workflow runs at various stages of processing.
+    These include only changes that are not expected based on changes in cell counts.<br>
```
allyhawkins, line 507: I agree with only showing those that have a different count but don't have a change in cell number.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -422,39 +427,182 @@ report_table(
...
+# warn if any unfiltered read counts have changed
+unfiltered_read_changes <- changed_read_counts |>
+  dplyr::filter(stringr::str_detect(field, "^unfiltered_")) |>
+  nrow()
 
-We will also include changes in altExps using the following fields:
+if (unfiltered_read_changes > 0) {
+  glue::glue("
+    <div class=\"alert alert-danger\">
+    **Warning**: Unfiltered cell counts have changed between the reference and comparison runs.
```
allyhawkins, line 525:
```suggestion
    **Warning**: Total reads for unfiltered cells have changed between the reference and comparison runs.
```
Since we're talking reads here and not cell counts.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -422,39 +427,182 @@ report_table(
...
+    library_change_frac,
+    mean_change_frac
+  )
 
-Again, we would report only changed metrics, starting with a summary table:
+report_table(
+  read_summary,
+  colnames = c("Metric", "Libraries", "Number of libraries changed", "% of libraries changed", "Mean % change")
+) |>
+  DT::formatPercentage(c("library_change_frac", "mean_change_frac"), 2)
```
allyhawkins, line 564: One thing I noticed is that we have a lot of 0.00% change for the change and percent change in the detail and then the mean % change in the summary. I think that's to be expected here since the changes are very small in comparison to the total number of reads. We could extend the number of digits we show? Or not worry about it, but I just wanted to point it out.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -422,39 +427,182 @@ report_table(
...
+```
 
-| Metric | Number of samples changed | % of samples changed | Mean % change |
-| ----- | --------- | --------- | ---------------- |
-| `scpca_filter_count` | 100 | 20 | -10 |
+```{r eval=has_read_changes}
```
allyhawkins, line 567: I might alter the width of these plots since some of the x-axis labels are getting cut off.

allyhawkins (approved):
I think this looks pretty good. I had only small comments, but I agree with your decisions and I don't need to see this again.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: I think for now not worrying about it... Curious what things look like with real data: I expect that we mostly won't have changes in this section, except perhaps tiny changes in the normalized counts.

allyhawkins: > I think for now not worrying about it... Curious what things look like with real data: I expect that we mostly won't have changes in this section, except perhaps tiny changes in the normalized counts.

Yea I totally agree. I was trying to think of a scenario where we would see read changes but no count changes and that was all I could think of. I hope we don't see large changes without changes in cell counts 😬

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: This was a bit unpredictable in my testing; I think for now I'm just going to leave it until we have real data.

Issue #867, opened 2025-05-09
[BUG] Compatibility with nextflow 25


jashapiro:
**Describe the bug**

There seems to be a possibly bug related to how we search for files that appeared when updating to nextflow 25.4. It is possible the is a nextflow bug, but we will need to investigate.

**To reproduce**
The error occurred in the following GHA run:
https://github.com/AlexsLemonade/scpca-nf/actions/runs/14930776628/job/41946213367#step:4:1

 and produced the following error
```
Run nextflow -log stub-run.log run main.nf -stub -profile stub -ansi-log false
  nextflow -log stub-run.log run main.nf -stub -profile stub -ansi-log false
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Zulu_jdk/17.0.15-6/x64
    JAVA_HOME_17_X64: /opt/hostedtoolcache/Java_Zulu_jdk/17.0.15-6/x64
    CAPSULE_LOG: none
N E X T F L O W  ~  version [2](https://github.com/AlexsLemonade/scpca-nf/actions/runs/14930776628/job/41946213367#step:4:2)5.04.0
Launching `main.nf` [boring_faggin] DSL2 - revision: ab2a2abcff
Executing workflow for all runs in the run metafile.
ERROR ~ The expected input files or Space Ranger results files for STUBR07 are missing.

 -- Check 'stub-run.log' file for details
ERROR ~ No such file or directory: test/runs/STUBR08/*_{R2,R2_*}.fastq.gz

 -- Check script 'modules/bulk-salmon.nf' at line: 167 or see 'stub-run.log' file for more details
ERROR ~ No such file or directory: test/runs/STUBR08/*_{R2,R2_*}.fastq.gz

 -- Check script 'modules/bulk-star.nf' at line: 42 or see 'stub-run.log' file for more details
WARN: [map_quant_feature:index_feature (1)] Unable to resume cached task -- See log file for details
WARN: [genetic_demux_vireo:starsolo_map:starsolo (1)] Unable to resume cached task -- See log file for details
WARN: [bulk_quant_rna:fastp (1)] Unable to resume cached task -- See log file for details
WARN: [map_quant_feature:index_feature (2)] Unable to resume cached task -- See log file for details
WARN: [map_quant_rna:alevin_rad (4)] Unable to resume cached task -- See log file for details
[...]

PR #868, opened 2025-05-09
Pin nextflow version

jashapiro:
As a temporary fix for #867, we want to prevent the action from using the latest version of Nextflow, so here I am pinning to the last working version of Nextflow until we can do some debugging.


allyhawkins (approved):
👍

Issue #854, opened 2025-04-21
Add Analysis result changes to change notebook


jashapiro:
Changes to analysis results will be added to the **Analysis Changes** section of the change report notebook, following the outline in the notebook itself. This will include reports on changes to the following sets of analyses: 

- Highly variable genes
- Clustering
- Cell type assignments
  - singleR
  - cellAssign
  - consensus

jashapiro:
closed by #878

PR #871, opened 2025-05-13
Add changes in HVGs and clustering to change report

jashapiro:
Part of #854 

Here I added sections for HVG changes and changes in clustering to the report. This started to get away from me, so there are kind of a lot of changes, more than I would have preferred. 

I didn't include summary tables for these, instead just opting to report the overall changes in text. I did however, make histograms for many of the numeric changes, and a stacked bar plot to illustrate changes in clustering visually.

The main challenge I ran into was the fact that I realized as I started to code this that I was not going to be able to calculate ARI among clusterings as I had planned, because to do that I would need the actual cluster assignments for each cell, which we did not include in the metrics file. I thought about filing issues to include all of those, but it seemed like overkill; we will almost certainly need to investigate more thoroughly anyway for these, so it didn't seem worth making all of the files that much larger (and having to regenerate everything).

So instead I am calculating what I am calling `maxARI`; which _should_ be the approximatelu possible ARI. I do this by pairing of clusters based on their sizes. I'm not sure that the algorithm I'm using is provably correct as the best pairing, but it seems to work as expected: it picks pairs based on finding the ones that match best first, then the next best match, etc.  In general, this seems to work pretty well. 

At @jaclyn-taroni's suggestion, I made stacked bar plots to be able to look at the changes in clustering between the two examples. I like how these came out, but it isn't always obvious how the maxARI calculation corresponds to the visualization, if I'm honest. I'm sorting these from lowest to highest ARI here and selecting the lowest 6, but the main driver seems to be the total number of cells in these cases (and the fact that there are only 3 values that can be calculated). 

The HVG calculations are a bit more straightforward: I am calculating two main measures: the Jaccard index to track changes in the included genes and a rank correlation to capture changes in gene order. The plots here should be pretty straightforward, but I'm not sure that my explanations of what I did are great, so do look at that. I might also want to change these plots from faceting to patchwork since the way I have it now I can't really label the x axis accurately.  

I'm attaching the most recent render: 
[compare-metrics.html.zip](https://github.com/user-attachments/files/20195409/compare-metrics.html.zip)


scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -86,6 +88,81 @@ report_table <- function(df, options = list(), ...) {
...
+      used = FALSE
+    ) |>
+    dplyr::arrange(desc(share_frac))
+
+  # pre-allocate result vectors
+  new_a <- integer(min(length(a_sizes), length(b_sizes)))
+  new_b <- integer(min(length(a_sizes), length(b_sizes)))
+  pair_count <- 0
+  # loop until all pairs are used
+  while (!all(match_df$used)) {
```
allyhawkins, line 122: Generally, this took me a while to figure out exactly what's going on here, so can you add some more descriptive comments?

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -86,6 +88,81 @@ report_table <- function(df, options = list(), ...) {
...
+  # then choose from smallest to largest, eliminating used pairs as you go
+
+  match_df <- expand.grid(
+    a_i = seq_along(a_sizes),
+    b_i = seq_along(b_sizes)
+  ) |>
+    dplyr::mutate(
+      a = a_sizes[a_i],
+      b = b_sizes[b_i],
+      share_frac = pmin(a, b) / pmax(a, b),
```
allyhawkins, line 112:
```suggestion
```
I don't think you use this at all in this function?

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -86,6 +88,81 @@ report_table <- function(df, options = list(), ...) {
...
+    match_df$used[used_a | used_b] <- TRUE
+  }
+  # add any missing indexes to the end
+  new_a <- c(new_a, setdiff(seq_along(a_sizes), new_a))
+  new_b <- c(new_b, setdiff(seq_along(b_sizes), new_b))
+  # return values
+  return(list(a_sizes[new_a], b_sizes[new_b]))
+}
+
+max_ari <- function(a_sizes, b_sizes) {
```
allyhawkins, line 142: This approach to calculating the max ARI seems reasonable to me. I don't really have a better idea and I think we will know more once we put this into practice if it's a good metric.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -151,6 +228,10 @@ lib_changes <- metrics_df |>
     library_id,
     added_removed
   )
+
+# filter metrics down libraries in both sets for the remainder of the notebook
```
allyhawkins, line 232:
```suggestion
# filter metrics down to libraries in both sets for the remainder of the notebook
```

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -606,23 +688,318 @@ report_table(
...
+  dplyr::mutate(
+    cluster_setequal = purrr::map2_lgl(cluster_sizes.ref, cluster_sizes.comp, setequal),
+    cluster_n_change = purrr::map2_int(
+      cluster_sizes.ref,
+      cluster_sizes.comp,
+      \(x, y) {
+        length(y) - length(x)
+      }
+    ),
+    cluster_n_change_frac = cluster_n_change / purrr::map_int(cluster_sizes.ref, length),
```
allyhawkins, line 839: Is this the total number of clusters that have changed divided by the total number of clusters?

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -606,23 +688,318 @@ report_table(
...
+
+if (!has_cluster_changes) {
+  knitr::asis_output(
+    "There are no changes in clustering between the reference and comparison runs."
+  )
+} else {
+  knitr::asis_output(glue::glue_data(
+    cluster_summary,
+    "Of the **{n}** libraries, **{n_changed}** {ifelse(n_changed ==1, 'has', 'have')} changes in clustering.<br>
+    Among the libraries with changes, the mean change in the number of clusters is **{mean_n_change}**, which corresons  a mean percentage change of **{round(mean_n_change_frac * 100, digits = 2)}%**.
```
allyhawkins, line 877:
```suggestion
    Among the libraries with changes, the mean change in the number of clusters is **{mean_n_change}**, which corresonds to a mean percentage change of **{round(mean_n_change_frac * 100, digits = 2)}%**.
```

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -606,23 +688,318 @@ report_table(
...
+```
+
+```{r}
+# cluster change overall summary
+cluster_summary <- cluster_stats |>
+  dplyr::summarise(
+    n = dplyr::n(),
+    n_changed = n - sum(cluster_setequal),
+    mean_n_change = mean(abs(cluster_n_change[which(!cluster_setequal)])),
+    mean_n_change_frac = mean(abs(cluster_n_change_frac[which(!cluster_setequal)])),
```
allyhawkins, line 865: I think I need some clarity on what % change and mean % change is here? Is it the number of clusters that have different sizes? Or is it how different are the clusters with different sizes?

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -606,23 +688,318 @@ report_table(
...
+  theme(legend.position = "none")
+```
+
+
+```{r, eval=has_cluster_changes}
+knitr::asis_output("#### Detail")
+```
+
+```{r, eval=has_cluster_changes}
+# Detail table of cluster changes
```
allyhawkins, line 961: It's confusing to me how libraries that have 0 clusters changed are showing up in this table? I would expect only those with a different number of clusters changed?

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -606,23 +688,318 @@ report_table(
...
+  theme_bw()
+
+suppressWarnings(print(cluster_plot))
+```
+
+```{r}
+# plot top changed clusters
+plot_df <- cluster_changes |>
+  dplyr::arrange(desc(cluster_max_ari)) |>
+  head(6) |> # only plot the top 6
```
allyhawkins, line 920: I wonder here if it actually makes sense to show all the libraries that have changes in the histogram but only the top ones in the stacked bar chart? I feel like this could change slightly across the board for a lot of libraries, so it would be helpful to see if there are libraries that are outliers or not.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -606,23 +688,318 @@ report_table(
...
+    mean_jaccard = mean(hvg_jaccard[which(!hvg_identical)]),
+    mean_cor = mean(hvg_rank_cor[which(!hvg_identical)])
+  )
+
+knitr::asis_output(glue::glue_data(
+  hvg_summary,
+  "Of the **{n}** libraries, **{n_changed}** {ifelse(n_changed == 1, 'has', 'have')} changes in highly variable genes.
+
+
+  Among the libraries with changes, the mean number of genes changed is **{mean_change}**, which is a mean percentage change of **{round(mean_change_frac * 100, digits = 2)}%**.
```
allyhawkins, line 749: I think we might want to update this text (and maybe the table to?) to be explicit about which libraries have changes in the genes themselves vs. which libraries only have changes in the order.

allyhawkins (commented):
I think this is generally looking good considering we don't know exactly what the real data will look like. I like the choices that you made regarding HVGs and clustering, but generally the clustering functions could probably use some more comments and description about exactly what's going on.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: I do; that is what I am sorting by to decide which samples to pick first.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: Yes, I will add some more comments.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: This is kind of present in the table; if the Jaccard index is 1, then the gene set is unchanged and only the order has changed. I kind of doubt this will ever happen in practice; I expect the bottom end of the list to change with some frequency. One thing I had thought about (and never implemented, as I'm only remembering the passing thought now) was calculating these stats for subsets of the HVGs; I can imagine that the top 500 or 1000 genes might be more stable, which might be nice to see at a glance.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -606,23 +688,318 @@ report_table(
...
+) |>
+  DT::formatPercentage(c("hvg_change_frac"), 2)
+```
 
-### Clustering
+### Clustering {.tabset}
 
 Report changes in the number of clusters, as well as changes in the sizes of the assigned clusters.
 We will ignore changes in the ordering of clusters by sorting by cluster size before comparisons.
 As a summary statistic, we might use adjusted Rand index (ARI) or adjusted mutual information (AMI) to compare the two sets of clusters, reporting when we are well below 1.
```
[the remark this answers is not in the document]
jashapiro: I sure meant to!

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: This is the change in the number of clusters, relative to the starting number of clusters. I definitely need to clarify!

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: Can you clarify what you mean here? The current histograms are all changes, and the stacked plots are only the ones with the highest max ARI. There definitely needs to be some text explaining that, but I was wondering if you meant that we should _not_ do that?

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: This is a labeling problem. It should say "Change in number of clusters". Or "change in cluster number" or something like that.

jashapiro (commented):
Thanks for your review @allyhawkins! You definitely got a lot of places where my brain had been too deep in the code to actually see what was clear and what was very much not. I'll work on some of the more clear changes now, but I did have a few questions for you if you have a chance before I finish up.

scripts/compare-metrics/compare-metrics-template.rmd
allyhawkins: Wow I'm blind and clearly didn't see the next line.

scripts/compare-metrics/compare-metrics-template.rmd
allyhawkins: > This is kind of present in the table; if the Jaccard index is 1, then the gene set is unchanged and only the order has changed. I kind of doubt this will ever happen in practice; I expect the bottom end of the list to change with some frequency. One thing I had thought about (and never implemented, as I'm only remembering the passing thought now) was calculating these stats for subsets of the HVGs; I can imagine that the top 500 or 1000 genes might be more stable, which might be nice to see at a glance.

Maybe we should add a note that the table shows only those that have changes, either in the gene content or the order of the genes as a reminder. Unless we have that stated somewhere already?

scripts/compare-metrics/compare-metrics-template.rmd
allyhawkins: I think I was confused because the histograms only show 3 libraries for the `cluster_max_ari`. Is it just that the number of libraries with changes is only 6 here anyways? Or did you accidentally filter them out somewhere?

allyhawkins: I see that there are only 3 for the ARI because some of them have NA, do we expect that?

jashapiro: Yeah, there are only 6 libraries with changes. And yes, we expect ARI to be `NA` for some libraries, because we can't calculate ARI if the number of cells has changed. Well, I'm sure I could do something but I didn't have any good ideas.

allyhawkins: > And yes, we expect ARI to be NA for some libraries, because we can't calculate ARI if the number of cells has changed. 

Also worth a note in the text!

allyhawkins: Alternatively, if the number of cells has changed then the clustering should have definitely changed so we could remove those as we do with the reads table?

jashapiro: I think we want to keep something there, because if we have a small change in the number of cells we would still want to know if the clustering was affected. This makes me think I should still try to do something if the cell count isn't equal. Maybe scaling down the counts somehow...

jashapiro:
I updated a the text and a number of plots to try to make things more clear throughout. Part of that was switching to patchwork rather than faceting when plotting metrics that are not on the same scale, which allows better axis labeling. 

I updated comments in the pairing function, and in [f8e7822](https://github.com/AlexsLemonade/scpca-nf/pull/871/commits/f8e7822f5416df6f8756ba97639acd9eedf40b91) I added rescaling the cluster sizes down to the smaller set of cells so we can always calculate a maxARI. I think that also helps avoid some of the confusion you saw. In general I think this will work out, if there are cases where filtering changes by a few cells here and there. On the other hand, I can imagine a case where changing filtering could eliminate a class of cells, which this might over-highlight, but since this is really only designed as a first pass screen for major changes, I think that will be okay. Investigating individual changes will always have to be part of the followup. 

Updated file below:
[compare-metrics.html.zip](https://github.com/user-attachments/files/20232535/compare-metrics.html.zip)

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -623,6 +1037,22 @@ This will include SingleR, CellAssign and consensus cell types, if present.
...
+
+celltypes_changes <- function(ref_celltypes, comp_celltypes) {
+  type_matrix <- dplyr::bind_rows(a, b) |> as.matrix()
+  rownames(type_matrix) <- c("ref", "comp")
+  type_matrix[is.na(type_matrix)] <- 0
+
+  added <- colnames(type_matrix)[which(type_matrix["ref"] == 0 && type_matrix["comp"] > 0)]
+  removed <- colnames(type_matrix)[which(type_matrix["comp"] == 0 && type_matrix["ref"] > 0)]
+  n_changes <- dist(type_matrix, method = "manhattan")[1]
+  euclidean_distance <- dist(type_matrix, method = "euclidean")[1]
```
allyhawkins, line 1051: I don't know if you meant to add this in here? I think I would remove it.

allyhawkins (approved):
This all looks good to me and as you said, its a starting point to then investigate further. I think there's just one chunk of code that shouldn't be here but other than that this looks good.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: I kind of did... I was originally doing all three of these parts together, but then decided it was better to finish the first two. But I'll remove it before merging you you can look at it fresh in the next PR

PR #877, opened 2025-05-19
Account for 10X flex example data in metadata

### LINKED ISSUE #872: Obtain example data and example metadata files for 10X flex data
Before we can get started on #870, we will need to obtain some example data and create example versions of the sample and library metadata files. We should probably obtain example data for both single and multiplexed since we will need to support both. I think we can use the following example datasets: 

- [Singleplexed example data](https://10x.vercel.app/datasets/Human_Kidney_4k_GEM-X_Flex)
- [Multiplexed example data](https://10x.vercel.app/datasets/80k_Human_PBMCs_PTG_MultiproPanel_IC_4plex) 

The FASTQ files should be saved to S3 in the `s3://scpca-references/example-data/example_fastqs` folder and organized by run ID. 

We will then want to update the `example_run_metadata.tsv` and `example_sample_metadata.tsv` to add in these samples using `10X_flex_single` and `10X_flex_multi` as the `technology` (exact tech names TBD, but I think I like these options). 

I think we could also then update the `example_multiplex_pools.tsv` based on any multiplexing barcodes. 

Once we do that, we should be able to proceed with development and use the `example`  profile for testing and development of this new module.

allyhawkins:
Closes #872 

Here I'm adding example 10X flex data to the folder of example FASTQ files and accounting for these two new libraries in the example metadata files. 

- I downloaded the input FASTQ files for a [singleplexed example](https://10x.vercel.app/datasets/Human_Kidney_4k_GEM-X_Flex) and [multiplexed example](https://10x.vercel.app/datasets/80k_Human_PBMCs_PTG_MultiproPanel_IC_4plex) and copied the FASTQ files to `s3://scpca-references/example-data/example_fastqs`. 
  - These are both examples that use the v1.1.0 probe set for 10X flex. 
  - This was the smallest multiplexed example that I could find that also uses the most recent probe set. It does have feature barcoding too, but that information is in separate FASTQ files. So I just copied over the gene expression FASTQ files and treated that sample as if we don't have any antibody information. 
- I accounted for these samples in the example run, sample, and multiplex metadata files. 
  - For technology, I am using `10x_flex_single` and `10x_flex_multi`. Are we okay with these names or would we prefer something else? 
  - I just filled in the sample metadata with dummy data as we do for the other example data. 
  - I filled in the `barcode_id` for the multiplex pools file with the probe IDs specified in the [experimental design for the multiplexed example](https://cf.10xgenomics.com/samples/cell-exp/9.0.0/80k_Human_PBMCs_PTG_MultiproPanel_IC_4plex_PBMC_BC1_AB1_Rested_Unstained/80k_Human_PBMCs_PTG_MultiproPanel_IC_4plex_PBMC_BC1_AB1_Rested_Unstained_web_summary.html). 

With these updates, we should now be able to run the workflow with `-profile example --run_ids run08,run09` to test the 10X flex set up.

### DIFF
--- a/examples/example_run_metadata.tsv
+++ b/examples/example_run_metadata.tsv
@@ -6,5 +6,5 @@ run04	library03	sample03;sample04	project01	10Xv3.1	EFO:0009922	cell	Mus_musculu
 run05	library03	sample03;sample04	project01	cellhash_10Xv3.1	EFO:0030012	cell	Mus_musculus.GRCm39.104	example_fastqs/run05	example_barcode_files/cellhash_barcodes.tsv	2[1-10]	NA	NA	NA
 run06	library04	sample05	project01	paired_end	EFO:0003747	bulk	Mus_musculus.GRCm39.104	example_fastqs/run06	NA	NA	NA	NA	NA
 run07	library05	sample06	project01	10Xv3.1	EFO:0009922	cell	Homo_sapiens.GRCh38.104	example_fastqs/run07	NA	NA	NA	NA	example_metadata_files/submitter_celltypes.tsv
-run08	library06	sample07	project02	10X_flex_single	EFO:0022606	nucleus	Homo_sapiens.GRCh38.110	example_fastqs/run08	NA	NA	NA	NA	NA
-run09	library07	sample08;sample09;sample10;sample11	project02	10X_flex_multi	EFO:0022606	cell	Homo_sapiens.GRCh38.110	example_fastqs/run09	NA	NA	NA	NA	NA
+run08	library06	sample07	project02	10Xflex_v1.1_single	EFO:0022606	nucleus	Homo_sapiens.GRCh38.110	example_fastqs/run08	NA	NA	NA	NA	NA
+run09	library07	sample08;sample09;sample10;sample11	project02	10Xflex_v1.1_multi	EFO:0022606	cell	Homo_sapiens.GRCh38.110	example_fastqs/run09	NA	NA	NA	NA	NA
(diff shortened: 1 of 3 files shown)


examples/example_run_metadata.tsv
```diff
@@ -6,3 +6,5 @@ run04	library03	sample03;sample04	project01	10Xv3.1	EFO:0009922	cell	Mus_musculu
 run05	library03	sample03;sample04	project01	cellhash_10Xv3.1	EFO:0030012	cell	Mus_musculus.GRCm39.104	example_fastqs/run05	example_barcode_files/cellhash_barcodes.tsv	2[1-10]	NA	NA	NA
 run06	library04	sample05	project01	paired_end	EFO:0003747	bulk	Mus_musculus.GRCm39.104	example_fastqs/run06	NA	NA	NA	NA	NA
 run07	library05	sample06	project01	10Xv3.1	EFO:0009922	cell	Homo_sapiens.GRCh38.104	example_fastqs/run07	NA	NA	NA	NA	example_metadata_files/submitter_celltypes.tsv
+run08	library06	sample07	project02	10X_flex_single	EFO:0022606	nucleus	Homo_sapiens.GRCh38.110	example_fastqs/run08	NA	NA	NA	NA	NA
```
jashapiro, line 9: I think it is probably worth encoding the version here too. Maybe `10Xflex_v1.1_single`?

jashapiro (commented):
This looks fine, but I think we do want the version number in the technology column?

examples/example_run_metadata.tsv
allyhawkins: > I think it is probably worth encoding the version here too. Maybe `10Xflex_v1.1_single`?

Ah yes! I had meant to add this in, so thank you. Versions are added in https://github.com/AlexsLemonade/scpca-nf/pull/877/commits/a27ae6e89800f93c2ced04fdf6498079a9ea0b5e.

allyhawkins:
> I did have a question though which came up in some separate testing... do we have versions of the example_metadata_files and example_barcode_files floating around somewhere that we use for processing the example data? The docs say you have to provide your own, but I would assume we have a version we use somewhere?

No we don't. We only describe those in the docs and link to those sections in the example directory. 

Just as a side note, the only run besides these two that can actually be used in real life for processing is `run01`. We don't have the FASTQs or any other data available for the other samples. Mainly because we instruct people to download their own example data rather than running the workflow with the data that's on S3. We could update the example directory as a whole to document being able to use the example profile to process example data and ensure that any samples in the example metadata file actually have corresponding data, but that is something we should do separately from the 10X flex support if we want to do it.

jashapiro:
> > I did have a question though which came up in some separate testing... do we have versions of the example_metadata_files and example_barcode_files floating around somewhere that we use for processing the example data? The docs say you have to provide your own, but I would assume we have a version we use somewhere?
> 
> No we don't. We only describe those in the docs and link to those sections in the example directory.
> 
> Just as a side note, the only run besides these two that can actually be used in real life for processing is `run01`. We don't have the FASTQs or any other data available for the other samples. Mainly because we instruct people to download their own example data rather than running the workflow with the data that's on S3. We could update the example directory as a whole to document being able to use the example profile to process example data and ensure that any samples in the example metadata file actually have corresponding data, but that is something we should do separately from the 10X flex support if we want to do it.

That makes some sense. I realized that was probably the case when I looked at the example profile and saw that it had that run specified. I was mostly confused I think by `run01` having a CITE-seq run associated with it. It might make sense to have that be a separate sample.

PR #881, opened 2025-05-22
Add probe set for 10X flex v1.1.0

### LINKED ISSUE #875: Account for probe set reference CSV
`cellranger multi` requires a probe set CSV file as input alongside the optional reference transcriptome (see #874). This issue should be used to track both adding that reference CSV to S3 and accounting for accessing it in the workflow. I think we should approach this as follows: 

- Add the appropriate [probe set](https://10x.vercel.app/support/flex-gene-expression/documentation/steps/probe-sets/chromium-frp-probe-set-files#probe_set_csv_file) to `s3://scpca-references`. I believe this should be v1.1.0. 
- Ensure that the version for the probe set is accounted for in the `technology` column, e.g., `10X_flex_v1.1.0_single` and `10X_flex_v1.1.0_pooled`. 
- Assign the path to the probe set based on the version in the technology string. Similar to how we handle defining the paths to 10X barcode lists. 
  - Alternatively, we could add an entry to the [scpca-refs.json](https://github.com/AlexsLemonade/scpca-nf/blob/main/references/scpca-refs.json) for `probe_set`. This would only be included for Ensembl versions that are compatible with Cell Ranger (e.g., 110 which is being added in #874) and for others would be set to NA. This would make sure we don't ever try to run a version of the probe set that is incompatible with the transcriptome reference. Even if we do this, I think we should still track the probe set version in the `technology` column.

allyhawkins:
Closes #875 

Here I am adding the information about the 10X flex probe set to our reference metadata and json files and reading in the path to that file in the main workflow. Because probe sets are attached to transcriptome reference versions, I think it makes sense to keep the probe and transcriptome versions and information together, see the screenshot from [10x FAQs below](https://10x.vercel.app/support/flex-gene-expression/documentation/steps/probe-sets/chromium-frp-probe-sets-overview): 
Screenshot 2025-05-22 at 2 15 57 PM

So to do that I made the following adjustments: 

- Added the probe CSV file for humans to `s3://scpca-references/homo_sapiens/ensembl-110/10X_flex_probes`. 
- Added a column to `refs-metadata.tsv` to specify the file name for that probe CSV. If there is no probe file associated with the Ensembl version or reference then it just gets `NA`. 
- Adjusted the script to generate the `scpca-refs.json` file to add an entry for the path to the 10X flex probe set file. Again, this will be null if the filename isn't provided. 
- Updated the mapping of metadata in `main.nf` to account for some times the ref paths being `null` for a specific organism. 
- Added the path to the probe file to the metadata map. 
- Included some documentation on how to add probe sets for using 10X flex. 

The caveat with this approach is that it relies on there being only one probe set version per reference transcriptome. I am guessing that will be the case for the foreseeable future, so maybe we don't need to worry about it. The other option is that we could create a different metadata file or some sort of map in the workflow that maps the ensemble version to the probe set files, but I felt like that was less robust? 

I also don't actually track the version of the probe set with the probe set file, but it is in the file name. We might consider adding a column that just has the version and keeping that in the ref json too? I think if we keep this approach we probably want to do that.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -29,7 +29,6 @@ all_techs = single_cell_techs + bulk_techs + spatial_techs
 rna_techs = single_cell_techs.findAll{it.startsWith('10Xv')}
 citeseq_techs = single_cell_techs.findAll{it.startsWith('CITEseq')}
 cellhash_techs = single_cell_techs.findAll{it.startsWith('cellhash')}
-flex_techs = single_cell_techs.findAll{it.startsWith('10Xflex')}
 
 // report template paths
 report_template_dir = file("${projectDir}/templates/qc_report", type: 'dir', checkIfExists: true)
--- a/main.nf
+++ b/main.nf
@@ -15,6 +15,11 @@ cell_barcodes = [
   'cellhash_10Xv3.1': '3M-february-2018.txt'
 ]
 
+// 10X flex probe set files
+probe_set_files = [
+  'v1.1.0' = 'Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv'
+]
+
 // supported technologies
 single_cell_techs = cell_barcodes.keySet()
 bulk_techs = ['single_end', 'paired_end']
@@ -23,6 +28,7 @@ all_techs = single_cell_techs + bulk_techs + spatial_techs
 rna_techs = single_cell_techs.findAll{it.startsWith('10Xv')}
 citeseq_techs = single_cell_techs.findAll{it.startsWith('CITEseq')}
 cellhash_techs = single_cell_techs.findAll{it.startsWith('cellhash')}
+flex_techs = single_cell_techs.findAll{it.startsWith('10Xflex')}
 
 // report template paths
 report_template_dir = file("${projectDir}/templates/qc_report", type: 'dir', checkIfExists: true)
@@ -141,18 +147,17 @@ workflow {
         slide_serial_number: Utils.parseNA(it.slide_serial_number),
         slide_section: Utils.parseNA(it.slide_section),
         ref_assembly: it.sample_reference,
-        ref_fasta: params.ref_rootdir + "/" + sample_refs["ref_fasta"],
-        ref_fasta_index: params.ref_rootdir + "/" + sample_refs["ref_fasta_index"],
-        ref_gtf: params.ref_rootdir + "/" + sample_refs["ref_gtf"],
-        // need to account for the refs sometimes being null 
-        salmon_splici_index: sample_refs?.splici_index ? "${params.ref_rootdir}/${sample_refs.splici_index}" : null,
-        t2g_3col_path: sample_refs?.t2g_3col_path ? "${params.ref_rootdir}/${sample_refs.t2g_3col_path}" : null,
-        mito_file: sample_refs?.mito_file ? "${params.ref_rootdir}/${sample_refs.mito_file}" : null,
-        salmon_bulk_index: sample_refs?.salmon_bulk_index ? "${params.ref_rootdir}/${sample_refs.salmon_bulk_index}" : null,
-        t2g_bulk_path: sample_refs?.t2g_bulk_path ? "${params.ref_rootdir}/${sample_refs.t2g_bulk_path}" : null,
-        cellranger_index: sample_refs?.cellranger_index ? "${params.ref_rootdir}/${sample_refs.cellranger_index}" : null,
-        star_index: sample_refs?.star_index ? "${params.ref_rootdir}/${sample_refs.star_index}" : null,
-        flex_probe_ref: sample_refs?.flex_probe_ref ? "${params.ref_rootdir}/${sample_refs.flex_probe_ref}" : null,
+        ref_fasta: "${params.ref_rootdir}/${sample_refs.ref_fasta}",
+        ref_fasta_index: "${params.ref_rootdir}/${sample_refs.ref_fasta_index}",
+        ref_gtf: "${params.ref_rootdir}/${sample_refs.ref_gtf}",
+        mito_file: "${params.ref_rootdir}/${sample_refs.mito_file}",
+        // account for the refs sometimes being null 
+        salmon_splici_index: sample_refs.splici_index ? "${params.ref_rootdir}/${sample_refs.splici_index}" : null,
+        t2g_3col_path: sample_refs.t2g_3col_path ? "${params.ref_rootdir}/${sample_refs.t2g_3col_path}" : null,
+        salmon_bulk_index: sample_refs.salmon_bulk_index ? "${params.ref_rootdir}/${sample_refs.salmon_bulk_index}" : null,
+        t2g_bulk_path: sample_refs.t2g_bulk_path ? "${params.ref_rootdir}/${sample_refs.t2g_bulk_path}" : null,
+        cellranger_index: sample_refs.cellranger_index ? "${params.ref_rootdir}/${sample_refs.cellranger_index}" : null,
+        star_index: sample_refs.star_index ? "${params.ref_rootdir}/${sample_refs.star_index}" : null,
         scpca_version: workflow.revision ?: workflow.manifest.version,
         nextflow_version: nextflow.version.toString()
       ]
--- a/main.nf
+++ b/main.nf
@@ -144,13 +144,15 @@ workflow {
         ref_fasta: params.ref_rootdir + "/" + sample_refs["ref_fasta"],
         ref_fasta_index: params.ref_rootdir + "/" + sample_refs["ref_fasta_index"],
         ref_gtf: params.ref_rootdir + "/" + sample_refs["ref_gtf"],
-        salmon_splici_index: params.ref_rootdir + "/" + sample_refs["splici_index"],
-        t2g_3col_path: params.ref_rootdir + "/" + sample_refs["t2g_3col_path"],
-        mito_file: params.ref_rootdir + "/" + sample_refs["mito_file"],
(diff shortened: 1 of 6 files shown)


main.nf
```diff
@@ -144,13 +144,15 @@ workflow {
...
-        star_index: params.ref_rootdir + "/" + sample_refs["star_index"],
+        // need to account for the refs sometimes being null 
+        salmon_splici_index: sample_refs?.splici_index ? "${params.ref_rootdir}/${sample_refs.splici_index}" : null,
+        t2g_3col_path: sample_refs?.t2g_3col_path ? "${params.ref_rootdir}/${sample_refs.t2g_3col_path}" : null,
+        mito_file: sample_refs?.mito_file ? "${params.ref_rootdir}/${sample_refs.mito_file}" : null,
+        salmon_bulk_index: sample_refs?.salmon_bulk_index ? "${params.ref_rootdir}/${sample_refs.salmon_bulk_index}" : null,
+        t2g_bulk_path: sample_refs?.t2g_bulk_path ? "${params.ref_rootdir}/${sample_refs.t2g_bulk_path}" : null,
+        cellranger_index: sample_refs?.cellranger_index ? "${params.ref_rootdir}/${sample_refs.cellranger_index}" : null,
+        star_index: sample_refs?.star_index ? "${params.ref_rootdir}/${sample_refs.star_index}" : null,
+        flex_probe_ref: sample_refs?.flex_probe_ref ? "${params.ref_rootdir}/${sample_refs.flex_probe_ref}" : null,
```
jashapiro, line 155: I think you can simplify this just a touch by removing the first `?` since that would be testing whether `sample_refs` exists, not the subfield. So just:
```
salmon_splici_index: sample_refs.splici_index ? "${params.ref_rootdir}/${sample_refs.splici_index}" : null,
```

But while you are editing all of these, can you make the three lines before more parallel? In other words change 
```
params.ref_rootdir + "/" + sample_refs["ref_gtf"]
```
to 
```
"${params.ref_rootdir}/${sample_refs.ref_gtf}"
```

I don't like the syntax changing midway.

Oh, and is the `mito_file` really optional? I don't think it should be, as that does not depend on mapping technology.

allyhawkins:
> > The caveat with this approach is that it relies on there being only one probe set version per reference transcriptome. I am guessing that will be the case for the foreseeable future, so maybe we don't need to worry about it. The other option is that we could create a different metadata file or some sort of map in the workflow that maps the ensemble version to the probe set files, but I felt like that was less robust?
> > I also don't actually track the version of the probe set with the probe set file, but it is in the file name. We might consider adding a column that just has the version and keeping that in the ref json too? I think if we keep this approach we probably want to do that.
> 
> Thinking about this a bit more, I am not sure this is the approach we want to take. I think we want to tie the probe reference to the `technology` column, more as we do for the barcode file. I think there is a good change that they will release a probe set 1.1.1, as they did with 1.0.1, which would mean there would be more than one probe set per reference. So I think we do need to have a map from technology to probe versions and the associated references. We can fill that in with the three current possible probe sets for human.

Okay yea, you're right. I think I was trying to make this too complicated. I reverted some of the changes to the ref metadata and json files and instead added a param to specify the path to the probe refs, as we do with barcode files. Then in `main.nf`, I just added a map with the probe versions and file names. We can then access it the same way we do with barcodes in the process that runs `cellranger multi`. 

I also moved the probe set file to `s3://scpca-references/flex-probe-refs`.

main.nf
```diff
@@ -15,6 +15,11 @@ cell_barcodes = [
   'cellhash_10Xv3.1': '3M-february-2018.txt'
 ]
 
+// 10X flex probe set files
+probe_set_files = [
+  'v1.1.0' = 'Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv'
+]
```
jashapiro, line 21: I think you want these named to match the technology column?And since "probe set" is a bit ambiguous, lets be more specific with the name, perhaps

```suggestion
// 10X flex probe set files
flex_probesets = [
  '10Xflex_v1.1_single': 'Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv',
  '10Xflex_v1.1_multi': 'Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv'
]
```

main.nf
```diff
@@ -23,6 +28,7 @@ all_techs = single_cell_techs + bulk_techs + spatial_techs
 rna_techs = single_cell_techs.findAll{it.startsWith('10Xv')}
 citeseq_techs = single_cell_techs.findAll{it.startsWith('CITEseq')}
 cellhash_techs = single_cell_techs.findAll{it.startsWith('cellhash')}
+flex_techs = single_cell_techs.findAll{it.startsWith('10Xflex')}
```
jashapiro, line 31: I think you want to get this list from the defined dictionary above.
```suggestion
flex_techs = flex_probesets.keySet()
```

But then you probably want to add that to the `all_techs` list and maybe to the `single_cell_techs` as well. 

So maybe you could just leave this and instead set
```
single_cell_techs = cell_barcodes.keySet() + flex_probesets.keySet()
```

I think we need to know what the next step is going to be before deciding whether you want flex to be considered part of `single_cell_techs` or just its own separate group.

jashapiro (commented):
Closer, but I think this would run into some errors with processing down the line. There is some chance it would make sense to just make the flex files part of the `barcodes/10X`directory and `cell_barcodes` array.

One thing to note is that we include a license with the barcodes directory; do we need to do that with these probes as well?

main.nf
allyhawkins: Sorry, I was just playing around with this and didn't mean to commit it. This will be part of the next PR!

allyhawkins:
> One thing to note is that we include a license with the barcodes directory; do we need to do that with these probes as well?

I don't think so... or at least I can't find anything that states that.

allyhawkins:
> There is some chance it would make sense to just make the flex files part of the barcodes/10Xdirectory and cell_barcodes array.

I'm iffy on this. I feel like it's not really a cell barcode file, so it makes sense to keep it separate to me. But I don't have super strong feelings about it so if you want me to add it to the cell barcodes list I can.

jashapiro:
> > There is some chance it would make sense to just make the flex files part of the barcodes/10Xdirectory and cell_barcodes array.
> 
> I'm iffy on this. I feel like it's not really a cell barcode file, so it makes sense to keep it separate to me. But I don't have super strong feelings about it so if you want me to add it to the cell barcodes list I can.

I think it is fine not to. Really kind of depends on how many things we want to keep track of and how.

main.nf
```diff
@@ -141,16 +147,17 @@ workflow {
...
+        ref_fasta_index: "${params.ref_rootdir}/${sample_refs.ref_fasta_index}",
+        ref_gtf: "${params.ref_rootdir}/${sample_refs.ref_gtf}",
+        mito_file: "${params.ref_rootdir}/${sample_refs.mito_file}",
+        // account for the refs sometimes being null 
+        salmon_splici_index: sample_refs.splici_index ? "${params.ref_rootdir}/${sample_refs.splici_index}" : null,
+        t2g_3col_path: sample_refs.t2g_3col_path ? "${params.ref_rootdir}/${sample_refs.t2g_3col_path}" : null,
+        salmon_bulk_index: sample_refs.salmon_bulk_index ? "${params.ref_rootdir}/${sample_refs.salmon_bulk_index}" : null,
+        t2g_bulk_path: sample_refs.t2g_bulk_path ? "${params.ref_rootdir}/${sample_refs.t2g_bulk_path}" : null,
+        cellranger_index: sample_refs.cellranger_index ? "${params.ref_rootdir}/${sample_refs.cellranger_index}" : null,
+        star_index: sample_refs.star_index ? "${params.ref_rootdir}/${sample_refs.star_index}" : null,
```
jashapiro, line 160: I was looking at the broader code here, and I see that for some reason I have not been using `null` for missing values, but rather empty strings (that is what `Utils.parseNA()` does anyway). I think that may well be what we want to do for consistency here too. 

In most practice, it won't make much difference, as either is "falsey", but null values can be a bit more of a pain to deal with, and it seems good to keep things as consistent as possible.

I'm less sure about the `scpca-refs.json` file, but I kind of think we might want to do the same there too, again just for consistency.
```suggestion
        salmon_splici_index: sample_refs.splici_index ? "${params.ref_rootdir}/${sample_refs.splici_index}" : '',
        t2g_3col_path: sample_refs.t2g_3col_path ? "${params.ref_rootdir}/${sample_refs.t2g_3col_path}" : '',
        salmon_bulk_index: sample_refs.salmon_bulk_index ? "${params.ref_rootdir}/${sample_refs.salmon_bulk_index}" : '',
        t2g_bulk_path: sample_refs.t2g_bulk_path ? "${params.ref_rootdir}/${sample_refs.t2g_bulk_path}" : '',
        cellranger_index: sample_refs.cellranger_index ? "${params.ref_rootdir}/${sample_refs.cellranger_index}" : '',
        star_index: sample_refs.star_index ? "${params.ref_rootdir}/${sample_refs.star_index}" : '',
```

allyhawkins: > I'm less sure about the scpca-refs.json file, but I kind of think we might want to do the same there too, again just for consistency.

Do you mean we should just use empty strings in the json file? I had tried to get it to write out NA when making the JSON but was unsuccessful, which is why I had left the null values. I can update to use empty strings though or if you know how to make it keep the NA?

jashapiro (approved):
This looks good to me. 

I think I would change the NA values to empty strings in the JSON too. Nulls just too often end up causing trouble. But it does not need to be done right now.

PR #882, opened 2025-05-23
Use empty strings instead of null in refs json

allyhawkins:
Based on https://github.com/AlexsLemonade/scpca-nf/pull/881#pullrequestreview-2865394976, I went ahead and updated the json file and the script to make the json to use empty strings instead of Null.

### DIFF
--- a/references/scripts/create-reference-json.R
+++ b/references/scripts/create-reference-json.R
@@ -64,7 +64,7 @@ create_ref_entry <- function(
       annotation_dir,
       glue::glue("{reference_name}.mitogenes.txt")
     ),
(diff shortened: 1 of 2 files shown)


references/scripts/create-reference-json.R
```diff
@@ -65,12 +65,12 @@ create_ref_entry <- function(
       glue::glue("{reference_name}.mitogenes.txt")
     ),
     # fill in optional entries with NA
```
jashapiro, line 67:
```suggestion
    # fill in optional entries with empty strings
```

jashapiro (approved):
Just a comment update

PR #889, opened 2025-06-11
Process for running cellranger multi with multiplexed samples

### LINKED ISSUE #876: Initiate module for running `cellranger multi`

allyhawkins:
Closes #876 

This PR adds a new process for running `cellranger multi` with a library that contains multiplexed samples to the flex workflow. The main difference here is that we need to add a `[samples]` section to the config file that contains the sample Id and the probe barcode ID for each sample. Everything else is the same between single and multiplexed runs. So I adjusted the config creation script to read in an optional multiplexed pools file and library ID arguments. If provided, then a `[samples]` section is created. One thing to note is that the first version of this used `pandas`, but our Docker image doesn't have `pandas` sadly so I converted to using `csv` in python. I think that's fine for this purpose. 

Within Nextflow, I added a process for the multiplexed samples that I pass the multiplex pools file to. As it is right now, I'm passing the entire pools file without any prior filtering to the python script and then doing the filtering within the Python script. I probably could do some filtering in Nextflow, but it's a small TSV file, so I'm not too concerned about passing the whole file to the process every time. Let me know if you prefer a different approach. 

I made a few other smaller changes while I was here: 

- I updated how we handle the output directories to follow the same format we do with spaceranger where we have output organized as: `checkpoints/cellranger-multi/libraryid/runid-cellranger-multi`. This meant defining both the publish directory and the results directory, again following the same approach we use when running spaceranger. 
- In looking at the outputs, the `SC_MULTI_CS` folder, which contains a bunch of intermediate files, was taking up a lot of space, so I decided to remove it. For the multiplexed run, it was ~80 GB out of the total ~92 GB for the output folder. So now the size of that folder is much more reasonable. 
- I fixed a typo where we were missing a `\` in the singleplexed process.

### DIFF
--- a/modules/cellranger-flex.nf
+++ b/modules/cellranger-flex.nf
@@ -27,7 +27,7 @@ process cellranger_flex_single {
     # run cellranger multi
     cellranger multi \
       --id=${out_id} \
-      --csv=flex_config.csv
+      --csv=flex_config.csv \
       --localcores=${task.cpus} \
       --localmem=${task.memory.toGiga()}
 
--- a/modules/cellranger-flex.nf
+++ b/modules/cellranger-flex.nf
@@ -12,7 +12,7 @@ process cellranger_flex_single {
   output:
     tuple val(meta), path(out_id)
   script:
-    out_id = file(meta.cellranger_multi_publish_dir).name
+    out_id = file(meta.cellranger_multi_results_dir).name
     meta += Utils.getVersions(workflow, nextflow)
     meta_json = Utils.makeJson(meta)
     """
@@ -33,9 +33,12 @@ process cellranger_flex_single {
 
     # write metadata
     echo '${meta_json}' > ${out_id}/scpca-meta.json
+
+    # remove Cell Ranger intermediates directory
+    rm -rf ${out_id}/SC_MULTI_CS
     """
   stub:
-    out_id = file(meta.cellranger_multi_publish_dir).name
+    out_id = file(meta.cellranger_multi_results_dir).name
     meta += Utils.getVersions(workflow, nextflow)
     meta_json = Utils.makeJson(meta)
     """
@@ -56,7 +59,7 @@ process cellranger_flex_multi {
   output:
     tuple val(meta), path(out_id)
   script:
-    out_id = file(meta.cellranger_multi_publish_dir).name
+    out_id = file(meta.cellranger_multi_results_dir).name
     meta += Utils.getVersions(workflow, nextflow)
     meta_json = Utils.makeJson(meta)
     """
@@ -82,9 +85,12 @@ process cellranger_flex_multi {
 
     # write metadata
     echo '${meta_json}' > ${out_id}/scpca-meta.json
+
+    # remove Cell Ranger intermediates directory
+    rm -rf ${out_id}/SC_MULTI_CS
     """
   stub:
-    out_id = file(meta.cellranger_multi_publish_dir).name
+    out_id = file(meta.cellranger_multi_results_dir).name
     meta += Utils.getVersions(workflow, nextflow)
     meta_json = Utils.makeJson(meta)
     """
@@ -107,6 +113,7 @@ workflow flex_quant{
       .map{
         def meta = it.clone();
         meta.cellranger_multi_publish_dir =  "${params.checkpoints_dir}/cellranger-multi/${meta.library_id}";
+        meta.cellranger_multi_results_dir = "${meta.cellranger_multi_publish_dir}/${meta.run_id}-cellranger-multi";
         meta.flex_probeset = "${params.probes_dir}/${flex_probesets[meta.technology]}";
         meta // return modified meta object
       }
@@ -115,11 +122,11 @@ workflow flex_quant{
           // input files exist
           it.files_directory && file(it.files_directory, type: 'dir').exists() && (
             params.repeat_mapping
-            || !file(it.cellranger_multi_publish_dir).exists()
-            || Utils.getMetaVal(file("${it.cellranger_multi_publish_dir}/scpca-meta.json"), "ref_assembly") != "${it.ref_assembly}"
+            || !file(it.cellranger_multi_results_dir).exists()
+            || Utils.getMetaVal(file("${it.cellranger_multi_results_dir}/scpca-meta.json"), "ref_assembly") != "${it.ref_assembly}"
           )
         )
-        has_cellranger_flex: file(it.cellranger_multi_publish_dir).exists()
+        has_cellranger_flex: file(it.cellranger_multi_results_dir).exists()
         missing_inputs: true
       }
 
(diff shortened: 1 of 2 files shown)


allyhawkins:
Okay this should now be ready for re-review. In testing this, I realized that our example sample IDs don't have SCPCS as the prefix, so I made it more flexible to account for any sample ID that is present in the `per_sample_outs` folder. I'm pretty sure it's not technically possible for `cellranger multi` to output a folder for a sample ID that is not provided in the original config file, but just in case I have a check that the provided sample ID in the output folder is in the original meta. If it's not, I print out a warning. This was helpful for debugging since I could see when things weren't matching up as expected. 

The other thing I did was change this to output the paths to the sample directory and also the path to the full output directory. This is so everything gets copied over to the `checkpoints` directory except that really large intermediate folder. I think it could be helpful to have some of the other files for people running the workflow, but we could be more specific about what files to keep. For now, keeping the entire directory seemed reasonable.

allyhawkins:
> This looks good to me overall. I had one slightly different suggestion about how to do the outputs for the multiplexed outputs, because the idea of publishing both a folder and some of its subfolders made me a bit queasy. I think making separate named ouput channels also makes the downstream potentially a bit more clear. A bit. I think the only thing besides the multi and per_sample_outs is the config file, but please update if there was anything else, or if you think the config file should be included in other channels for any reason.

I understand that this wasn't ideal, so I updated it to use the multiple outputs that you had suggested. I think those are the only files that we should really need and encompasses most things. The one other change I made though is to emit the same folders from the single and multiplexed processes. For the singleplexed, the raw data is in the `multi` folder, but the filtered data is in the `per_sample_outs` folder, so I think we want to keep both even if we only use the raw folder in the actual workflow. 

I also updated some of the comments around the skipping mapping steps. Let me know if there are still things there that need to be cleared up.

modules/cellranger-flex.nf
```diff
@@ -95,17 +163,76 @@ workflow flex_quant{
...
 
-    // need to join back with skipped reads before outputting
-    flex_quants_ch = flex_channel.has_cellranger_flex
+        // update sample ID and return new meta with path to h5 file for each sample
+        updated_meta.sample_id = sample_id;
+        return [updated_meta, file("${out_dir}/count/sample_raw_feature_bc_matrix.h5")]
+      }
+
+    // combine single and multi outputs
+    // for singleplexed data, the raw H5 is in the multi folder
```
jashapiro, line 189: Thank you for this comment, as boy that is going to be confusing when you come back to it!

modules/cellranger-flex.nf
```diff
@@ -95,17 +163,76 @@ workflow flex_quant{
...
+    // define output H5 file and meta data using existing meta for singleplexed
+    has_flex_single_ch = has_flex_ch.single
       .map{meta -> tuple(
-        Utils.readMeta(file("${meta.cellranger_multi_publish_dir}/scpca-meta.json")),
-        file(meta.cellranger_multi_publish_dir, type: 'dir')
+        Utils.readMeta(file("${meta.cellranger_multi_results_dir}/scpca-meta.json")),
+        file("${meta.cellranger_multi_results_dir}/outs/multi/count/raw_feature_bc_matrix.h5")
+      )}
+
+    // define output H5 file for multiplexed
```
jashapiro, line 213:
```suggestion
    // define output H5 files for multiplexed
```

modules/cellranger-flex.nf
```diff
@@ -95,17 +163,76 @@ workflow flex_quant{
...
+        Utils.readMeta(file("${meta.cellranger_multi_results_dir}/scpca-meta.json"))
       )}
+      .transpose() // [sample id, meta]
+      // replace existing sample id and define path to existing directory with H5 file
+      .map{ sample_id, meta -> 
+        def updated_meta = meta.clone();
+        def demux_h5_file = file("${meta.cellranger_multi_results_dir}/outs/per_sample_outs/${sample_id}/count/sample_raw_feature_bc_matrix.h5");
+        updated_meta.sample_id = sample_id;
+        return [updated_meta, demux_h5_file]
+      }
```
jashapiro, line 229: I was wondering if you could do the whole thing without splitting for a bit less duplication
```suggestion
    has_flex_ch = flex_channel.has_cellranger_flex
      // [tuple(sample_ids), meta]
      .map{ meta -> tuple(
        meta.sample_id.tokenize(","),
        Utils.readMeta(file("${meta.cellranger_multi_results_dir}/scpca-meta.json"))
      )}
      .transpose() // [sample id, meta]
      // replace existing sample id and define path to existing directory with H5 file
      .map{ sample_id, meta ->
        def updated_meta = meta.clone();
        // path depends on whether singleplex or multiplex
        def demux_h5_file = if(meta.technology.contains("single")){
            file("${meta.cellranger_multi_results_dir}/outs/multi/count/raw_feature_bc_matrix.h5")
          } else if(meta.technology.contains("multi")) {
            file("${meta.cellranger_multi_results_dir}/outs/per_sample_outs/${sample_id}/count/sample_raw_feature_bc_matrix.h5")
          };
        updated_meta.sample_id = sample_id;
        return [updated_meta, demux_h5_file]
      }
```

modules/cellranger-flex.nf
```diff
@@ -95,17 +163,76 @@ workflow flex_quant{
...
+
+    // create channel of skipped libraries
+    flex_quants_ch = has_flex_single_ch.mix(has_flex_multi_ch)
```
jashapiro, line 232: If my previous suggestion works, you would not need this.
```suggestion
```

modules/cellranger-flex.nf
```diff
@@ -95,17 +163,76 @@ workflow flex_quant{
...
 
-    // TODO: Make sure both single and multi are incldued before mixing
-    all_flex_ch = cellranger_flex_single.out.mix(flex_quants_ch)
+    // Combine single, multi, and skipped libraries
+    all_flex_ch = cellranger_flex_ch.mix(flex_quants_ch)
```
jashapiro, line 235:
```suggestion
    all_flex_ch = cellranger_flex_ch.mix(has_flex_ch)
```

jashapiro (commented):
Looks good. My suggestion below is partly a question... Since the splitting the skipped channel was what was confusing me at first, I wonder if we can get away without splitting it. I added an `it` to the file name creation, but I think the rest works whether the sample is multiplexed or not? So that can remove a split/mix, which I think makes things a bit easier to follow?

allyhawkins:
@jashapiro I combined the skipped library channel like you suggested and made a few syntax fixes so Nextflow didn't complain. This change is much more straight forward to follow!

jashapiro (approved):
LGTM!

PR #898, opened 2025-06-18
Read in Cellranger output as SCE 

### LINKED ISSUE #884: Process for reading in cellranger multi output and generate SCE
Once we add support for running cellranger multi for both single and multiplexed samples in #876, we will need to add a process for reading in the cellranger output and generating an unfiltered SCE object. I haven't fully figured out how this will look since I want to dig into the exact outputs from running cellranger first, but I wanted to file this to track that this will be the next step.

allyhawkins:
Closes #884 

Here I am modifying the workflow to add a process for generating an unfiltered SCE object from the cellranger outputs. I made a few choices in doing this that I hope will help keep things modular while minimizing changes to how we process other data. 

The biggest change I made was breaking up what was previously the `generate_unfiltered_sce.R` script into two scripts: one script to actually read in the alevin data and create the SCE object and one script to do the formatting of the object which includes adding cell QC stats and sample metadata. Then I wrote a new script that is used to read in the cellranger data and create the initial SCE object, and pass that as input to the formatting script. Alternatively, I could have modified the existing script to read in either alevin or cellranger data, but that didn't seem ideal. I also wanted to minimize the amount of code that was duplicated between the scripts, which is why I chose to broke it up. 

I added a new workflow that emits the unfiltered and filtered SCE from cellranger. This output then gets joined with the other unfiltered and filtered objects and passed through the remaining part of the workflow. This also includes a new process for generating the unfiltered SCE from cellranger, which runs the two scripts - generate cellranger SCE and format SCE. 

A few other things to note:

- I was originally use the H5 file, but that was creating an SCE with a weirdly formatted matrix that was stored as a separate matrix named `seed` inside the `counts` assay. I tried to fix this by using `type = "sparse"` but you can only use that with the folder and not the H5 file. So I decided to just switch to using the folder rather than H5. 
- There are a few pieces of additional metadata that we add to objects processed with alevin-fry including the alevin-fry version. I feel like having the cellranger version is helpful, so I updated the output of the cellranger multi processes to include the `_versions` file and then read that into the R script. We also report the name of the index folder used for mapping with alevin, so I did the same thing here. 
- Related to metadata, I left a TODO about adding in getting the total and mapped reads from the cellranger output. With each run there is a `metrics_summary.csv` file that contains the total number of reads and then the percentage of reads mapped. I think we could get this information from there, but this PR started getting big so I thought that could be a next steps item. I'm attaching an example of the summary file for a multiplexed library: 
[metrics_summary.csv](https://github.com/user-attachments/files/20803827/metrics_summary.csv)


modules/cellranger-flex.nf
```diff
@@ -166,10 +170,18 @@ workflow flex_quant{
     // run cellranger multiplexed
     cellranger_flex_multi(flex_reads.multi, file(pool_file))
 
+    // get version files from output
+    // first mix multi and single versions back together and then will combine with cellranger output
+    versions_ch = cellranger_flex_single.out.versions.mix(cellranger_flex_multi.out.versions)
+      .map{ meta, versions_file -> tuple(
+        meta.library_id,
+        versions_file
+      )}
```
jashapiro, line 179: If we need the versions files, I think it makes sense to include them in the tuple output to save this effort?

modules/cellranger-flex.nf
```diff
@@ -64,6 +66,7 @@ process cellranger_flex_multi {
     tuple val(meta), path("${out_id}/outs/per_sample_outs/*", type: 'dir'), emit: "per_sample"
     tuple val(meta), path("${out_id}/outs/multi"), emit: "multi"
```
jashapiro, line 67: Since you are using the versions at least, it probably makes sense to  have just the two outputs, so you can keep everything together more easily without rejoining. The added files are small! You probably will want to update the singleplex to match this though. 
```suggestion
    tuple val(meta), 
      path("${out_id}/outs/per_sample_outs/*", type: 'dir'), 
      path("${out_id}/_versions"),
      path("${out_id}/outs/config.csv"),
      emit: "per_sample"
    tuple val(meta),
      path("${out_id}/outs/multi", type = 'dir'), 
      path("${out_id}/_versions"),
      path("${out_id}/outs/config.csv"),
      emit: "multi"
```

modules/cellranger-flex.nf
```diff
@@ -166,10 +170,18 @@ workflow flex_quant{
...
+        meta.library_id,
+        versions_file
+      )}
+
     // transpose cellranger multi output to have one row per output folder
-    // for multiplexed data, the raw H5 is in the per_sample_outs folder
+    // for multiplexed data, the directory with cellranger output is in the per_sample_outs folder
     cellranger_flex_multi_flat_ch = cellranger_flex_multi.out.per_sample
-      .transpose()
+      .transpose() // [meta, out_dir]
```
jashapiro, line 184: If you use the outputs as suggested, this would be:
```suggestion
      .transpose() // [meta, out_dir, versions, config]
```

jashapiro (commented):
I didn't have a chance to fully look through the scripts yet, but I think the overall pattern there looks fine. But I did notice you had to go to a fair bit of effort to rejoin the `_versions` directory, so I made a suggestion to keep that as part out the output tuple, so it will be carried along after transpose and you won't have to worry about joining channels.

PR #900, opened 2025-06-25
Add some documentation around 10x flex support

### LINKED ISSUE #888: Update documentation for running and maintaining example data for 10x flex
We should probably document in the internal documentation how we are using example 10x flex data for testing that workflow. Specifically, we should include some information on where the datasets were obtained from and instructions on how to run the workflow to process that example data.

allyhawkins:
Closes #888 

Here I am adding some documentation around running samples with 10x flex. 

- For internal instructions, I just added a section that mentions the example datasets and how to run the workflow to process those samples. I didn't think it needed much, but I just wanted to note where the example datasets came from and also how to actually run those samples with the `example` profile. 
- While I was here, I updated the external documentation to note that flex samples can be used. 
  - I updated the `technology` column options to include 10x flex.
  - I added a section within the special considerations section for using 10x flex data. The main thing I wanted to note is that `cellranger multi` is used and how multiplexed samples are dealt with. Let me know if you think there is more that we should add here?

### DIFF
--- a/components/dictionary.txt
+++ b/components/dictionary.txt
@@ -57,7 +57,6 @@ SCE
 scpca
 ScPCA
 Seqera
-Singleplexed
 singleplexed
 SingleR
 subdiagnosis
--- a/components/dictionary.txt
+++ b/components/dictionary.txt
@@ -58,6 +58,7 @@ scpca
 ScPCA
 Seqera
 Singleplexed
+singleplexed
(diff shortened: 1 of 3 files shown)


jashapiro (approved):
LGTM! I had one little comment about putting the Flex before Spatial, but it is really up to you.

PR #903, opened 2025-06-26
Update QC report tables to account for cellranger-multi processing 

### LINKED ISSUE #896: Update QC report to include cellranger metadata when technology is flex
The current QC report has some alevin-fry specific metrics that are reported in the summary tables. It probably doesn't make sense to include those for 10x flex data. We should pull out all the alevin-fry specific metrics and add them if the `mapping_tool` is `alevin-fry`. 

We should then have a separate set of metrics that get added if the `mapping_tool` is cellranger multi.

allyhawkins:
Closes #896 

This PR includes changes to the tables shown in the QC report to account for samples that were processed with `cellranger-multi` instead of `alevin-fry`. I ended up grabbing the `mapping_tool` from the metadata and modifying the raw library metrics and pre-processing information tables depending on which mapping tool is used. 
Here's an example of the updated report with a cellranger run:
[library06_qc 2.html.zip](https://github.com/user-attachments/files/20929558/library06_qc.2.html.zip)

One thing that we display in the table is the total reads and the percentage of reads mapped to the transcriptome. This information lives in the `metrics_summary.csv` file output in the `per_sample_outs` folder by `cellranger-multi`. So I made a few changes to the actual workflow to ensure we grab that file and read that into the script used to generate the unfiltered SCE object. Because it lives in the `per_sample_outs` regardless of library type and the raw file is either in `multi` or `per_sample_outs`, I had to emit those folders in the same output for singleplexed samples to avoid a bunch of joining. I then realized we weren't actually using the separate outputs, so I just combined them which I think is totally fine since we then select the files we need for downstream later. 

I then use the `metrics_summary.csv` to grab the total reads and the mapped reads. Note that in `alevin-fry` they report a number of reads for the mapped reads, but `cellranger-multi` reports a percentage. I ended up just keeping the % rather than converting to the actual number of reads since in the report we show the %, but let me know if you would like me to change it? 

Also, grabbing the metrics themselves was a little weird... There are a lot of very similarly worded names for the metrics in the summary file. The cellranger docs direct you to look at the web summary for definitions of each of the metrics. 
Screenshot 2025-06-26 at 11 02 47 AM

So I used the web summary from one of the libraries to find the row with the description of what I think is the total percentage of reads that were mapped. 
Screenshot 2025-06-26 at 11 03 03 AM
Here's an example of the metrics file for reference. 
[metrics_summary.csv](https://github.com/user-attachments/files/20929516/metrics_summary.csv)

So I chose to use the `Confidently mapped reads in cells` number. This metric appears twice in the metrics summary, but once where `Group Name` is empty and once where it has `GEX_1`. For singleplexed the number is the same each time it shows up, but for multiplexed the number is different and the `GEX_1` number corresponds to what's shown in the summary report... So I filtered to that group before grabbing the metrics.

### DIFF
--- a/modules/cellranger-flex.nf
+++ b/modules/cellranger-flex.nf
@@ -13,13 +13,8 @@ process cellranger_flex_single {
     tuple val(meta), 
       path("${out_id}/outs/per_sample_outs/*", type: 'dir'), 
       path("${out_id}/_versions"),
-      path("${out_id}/outs/config.csv"),
-      emit: "per_sample"
-    tuple val(meta),
-      path("${out_id}/outs/multi", type: 'dir'), 
-      path("${out_id}/_versions"),
-      path("${out_id}/outs/config.csv"),
-      emit: "multi"
+      path("${out_id}/outs/per_sample_outs/${meta.library_id}/metrics_summary.csv"),
+      path("${out_id}/outs/config.csv")
   script:
     out_id = file(meta.cellranger_multi_results_dir).name
     meta += Utils.getVersions(workflow, nextflow)
@@ -51,8 +46,9 @@ process cellranger_flex_single {
     meta += Utils.getVersions(workflow, nextflow)
     meta_json = Utils.makeJson(meta)
     """
-    mkdir -p ${out_id}/outs/per_sample_outs/${meta.sample_id}
+    mkdir -p ${out_id}/outs/per_sample_outs/${meta.library_id}
     mkdir -p ${out_id}/outs/multi
+    touch ${out_id}/outs/per_sample_outs/${meta.library_id}/metrics_summary.csv
     touch ${out_id}/outs/config.csv
     touch ${out_id}/_versions
     echo '${meta_json}' > ${out_id}/scpca-meta.json
@@ -74,11 +70,6 @@ process cellranger_flex_multi {
       path("${out_id}/_versions"),
       path("${out_id}/outs/config.csv"),
       emit: "per_sample"
-    tuple val(meta),
-      path("${out_id}/outs/multi", type: 'dir'), 
-      path("${out_id}/_versions"),
-      path("${out_id}/outs/config.csv"),
-      emit: "multi"
   script:
     out_id = file(meta.cellranger_multi_results_dir).name
     meta += Utils.getVersions(workflow, nextflow)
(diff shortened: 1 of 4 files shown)


modules/cellranger-flex.nf
```diff
@@ -198,18 +190,24 @@ workflow flex_quant{
...
+    cellranger_flex_ch = cellranger_flex_single.out
       // get path to raw output directory for singleplexed 
-      .map{ meta, out_dir, versions, config -> tuple(
+      .map{ meta, out_dir, versions, metrics, config -> tuple(
         meta, 
-        file("${out_dir}/outs/multi/count/raw_feature_bc_matrix", type: 'dir'),
-        file(versions)
+        file("${out_dir}/count/raw_feature_bc_matrix", type: 'dir'),
+        file(versions),
+        file(metrics)
```
jashapiro, line 210: These are coming in as files already, so you should not need to modify them
```suggestion
        versions,
        metrics
```

modules/cellranger-flex.nf
```diff
@@ -198,18 +190,24 @@ workflow flex_quant{
...
     // for singleplexed data, the raw output is in the multi folder
-    cellranger_flex_ch = cellranger_flex_single.out.multi
+    cellranger_flex_ch = cellranger_flex_single.out
       // get path to raw output directory for singleplexed 
-      .map{ meta, out_dir, versions, config -> tuple(
+      .map{ meta, out_dir, versions, metrics, config -> tuple(
         meta, 
-        file("${out_dir}/outs/multi/count/raw_feature_bc_matrix", type: 'dir'),
-        file(versions)
+        file("${out_dir}/count/raw_feature_bc_matrix", type: 'dir'),
```
jashapiro, line 208: I'm not 100% sure this will work, but since `out_dir` is already a file path, you should be able to modify it with the `/` operator. I don't know this is actually any more clear though 
```suggestion
        out_dir / "count/raw_feature_bc_matrix" ,
```

jashapiro (commented):
I left some comments below, but I'd say I am a little confused at the moment between the code as it stands and your comments later. 

The main thing is that it looks from the code like you are not using the `outs/multi` output directory at all now, which could be fine, but I would still like it to be kept as an output in a second channel, just to be sure it is saved to the checkpoints directory (in case we ever want to come back to demux differently). But in your comments you said that you needed a file from it, which I am not seeing in the code. To me it looks like in your code the `metrics_summary.csv` is in the `per_sample_outs/`? In which case we might not need it to be a separate output? Of course, the file is small, so the separate output is just fine; I'm just not sure it is needed.

Or I might just be confused?

Issue #902, opened 2025-06-26
Test processing 10x flex data with --perform_celltyping


allyhawkins:
So far, I have been able to run the workflow with the 10x flex example data successfully, but have been doing all runs with `--perform_celltyping false`. We should do a test run with the `--perform_celltyping` flag set to true (default in the example profile), just to be sure that there aren't any issues with doing cell typing with the 10x flex data. If there are any further modifications that we need to make we should do that before closing this.

allyhawkins:
It looks like we are going to need to adjust the models we use with `SingleR` to account for a different gene set for flex data... I'm seeing this error for the `SingleR` process which indicates there are genes in the model that are not found in the input gene set. 

```
'rownames(test)' does not contain all genes used in 'trained'
  Calls: <Anonymous> -> lapply -> FUN
```

In looking at the [docs for `SingleR`](https://bioconductor.org/books/release/SingleRBook/advanced-options.html#preconstructed-indices), I think all genes in the object have to be present in the model. 

> The resulting data structure can then be re-used across multiple classifications with different test datasets, provided the gene annotation in the test dataset is identical to or a superset of the genes in the training set. 

We need to build a new model, which should probably happen in [`build-celltype-ref.nf`](https://github.com/AlexsLemonade/scpca-nf/blob/main/build-celltype-ref.nf). I think we probably want a separate process that trains the models using the cellranger index rather than the alevin-fry index and then we should make sure that the file names for the models reflect the mapping tool they should be used with. 

Note that for `CellAssign` we have a step that filters the genes to only shared genes prior to running it, so we should not have the same issues.

allyhawkins:
I was able to run through a 10x flex sample with cell typing so closing this (See #911 for the last change).

PR #904, opened 2025-06-26
Train SingleR model to use with 10x flex

allyhawkins:
Related to #902 

Here I am adding a process to the `build-celltype-ref.nf` workflow to create a SingleR model file for use with the 10X flex datasets. For running cell typing, we need to have models that were trained using the appropriate gene list, which in this case is the genes that are present in the probe set for 10x flex. 

To do this, I ended up creating a new process that runs the script for training models but provides the probe set reference from 10x rather than the tx2gene file from the transcriptome used with `alevin-fry`. I modified the existing script to take in either the tx2gene or probeset to create the list of gene ids used to build the model. 

I was really trying to minimize the amount of changes here, which meant keeping everything organized in the same way but just adding new model files that can be used with flex. But we need to distinguish the filenames in some way to indicate they are for flex. I also noticed when revisiting the models we built previously that we don't account for the version of the transcriptome used to create the model. We only document the celldex version, but the model itself is restricted to the set of genes in the transcriptome used when mapping with alevin-fry. I think it might be useful to have that information. 

I think the best way to do this is to add a reference gene source to the file name so that they are now: `<ref_name>_<source>_<version>_<genes_ref>_model.rds`. 

For now, I'm making the changes necessary to accommodate these new file names, which were pretty minimal. How do we feel about changing the file names? We should probably keep the old files there, too, since skipping `SingleR` relies on the model file name being the same. We could also just add the information about `flex` to the new files and not re-create any of the other model files? I see benefits of both options so curious what others think.

### DIFF
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -612,7 +612,12 @@ We have provided an example multiplex pool file for reference that can be found 
 Libraries processed with the [GEM-X Flex Gene Expression protocol from 10x Genomics](https://www.10xgenomics.com/products/flex-gene-expression) using either single or multiplexing will be quantified using [`cellranger multi`](https://www.10xgenomics.com/support/software/cell-ranger/latest/analysis/running-pipelines/cr-flex-multi-frp) instead of `salmon` and `alevin-fry`. 
 *Note:* Currently only libraries processed using the v1.1.0 probe set are supported. 
 
-There are no special considerations for singleplexed libraries other than indicating the appropriate `technology` in the `run_metadata.tsv` file, `10Xflex_v1.1_single`. 
+You will need to provide a [docker image](https://docs.docker.com/get-started/) that contains the [Cell Ranger software from 10X Genomics](https://www.10xgenomics.com/support/software/cell-ranger/downloads).
+For licensing reasons, we cannot provide a Docker container with Cell Ranger for you.
+As an example, the Dockerfile that we used to build Cell Ranger can be found [here](https://github.com/AlexsLemonade/alsf-scpca/tree/main/images/cellranger).
+After building the docker image, you will need to push it to a [private docker registry](https://www.docker.com/blog/how-to-use-your-own-registry/) and set `params.CELLRANGER_CONTAINER` to the registry location and image ID in the `user_template.config` file.
+
+There are no special considerations for singleplexed libraries other than indicating the appropriate `technology` in the `run_metadata.tsv` file, `10Xflex_v1.1_single`.
 
 If the libraries are multiplexed, the appropriate `technology` term, `10Xflex_v1.1_multi`, will need to be indicated in the `run_metadata.tsv` file and an additional TSV file, `cellhash_pool_file`, must be provided. 
 When processing multiplexed libraries, demultiplexing will be performed by `cellranger multi`, so the quantified gene expression data for each sample will be output separately.  
--- a/modules/export-anndata.nf
+++ b/modules/export-anndata.nf
@@ -8,11 +8,18 @@ process export_anndata {
     input:
       tuple val(meta), path(sce_file), val(file_type)
     output:
-      tuple val(meta), path("${meta.library_id}_${file_type}_*.h5ad"), val(file_type)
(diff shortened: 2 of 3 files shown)


bin/make_celltype_ref_table.R
```diff
@@ -4,7 +4,7 @@
 # and output a table with the reference name, source, and version
 
 # File names must be of the format:
-#  <reference_name>_<source>_<version><optional_extra.ext>
+#  <reference_name>_<source>_<version>_<gene_ref><optional_extra.ext>
```
jashapiro, line 7:
```suggestion
#  <reference_name>_<source>_<version>_<gene_set>_<optional_extra.ext>
```

build-celltype-ref.nf
```diff
@@ -47,7 +49,38 @@ process train_singler_models {
...
+  label 'cpus_4'
+  label 'mem_16'
+  input:
+    tuple val(ref_name), path(ref_file)
+    path flex_probeset
+  output:
+    path celltype_model
+  script:
+    ref_file_basename = file("${ref_file}").baseName
+    celltype_model = "${ref_file_basename}_v1.1-flex-probes_model.rds"
```
jashapiro, line 71: I don't like the flex version being hardcoded here!

jashapiro (commented):
I haven't really looked too closely at the code here, because I wanted to just think about the correct approach in the abstract first...

Do we want to have separate gene sets for the assignment that depend on the type of input data? Or should we restrict to the common genes across the two transcriptome references? I guess part of the question is how different the gene sets are. 

I think what I am saying is that we might just want to update the model. That will require reprocessing everything. But I think we will want to do that anyway even if it just a name change. Which may not be since the celldex version has been updated. I suspect we will want to keep the old model files are around either way.

Either way, let's think about what the names going forward should be, as I do think we want to get some version info into the names that might not have been present to start. 

So the current names are things like:
`BlueprintEncodeData_celldex_1-10-1_model.rds`

And the new proposal is: 
`BlueprintEncodeData_celldex_1-14-0_Homo-sapiens.GRCh38.104_model.rds`
`BlueprintEncodeData_celldex_1-14-0_v1.1-flex-probes_model.rds`

Overall, I think that is mostly fine, but a few suggestions:
- a bit more consistent in the punctuation (convert periods to dashes as we do for the celldex version)
- removing `Homo-sapiens` because `GRCh38` already tells you that. you could also probably just reduce it to `ensembl-104` with no loss of specificity.
- move version string to end of flex string (and get the version from input, rather than hardcoding it in the process)
- add the creation date for versioning? Maybe that is overkill though.

So I would propose something like: 
`BlueprintEncodeData_celldex_1-14-0_GRCh38-104_2025-06-27_model.rds`
`BlueprintEncodeData_celldex_1-14-0_10xFlex-v1-1_2025-06-27_model.rds`

jashapiro:
> For some added context, the alevin-fry tx2gene has 60,000 genes in it and the probe set has 19,000 genes in it.

The gene sets in celldex are closer to the 19K genes, so I suspect that the difference in gene sets after taking the intersections is much smaller as well. Which is to say I expect that `instersection(celldex,  tx2gene)` to be pretty similar to `intersection(celldex, flex)`, even though tx2gene and flex are quite distinct.

config/reference_paths.config
```diff
@@ -17,6 +17,10 @@ probes_dir = "${params.ref_rootdir}/flex-probe-refs"
 // name must match one of the keys in ref_json
 celltype_organism = "Homo_sapiens.GRCh38.104"
 
+// parameters for creating references for cell typing with 10x flex
+// version of probeset used, will be set as the name in the SingleR model file
+celltype_probeset_version = "10XFlex-v1-1"
```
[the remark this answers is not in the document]
jashapiro: > > Not I also changed to 10xFlex, for "correctness" but also because it is easier to read?
> 
> I was using 10XFlex because it matches how we refer to the technology in the metadata columns... We use capital X for all the techs which was probably just a bad move on our part many years ago but we can't change it now. So I can change it here, but just wanted to provide some rationale as to why I'm using capital X.

Yeah, I realized that in my next review. But then we should use `10Xflex` to match.

PR #908, opened 2025-07-01
Update 10x flex multiplexed filename outputs and document docker requirements

### LINKED ISSUE #901: Ensure demultiplexed output files have unique filenames for 10x flex runs
I noticed when looking at the outputs for the example 10x flex multiplexed data, that the rds files are labeled with the library ID and therefore are duplicated across all samples from the same library. I think having the files named with just the library ID and not the sample ID could be confusing, even if they are in separate folders for each sample. 

It might be worth updating the file names for these files to be `<library id>-<sample id>_unfiltered.rds`. 

Eventually these files for 10X flex samples will go on the Portal, so we should be cognizant of the fact that changing the filename could affect more than just this workflow. 

Here's an example of what the output looks like right now: 

Image

### LINKED ISSUE #907: Add instructions about creating docker image for using 10x flex
We should make a note in the external instructions about how you need to create your own docker image with cellranger to run any 10x flex samples. We mention it in the [spatial section](https://github.com/AlexsLemonade/scpca-nf/blob/10x-flex-support/external-instructions.md#spatial-transcriptomics-libraries), so just need to add some similar wording to the 10x flex section.

allyhawkins:
Closes #901 
Closes #907 

Here I'm updating the two processes that publish the SCE and AnnData files to rename any multiplexed 10x flex samples to have `<library_id>-<sample_id>` as the file prefix instead of just library id. This was a pretty simple fix and I just checked for the 10X flex technology before defining the prefix to use for the files. Is there a different approach that would be preferred? 

While I was here, I went ahead and updated the instructions to use Docker to address #907. I mostly copied the same language we have in the spatial section, but updated it for cellranger.

### DIFF
--- a/modules/export-anndata.nf
+++ b/modules/export-anndata.nf
@@ -11,7 +11,7 @@ process export_anndata {
       tuple val(meta), path("${file_prefix}_${file_type}_*.h5ad"), val(file_type)
     script:
       // set output file names based on having 10x flex multiplexed or not 
-      if (meta.technology == "10Xflex_v1.1_multi"){
+      if (meta.technology in ["10Xflex_v1.1_multi"]){
         file_prefix = "${meta.library_id}-${meta.sample_id}"
       } else {
         file_prefix = "${meta.library_id}"
@@ -42,7 +42,7 @@ process export_anndata {
       """
     stub:
       // set output file names based on having 10x flex multiplexed or not 
-      if (meta.technology == "10Xflex_v1.1_multi"){
+      if (meta.technology in ["10Xflex_v1.1_multi"]){
         file_prefix = "${meta.library_id}-${meta.sample_id}"
       } else {
         file_prefix = "${meta.library_id}"
--- a/modules/export-anndata.nf
+++ b/modules/export-anndata.nf
@@ -8,11 +8,18 @@ process export_anndata {
     input:
       tuple val(meta), path(sce_file), val(file_type)
     output:
-      tuple val(meta), path("${meta.library_id}_${file_type}_*.h5ad"), val(file_type)
+      tuple val(meta), path("${file_prefix}_${file_type}_*.h5ad"), val(file_type)
     script:
-      rna_h5ad_file = "${meta.library_id}_${file_type}_rna.h5ad"
-      feature_h5ad_file = "${meta.library_id}_${file_type}_${meta.feature_type}.h5ad"
-      pca_meta_file = "${meta.library_id}_${file_type}_pca.tsv"
+      // set output file names based on having 10x flex multiplexed or not 
+      if (meta.technology == "10Xflex_v1.1_multi"){
+        file_prefix = "${meta.library_id}-${meta.sample_id}"
+      } else {
+        file_prefix = "${meta.library_id}"
+      }
+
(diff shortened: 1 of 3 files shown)


modules/export-anndata.nf
```diff
@@ -8,11 +8,18 @@ process export_anndata {
...
       tuple val(meta), path(sce_file), val(file_type)
     output:
-      tuple val(meta), path("${meta.library_id}_${file_type}_*.h5ad"), val(file_type)
+      tuple val(meta), path("${file_prefix}_${file_type}_*.h5ad"), val(file_type)
     script:
-      rna_h5ad_file = "${meta.library_id}_${file_type}_rna.h5ad"
-      feature_h5ad_file = "${meta.library_id}_${file_type}_${meta.feature_type}.h5ad"
-      pca_meta_file = "${meta.library_id}_${file_type}_pca.tsv"
+      // set output file names based on having 10x flex multiplexed or not 
+      if (meta.technology == "10Xflex_v1.1_multi"){
```
jashapiro, line 14: Just because I commented in the last PR, we should standardize capitalization here

But for now I am changing it to `in` for potential future expansion.
```suggestion
      if (meta.technology in ["10Xflex_v1.1_multi"]){
```

modules/export-anndata.nf
```diff
@@ -34,8 +41,15 @@ process export_anndata {
 
       """
     stub:
-      rna_h5ad_file = "${meta.library_id}_${file_type}_rna.h5ad"
-      feature_h5ad_file = "${meta.library_id}_${file_type}_${meta.feature_type}.h5ad"
+      // set output file names based on having 10x flex multiplexed or not 
+      if (meta.technology == "10Xflex_v1.1_multi"){
```
jashapiro, line 45:
```suggestion
      if (meta.technology in ["10Xflex_v1.1_multi"]){
```

modules/publish-sce.nf
```diff
@@ -18,18 +18,25 @@ process qc_publish_sce {
     workflow_url = workflow.repository ?: workflow.manifest.homePage
     workflow_version = workflow.revision ?: workflow.manifest.version
 
+    // set output file names based on having 10x flex multiplexed or not 
+    if (meta.technology == "10Xflex_v1.1_multi"){
```
jashapiro, line 22:
```suggestion
    if (meta.technology in ["10Xflex_v1.1_multi"]){
```

modules/publish-sce.nf
```diff
@@ -72,15 +79,21 @@ process qc_publish_sce {
       --metrics_json "${metrics_json}"
     """
   stub:
-    unfiltered_out = "${meta.library_id}_unfiltered.rds"
-    filtered_out = "${meta.library_id}_filtered.rds"
-    processed_out = "${meta.library_id}_processed.rds"
-    qc_report = "${meta.library_id}_qc.html"
-    metadata_json = "${meta.library_id}_metadata.json"
-    metrics_json = "${meta.library_id}_metrics.json"
+    if (meta.technology == "10Xflex_v1.1_multi"){
```
jashapiro, line 82:
```suggestion
    if (meta.technology in ["10Xflex_v1.1_multi"]){
```

jashapiro (approved):
LGTM, with just a touch of future-proofing.

Issue #893, opened 2025-06-13
Add profile for new data lab AWS account batch


jashapiro:
Following the completion of https://github.com/AlexsLemonade/scpca-nf-infra/issues/6, we will want to create a profile that allows running on our updated infrastructure. 

If we have decided to use Wave containers and the fusion file system for this iteration, we will want to be sure to set that option in the configuration.

jashapiro:
This should also use the ECR described in https://github.com/AlexsLemonade/scpca-nf-infra/issues/7

PR #916, opened 2025-07-14
Update Nextflow configs to use recommended structure

jashapiro:
While working on #893, I ran into a number of validation errors while using the Nextflow VScode plugin. The crux of these was that we were apparently using `includeConfig` a bit more liberally than they would have liked, such that we were including "fragments" of config files that would not be valid on their own.

From https://www.nextflow.io/docs/latest/config.html:
> Config includes can also be specified within config blocks. However, config files should only be included at the top level or in a profile so that the included config file is valid on its own and in the context in which it is included.

This PR just includes the cleanup to pass validation, which mostly means moving the include blocks up a level to either the top level or the top level of a profile, and adding the appropriate wrapping blocks to the included files.


allyhawkins (approved):
LGTM

PR #919, opened 2025-07-15
scpcaTools v0.4.3 - main #2

allyhawkins:
Duplicate of #917 
Let's try this again... This should merge in the changes needed for the release to `main`.

### DIFF
--- a/.github/workflows/nextflow-stub-check.yaml
+++ b/.github/workflows/nextflow-stub-check.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Nextflow
         uses: nf-core/setup-nextflow@v2
         with:
-          version: 24.10.6
+          version: latest
 
       - name: Check Nextflow workflow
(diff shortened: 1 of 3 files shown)


.github/workflows/nextflow-stub-check.yaml
```diff
@@ -14,7 +14,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Nextflow
-        uses: nf-core/setup-nextflow@v2.0.0
+        uses: nf-core/setup-nextflow@v2
+        with:
+          version: 24.10.6
```
sjspielman, line 19: So why is it working with the `latest` in `development` but not here is my question, do you have a sense..?

https://github.com/AlexsLemonade/scpca-nf/blob/838754d5bf1caabe07eed27c6fc6eb3892790299/.github/workflows/nextflow-stub-check.yaml#L16-L19

sjspielman (approved):
🐋

PR #921, opened 2025-07-16
Small fixes for scpca-nf

jashapiro:
A couple of small fixes for the scpca-nf running in the new account.

The main one is the change to accommodate the new style of cell type reference paths. We might need to incorporate that one into the dev branch sooner rather than later, but probably not right away. I mostly started from copying the version in the flex branch, but I made a few changes to deal with the potential for either file name format. I also added the extra fields to the metadata, because why not. 

The other change is just fixing the AWS secret name.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -32,7 +32,7 @@ option_list <- list(
     opt_str = c("--singler_model_file"),
     type = "character",
     help = "Name of file containing a single model generated for SingleR annotation.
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
LGTM, but maybe we should file an issue or note somewhere that we need to update the reference file name handling in `development`?

jashapiro:
> LGTM, but maybe we should file an issue or note somewhere that we need to update the reference file name handling in `development`?

Filed a PR: #922

PR #923, opened 2025-07-17
Update documentation for running scpca-nf on Cavatica

### LINKED ISSUE #890: Add documentation about sbmanifest usage for converting metadata files
While I have a script for using the `sbmanifest` program to convert the run metadata file for use on Cavatica as part of #865 , this is severely underdocumented. We need to expand this documentation to explain the process for updating the file with the required paths, as well as updating any additional files that might be needed. 

Briefly, any table that refers to files that are stored in Cavatica will need to be updated: this issue should include cataloging the files where this is most likely to be needed in the short and medium term (I think we can restrict support to our set of references for now, for example).

### LINKED ISSUE #891: Add documentation for running scpca-nf on Cavatica
With #865 nearly complete, we will also need to have documentation about using `scpca-nf` on Cavatica. We will want to include basic instructions in the `sb_nextflow_schema.yaml` file docs, so it is visible on the website, but we may want additional info that we can link to as well.

jashapiro:
Closes #891 
Closes #890

I revamped the docs for Cavatica, making the main instructions a separate, somewhat longer, document. I shortened up the instructions that will display on Cavatica itself, with links to the main docs (that don't work now but will once we have this merged).

The biggest thing I ran into was that the `sbmanifest` script does not work well with `NA` values, and so I added a little bit about how to deal with that. Sadly, it also breaks validation, but I can't figure out a way around that really, without rewriting the whole script. I plan to file an issue about it, but I don't have much faith that it will be addressed with any rapidity.

Finally, I was having trouble with spell checking with emoji and pre-commit, so I took that out for now. We'll see if it passes the GHA check. Or I might have to abandon my emoji.


allyhawkins (approved):
LGTM

Issue #929, opened 2025-07-28
Add module to detect doublets


sjspielman:
This issue tracks adding a submodule to run the doublet detection script being developed in #928.

- The module should be structured about the same as our clustering module, for example: https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/modules/cluster-sce.nf
- We can add it into `main.nf` here right after clustering: https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/main.nf#L301-L302

sjspielman:
Thinking again, maybe we don't want to slot this near the clustering process since we should probably be performing this detection on the _filtered_ (not the processed) object, since this corresponds more in practice to when someone would run this step? @jashapiro @allyhawkins sorry missed this bit in DSTM, but let me know if you have thoughts :)

sjspielman:
> I think we want to use the filtered objects since those have all cells in them before any further filtering? See the screenshot below from the GitHub page.

Yeah, that's where I am too.

sjspielman:
Closed by #938

PR #936, opened 2025-07-29
Add script to detect doublets

### LINKED ISSUE #928: Add script to detect doublets
This issue tracks adding a script to `bin/` to detect doublets using `scDblFinder`. The code should be adapted from this script in `OpenScPCA-nf`: https://github.com/AlexsLemonade/OpenScPCA-nf/blob/5b04808c23ebe5a127c132305320929d4b94714f/modules/doublet-detection/resources/usr/bin/run_scdblfinder.R

sjspielman:
Closes #928 

This PR adds a script to detect doublets on a given SCE file*. There are a few main things to be aware of:
- I run `scDblFinder` at defaults which seems fine to me for the goal of this addition
- I save 3 bits of information to the SCE:
  - the crux values are saved in the `colData` directly: the classification and the associated doublet score
  - the full table that `scDblFinder` exports is saved in `metadata`. It's a not-huge DataFrame and seemed worth including for transparency in the same way that we include the full SingleR output.

*Note I was agnostic about the process stage of this input file in how I named this argument, because I wondered if we want to be running this on filtered and not processed? Feel free to weigh in as part of review! https://github.com/AlexsLemonade/scpca-nf/issues/929#issuecomment-3133020583

### DIFF
--- a/bin/detect_doublets.R
+++ b/bin/detect_doublets.R
@@ -50,7 +50,7 @@ stopifnot(
 
 # set up multiprocessing params
 if (opt$threads > 1) {
-  bp_param <- BiocParallel::MulticoreParam(opt$threads, RNGSeed = opt$seed)
+  bp_param <- BiocParallel::MulticoreParam(opt$threads)
 } else {
   bp_param <- BiocParallel::SerialParam()
 }


sjspielman:
> I'm on the fence about including the full result though. I know it's small, but what information are we getting from that data frame that someone might use?

The main information you get is all the per-cell quantities calculated as intermediate steps in scDblFinder, and all the information about the artificial doublets that were simulated as part of the classification. I don't think any of these items are strictly needed though, so I'm ok removing this.

jashapiro:
> I updated this as requested, and also!! While looking through scDblFinder docs to make sure that those table quantities aren't really necessarily to provide, I noticed this 😬 https://bioconductor.org/packages/devel/bioc/vignettes/scDblFinder/inst/doc/scDblFinder.html#how-can-i-make-this-reproducible So, I updated to use `RNGseed` in a couple scripts that seem they could use it. Let me know if you disagree with that change and/or prefer to make it in a separate PR perhaps?

Were you seeing changes? My read of those docs is that it only applies if you use the `samples` argument, which we were not. Not a problem to add it though, as it can't hurt.

allyhawkins:
> I updated this as requested, and also!! While looking through scDblFinder docs to make sure that those table quantities aren't really necessarily to provide, I noticed this 😬 https://bioconductor.org/packages/devel/bioc/vignettes/scDblFinder/inst/doc/scDblFinder.html#how-can-i-make-this-reproducible So, I updated to use `RNGseed` in a couple scripts that seem they could use it. Let me know if you disagree with that change and/or prefer to make it in a separate PR perhaps?

I'm confused because the first sentence of that FAQ you linked says to use `set.seed` and then it says if processing multiple samples to use `RNGseed` and we are only processing one sample. And then it says if processing the samples serially to use it, so I'm not sure what to do if processing a single sample with `MulticoreParam()`. I don't think we need to change it?

allyhawkins (approved):
LGTM

PR #937, opened 2025-07-29
Make sure celltype checkpoint files have unique names for flex multiplexed libraries

### LINKED ISSUE #935: Make sure checkpoints for cell typing with flex multi samples have unique names
Once we complete cell typing for a library, we save the results to the `checkpoints` folder and name the folder by the library ID. This won't work when we have flex samples that have been multiplexed because multiple samples have the same library ID but need to be cell typed separately. Right now this means the results are being overwritten for each sample and there are joining errors when re-running the workflow and using previously generated results for cell typing.

allyhawkins:
Closes #935 

We have been saving the cell type checkpoints results to `library_id/library_id-cellassign` or `library_id/library_id-singler`, which will not work for any flex libraries that are multiplexed. We need to make sure the folder names are unique and use both the library and sample ID as we do with the final published files. Additionally, we were using the library ID for all joins when skipping cell typing that already exists. When we have multiple samples with the same library ID this causes an error. 

To deal with those issues, I added a `unique_id` to `meta` at the very beginning of the cell typing workflow. This ID is a library ID unless the library is multiplexed with 10x flex, and then I used the library ID and the sample ID. I then use that unique ID to name the output folder. 

We also have a lot of logic in this workflow that was using the library ID to join things back together after skipping or not having certain files. I updated that to use the unique ID instead so this should solve any joining errors. 

Lastly, I noticed in the example data we had the `is_cell_line` set to `TRUE` for some of the multiplexed samples. We want to be able to test cell typing with those samples, so we should set that to `FALSE`. 

I tested this with the multiplexed library that we have in the example data with `SingleR`, but did not run `CellAssign`.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -309,18 +309,18 @@ workflow {
   // combine all SCE outputs
   // Make channel for all library sce files
   all_sce_ch = sce_ch.no_genetic.mix(flex_sce_ch.continue_processing, genetic_demux_sce.out)
-    .map{ meta_in, unfiltered_sce, filtered_sce -> 
-        def meta = meta_in.clone(); // clone meta before adding in unique id
-        // define a unique ID that will be used to label output folders and join skipped libraries throughout workflow
-        // we can't use library ID for flex multiplexed so will use sample and library ID for those samples only 
-        meta.unique_id = (meta.technology in ["10Xflex_v1.1_multi"]) ? "${meta.library_id}-${meta.sample_id}" : "${meta.library_id}";
-        // return updated meta and sce files 
-        return [
-          meta,
-          unfiltered_sce,
-          filtered_sce
-        ]
-      }
+    // add  unique ID  to metadata that will be used to label output folders and join skipped libraries throughout workflow
+    .map{ meta_in, unfiltered_sce, filtered_sce ->
+      def meta = meta_in.clone(); // clone meta before adding in unique id
+      // we can't use library ID for flex multiplexed so will use sample and library ID for those samples only
+      meta.unique_id = (meta.technology in ["10Xflex_v1.1_multi"]) ? "${meta.library_id}-${meta.sample_id}" : "${meta.library_id}";
+      // return updated meta and sce files
+      return [
+        meta,
+        unfiltered_sce,
+        filtered_sce
+       ]
+    }
   post_process_sce(all_sce_ch)
 
 


modules/classify-celltypes.nf
```diff
@@ -162,6 +162,19 @@ workflow annotate_celltypes {
 
     // branch to cell type the non-cell line libraries only
     sce_files_channel_branched = sce_files_channel
+      .map{ meta_in, unfiltered_sce, filtered_sce, processed_sce -> 
+        def meta = meta_in.clone(); // clone meta before adding in unique id for tracking cell type results
+        // define a unique ID that will be used to label output folders and join skipped libraries
+        // we can't use library ID for flex multiplexed so will use sample and library ID for those samples only 
+        meta.unique_id = (meta.technology in ["10Xflex_v1.1_multi"]) ? "${meta.library_id}-${meta.sample_id}" : "${meta.library_id}";
```
jashapiro, line 169: I think I'd like to see this change in the main workflow, right after making the SCE files? Then we would have it available more generally?

jashapiro (commented):
This looks good to me.

My only thought is that we might want to make the unique_id part of the `meta` in the main workflow so that it can be used elsewhere for consistency as needed. At this stage, I'm not sure where exactly that would be, but it seems like something we would like to standardize if we can. I might file an issue about extending the use across other modules, but not otherwise spend time on those modifications now (aside from adding `unique_id` earlier).

allyhawkins:
> My only thought is that we might want to make the unique_id part of the meta in the main workflow so that it can be used elsewhere for consistency as needed. At this stage, I'm not sure where exactly that would be, but it seems like something we would like to standardize if we can. I might file an issue about extending the use across other modules, but not otherwise spend time on those modifications now (aside from adding unique_id earlier).

This definitely seems reasonable to me. For now, I moved it to right before post processing and after generating the unfiltered and filtered SCEs. That's the earliest point in the workflow where everything is all in one channel? I can file an issue to track making updates to actually use that unique ID in other places.

main.nf
```diff
@@ -309,6 +309,18 @@ workflow {
...
+        // define a unique ID that will be used to label output folders and join skipped libraries throughout workflow
+        // we can't use library ID for flex multiplexed so will use sample and library ID for those samples only 
+        meta.unique_id = (meta.technology in ["10Xflex_v1.1_multi"]) ? "${meta.library_id}-${meta.sample_id}" : "${meta.library_id}";
+        // return updated meta and sce files 
+        return [
+          meta,
+          unfiltered_sce,
+          filtered_sce
+        ]
+      }
```
jashapiro, line 323: Just moving the initial comment and updating indentation
```suggestion
    // add  unique ID  to metadata that will be used to label output folders and join skipped libraries throughout workflow
    .map{ meta_in, unfiltered_sce, filtered_sce ->
      def meta = meta_in.clone(); // clone meta before adding in unique id
      // we can't use library ID for flex multiplexed so will use sample and library ID for those samples only
      meta.unique_id = (meta.technology in ["10Xflex_v1.1_multi"]) ? "${meta.library_id}-${meta.sample_id}" : "${meta.library_id}";
      // return updated meta and sce files
      return [
        meta,
        unfiltered_sce,
        filtered_sce
       ]
    }
```

jashapiro (approved):
LGTM, with just a minor formatting update.

PR #938, opened 2025-07-29
Incorporate doublet detection into workflow

### LINKED ISSUE #929: Add module to detect doublets

sjspielman:
Closes #929 

In this PR, I am adding code to run the `detect_doublets.R` script into the workflow. Given that we should probably be running this on filtered instead of processed files, I decided for a first stab to take the path of absolute least resistance and just add the script to run inside our existing `filter_sce` process, which does use the `slim` container (for which `scDblFinder` inclusion is still pending in https://github.com/AlexsLemonade/scpcaTools/pull/305). This code is therefore analogous to how we add in submitter annotations in the unfiltered processes, e.g.:

https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/modules/sce-processing.nf#L4-L45

It reads in an SCE, adds some colData, and exports the SCE to the same input file. As such, I updated the R script to have only a single `sce_file` argument. I also reverted a smidge and renamed the seed to `random_seed` since that better matches what's in this area of the workflow, it turns out....

I'm opening this as a draft until https://github.com/AlexsLemonade/scpcaTools/pull/305 is completed, but in the meantime people can feel free to weigh in, or let me know if a bona fide module would be better here for (as the name suggests...) workflow modularity :)

### DIFF
--- a/bin/detect_doublets.R
+++ b/bin/detect_doublets.R
@@ -62,7 +62,7 @@ sce <- readr::read_rds(opt$input_sce_file)
 doublet_result <- scDblFinder::scDblFinder(
   sce,
   BPPARAM = bp_param,
-  resultType = "table" # keep as table in case we eventually want to provide additional output
+  returnType = "table" # keep as table in case we eventually want to provide additional output
 )
 
 # store the `score` and `class` columns in the colData
--- a/bin/detect_doublets.R
+++ b/bin/detect_doublets.R
@@ -4,7 +4,7 @@
 #
 # This script reads in an RDS file containing an SCE and runs scDblFinder.
 # Classifications and doublet scores are stored in the SCE's colData slot.
-# The SCE is then exported to the same file as was inputted.
+# The SCE is then exported.
 
 # import libraries
 library(optparse)
@@ -13,10 +13,16 @@ library(SingleCellExperiment)
 # Set up optparse options
 option_list <- list(
   make_option(
-    opt_str = c("--sce_file"),
+    opt_str = c("--input_sce_file"),
     type = "character",
     default = "",
     help = "Path to RDS file that contains the SCE object to perform doublet detection on."
+  ),
+  make_option(
+    opt_str = c("--output_sce_file"),
+    type = "character",
+    default = "",
+    help = "Path to output RDS file to save SCE with detected doublets."
   ),
   make_option(
     opt_str = c("-t", "--threads"),
@@ -36,8 +42,11 @@ option_list <- list(
 opt <- parse_args(OptionParser(option_list = option_list))
 set.seed(opt$random_seed)
 
-# check SCE file
-stopifnot("`sce_file` is missing." = file.exists(opt$sce_file))
+# check SCE files
+stopifnot(
+  "`input_sce_file` is missing." = file.exists(opt$input_sce_file),
+  "`output_sce_file` should end in .rds" = stringr::str_ends(opt$output_sce_file, "\\.rds")
+)
 
 # set up multiprocessing params
 if (opt$threads > 1) {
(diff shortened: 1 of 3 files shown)


bin/detect_doublets.R
```diff
@@ -4,7 +4,7 @@
 #
 # This script reads in an RDS file containing an SCE and runs scDblFinder.
 # Classifications and doublet scores are stored in the SCE's colData slot.
-# The SCE is then exported.
+# The SCE is then exported to the same file as was inputted.
```
jashapiro, line 7: Let's not overwrite the file. We can give it a new name and change the output file.

modules/sce-processing.nf
```diff
@@ -129,6 +129,11 @@ process filter_sce {
       ${adt_present ? "--adt_barcode_file ${feature_barcode_file}":""} \
       --prob_compromised_cutoff ${params.prob_compromised_cutoff} \
       ${params.seed ? "--random_seed ${params.seed}" : ""}
+
+    detect_doublets.R \
+      --sce_file ${filtered_rds} \
+      ${params.seed ? "--random_seed ${params.seed}" : ""} \
+      --threads ${task.cpus}
```
jashapiro, line 136: Rather than overwriting the same SCE file, make the intermediate file have a fixed name. 
```suggestion
    detect_doublets.R \
      --input_sce_file filtered.rds \
      --output_sce_file ${filtered_rds} \
      ${params.seed ? "--random_seed ${params.seed}" : ""} \
      --threads ${task.cpus}
```

(I can't suggest the other edit, but it woudl be to change the `filter_sce.R` call to start:
```
filter_sce.R \
      --filtered_file filtered.rds \
```

jashapiro (commented):
I think doing this in the same process is fine; I just made a suggestion to keep input and output files separate for clarity.

One thing I would suggest as part of this though would be to add an option to `filter_sce.R` so you can write out an uncompressed RDS file to save a fair amount of processing time for the intermediate file read and write.

sjspielman:
This is working on my end and should be ready for a real review now!

bin/filter_sce.R
```diff
@@ -32,6 +32,12 @@ option_list <- list(
     default = FALSE,
     help = "flag to set the miQC's enforce_left_cutoff option to TRUE"
   ),
+  make_option(
+    opt_str = c("--no_sce_compression"),
+    action = "store_false",
+    default = TRUE,
+    help = "flag to turn off compression of the output SCE object"
+  ),
```
jashapiro, line 40: I appreciate that this is may be less prone to user error, but I would just pass along the compression flag and then check that it is valid (one of `c("none", "gz", "bz2", "xz")`)


```suggestion
  make_option(
    opt_str = c("--sce_compression"),
    type = "character",
    default = "bz2",
    help = "compression method for writing SCE as an RDS file."
  ),
```

bin/detect_doublets.R
```diff
@@ -19,10 +19,10 @@ option_list <- list(
     help = "Path to RDS file that contains the SCE object to perform doublet detection on."
   ),
   make_option(
-    opt_str = c("-o", "--output_sce_file"),
```
jashapiro, line 22: What do you have against short options? 😃

bin/filter_sce.R
sjspielman: > so is probably not worth the effort, given we are not likely to need the flexibility.

My sense is if we want to change compression in general, there's a fair bit of changes across the code base to implement as `bz2` is hardcoded all over the place. So, I think this is something to update to if/when we end up needing this flexibility

bin/detect_doublets.R
sjspielman: Huh, who knows! Let's bring it back!

PR #943, opened 2025-07-30
Update UMAPs to include consensus cell type annotations

### LINKED ISSUE #933: Update UMAPs to include consensus cell type annotations
We need to update the UMAP section to have a faceted UMAP showing the consensus cell type annotations in the main report. This should appear after submitter annotations. 

When doing this we also want to move the UMAPs for SingleR and CellAssign from the main report to the supplemental report.

allyhawkins:
Closes #933 

I added the UMAP for the consensus cell types to the main report and the supplemental report. Just like with the tables, I show the SingleR/CellAssign UMAPs in the main report if there are no consensus cell types. 

The only tricky thing here was that I didn't love the text we show if we are only showing consensus cell types in the main report (if we don't have submitter). I made a small change to just remove one of the lines describing multiple UMAPs if we are only showing consensus.

Here's the example reports again: 
[celltype_reports.zip](https://github.com/user-attachments/files/21517177/celltype_reports.zip)

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -448,16 +448,25 @@ if (length(levels(umap_df$cluster)) <= 8) {
 }
 ```
 
+
 ```{r, eval = has_umap && has_celltypes }
-knitr::asis_output(
+# only mention multiple annotations if we're actually showing multiple annotations
+if (has_consensus && !has_submitter && !is_supplemental) {
+  umap_text <- ""
+} else {
+  umap_text <- "For each cell typing method, we show a separate faceted UMAP."
+}
+
+
+knitr::asis_output(glue::glue(
   'Next, we show UMAPs colored by cell types.
-For each cell typing method, we show a separate faceted UMAP.
+  {umap_text}
 In each panel, cells that were assigned the given cell type label are colored, while all other cells are in grey.
 
 For legibility, only the seven most common cell types are shown.
 All other cell types are grouped together and labeled "All remaining cell types" (not to be confused with "Unknown cell type" which represents cells that could not be classified).
   '
-)
+))
 ```
 
 <!-- Now, UMAPs of cell types, where present -->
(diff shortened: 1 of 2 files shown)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -448,16 +448,25 @@ if (length(levels(umap_df$cluster)) <= 8) {
...
 
+
 ```{r, eval = has_umap && has_celltypes }
-knitr::asis_output(
+# only mention multiple annotations if we're actually showing multiple annotations
+if (has_consensus && !has_submitter && !is_supplemental) {
+  umap_text <- ""
+} else {
+  umap_text <- "For each cell typing method, we show a separate faceted UMAP."
+}
```
sjspielman, line 458: This makes sense to me

templates/qc_report/celltypes_qc.rmd
```diff
@@ -484,6 +493,28 @@ faceted_umap(
...
+
+```{r, eval=has_consensus && has_umap, message=FALSE, warning=FALSE, fig.width = consensus_dims[1], fig.height = consensus_dims[2], fig.align = "center"}
+# umap for cell assign annotations
+faceted_umap(
+  umap_df,
+  consensus_n_celltypes,
+  consensus_celltype_annotation_lumped,
+  point_size = umap_facet_point_size
+) +
+  ggtitle("UMAP colored by Consensus cell type annotations")
```
sjspielman, line 515: for consistency with submitter umap which is lower case
```suggestion
  ggtitle("UMAP colored by consensus cell type annotations")
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -484,6 +493,28 @@ faceted_umap(
...
+} else {
+  # set fake dims for evaluating next chunk
+  consensus_dims <- c(1, 1)
+}
+```
+
+
+```{r, eval=has_consensus && has_umap, message=FALSE, warning=FALSE, fig.width = consensus_dims[1], fig.height = consensus_dims[2], fig.align = "center"}
+# umap for cell assign annotations
+faceted_umap(
```
sjspielman, line 509: I realized looking at these UMAPs we seem to have a "bug" in the UMAP code here. See the order of the legend vs the order of the panels for faceted UMAPs where the legend is on the bottom (and not in a "panel nook") doesn't match cleanly? 
I'm pretty sure the fix here is to add `guides(color = guide_legend(byrow = TRUE))` to the `else` theming here:
https://github.com/AlexsLemonade/scpca-nf/blob/764622b7bc3f568f1360c831b1c8feccb7ad72b6/templates/qc_report/celltypes_qc.rmd#L215-L221

Can you try this out and see if it does the trick?

sjspielman (commented):
Looks good overall! While I was looking at UMAPs though, I caught some not-great legend behavior which should be fixable here.

PR #945, opened 2025-08-01
Add dot plot of marker gene expression for consensus cell types

### LINKED ISSUE #934: Dot plot of marker gene expression for consensus cell types to main report
We will want a dot plot showing the expression of marker genes for consensus cell types in the main section of the report. I think to do this we want to copy the marker gene table from `OpenScPCA-analysis` and save it to this repo like we do with the consensus cell type reference. Then we can read it in to the report as a parameter to create the dot plot.

allyhawkins:
Closes #934 

Here I'm adding in the dot plot to both the main and supplemental reports. As discussed in the outline, the dotplots appear after the UMAPs. 

There's a few reference files that we use to make these dot plots (as mentioned in https://github.com/AlexsLemonade/scpca-nf/issues/934#issuecomment-3141249391) that I am adding to the repo here. I also made all these files paramters to the acutal workflow. I placed the validation groups and marker gene tsvs in `references`. For the palette, I made a new folder, `templates/palettes` since these are specifically used in the report and I didn't think it made sense to add to `references`? 

Just a side note that when we update consensus cell types to consider `SCimilarity` annotations, these files may need to be updated, but that will happen first in `OpenScPCA-analysis` and then we will port the changes over to here. 

I then added an argument to the `sce_qc_report.R` script and parameters to the main and supplemental reports to pass in the three files. I decided to pass the file paths to the report rather than read them in as data frames, since we only need to read them in if consensus cell types are present. I think Nextflow will be okay with this, but I haven't been able to fully test since I need to wait on https://github.com/AlexsLemonade/scpcaTools/pull/310. 

The actual dot plot code is copied from how I made the plot in the [`ews-nf` report](https://github.com/AlexsLemonade/ews-nf/blob/0125c5f1088781b530cd08b11f041d28bb6016ba/templates/utils/plotting-functions.Rmd#L56). The only main thing to note is that I am showing every cell type regardless of how many cells are present, which I think is the right move. 

Here's a copy of the main and supplemental reports: 
[reports.zip](https://github.com/user-attachments/files/21553654/reports.zip)

Also I was testing the workflow and hit some issues with the doublet detection that I fixed. One of those is to change to using the `edge` tag for all the `scpcaTools` images.


sjspielman (approved):
LGTM!

PR #950, opened 2025-08-05
Update heatmaps to include consensus cell types 

### LINKED ISSUE #946: Add a heatmap comparing submitter annotations to consensus annotations
We would like to include a heatmap in the main report that directly compares submitter annotations (if present) to consensus annotations.

### LINKED ISSUE #947: Update heatmaps in supplemental to compare all assignments to consensus
We will want to update the comparison between annotations section in the supplemental report to include just one heatmap that compares all annotations to consensus annotations. This heatmap should be a stacked heatmap with consensus on the x axis and submitter, SingleR, CellAssign, and eventually SCimilarity on the y-axis.

allyhawkins:
Stacked on #945 
Closes #946 
Closes #947 

Here I'm updating the heatmaps in the main and supplemental reports. After thinking about it some more, I ended up making a few changes outside of what we talked about, so let me know if you disagree with any of these. 

- The main report now includes a heatmap comparing consensus annotations to submitter annotations if both of these are present. 
  - This heatmap is _only_ in the main report and not in the supplemental report and only shown if both submitter and consensus cell types are present. If submitter is present and consensus is not present then we don't show anything in the main report. 
- The supplemental report now includes two heatmaps. One comparing all annotation types to submitter, if present and then one just looking at all automated methods vs. consensus. I know originally we said to stack everything up vs. consensus but I think I like seeing how each of the methods compare to submitter if present. Then the second heatmap looks at automated methods that were used to generate consensus labels. What do we think? 

Because we now make heatmaps in both reports, I moved the heatmap functions to the `utils` file. 

Here's a copy of both reports if submitter is present: 
[reports.zip](https://github.com/user-attachments/files/21606275/reports.zip)

### DIFF
--- a/templates/qc_report/utils/celltype_functions.rmd
+++ b/templates/qc_report/utils/celltype_functions.rmd
@@ -254,7 +254,6 @@ create_single_heatmap <- function(
     col_fun = heatmap_col_fun,
     column_names_side = "bottom",
     ...) {
-  
   heat <- ComplexHeatmap::Heatmap(
     t(mat), # transpose because matrix rows are in common & we want a vertical arrangement
     col = col_fun,
@@ -271,7 +270,7 @@ create_single_heatmap <- function(
     column_title = column_title,
     column_title_gp = grid::gpar(fontsize = 10),
     column_names_side = column_names_side,
-    column_names_gp = grid::gpar(fontsize = labels_font_size),    ### passed in args
+    column_names_gp = grid::gpar(fontsize = labels_font_size), ### passed in args
     ...,
     ## Legend parameters
     heatmap_legend_param = list(
@@ -307,17 +306,16 @@ create_heatmap_list <- function(
   heatmap_list <- matrix_list |>
     purrr::imap(
       \(mat, name) {
-        
         # get the index of the list for mat
         idx <- which(names(matrix_list) == name)
-        
-        # if its the first one set the names to the top, otherwise bottom 
-        if(idx == 1){
+
+        # if its the first one set the names to the top, otherwise bottom
+        if (idx == 1) {
           column_names_side <- "top"
         } else {
           column_names_side <- "bottom"
         }
-        
+
         create_single_heatmap(
           mat,
           name,
@@ -327,7 +325,8 @@ create_heatmap_list <- function(
           col_fun,
           column_names_side,
           ...
-        )}
+        )
+      }
     ) |>
     # concatenate vertically into HeatmapList object
     purrr::reduce(ComplexHeatmap::`%v%`)


templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -437,131 +231,101 @@ if (!has_multiplex) {
...
+Note that due to different annotations references, these methods may use different names for similar cell types.
+")
 ```
 
-```{r, eval = has_submitter & (has_cellassign | has_singler)}
-# calculate matrices comparing to submitter
-jaccard_submitter_matrices <- available_celltypes |>
+```{r, eval = has_consensus}
+# having consensus celltypes implies that both SingleR and CellAssign exist
+# TODO: Update to handle adding in SCimilarity
```
sjspielman, line 298: also for future, `has_consensus` (used for chunk eval) won't imply that all 3 exist, just at least 2.

sjspielman (commented):
Overall, these changes look totally fine to me and I'm on board with the decisions you've made! 

That said with fresh eyes over here, I am wondering _more generally_ about our stacked heatmaps. I'm just not able to see the "full view" on my screen at once for the stacked heatmaps here, and with SCimilarity added into the mix that will only get worse! I'm trying to think of what plotting strategies might help to get around this. Some thoughts in no particular order, and these are not necessarily mutually exclusive. A bigger change here would be a separate PR too.
- change how we determine heatmap sizing to get it to be shorter, which may also involve text sizing changes
- add the x-axis labeling (the main set of cell types all else are being compared to) to the top of the heatmap also, so we can see it
- ...tabsets instead of stacking? one per comparison to the main group per tab? of course this ruins our ability to compare multiple annotations at once, but I'm struggling to do that anyways given how tall it all is
  - very possibly I'm just too excited about tabsets though 😉

sjspielman:
> I'm a little iffy about this though because I think the purpose of these plots is to be detailed and look across all the cells.

I echo this hesitation

> I don't love this idea because as you said it ruins the ability to look at multiple annotations at once. I like being able to look at the annotation on the x-axis and see what each method annotates them as.

Yes, I know the tabsets are bad I was grasping at straws 😭 

> I feel like we spent a lot of time optimizing this so you can read labels so I'm not sure we want to do this. I think we start with showing labels at the top and bottom of the x-axis.

I'm convinced, let's just add labels at the top too and leave it alone unless someone tells us otherwise.

allyhawkins:
@sjspielman I added labels to the top and bottom of the heatmaps. This should be ready for another look. 
[celltypes_supplemental_report.html.zip](https://github.com/user-attachments/files/21673174/celltypes_supplemental_report.html.zip)

PR #951, opened 2025-08-06
Move clustering plots to a separate section

### LINKED ISSUE #948: Move clustering plots to their own section in the supplemental report
> I wonder if we should move the comparison of cell types to clusters to only be in the supplemental report in their own section? This would include moving the UMAP of clusters and then having the stacked heatmap with a bunch of warnings that this is default clustering?
> 
> I think that's a good plan actually! Having clusters in their own section I think helps a lot to emphasize the the caveats, and keeping the main report focused on the more robust analysis makes sense to me. Either way at a minimum, adding a warning that clusters were not evaluated, and state the parameters explicitly. I see we mostly do that already, but we could add in nearest neighbors too (of course these spots will change altogether with this reorg, so just showing what we currently say):
> 
> https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/templates/qc_report/celltypes_supplemental_report.rmd#L451
> 
> https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/templates/qc_report/celltypes_qc.rmd#L388
>  

 _Originally posted by @sjspielman in [#910](https://github.com/AlexsLemonade/scpca-nf/issues/910#issuecomment-3150699916)_

allyhawkins:
Stacked on #950 
Closes #948 

Here I'm moving the clustering UMAP and heatmap to a separate `Unsupervised clustering` section to be shown in the supplemental report. This section appears after the cell label comparisons (see #950) and before the diagnostic plots. I included a statement warning users to interpret the clusters with caution. 

At first, I was skeptical about moving the UMAP with clusters out of the main report since people love to look at colorful UMAPs, BUT I don't think it's entirely helpful where it currently is in the main report. We include it within the cell type section as a UMAP before the faceted UMAPs. I think _if_ we include the UMAP in the main report, it should be its own section outside of the cell type section. Because we don't really trust these clusters anyways, I don't love having a whole section in the main report on it, but feel free to convince me otherwise. 

So all of that led me back to just having the one section in the supplemental report. What do we think? 

Here's a copy of both reports: 
[reports.zip](https://github.com/user-attachments/files/21634074/reports.zip)

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -341,14 +341,14 @@ jaccard_consensus_matrices |>
 knitr::asis_output(glue::glue("
 # Unsupervised clustering
 
-In this section, we look at unsupervised cluster assignments. 
-The first plot shows the cluster assignments on a UMAP and the second plot shows a heatmap comparing cluster assignments to cell type annotations. 
+In this section, we look at unsupervised cluster assignments.
+The first plot shows the cluster assignments on a UMAP and the second plot shows a heatmap comparing cluster assignments to cell type annotations.
 The heatmap is colored by the Jaccard similarity index as described in the `Cell label comparisons` section above.
 
 Clusters were calculated using the graph-based {metadata(processed_sce)$cluster_algorithm} algorithm with  {metadata(processed_sce)$cluster_weighting} weighting.
 
-**Caution**: Clusters are computed without performing any evaluation or optimization of clustering parameters. 
-These results should be interpreted with caution. 
+**Caution**: Clusters are computed without performing any evaluation or optimization of clustering parameters.
+These results should be interpreted with caution.
 "))
 ```
 
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -373,23 +373,11 @@ glue::glue("
 
 
 ```{r eval = has_umap && has_clusters, message=FALSE, warning=FALSE}
-clusters_plot <- plot_umap(
-  umap_df,
-  cluster,
-  "Cluster",
-  point_size = umap_point_size
-) +
-  ggtitle("UMAP colored by cluster identity")
-
-# Determine palette based on number of levels.
-# If we have <=8, we can use a CVD-friendly palette (they generally don't have more than 8 colors).
-#  Otherwise, we will use the default palette.
-if (length(levels(umap_df$cluster)) <= 8) {
-  clusters_plot +
-    scale_color_brewer(palette = "Set2")
-} else {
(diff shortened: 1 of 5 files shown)


sjspielman:
> Because we don't really trust these clusters anyways, I don't love having a whole section in the main report on it, but feel free to convince me otherwise.

I agree with this and won't be trying to convince you otherwise (as the originating issue #948 suggests!), but let's give @jashapiro a chance to play devil's advocate here too?

templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -336,6 +336,61 @@ jaccard_consensus_matrices |>
...
+knitr::asis_output(glue::glue("
+# Unsupervised clustering
+
+In this section, we look at unsupervised cluster assignments.
+The first plot shows the cluster assignments on a UMAP and the second plot shows a heatmap comparing cluster assignments to cell type annotations.
+The heatmap is colored by the Jaccard similarity index as described in the `Cell label comparisons` section above.
+
+Clusters were calculated using the graph-based {metadata(processed_sce)$cluster_algorithm} algorithm with  {metadata(processed_sce)$cluster_weighting} weighting.
+
+**Caution**: Clusters are computed without performing any evaluation or optimization of clustering parameters.
```
sjspielman, line 350: either way, 💯

allyhawkins:
@sjspielman I updated this PR based on our discussion in pre-planning. The clusters UMAP is now in the dimensionality reduction and appears before the total genes detected and faceted HVGs. When we add tabs to the report, I imagine each of these will be its own tab. 

The supplemental report now has a clustering section with both the UMAP and the comparison heatmaps. 

When I was doing this I noticed that we had a function `plot_umap` that we were only using to create the clusters umap. I updated that to be `plot_clusters_umap` and included all the code to produce the UMAP from the processed SCE rather than the `umap_df`. This is because we don't make `umap_df` unless we have cell types, and it includes all the cell type information as well as clusters and UMAPs. Since this is now being created regardless of cell types, it made sense to just pull it from the `processed_sce` directly, like we do for the other UMAPs. 

Here's a new copy of the reports: 
[reports.zip](https://github.com/user-attachments/files/21722444/reports.zip)

allyhawkins:
@sjspielman I updated the text in the points you noted. I did not update the function though since I don't think it's worth the effort as mentioned in https://github.com/AlexsLemonade/scpca-nf/pull/951#discussion_r2270348799. This should be ready for another look.

sjspielman (approved):
LGTM!

allyhawkins:
Noting that I cherry-picked commits from #957 that @jashapiro used to fix the GHA since those changes are going into a different branch.

PR #954, opened 2025-08-08
Make 10xflex branch stricter and merge in `development` changes

jashapiro:
This PR is merging the changes in https://github.com/AlexsLemonade/scpca-nf/pull/953 as well as some of the other changes in the development branch into the 10xFlex support branch. 

We probably should be merging in `development` a bit more often just to avoid big PRs like this (or just wait for one big merge at the end), but for now here we are... With some of the structural changes, I think it is worth doing now.


allyhawkins (commented):
This is fine, but just noting that there's a small bug in the report that's included as part of #951 right now. I want to wait until that gets in to update this and merge it so that we don't introduce a temporary bug into the 10x-flex-support branch while Tony is using it to run samples.

allyhawkins (approved):
LGTM

PR #962, opened 2025-08-13
Add tabs and DT to the main report

allyhawkins:
Part 1 of #371 

Here I'm making a number of small changes to the main report to make things prettier. Mostly, I added tabs where I thought they helped de-clutter things without hiding too much and converted some tables to use `DT::datatable()`. 

Here's a zip file with three different versions of the main report - a regular single cell library with submitter annotations, a library with CITE-seq, and a multiplexed library: 
[reports.zip](https://github.com/user-attachments/files/21761997/reports.zip)

Here's an overview of the changes that I made: 

- I added tabs in places where we had related tables or plots. I find this really helpful for viewing the cell type tables and UMAPs!
  - This includes a tab with the QC plots (knee plot, miQC etc) and I was a little unsure if we wanted to do this but in general I think I like the tabs here? Did I add too many tabs? 
- To add all the tabs I did have to update some headers. While I was here I noticed we were all over the place with our capitalization. So I updated anything that was part of a tab or wasn't header 1 to now be sentence case instead. 
- I also moved the text so that it's consistently above plots rather than below. 
- I replaced the cell type annotation tables with `DT::datatable()` so these tables are now interactive. I also did this for the ADT and HTO statistics, where we show all the ADTs/HTOs and their mean percentage. 
  - I did not update any of the basic summary tables since I think DT would be too much for those tables? If the contrasting styles are too much between the tables I can make a simplified DT option that we can use for those tables, but generally I don't think we need to do that? 

I started with just making these changes in the main report and once we have this updated I'll go through the supplemental report. Those changes should be pretty minor since I included the tables and UMAPs for cell types that appear in the main report here.

### DIFF
--- a/components/dictionary.txt
+++ b/components/dictionary.txt
@@ -47,7 +47,6 @@ miQC
 Nextflow
 Nextflow's
 nf
-Oligos
 oligos
 PanglaoDB
 parallelizing
--- a/components/dictionary.txt
+++ b/components/dictionary.txt
@@ -48,6 +48,7 @@ Nextflow
 Nextflow's
 nf
 Oligos
+oligos
 PanglaoDB
 parallelizing
 pixi
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -1,4 +1,4 @@
-# Cell type Annotation Summary
+# Cell type annotation summary
 
 <!--
 This file is meant to be run as a child report within either `main_qc_report.rmd` or `celltypes_supplemental_report.rmd`. 
@@ -244,8 +244,14 @@ glue::glue("
 ")
 ```
 
+```{r}
+if(has_consensus & !has_submitter & !is_supplemental){
+  knitr::asis_output('## Statistics')
+} else {
+  knitr::asis_output('## Statistics {.tabset}')
+}
+```
 
-## Statistics
 
 ```{r, warning = FALSE}
 # Create data frame of cell types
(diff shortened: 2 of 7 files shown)


components/dictionary.txt
```diff
@@ -48,6 +48,7 @@ Nextflow
 Nextflow's
 nf
 Oligos
+oligos
```
sjspielman, line 51: I think we only need one of these, and since you added the lowercase I guess we don't need the uppercase anymore? Much like removing a ggplot legend, I am entirely incapable apparently of remembering which one is needed to cover the other 😑

sjspielman (commented):
> This includes a tab with the QC plots (knee plot, miQC etc) and I was a little unsure if we wanted to do this but in general I think I like the tabs here? Did I add too many tabs?

I am _not_ the person who's going to tell you no, I love the tabs!! I think the report is overall a lot cleaner now. The only reason to _not_ include the tabs is if having plots side-by-side is critical for interpretation. I think it can help but the tabbing doesn't really _harm_ interpretation so I like it! I also think it's _good_ to use DT only for tables with potential to be too long, since DT tables themselves are a little visually noisy. For small tables the bells and whistles are distracting imo. 

One question I have though, can you render a version of this report that has _no consensus_? In this case the singler/cellassign tables would be in the main report too, so just want to see how that looks (unless some of that logic is forthcoming in next PRs).

allyhawkins:
> One question I have though, can you render a version of this report that has no consensus? In this case the singler/cellassign tables would be in the main report too, so just want to see how that looks (unless some of that logic is forthcoming in next PRs).

Yes here you go! 
[main_qc_report.html.zip](https://github.com/user-attachments/files/21776334/main_qc_report.html.zip)

templates/qc_report/celltypes_qc.rmd
```diff
@@ -318,7 +324,7 @@ knitr::asis_output('
 In this table, cells labeled "Submitter-excluded" are those for which submitters did not provide an annotation.
 ')
 create_celltype_n_table(celltype_df, submitter_celltype_annotation) |>
-  format_celltype_n_table()
+  DT::datatable(rownames = FALSE)
```
jashapiro, line 327: Overall, I think this looks good. The tabs are quite nice to clean things up!

One question that is hard to see here: Do we still use the `format_celltype_n_table()` function at all? If not, it should probably be removed?

Or we might want to make a different function with more settings to customize the DT display like I did here:  https://github.com/AlexsLemonade/scpca-nf/blob/2b621752f362ad77ca69c6cd07cad56f7a7eca60/scripts/compare-metrics/compare-metrics-template.rmd#L53-L91

There I use scrolling tables to reduce some of the excess chrome, which I liked (though it can occasionally hijack the scroll wheel to scroll the table when you mean to scroll the page). This also means you don't need any custom logic to hide elements when the table only has a few rows.

allyhawkins:
@jashapiro I updated this to remove that extra function and instead include a formatting function for any of the datatables in `report_functions.Rmd`. This includes scrolling support and I do think looks better! 

Here's an updated version of the CITE-seq report with the scrolling: 
[main_qc_report.html.zip](https://github.com/user-attachments/files/21780098/main_qc_report.html.zip)

sjspielman (approved):
🚀

PR #963, opened 2025-08-14
Create a separate cell browser workflow

jashapiro:
Because cell browser will usually need to be run on all projects and samples at once, I moved it to a separate workflow. I also rearranged the order of the workflow itself to reflect my improved understanding of how the different parts will interact. 

I'm also adding here some template config files that are used in the workflow, as well as additional parameters that will be used when building the site.


jashapiro:
(I merged in the changes to the report from development as well.)

allyhawkins (approved):
LGTM

PR #968, opened 2025-08-14
Complete initial Cell Browser site build

### LINKED ISSUE #956: Add python script for creating site and project config and abstract files
The site config and project config files will be created based on information in the project metadata file. We will want a python script that reads that tsv file (using the standard library `csv` module) and creates the required output files, possibly using template files for additional content

### LINKED ISSUE #959: Add import of h5ad files
Each library will need to be imported separately to create the data formats for input.  This should be done with the `cbImportScanpy` CLI, but with manually created (by our own script) `desc.conf` and `cellbrowser.conf` files.

### LINKED ISSUE #960: Build full cellbrowser site
After all data is in place, we will build the full site with `cbBuild`

jashapiro:
closes #956 
closes #959
closes #960  

⚠️ stacked on #963 

This PR completes an initial version of the Cell Browser site generation workflow.

The parts are these:
- A `cellbrowser_config.py` script which creates the required config files for individual libraries and projects as a whole. In many ways, this could be two totally separate scripts, but since both would be pretty short, I just kept them together with a `--conf_type` argument to say whether to produce library conf files or project conf files, which are handled by separate functions. 
- Processes for running the above script, followed by the appropriate cell browser commands
  - for the individual libraries, I use `cbImportScanpy`, which also copies over the h5ad file (with the expectation of allowing downloads), so I remove that copy to save space and runtime. (couldn't find an option to do that)
  - The project process takes all of the created library directories from above, then creates all of the individual project config files, and moves the library files to their respective project directories. After this, it runs `cbBuild` to create the site. 
  - Only the site directory is published as output

If you want to view the test site, you can do so by downloading from `s3://nextflow-ccdl-results/scpca/processed/cellbrowser` and then serving the directory with a local server (installing on the fly with `uvx` below:

```
aws s3 sync s3://nextflow-ccdl-results/scpca/processed/cellbrowser cellbrowser
uvx --from cellbrowser cbUpgrade -o cb_site -p 8880
```

Some changes from previous structure and thoughts to the future:
- the output directory is now just a name rather than a full path, and is created under `outdir`. I think I could change this back, but it seemed the simplest this way, so I could export the whole directory. 
- At the moment, each RNA library is presented separately, which I think is what we want, but I label them with both the sample and library ids to make it easier for users.
- The default coloring is by cluster, which is set by a param. I think we ought to be able to change that to the consensus cell type, and we might be able to make the list of enums such that there are multiple options (all cell type columns + cluster?), but I wanted to get a working version first, and we should be able to revisit that in a future PR as needed.
- If you note other parts of the site that look like they might be customizable, let me know, and we can file issues to try to tweak it!


jashapiro:
One other future thought: Right now I am building the whole site at once, which may be slow. I _think_ it should be possible split the current version to do one project at a time, and then a final rebuild with the full site.

PR #972, opened 2025-08-15
Fix some tabs in the supplemental cell type report

### LINKED ISSUE #371: Update QC report tables

allyhawkins:
Closes #371 

I went to add tabs and such to the supplemental cell type report, but there really aren't any places where I would do that outside of the tables and the UMAPs which already have tabs. I did notice though that if you have all 4 annotations (submitter, consensus, cellassign, and singleR) that the tabs don't fit on one line and it looks bad. So I added some css to make sure that the tabs are always on one line and I think it looks much better. This should only be a problem when you have all the annotations in the tables section which only appears in the supplemental report, but just in case we ever expand on tabs (such as adding more annotations) I added the code to the main report too. 

I also shortened the tab headers in the tables section to match the headers in the UMAP section. 

Here's a zip file with lots of different versions of the reports for reference: 
[reports.zip](https://github.com/user-attachments/files/21799194/reports.zip)

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -302,8 +302,7 @@ if (length(unclassified_methods) > 0) {
 
 
 ```{r, eval = has_submitter}
-knitr::asis_output('
-### Submitter-provided cell type annotations\n
+knitr::asis_output('### Submitter-provided annotations
 
 In this table, cells labeled "Submitter-excluded" are those for which submitters did not provide an annotation.
 ')
@@ -312,17 +311,18 @@ create_celltype_n_table(celltype_df, submitter_celltype_annotation) |>
 ```
 
 ```{r, eval = has_consensus}
-knitr::asis_output('### Consensus cell type annotations\n
+knitr::asis_output('### Consensus cell type annotations
 
 In this table, cells labeled "Unknown cell type" are those whose `SingleR` and `CellAssign` annotations did not agree.
 In the processed result files, these cells are labeled `Unknown`.
 ')
+
 create_celltype_n_table(celltype_df, consensus_celltype_annotation) |>
   format_datatable()
 ```
 
 ```{r, eval = has_singler && (!has_consensus || is_supplemental)}
-knitr::asis_output('### `SingleR` cell type annotations\n
+knitr::asis_output('### `SingleR` annotations
 
 In this table, cells labeled "Unknown cell type" are those which `SingleR` pruned due to low-quality assignments.
 In the processed result files, these cells are labeled `NA`.
@@ -332,7 +332,7 @@ create_celltype_n_table(celltype_df, singler_celltype_annotation) |>
 ```
 
 ```{r, eval = has_cellassign && (!has_consensus || is_supplemental)}
-knitr::asis_output('### `CellAssign` cell type annotations\n
+knitr::asis_output('### `CellAssign` annotations
 
 In this table, cells labeled "Unknown cell type" are those which `CellAssign` could not confidently assign to a label in the reference list.
 In the processed result files, these cells are labeled `"other"`.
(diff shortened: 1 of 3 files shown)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -312,17 +311,18 @@ create_celltype_n_table(celltype_df, submitter_celltype_annotation) |>
...
 In the processed result files, these cells are labeled `Unknown`.
 ')
+
 create_celltype_n_table(celltype_df, consensus_celltype_annotation) |>
   format_datatable()
 ```
 
 ```{r, eval = has_singler && (!has_consensus || is_supplemental)}
-knitr::asis_output('### `SingleR` cell type annotations\n
+knitr::asis_output('### `SingleR` annotations
```
sjspielman, line 325: For the UMAP version of these labels, we don't have `backticks`. We should add them in to be consistent.

One question I have is how 5 tabs will look once we add in SCimilarity, but we'll cross that bridge later! We'll also cross the submitter + openscpca (or 2x openscpa...) bridge even more later, and with any luck never!

sjspielman (commented):
Looks good to me (I love the tabs so much!!) and the CSS works clearly works, though I'm not really able to review it more in depth beyond it makes the tabs look nice - I'd tag an engineer if you want expert eyes on that code!

While I was looking over the report, I noticed one additional tiny additional discrepancy we might as well address while we're here. When we refer to the `Unknown cell type` label, we _don't_ refer to it with backticks in the cell type table descriptions but we do use backticks (I think?) everywhere else. We might want to make that consistent and backtick all around.

Also, the report you included `no_consensus_celltypes` does actually have consensus, but I have faith the actual version of that looks fine :)

allyhawkins:
I updated this so that we use "Unknown cell type" in all descriptions since that's what we use the majority of the time and I added backticks to SingleR and CellAssign in the tab labels. I also simplified the CSS so I think we're good to go there. 

This should be ready for another look @sjspielman!

Also whoops... sorry about that with the reports!

templates/qc_report/celltypes_qc.rmd
```diff
@@ -479,8 +479,8 @@ We expect that each cell type group should show high expression of marker genes
 
 Note that dots are only shown if the gene has mean expression greater than 0 and is expressed in at least 10% of the cells in the given cell type.
 
-If all consensus cell types are labeled as `Unknown cell type`, then no dot plot will be shown.
-  ")
+If all consensus cell types are labeled as "Unknown cell type", then no dot plot will be shown.
```
sjspielman, line 482: 👍

templates/qc_report/main_qc_report.rmd
```diff
@@ -24,6 +24,18 @@ output:
...
+ <!--Make sure tabs always fit on one line--> 
+<style>
+.nav-tabs {
+  display: flex !important;
+  flex-wrap: nowrap !important;
+  flex: 1 1 auto !important;
+  text-align: center;
+  white-space: normal !important;
+}
+</style>
```
sjspielman, line 36: I'm not sure the right syntax _style_ for this, but I might tab in `.nav-tabs`? Same for other css block in the other report.

sjspielman (approved):
LGTM, also found one more spot to update to quotes but don't need to see again!

https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/templates/qc_report/celltypes_supplemental_report.rmd#L619

PR #975, opened 2025-08-19
add 10Xv3_5prime chemistry

### LINKED ISSUE #958: 10Xv3_5prime chemistry
I found that the current workflow does not handle `10Xv3_5prime chemistry`. In the `alevin-fry generate-permit-list` step, it only generated a few hundred cells. When I faked the run_metadata to use 10Xv3.1, the number increased to ~2,000 cells, and forcing --expected-ori `rc` also gave similar results. However, my sample should have ~14,000 cells if processed with Cell Ranger.

For 10Xv2_5prime using an old barcode whitelist, the workflow also failed to detect many cells.

Can you modify the workflow to also handle `10Xv3_5prime chemistry`?

**Additional context**

jashapiro:
closes #958

Here I am adding support for 10Xv3 5prime chemistry, as well as 10Xv4. As these both have the same geometry as the  existing 10Xv3 chemistries, I was able to accomplish this by adding the relevant barcode files to our S3 references bucket and updating the dictionaries to include support for those. 

Since the files were distributed within Cell Ranger as `.gz` files, I left them that way, and added a little block to decompress them as needed by alevin-fry. 

During testing, I ran into a problem of my own making where I forgot to include the correct references in my run file, which made for a confusing error because we were filtering based on the reference existing. So I split that out a bit using a `.branch` so we can warn for missing/mismatched references. 

I tested this with one of the example datasets from 10x, and obtained what appeared to be pretty comparable results (though the miQC filtering looks not so great!) I am attaching our QC and the 10X QC docs for comparison: 
[QC_docs.zip](https://github.com/user-attachments/files/21859715/QC_docs.zip)

This PR will ultimately need to be followed up with a docs update.


allyhawkins (approved):
LGTM 🚀

PR #984, opened 2025-08-26
Create a Docker image for Cell Browser

jashapiro:
This PR is to create a docker image to use with the UCSC Cell Browser actions that we are adding. 

To make this I defined the environment with pixi, and used the `pixi` base image to perform the installation of the packages from the lock file. Following the suggestion at https://github.com/prefix-dev/pixi-docker, I used a multi-step build so that the cache and `pixi` executable would not actually be included in the production image. 

I also created a GHA to build and push the image. I am submitting this as a draft until I confirm that the build works as expected. (It does, at least the build part; we'll test the push when approved)

I did include support in the `pixi` environment for `osx-arm64`, but that is probably not really necessary, and it does add some complexity since the `procps-ng` package is not available on `osx-arm64`. If you think it makes sense to just include linux support in the `pixi.toml` file, let me know (it builds just fine either way).


allyhawkins (approved):
This looks good to me, just some minor documentation udpates.

PR #987, opened 2025-08-27
Set up for inferCNV and count normal cells

### LINKED ISSUE #983: Lay groundwork for inferCNV module
_Broken out of https://github.com/AlexsLemonade/scpca-nf/issues/965_

The goal of this issue is to begin laying the groundwork to run `inferCNV` in the workflow. This issue tracks the following additions:

- Add workflow parameters, including associated docs in `nextflow_schema.json` this module will need:
	- file paths; note that the first two should use URLs to `OpenScPCA-analysis` for now since these files are subject to change with incoming `SCimilarity`-related changes. We can circle back and update later to add these files into our references directly.
		- TSV mapping diagnoses to validation groups
		- TSV mapping validation groups to cell types
		- gene order file
	- threshold for how many normal cells are needed to run `inferCNV`; we plan to use 100
	- booleans for whether we should run `inferCNV`
		- `repeat_infercnv` and `perform_infercnv` , both set to  `false` by default
- Update `add_celltypes_to_sce.R` to export an environment variable of how many normal reference cells are present, and then save this value to `meta`
	- See some more implementation details for this here: https://github.com/AlexsLemonade/scpca-nf/issues/965#issuecomment-3221429029

sjspielman:
Closes #983 

This PR gets going for inferCNV in two main ways:

- Adds workflow parameters we'll need (and need already, even!) and documents them in the schema json
  - The diagnosis TSVs are not yet final because of SCimilarity changes, so those are URLs for now until we can port them into this repo
- Adds functionality to count the number of normal reference cells. I left a whole bunch of breadcrumbs in the code for review! Overview:
  - There are three new TSV files needed to do the counting (mapping files: validation groups -> consensus cell types, broad diagnosis -> validation groups, broad -> sample diagnosis). (This makes sense of course, as the number of the counting shall be 3 🐇)
  - The script will only do the counting if all three of these files exist, and if consensus cell types exist. 
    - These TSVs are only passed into the script _if_ both the diagnosis TSVs exist. I pass these files into the process by checking if they exist and change them to `"NO_FILE"` if they are falsey to enable this check.
    - Note that the validation file is guaranteed to exist - we have it as a parameter check at the beginning of `main`, so we don't need to include it in that check. 
  - The script is step up to _always_ export a normal count file: It will either be empty (counting not performed), or it will have a value (counting performed). I chose to always export something rather than having nextflow check if the file exists (e.g. in a scenario where only exported if counting performed), to avoid more `if`s in nextflow. But, one could do it that way too I think!
    - I store _either_ this value in `meta`, or the value false. I don't love this choice, as I elaborate on in breadcrumb review comments in the code! In `stub`, I just pop in a larger-than-default value to get that logic tested


sjspielman:
Just kidding, this approach to capture the number of cells doesn't work because it doesn't like the groovy code in the script block after the bash code and won't actually read the file to store the value in `meta`. I am playing around with exporting the value into `stdout` in my free time, but this may just wait as a draft, tbd!

modules/classify-celltypes.nf
```diff
@@ -130,19 +135,26 @@ process add_celltypes_to_sce {
...
+      ${check_normal_cells ? "--diagnosis_celltype_ref ${diagnosis_celltypes_file}" : ''} \
+      ${check_normal_cells ? "--diagnosis_map_ref ${diagnosis_map_file}" : ''} \
+      --normal_cells_file "normal_cell_count.txt"
     """
+    // this sets normal_cell_count to the number (if calculated in the script) or false (not calculated in the script, empty file spit out instead).
+    // i don't love this difference in types? so..should we just set up a logical here for whether to run infercnv and branch on that later, since we're getting false out of this anyways?
+    // or, what could we put in this field besides false if the info is NOT calculated? just not have a field?
+    normal_cell_count = file("normal_cell_count.txt").text
+    meta['normal_cell_count'] = normal_cell_count ? x.toInteger() >= params.infercnv_min_normal_cells : false
   stub:
```
allyhawkins, line 149: I'm very briefly looking at this, but I think I would modify this to output the file as part of the output of this process and then add to the `meta` in the main workflow block rather than here.

modules/classify-celltypes.nf
```diff
@@ -112,14 +112,19 @@ process add_celltypes_to_sce {
...
-    tuple val(meta), path(annotated_rds)
+    tuple val(meta), path(annotated_rds), env("NORMAL_CELL_COUNT")
   script:
     annotated_rds = "${meta.library_id}_processed_annotated.rds"
     singler_present = "${singler_dir.name}" != "NO_FILE"
     singler_results = "${singler_dir}/singler_results.rds"
     cellassign_present = "${cellassign_dir.name}" != "NO_FILE"
     cellassign_predictions = "${cellassign_dir}/cellassign_predictions.tsv"
+    // we only check for normal cells if we have both diagnosis mapping files
+    check_normal_cells = "${diagnosis_groups_file.name}" != "NO_FILE" &&  "${diagnosis_celltypes_file.name}" != "NO_FILE"
```
jashapiro, line 127: If you change to passing an empty list for missing files, this will need to change, possibly to just (untested):
```suggestion
   count_normal_cells = diagnosis_groups_file  &&  diagnosis_celltypes_file
```

(I also updated the name to `count` since that is what we are doing)

modules/classify-celltypes.nf
```diff
@@ -130,19 +135,26 @@ process add_celltypes_to_sce {
       ${cellassign_present ? "--cellassign_ref_file ${meta.cellassign_reference_file}" : ''} \
       ${cellassign_present ? "--celltype_ref_metafile ${celltype_ref_metadata}" : ''} \
       ${cellassign_present ? "--panglao_ontology_ref ${panglao_ref_file}" : ''} \
-      --consensus_celltype_ref ${consensus_ref_file}
-
-
+      --consensus_celltype_ref ${consensus_ref_file} \
+      ${check_normal_cells ? "--consensus_validation_ref ${validation_ref_file}" : ''} \
+      ${check_normal_cells ? "--diagnosis_celltype_ref ${diagnosis_celltypes_file}" : ''} \
+      ${check_normal_cells ? "--diagnosis_groups_ref ${diagnosis_groups_file}" : ''} \
```
jashapiro, line 141: I'm just looking at all of these repeated tests and wondering if we can simplify (throughout) to a single `if`

```suggestion
      ${if(count_normal_cells)
      """\
      --consensus_validation_ref ${validation_ref_file} \
      --diagnosis_celltype_ref ${diagnosis_celltypes_file} \
      --diagnosis_groups_ref ${diagnosis_groups_file} \
      """}
```

main.nf
```diff
@@ -178,6 +178,7 @@ workflow {
         t2g_bulk_path: params.ref_rootdir + "/" + sample_refs["t2g_bulk_path"],
         cellranger_index: params.ref_rootdir + "/" + sample_refs["cellranger_index"],
         star_index: params.ref_rootdir + "/" + sample_refs["star_index"],
+        gene_order_file: params.ref_rootdir + "/" + sample_refs["gene_order_file"],
```
jashapiro, line 181: `gene_order_file` will need to be added to `test/stub-refs.json` along with an appropriate stub file.

bin/add_celltypes_to_sce.R
```diff
@@ -67,6 +70,31 @@ option_list <- list(
...
+  make_option(
+    opt_str = c("--consensus_validation_ref"),
+    type = "character",
+    help = "Path to file mapping consensus validation groups to consensus labels for counting normal reference cells intended as input to inferCNV",
+    default = ""
+  ),
+  make_option(
+    opt_str = c("--diagnosis_celltype_ref"),
+    type = "character",
+    help = "Path to file mapping broad diagnoses to consensus validation groups for counting normal reference cells intended as input to inferCNV",
```
jashapiro, line 83: I think we should say here what the expected format and columns are for these files. (I don't think you need to say that they are for inferCNV, though that is fine)

jashapiro (commented):
This looks like a good start, but there are a number of things that I expect are going to cause some trouble down the line, and I had a few questions as well. 

The biggest thing is our handling of missing files, which is a complex problem that the Nextflow dev teams has had plans to handle for a while but so far has not gotten to, so everything is kind of a hack. The biggest problem that we have run into in the past is that you can't pass two of the same file as inputs, so the `NO_FILE` solution only works for a single potentially missing file. We could have a whole set of empty files, but in one comment someone proposed passing an empty array in place of a `file` object, so we could test that (I have not tested it myself). 

My one other nagging thought is all of the references to "normal" cells, which I would love to see a better name for. It just makes me squeamish to know that the counts we are getting do not include all normal cells, so I don't want people to infer that. Maybe "reference" cells? or "cnv_reference" for more specificity?

bin/add_celltypes_to_sce.R
```diff
@@ -320,7 +349,53 @@ if (has_singler && has_cellassign) {
...
+      # get the consensus cell types
+      dplyr::left_join(consensus_validation_df, by = c("celltype_groups" = "validation_group_annotation")) |>
+      dplyr::pull(consensus_annotation) |>
+      unique()
+
+    normal_cells <- sum(celltype_df$consensus_annotation %in% reference_celltypes)
+  }
+} else {
+  # set normal cells to ""; we don't run inferCNV if there are no consensus cell types
+  normal_cells <- ""
```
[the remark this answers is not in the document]
sjspielman: Ah no, this is just in the wrong spot; I meant to have this on the other end of `if (has_singler && has_cellassign) {`.

modules/classify-celltypes.nf
sjspielman: Can you comment more on this syntax? Your suggestion leads to:

```
    """
    add_celltypes_to_sce.R \
      --input_sce_file ${processed_rds} \
      --output_sce_file ${annotated_rds} \
      ${singler_present ? "--singler_results  ${singler_results}" : ''} \
      ${singler_present ? "--singler_model_file ${meta.singler_model_file}" : ''} \
      ${cellassign_present ? "--cellassign_predictions  ${cellassign_predictions}" : ''} \
      ${cellassign_present ? "--cellassign_ref_file ${meta.cellassign_reference_file}" : ''} \
      ${cellassign_present ? "--celltype_ref_metafile ${celltype_ref_metadata}" : ''} \
      ${cellassign_present ? "--panglao_ontology_ref ${panglao_ref_file}" : ''} \
      --consensus_celltype_ref ${consensus_ref_file} \
      ${if(count_normal_cells)
      """\
      --consensus_validation_ref ${validation_ref_file} \
      --diagnosis_celltype_ref ${diagnosis_celltypes_file} \
      --diagnosis_groups_ref ${diagnosis_groups_file} \
      """}
      --normal_cells_file "normal_cells.txt" \
  """
```

Is this for sure what you meant? The `"""` nesting is throwing me off a bit

modules/classify-celltypes.nf
```diff
@@ -277,18 +289,32 @@ workflow annotate_celltypes {
...
+    // outputs [meta, annotated processed rds, NORMAL_CELL_COUNT]
     add_celltypes_to_sce(
       assignment_input_ch.add_celltypes,
       file(params.celltype_ref_metadata), // file with CellAssign reference organs
       file(params.panglao_ref_file), // used for assigning ontology IDs for CellAssign results
-      file(params.consensus_ref_file) // used for assigning consensus cell types if both SingleR and CellAssign are used
+      file(params.consensus_ref_file), // used for assigning consensus cell types if both SingleR and CellAssign are used
+      file(params.validation_groups_file),  // maps consensus cell types to cell type groups, for counting normal reference cells
+      file(params.diagnosis_celltypes_file ?: empty_file, checkIfExists: true), // maps broad diagnoses to cell type groups, for counting normal reference cells
+      file(params.diagnosis_groups_file ?: empty_file, checkIfExists: true) // maps sample diagnoses to broad diagnoses, for counting normal reference cells
```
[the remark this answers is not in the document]
sjspielman: I'm wondering too if `stageAs()` could help here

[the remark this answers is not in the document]
jashapiro: > I'm wondering too if `stageAs()` could help here

I don't think so, as that will give the same name regardless of content. And I'm not sure it will let you stage the same file twice anyway.

sjspielman:
Noting I'm still responding to other reviews (giving this comment a bump! https://github.com/AlexsLemonade/scpca-nf/pull/987#discussion_r2319359466 :)) 

I was feeling very curious so I ended up testing if `stageAs` could work, and I believe it works fine! To determine whether to provide these files to the R script, I check their size instead of their name.

To test the code, I temporarily added the normal cell count into the meta export to check it, and instead of an empty string, I'm using the string "Not calculated" if not calculated. 

At commit `a53c50a`where I provided both diagnosis files, we got the expected correct count in the meta:
Run config (can you see this link?): https://cloud.seqera.io/user/stephanie-spielman/watch/3k1n2rnTLLliml/v2/configuration
[a53c50a-SCPCL000001_metadata.json](https://github.com/user-attachments/files/22139433/a53c50a-SCPCL000001_metadata.json)

At commit `8751052` where I did not provide either file, we got Not calculated:
Run config: https://cloud.seqera.io/user/stephanie-spielman/watch/1fLFYzkFOwB7gD/v2/configuration
[8751052-SCPCL000001_metadata.json](https://github.com/user-attachments/files/22139425/8751052-SCPCL000001_metadata.json)

At the same commit `8751052` I tested only providing one of the diagnosis files, and again it successfully gives us "Not calculated."
Run config: https://cloud.seqera.io/user/stephanie-spielman/watch/1fLFYzkFOwB7gD/v2/configuration

If you feel really strongly about empty lists instead I can test that instead, but this seems to be a fine approach!

modules/classify-celltypes.nf
```diff
@@ -112,14 +112,19 @@ process add_celltypes_to_sce {
     path(celltype_ref_metadata) // TSV file of references metadata needed for CellAssign only
     path(panglao_ref_file) // used for assigning ontology IDs for CellAssign results
     path(consensus_ref_file) // used for assigning consensus cell types if both SingleR and CellAssign are used
+    path(validation_ref_file) // maps consensus cell types to cell type groups, for counting normal reference cells
+    path(diagnosis_celltypes_file, stageAs: 'diagnosis_celltypes.txt') // maps broad diagnoses to cell type groups, for counting normal reference cells
+    path(diagnosis_groups_file, stageAs: 'diagnosis_groups.txt') // maps broad diagnoses to cell type groups, for counting normal reference cells
```
jashapiro, line 117: Glad to know that this works, even if I didn't expect it! I realized recently that `stageAs:` can be substituted with `name:`, which seems to be the current preference. In fact, you can simply put the file name as the first argument, but in your case having the variable name is still useful, so I would leave it as you have it.
```suggestion
    path(diagnosis_celltypes_file, name: 'diagnosis_celltypes.txt') // maps broad diagnoses to cell type groups, for counting normal reference cells
    path(diagnosis_groups_file, name: 'diagnosis_groups.txt') // maps broad diagnoses to cell type groups, for counting normal reference cells
```

modules/classify-celltypes.nf
```diff
@@ -130,19 +135,26 @@ process add_celltypes_to_sce {
       ${cellassign_present ? "--cellassign_ref_file ${meta.cellassign_reference_file}" : ''} \
       ${cellassign_present ? "--celltype_ref_metafile ${celltype_ref_metadata}" : ''} \
       ${cellassign_present ? "--panglao_ontology_ref ${panglao_ref_file}" : ''} \
-      --consensus_celltype_ref ${consensus_ref_file}
-
-
+      --consensus_celltype_ref ${consensus_ref_file} \
+      ${count_normal_cells ? "--consensus_validation_ref ${validation_ref_file}" : ''} \
+      ${count_normal_cells ? "--diagnosis_celltype_ref ${diagnosis_celltypes_file}" : ''} \
+      ${count_normal_cells ? "--diagnosis_groups_ref ${diagnosis_groups_file}" : ''} \
```
jashapiro, line 141: Now that you have named files, you can remove the variables here: 
```suggestion
      ${count_normal_cells ? "--consensus_validation_ref ${validation_ref_file}" : ''} \
      ${count_normal_cells ? '--diagnosis_celltype_ref diagnosis_celltypes.txt' : ''} \
      ${count_normal_cells ? '--diagnosis_groups_ref diagnosis_groups.txt' : ''} \
```

jashapiro (commented):
Given that named staging works, that seems fine to me. I just noted that once you have a fixed name you can use it directly!

Other comments include using `NA` for missing data and confirming that I did mean to nest `"""`, but I don't feel too strongly about it.

sjspielman:
We're cooking!

- normal -> reference naming update. I did this throughout, because consistency is better for future stephanie and therefore future josh, the reviewer of future stephanie's code
- `stageAs` -> `name`, and use said names
- I now check for empty files in the R script instead of nextflow, much cleaner! Other R script updates as requested too
- stub things added
- After code updates, I re-tested this providing and not providing the diagnosis files, and it's getting either the count or `"NA"` as expected

bin/add_celltypes_to_sce.R
```diff
@@ -320,7 +354,52 @@ if (has_singler && has_cellassign) {
...
+    opt$consensus_validation_ref,
+    opt$diagnosis_celltype_ref,
+    opt$diagnosis_groups_ref
+  )
+
+  # Only run calculation if file sizes are not NA or 0
+  # the only way this "check" passes is if files exist and have contents
+  input_file_sizes <- file.size(check_input_files)
+  run_calculation <- !any(is.na(input_file_sizes)) & !any(input_file_sizes == 0)
+  if (run_calculation) {
```
jashapiro, line 373: Simplify a bit? 
```suggestion
  input_file_sizes <- file.size(check_input_files) |> tidyr::replace_na(0)
  if (all(input_file_sizes > 0)) {
```

bin/add_celltypes_to_sce.R
```diff
@@ -67,6 +70,33 @@ option_list <- list(
...
+  make_option(
+    opt_str = c("--diagnosis_groups_ref"),
+    type = "character",
+    help = "Path to TSV file mapping broad diagnoses to individual diagnoses for counting normal reference cells intended as input to inferCNV.
+    This file should have columns `diagnosis_group` for the broad diagnosis and `submitted_diagnosis` with individual diagnoses in ScPCA",
+    default = ""
+  ),
+  make_option(
+    opt_str = c("--reference_cells_file"),
+    type = "character",
```
jashapiro, line 96: I prefer defaults to failure. But I can be overruled.

```suggestion
    type = "character",
    default = "reference_cell_count.txt"
```

modules/classify-celltypes.nf
```diff
@@ -130,19 +133,26 @@ process add_celltypes_to_sce {
...
-
-
+      --consensus_celltype_ref ${consensus_ref_file} \
+      --consensus_validation_ref ${validation_ref_file} \
+      --diagnosis_celltype_ref "diagnosis_celltypes.txt" \
+      --diagnosis_groups_ref "diagnosis_groups.txt" \
+      --reference_cells_file "reference_cells.txt"
+
+      # save so we can export as environment variable
+      REFERENCE_CELL_COUNT=\$(cat "reference_cells.txt")
```
jashapiro, line 143: something about `reference_cells` makes me think it will be a list of cells, so I am clarifying for myself.
```suggestion
      --reference_cells_file "reference_cell_count.txt"

      # save so we can export as environment variable
      REFERENCE_CELL_COUNT=\$(cat "reference_cell_count.txt")
```

config/reference_paths.config
```diff
@@ -43,6 +43,12 @@ params {
   validation_markers_file = "${projectDir}/references/validation-markers.tsv"
   validation_palette_file = "${projectDir}/templates/palettes/validation-group-palette.tsv"
 
+  // TODO: These files will be added to the references dir once we have tagged versions. For now we use URLs.
```
jashapiro, line 46: Just clarifying which reference directory (since we also have references on S3 and I was thinking about there first)
```suggestion
  // TODO: These files will be added to the `${projectDir}/references` dir once we have tagged versions. For now we use URLs.
```

bin/add_celltypes_to_sce.R
```diff
@@ -320,7 +354,52 @@ if (has_singler && has_cellassign) {
   # add consensus cell type and ontology to sce
   sce$consensus_celltype_annotation <- celltype_df$consensus_annotation
   sce$consensus_celltype_ontology <- celltype_df$consensus_ontology
+
+
+  # Count the number of normal reference cells -------------------------
+
+  # Recalculate if these optional reference files were provided
+  # they have no defaults, so they will be NULL if not provided
```
jashapiro, line 362:
```suggestion
  # Calculate if these optional reference files were provided and are not 0 bytes
```

bin/add_celltypes_to_sce.R
```diff
@@ -94,6 +94,7 @@ option_list <- list(
   make_option(
     opt_str = c("--reference_cells_file"),
     type = "character",
+    default = "reference_cell_count.txt"
```
jashapiro, line 97: oops
```suggestion
    default = "reference_cell_count.txt",
```

PR #995, opened 2025-09-04
Add vireo docker image

jashapiro:
Now that we have an established protocol for Docker images within the `scpca-nf` repository, it seemed to make sense to use it for the other image that we manage outside of scpcaTools. So here I copied the vireo image from alsf-scpca to this repository, with updated labels to match the Cell Browser image, and a GHA to build and push the image. 

Note, this is stacked on #994


jashapiro:
> LGTM. Should we do the same with the cellranger/space ranger images? I'm not sure if we can use a GHA to build and push a private image, but either way we might want to at least port over the `Dockerfile`.

Yes, I think we probably will want to port those over, but yeah, probably without a build action, which is part of why I didn't do that yet.

PR #1007, opened 2025-09-09
Add script to run inferCNV

### LINKED ISSUE #1001: Add script to run inferCNV
_Broken out of #965_

This issue tracks adding a script to run `inferCNV` on a given library. This script should export the PNG heatmap for later use in the QC report, as well as an updated SCE object with the following new contents:
- New metadata fields:
  - the full inferCNV table should be stored in the metadata
  - of the consensus cell types used in the reference, as we do for Cell Assign groups: https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/bin/add_celltypes_to_sce.R#L259
- New colData column `total_infercnv`. This should be calculated as in e.g. https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/2624fc57ca27dae7b4a9eabdb6a670a16df9c3d4/analyses/infercnv-consensus-cell-type/exploratory-notebooks/SCPCP000015/02_ewings-reference-cnv.Rmd#L34-L61

Note also that we'll want to wrap `inferCNV` in a tryCatch, since it might fail. In that case, we'll still want an empty PNG file for use in nextflow since we have to stage it for the report.

### Other inferCNV considerations to be discussed before implementing this:

- By default, inferCNV does not consider `chrX`, `chrY`, and `chrM`. We don't have `M` at all, but our `X` and Y` are now indicated with arms. As such, these chromsomes won't be excluded unless we pass in an updated vector giving the names for our chrs X/Y. Should we do this?
  - See the `chr_exclude` argument for context: https://rdrr.io/github/broadinstitute/inferCNV/man/CreateInfercnvObject.html 
- We will need to again pull out the normal reference cells using code similar/identical to what we previously implemented here: https://github.com/AlexsLemonade/scpca-nf/blob/aa09fc96df0c5c021884e9ba30d6c02ba3e8b757/bin/add_celltypes_to_sce.R#L372 
  This will involve staging 3 additional input files too (validation file, 2 diagnosis files). I see two approaches here, motivated by my disinterest in having exactly duplicated code because maintaining that is hard:
  - Add a function to scpcaTools to do this to avoid repeating code. This involves more overhead though!
  - My preference is to modify `add_celltypes_to_sce.R` to add a colData column with T/F for whether the cell is in the normal reference. In this scenario...
    - We'd also pass in the threshold to `add_celltypes_to_sce.R` and only add the column if the threshold is met
    - If inferCNV fails for whatever reason, we can go ahead and remove that column in the script that runs inferCNV

sjspielman:
Closes #1001 

This PR adds a script to run inferCNV, which I have locally tested with much success. The only code changes here are this R script. Here's a rundown:

- One thing I learned is that we don't actually need to _write out_the `annotations_file` as an intermediate step but it can be a data frame, so that's a nice upgrade from previous scripts that run inferCNV!
- I have wrapped running inferCNV in a tryCatch, and if it fails I touch a PNG file with the same name that inferCNV creates it has. I don't attempt any renaming but just take what inferCNV wants to call it
- I add three items to the SCE if infercnv runs successfully:
  - the wide table goes into metadata
  - a total_cnv column goes into coldata
  - I also added another metadata field `infercnv_options` with a list of info that comes out of infercnv that seemed useful to communicate, what do you think? Here's what's in it, for example:
  
```
> infercnv_result@options
$cutoff
[1] 0.1

$out_dir
[1] "infercnv-test"

$HMM
[1] TRUE

$HMM_type
[1] "i6"

$denoise
[1] TRUE

$num_threads
[1] 8

$save_rds
[1] FALSE

$smooth_method
[1] "pyramidinal"

$HMM_report_by
[1] "subcluster"

$analysis_mode
[1] "subclusters"

$tumor_subcluster_partition_method
[1] "leiden"

$hclust_method
[1] "ward.D2"

$leiden_function
[1] "CPM"

$z_score_filter
[1] 0.8

$tumor_subcluster_pval
[1] 0.1

$k_nn
[1] 20

$leiden_method
[1] "PCA"

$per_chr_hmm_subclusters
[1] FALSE

$cluster_by_groups
[1] TRUE

$max_centered_threshold
[1] 3

$remove_genes_at_chr_ends
[1] FALSE

$prune_outliers
[1] FALSE

$BayesMaxPNormal
[1] 0.5

$mask_nonDE_genes
[1] FALSE

$noise_filter
[1] NA

$sd_amplifier
[1] 1.5

$noise_logistic
[1] FALSE

$chr_exclude
[1] "chrXp" "chrXq" "chrYp" "chrYq"

$max_cells_per_group
NULL

$min_max_counts_per_cell
[1] 100 Inf

$counts_md5
[1] "d6b907c1b37e8a078cc023e5acca5f71"
```

- In the end, I went ahead and removed all the inferCNV outputs except the PNG file, let me know if you think this is overkill! There's just so many files 🫠 

The other change is here is uncommenting some params that should have been uncommented previously! But during testing in an earlier PR the commenting-out snuck through.

----

Note that in the nextflow process coming up, I'll plan to create the output directory before running the script similar to what we do for cell typing, e.g.: https://github.com/AlexsLemonade/scpca-nf/blob/731fd68b9d7c8526d0c76dda577a8163d867844e/modules/classify-celltypes.nf#L21-L29

And in the subworkflow, I will add this directory & file name to meta for later use to stage it for the report, similar to e.g.: https://github.com/AlexsLemonade/scpca-nf/blob/731fd68b9d7c8526d0c76dda577a8163d867844e/modules/classify-celltypes.nf#L206-L217

### DIFF
--- a/bin/run_infercnv.R
+++ b/bin/run_infercnv.R
@@ -1,9 +1,10 @@
 #!/usr/bin/env Rscript
 
-# This script runs inferCNV on a SingleCellExperiment object and saves results to the SCE object:
-# metadata field `infercnv_table`: (data frame) The full table output from the inferCNV HMM
-# metadata field `infercnv_options`: (list) The options used when calling inferCNV, from the @options slot of the output object
-# colData column `total_cnv`: The sum of CNV per cell, calculated from the HMM output
+# This script runs inferCNV on a SingleCellExperiment object and export result files:
+# - inferCNV heatmap png
+# - an RDS file with a list of information to save to the SCE file later:
+#  - wide metadata table with CNVs
+#  - list of inferCNV options
 
 suppressPackageStartupMessages({
   library(SingleCellExperiment)
@@ -18,15 +19,15 @@ option_list <- list(
     help = "Path to the SCE file to run inferCNV on"
   ),
   make_option(
-    opt_str = "--output_rds",
-    type = "character",
-    help = "Path to the output SCE file inferCNV output"
-  ),
-  make_option(
     opt_str = c("--output_dir"),
     type = "character",
     default = "",
     help = "Folder to save final infercnv results"
+  ),
+  make_option(
+    opt_str = "--output_rds",
+    type = "character",
+    help = "Path to the output RDS file to hold inferCNV results of interest"
   ),
   make_option(
     opt_str = c("--gene_order_file"),
@@ -54,9 +55,9 @@ opts <- parse_args(OptionParser(option_list = option_list))
 
 stopifnot(
   "input_sce_file does not exist" = file.exists(opts$input_sce_file),
-  "gene_order_file does not exist" = file.exists(opts$gene_order_file),
   "output_dir does not exist" = dir.exists(opts$output_dir),
-  "output_rds was not provided" = !is.null(opts$output_rds)
+  "output_rds was not provided" = !is.null(opts$output_rds),
+  "gene_order_file does not exist" = file.exists(opts$gene_order_file)
 )
 
 # set up files and paths -----------------
@@ -86,7 +87,6 @@ annotations_df <- data.frame(
   row.names = sce$barcodes
 )
 
-
 # run infercnv in a tryCatch in case of errors
 # this will automatically create the heatmap file in the output directory
 infercnv_result <- tryCatch(
@@ -112,6 +112,7 @@ infercnv_result <- tryCatch(
       )
   },
   error = function(e) {
+    message("failed")
     # If inferCNV failed, create an empty heatmap file
     system(glue::glue("touch {png_file}"))
 
@@ -119,10 +120,10 @@ infercnv_result <- tryCatch(
     NULL
   }
 )
-# confirm png file exists
+# confirm png file was created
 stopifnot("PNG file not created" = file.exists(png_file))
 
-# save results SCE if inferCNV ran successfully ------------------------------
+# save relevant results to RDS if inferCNV ran successfully -------------------
 if (!is.null(infercnv_result)) {
   # confirm final infercnv object exists; the metadata table is created from this file
   stopifnot(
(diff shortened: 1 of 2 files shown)


bin/run_infercnv.R
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env Rscript
+
+# This script runs inferCNV on a SingleCellExperiment object and saves results to the SCE object:
+# metadata field `infercnv_table`: (data frame) The full table output from the inferCNV HMM
+# metadata field `infercnv_options`: (list) The options used when calling inferCNV, from the @options slot of the output object
+# colData column `total_cnv`: The sum of CNV per cell, calculated from the HMM output
```
allyhawkins, line 6: I agree that we probably want two separate scripts (and two separate processes) so this should be broken out of here.

bin/run_infercnv.R
```diff
@@ -0,0 +1,174 @@
...
+    opt_str = "--output_rds",
+    type = "character",
+    help = "Path to the output SCE file inferCNV output"
+  ),
+  make_option(
+    opt_str = c("--output_dir"),
+    type = "character",
+    default = "",
+    help = "Folder to save final infercnv results"
+  ),
```
allyhawkins, line 30: In the new script I would update this to explicitly define the metadata tsv and heatmap file as inputs here.

bin/run_infercnv.R
```diff
@@ -0,0 +1,174 @@
...
+  sce,
+  opts$output_rds,
+  compress = "bz2"
+)
+
+
+# clean up: remove all non-heatmap files from the output directory ------
+remove_files <- list.files(opts$output_dir, full.names = TRUE)
+remove_files <- remove_files[remove_files != png_file]
+fs::file_delete(remove_files)
```
allyhawkins, line 174: You won't need this if you provide the path to the heatmap and metadata in Nextflow. Only what's specified in the output will actually get published.

allyhawkins (commented):
I agree with your comment about splitting this script up so that adding to the SCE happens separately. We should store the `infercnv` intermediate files (heatmap and metadata) in the `checkpoints` folder.

sjspielman:
Thanks for the review, @allyhawkins! FYI I've started working on splitting up this into 2 scripts already. I'll look into what you have for `ews-nf` and add in accordingly, and I'm going to stack the 2nd script (#1008) on this PR.

PR #1010, opened 2025-09-10
Add module for inferCNV

### LINKED ISSUE #965: Add submodule to run inferCNV
We plan to add functionality to run `inferCNV` across samples as part of `scpca-nf`. In this issue, I'm outlining a proposed flow for this. When we reach this development stage, this will be broken up into individual issues. 
Please feel free to leave comments for tweaking this strategy! (e.g., is 2 processes here too many I/Os?).

- To begin, we'll need the following additional workflow parameters:
  - URI to the diagnosis/normal cells TSV
  - Threshold for number of normal cells needed to run `inferCNV` (we'll use 100)
  - Arms gene ordering file
  - Booleans for `repeat` and `perform`

We'll add a submodule to run after the celltyping process which will follow this flow:

- First, enter a process that will determine if we should run `inferCNV`
  - Inputs: TSV, threshold parameter, processed SCE
  - Actions: Runs a script that counts the number of normal cells. If >= threshold, add a boolean column to the processed SCE with `infercnv_reference` and rewrite out the updated SCE.  Also, export a boolean we can use to branch into the `inferCNV` process next.
  - Outputs: Updated processed SCE file and updated meta json with boolean if we're running `inferCNV`
- Now we can branch into an `inferCNV` process
  - Inputs: arms gene order file, processed SCE  
  - Actions: Runs a script to run `inferCNV` and save results
    - We'll keep the PNG around in files to later include in the report
    - We'll add a `total_cnv` column as we've previously calculated to the `colData`
    - We'll add the full wide as heck metadata table output to the SCE metadata
 - Mix back channels following the branching, and head back into `main.nf`

sjspielman:
Closes #965 

This adds the inferCNV module. I'm opening as a draft because it's still undergoing some testing. Specifically, I know the nextflow logic is all sound because I've run this all the way through on batch in 378e97a06202ae74720eb2f60852ecf8650d4992, but I had the actual scripts commented out there and just `touched` files we needed to stage because hooooboy was testing while running inferCNV a pain! So, now that the processes and flow is all confirmed working, I'm currently testing it with the R scripts actually running and we'll find out if memory needs to be further bumped or there's a sneaky R bug still to catch.

---

Here's the rundown of the inferCNV module, which has some similarities to, but less complexity than!, the cell type module. It has two processes: one to run inferCNV and one to add results to the SCE, to support the `repeat` flag.

- First, we branch off libraries we're not going to run: cell lines, and too few reference cells. Similar to the cell type module, I leave unfiltered/filtered off to the side for most of the workflow and bring them back in the end.
- Of the eligible libraries, we branch to not repeat running inferCNV unless specified. Results are stored in `checkpoints/infercnv/<library>`
- We then run inferCNV, and publish the heatmap and results RDS (note this is an list I made that contains the stuff we want to store in the SCE only; it's not a full inferCNV result object) to the checkpoints dir
  - Noting that the temporary directory strategy for saving inferCNV files was borrowed from `ews-nf`: https://github.com/AlexsLemonade/ews-nf/blob/1ce6754a068d8fdcac37b62c88380df7e01f26ff/modules/infercnv/main.nf#L53-L55, thanks @allyhawkins for debugging this for the rest of us!
  - A note also that at an earlier iteration, I was popping the output files into a directory and publishing that directory to the checkpoints. That wasn't the move, since you ended up with too much nesting: `checkpoints/infercnv/libraryid/libraryid/stuff`. So, just the files get staged with a glob.
- Then, we mix up the skipped and the run channels to head into the 2nd process which add inferCNV results to the processed SCE
- Finally, we get back to the `[meta, unfiltered, filtered, processed]` channel structure with all libraries including the ineligible ones, and emit

Other changes:
- bug fix to `separate_longer_delim`. (I probably need a parallel bug fix for this code change over in `OpenScPCA-analysis`, but the notebook where that got added is going to be updated later anyways and it doesn't run in CI, so I'm leaving it be)
- I added some inferCNV checks to `main.nf`. I also have a slightly nested `if` situation for running cell types & infercnv if you have feedback there! I find it readable, but do others?
- I modified the code that save the reference cell count to meta to ensure it's an integer so we can do math on it later, which meant converting NA to 0. Open to other strategies! 
- I changed the param `perform_infercnv` to `run_infercnv` since that just worked better for my brain even though we use `perform_` for celltyping. I think this is fine though?

Questions:
~- I have a question as well about my use of the `library_id` tag. At various points I tried using `unique_id` instead but got errors that it didn't exist. I decided to not worry about that yet and just use `library_id`, but should I worry?~ EDIT: I just had a typo, fixed and using `unique_id` all around now!
- Naming was hard for this one, since I couldn't just call everything `run_infercnv`. Do we like what I did, which is `call_infercnv`? Other ideas?
- **UPDATE AND EDIT**: See this comment from @allyhawkins: https://github.com/AlexsLemonade/scpca-nf/pull/1013#discussion_r2342267176
  -  Currently, I am branching off to not put anything in the SCE objects that didn't run inferCNV, but I think I like this idea more. What do you think of going that direction instead? In this case, wouldn't branch before going into the `add_infercnv_to_sce` process but send all the eligible files in, and I'd use infercnv result file size in the R script to determine what to put into `infercnv_options` (actual options or "sorry it failed")

### DIFF
--- a/.github/workflows/nextflow-stub-check.yaml
+++ b/.github/workflows/nextflow-stub-check.yaml
@@ -28,10 +28,10 @@ jobs:
           sudo mv nextflow /usr/local/bin/
 
       - name: Check Nextflow workflow
-        run: nextflow -log stub-run.log run main.nf -stub -profile stub -ansi-log false --perform_cnv_inference
+        run: nextflow -log stub-run.log run main.nf -stub -profile stub -ansi-log false
 
       - name: Check Nextflow with checkpoints from previous run
-        run: nextflow -log checkpoint-run.log run main.nf -stub -profile stub -ansi-log false --perform_cnv_inference
+        run: nextflow -log checkpoint-run.log run main.nf -stub -profile stub -ansi-log false
 
       - name: Check Nextflow workflow for building cell type references
         run: nextflow -log celltype-ref-run.log run build-celltype-ref.nf -stub -profile stub -ansi-log false


jashapiro:
> One implementation note here is that I'm sending those files into the `run_infercnv` process, they just aren't getting run through the script.

Yeah, we'd really like to avoid that if we can. On a big run it might not matter too much, but spinning up a big instance just to no-op seems like a bad idea.

modules/call-cnvs.nf
```diff
@@ -0,0 +1,148 @@
...
+    run_infercnv(infercnv_input_ch.call_infercnv)
+
+    // prepare to add results for all eligible libraries: [meta, processed sce, results file]
+    add_infercnv_results_ch = infercnv_input_ch.skip_infercnv
+      .map{ meta, processed_sce, gene_order_file -> tuple(
+        meta,
+        processed_sce,
+        file("${meta.infercnv_results_file}", checkIfExists: true),
+        file("${meta.infercnv_heatmap_file}", checkIfExists: true)
+      )}
```
jashapiro, line 124: If you add logic to also skip infercnv for libraries that do not meet criteria, you could modify these lines to create any missing files:

```suggestion
    add_infercnv_results_ch = infercnv_input_ch.skip_infercnv
      .map{ meta, processed_sce, gene_order_file -> 
        infercnv_results = file("${meta.infercnv_results_file}")
        infercnv_heatmap = file("${meta.infercnv_heatmap_file}")
        // create empty files if they don't exist
        if (!infer_cnv_results.exists) infercnv_results.text = ''
        if (!infer_cnv_heatmap.exists) infercnv_heatmap.text = '' 
        // return tuple
        tuple(
          meta,
          processed_sce,
          infercnv_results,
          infercnv_heatmap
         )
      }
```

sjspielman: 🤯

sjspielman: Trying to implement this, the syntax works fine for me in `nextflow console` standalone (noting I grabbed this docs link for myself too https://www.nextflow.io/docs/latest/working-with-files.html#reading-and-writing-an-entire-file), but not from the nextflow workflow itself.

With syntax fixes I'm working with...

```
add_infercnv_results_ch = infercnv_input_ch.skip_infercnv
      .map{ meta, processed_sce, gene_order_file ->
        def infercnv_results = file("${meta.infercnv_results_file}") // also tried without def but i do think we want def for local definition?
        def infercnv_table   = file("${meta.infercnv_table_file}")
        def infercnv_heatmap = file("${meta.infercnv_heatmap_file}")
        // create empty files if they don't exist
        if (!infercnv_results.exists()) infercnv_results.text = '' // ERROR HERE!!
        if (!infercnv_table.exists())   infercnv_table.text = ''
        if (!infercnv_heatmap.exists()) infercnv_heatmap.text = ''
        // return tuple...etc below.
```

But persistently getting `ERROR ~ No such file or directory` at the denoted line (yeah, we know there's no such file...kinda the point half the time...). Any thoughts? If nextflow doesn't like this syntax in this particular context, my other move would be a "dummy process" called `touch_infercnv` or so, but I do want to try to get this working if we can since it's much simpler.

sjspielman: The file doesn't exist bug error I'm getting is because _directory the files are meant to live in_ doesn't exist locally; if I make the directory where they would live on my own beforehand, there's no error.

I was confused that this passed the stub workflow in CI https://github.com/AlexsLemonade/scpca-nf/pull/1010/commits/c2a4036d7f65c47589f29632063c7ec8e421d602, but realized that was only because this part of the code wasn't being tested.

I'm not sure about syntax for creating directories from a workflow - is there such a thing that you know if, without having to use a process w/ publishDir?

sjspielman: yep, it's fine! cleaned up a bit https://github.com/AlexsLemonade/scpca-nf/pull/1010/commits/16bffce20fd9c4f0831437753bdd6436e998fefc

jashapiro (commented):
Overall this looks good. I had a couple questions and some style suggestions, but I don't think anything too major!

modules/call-cnvs.nf
```diff
@@ -0,0 +1,171 @@
...
+    table_file="${meta.unique_id}_infercnv-table.txt"
+    heatmap_file="${meta.unique_id}_infercnv-heatmap.png"
+    """
+    # note that inferCNV fails, the script will output empty results/heatmap files
+    run_infercnv.R \
+      --input_sce_file ${processed_rds} \
+      --output_rds ${results_file} \
+      --output_table ${table_file} \
+      --output_heatmap ${heatmap_file} \
+      --temp_dir \$PWD \
```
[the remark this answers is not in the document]
sjspielman: I am confused by your confusion... 
InferCNV requires specify a directory where it will dump all of its output files. Similar to how we've run inferCNV in `ews-nf`, we're telling inferCNV here to dump all the output files in `pwd`, and then in the end the files get copied to their final names which in this case I suppose is just renaming. Maybe this is the confusion, that I'm copying instead of just renaming (left over approach from pre-porting into nextflow, since publishDir will take care of copying?).

modules/call-cnvs.nf
```diff
@@ -0,0 +1,171 @@
...
+        if (!infercnv_results.exists()) infercnv_results.text = ''
+        if (!infercnv_table.exists())   infercnv_table.text = ''
+        if (!infercnv_heatmap.exists()) infercnv_heatmap.text = ''
+        // return tuple
+        tuple(
+          meta,
+          processed_sce,
+          file(meta.infercnv_results_file, checkIfExists: true),
+          file(meta.infercnv_table_file, checkIfExists: true),
+          file(meta.infercnv_heatmap_file, checkIfExists: true)
```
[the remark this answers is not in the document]
jashapiro: We already know they exist, because we just checked and then made them if they didn't. So just:
```suggestion
          infercnv_results,
          infercnv_table,
          infercnv_heatmap
```

modules/call-cnvs.nf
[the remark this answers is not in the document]
sjspielman: noting that inferCNV will make it for us although not convinced `file.exists()` is what they should be using to check for it... https://github.com/broadinstitute/infercnv/blob/624feae9727dff74926aecd0d8945a21d61b572b/R/inferCNV_ops.R#L441-L444
I'll still create it to be safe though as is my wont. 

I'm going to do another `--repeat_cnv_inference` test run with these changes, eta 2 hours!

PR #1018, opened 2025-09-16
Add openscpca annotations to the report

### LINKED ISSUE #971: Account for OpenScPCA annotations in the report
Once we have added the OpenScPCA annotations to the processed objects, we will need to include them in the report. They should be added to the existing cell type sections of the report and treated similarly to how we handle submitter annotations. If they are present, they will be in the main report along with a comparison to consensus labels. 

In the description of these annotations we should be sure to add a link to the original module in `OpenScPCA-analysis`.

allyhawkins:
Closes #971 

Here I'm adding the OpenScPCA annotations to the cell type sections of the report. I treated these very similarly to submitter annotations so they are shown both in the main report section and the supplemental report if present. 

- Just like submitter annotations, these appear first in the tabs for tables and UMAPs. I think this is the right choice, but someone tell me if they would prefer to have consensus first in the order. 
- If present, we show a heatmap comparing OpenScPCA to consensus in the main report and then in the supplement OpenScPCA annotations are compared to all other annotations (just like we do for submitter). 
- The way this is setup, we theoretically could have both openscpca and submitter annotations and would just have two heatmaps, one comparing to submitter and one comparing to openscpca. I don't think this will actually happen, but I thought it made the most sense to future proof against that. 

Here's a zip with reports for libraries with and without openscpca annotations: 
[reports.zip](https://github.com/user-attachments/files/22365313/reports.zip)

I also fixed a minor bug I saw in the infercnv changes, but I noticed it was also committed to https://github.com/AlexsLemonade/scpca-nf/pull/1010 this morning.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -385,7 +385,7 @@ if (has_singler && has_cellassign) {
     # get the cell type groups to consider for this diagnosis
     reference_validation_groups <- diagnosis_celltype_df |>
       dplyr::filter(diagnosis_group == broad_diagnosis) |>
-      tidyr::separate_delim_longer(celltype_groups, sep = ",") |>
+      tidyr::separate_longer_delim(celltype_groups, delim = ",") |>
       dplyr::pull(celltype_groups) |>
       stringr::str_trim() # remove any leading or trailing spaces
 
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -190,6 +190,7 @@ cellassign_not_run <- check_cellassign_not_run(
 # define bullets glue string
 methods_bullets <- c(
   ifelse(has_submitter, "Submitter-provided", NA),
+  ifelse(has_openscpca, "OpenScPCA", NA),
   ifelse(has_consensus, "Consensus cell types", NA),
   ifelse(has_singler, "`SingleR`", NA),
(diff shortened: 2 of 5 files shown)


sjspielman (commented):
This looks good overall but there's one bug I caught - in the statistics tables, we still have the string "Unknown cell type" instead of "openscpca-excluded", see here:
Screenshot 2025-09-16 at 11 34 17 AM

Pretty sure it's coming from here: https://github.com/AlexsLemonade/scpca-nf/blob/289e6a59e03fbed03f2827dffd87f3bc5dcd273f/templates/qc_report/utils/celltype_functions.rmd#L88-L91
which is called from here for openscpca: https://github.com/AlexsLemonade/scpca-nf/blob/289e6a59e03fbed03f2827dffd87f3bc5dcd273f/templates/qc_report/utils/celltype_functions.rmd#L66-L72
We probably want to make `unknown_string` an argument in that function so we can specify the right one when prepping the OpenScPCA version so that the excluded string doesn't get overwritten.

allyhawkins:
> This looks good overall but there's one bug I caught - in the statistics tables, we still have the string "Unknown cell type" instead of "openscpca-excluded", see here:

This isn't a bug. There are unknown cells here. "openscpca-excluded" should only refer to cells that were not even attempted to be annotated by the module and there is no existing cell type annotation. unknown means we tried to annotate it and it's still unknown, just like with Singler, CellAssign, and consensus.

sjspielman (approved):
Rephrase is great, rest still looks good!

PR #1019, opened 2025-09-16
add target = "_blank" to cell browser abstract links

jashapiro:
Does what it says... Just checking that this is what you were looking for!


davidsmejia (approved):
LGTM

PR #1033, opened 2025-09-18
Fix a few minor nextflow issues

jashapiro:
While investigating https://github.com/AlexsLemonade/scpca-nf/pull/1026#issuecomment-3307756048, I was looking through my local copy of scpca-nf, and found one place where we aren't really tracking the `perform_celltyping` variable quite right in the workflow, so I fixed that, along with some other complaints that the linter was throwing at me.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -39,13 +39,8 @@ def check_parameters() {
     param_error = true
   }
 
-  perform_celltyping = params.perform_celltyping // create separate variable to use instead
   // CNV inference checks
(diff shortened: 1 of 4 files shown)


jashapiro:
> while we're here, can you point me to why nextflow likes us to use a different name in mapping (I assume the `_` prefix is just to keep variables different but identifiable)?

The `_` prefix tells the linter that you are not using the variable in the function (lambda here), otherwise it warns that you have an unused variable.

Issue #1025, opened 2025-09-17
Ensure inferCNV checkpoints have versioning information


sjspielman:
While working on #1005, I realized that we need to tweak how we handle the `inferCNV` checkpoint files. Specifically, the checkpoint files don't contain any version information. This means that if we update the mapping between diagnoses and cell types for creating the normal reference, `scpca-nf` would not know that the checkpoint files are out of date and it would not repeat inference. But, similar to cell type annotation, we would want the workflow to flag these checkpoints as out of date and not use them.

Since the diagnosis TSVs where this information comes are from `OpenScPCA-analysis`, we could use a version tag from that repo to track this. Then again, it's not clear that future versions of these TSV files will _definitely_ always be created/evaluated in that repo first, so we might need to figure out a different tagging scheme too.

sjspielman:
We plan to go with what is outlined (and commented - hashes, not a list!) in #1039 here, so we can close this issue now in favor of that one.

PR #1034, opened 2025-09-18
Add external instructions for running inferCNV

### LINKED ISSUE #1005: Add infercnv docs to external instructions
This is a placeholder issue tracking adding docs to external-instructions.md on how to run inferCNV. Details will be added when we get closed to implementing the docs.

sjspielman:
Closes #1005 

This PR adds external instructions and example metadata files. 

Writing this one was a bit of a journey! I ended up deviating from the initial organization I sketched out for myself in #1005 since I couldn't really figure out a way to explain/motivate the diagnosis metadata files without getting into how the sausage is made, but I tried to not get too technical with it - let me know if I succeeded! I'm very curious to hear thoughts about the organization and this approach to framing - I'm open to changing things up!

There are also some TODOs I left for future us in the repeat section, since we need to figure some more things out (#1025).

I also added the associated example metadata files. A couple comments on this:
- I decided to update the example metadata sample file to have fewer diagnoses which helps show that you don't have to repeat diagnoses in these new TSVs. Is this a good idea or overkill?
- I used "real" diagnosis groups and cell type groups, since I thought some biology was best for this example since it's a little tricky imo. Thoughts?
- I wanted to check in about our versions of these metadata files: first, they are named slightly differently because of how they were created in OpenScPCA-analysis, and also the broad diagnosis map TSV has additional columns which aren't relevant/used by scpca-nf. Since the instructions talk about how we have these existing files they can use, I wonder if it's confusing that this TSV has extra content. Overall, I am leaning towards renaming these files in this repo to more closely match both these example file names and their workflow parameter names, as well as stripping out those extra columns, and then documenting those manual steps in `references/README.md`. What do you think of this?


external-instructions.md
```diff
@@ -454,6 +459,136 @@ The cell type label file is a TSV file with the following required columns:
...
+`scpca-nf` uses results from its cell type annotation module to create this normal reference.
+As such, to perform CNV inference, you must also [perform cell type annotation, as described above](#cell-type-annotation).
+Note that if CNV inference is turned on, cell type annotation will automatically be turned on as well.
+
+You can turn on CNV inference by taking the following steps:
+
+1. Ensure your [prepared sample metadata file](#prepare-the-sample-metadata-file) contains a `diagnosis` column.
+   This column is **required** in your sample metadata file as `scpca-nf` uses this information to create the normal reference.
+2. [Determine which cell types](#determining-normal-reference-cell-types) should be used for the `inferCNV` normal reference.
+3. [Prepare the diagnosis metadata files](#preparing-diagnosis-metadata-files) which the workflow will use to create the normal cell type reference.Using the information from the previous steps,
```
allyhawkins, line 478:
```suggestion
3. [Prepare the diagnosis metadata files](#preparing-diagnosis-metadata-files) which the workflow will use to create the normal cell type reference using the information from the previous steps. 
```

external-instructions.md
```diff
@@ -454,6 +459,136 @@ The cell type label file is a TSV file with the following required columns:
...
+   You will need to provide the path/uri to this files as a workflow parameters, which you will need to define in your configuration file.
```
allyhawkins, line 479:
```suggestion
   You will need to provide the path/uri to these files as workflow parameters, which you will need to define in your configuration file.
```

external-instructions.md
```diff
@@ -454,6 +459,136 @@ The cell type label file is a TSV file with the following required columns:
...
+   For more information on adding parameters to your configuration file, see [Configuring scpca-nf for your environment](#configuring-scpca-nf-for-your-environment).
+4. Run the workflow with the `--perform_cnv_inference` flag.
+
+Once you have followed both the above steps and the [steps to prepare for cell type annotation](#cell-type-annotation) and you have added the path/uri to the diagnosis metadata files to your configuration file, you can use the following command to run the workflow with CNV inference:
```
allyhawkins, line 483:
```suggestion
Once you have followed the above steps, including the [steps to prepare for cell type annotation](#cell-type-annotation), and you have added the path/uri to the diagnosis metadata files to your configuration file, you can use the following command to run the workflow with CNV inference:
```

external-instructions.md
```diff
@@ -454,6 +459,136 @@ The cell type label file is a TSV file with the following required columns:
...
+
+
+### Prepare the diagnosis metadata files
+
+`scpca-nf` uses the consensus cell type labels derived from `SingleR` and `CellAssign` annotations to create the set of normal reference cells.
+The specific consensus cell types to use are determined by the sample's diagnosis.
+
+To manage many diagnoses and consensus cell types, the Data Lab has organized [individual sample diagnoses into broad diagnosis groups](references/broad-diagnosis-map.tsv), and similarly [consensus cell types into validation groups](references/consensus-validation-groups.tsv).
+
+`scpca-nf` uses these grouping to specify the normal reference cell types as follows:
```
allyhawkins, line 498:
```suggestion
`scpca-nf` uses these groupings to specify the normal reference cell types as follows:
```

external-instructions.md
```diff
@@ -454,6 +459,136 @@ The cell type label file is a TSV file with the following required columns:
...
+
+1. Given a sample's diagnosis, identify the corresponding broad diagnosis
+    * This step utilizes the provided `params.diagnosis_groups_file` TSV
```
allyhawkins, line 501: A thought I had while reading this is that we could make this file optional and if that's the case, then they just need to make sure the diagnosis in the sample metadata table matches the diagnosis in the column for broad diagnosis in `diagnosis_celltypes_file`. I feel like the broad groups for diagnoses are really just helpful for us and not for external users that are running samples independently. 

If we did that, then we should simplify all of these instructions to not even mention the broad diagnosis file.

external-instructions.md
```diff
@@ -454,6 +459,136 @@ The cell type label file is a TSV file with the following required columns:
...
+
+The Data Lab has prepared these metadata files for all sample diagnoses present in the ScPCA Portal, which `scpca-nf` uses by default:
+
+* The diagnosis groups metadata file is provided in [`references/broad-diagnosis-map.tsv`](references/broad-diagnosis-map.tsv)
+  * Broad diagnoses are listed in the `diagnosis_group` column, and sample-level diagnoses are listed in the `submitted_diagnosis` column
+* The diagnosis cell types metadata file is provided in [`references/diagnosis-celltype-groups.tsv`](references/diagnosis-celltype-groups.tsv)
+  * Broad diagnoses are listed in the `diagnosis_group` column, and consensus cell type validation groups are listed in the listed in the `celltype_groups` column
+
+You can inspect these files to determine if your sample's diagnosis is already present in the `submitted_diagnosis` column of the diagnosis groups metadata file.
+If your sample's diagnosis is present, you do not have to prepare your own metadata files; you can let `scpca-nf` make use of these existing metadata files.
```
allyhawkins, line 519: I would move this statement to the very top of this section. I also might refocus this and have the first section be just how to run inferCNV using existing metadata files and then have these sections that talk about actually making the metadata files. They only need to make this file if they have a diagnoses that is not present yet in our data.

external-instructions.md
allyhawkins: Related to this, maybe we should just refocus this on how to run inferCNV using our existing metadata files and then as a future TODO, ensure we are able to support someone running the workflow with a diagnoses that we don't have or a custom normal reference list.

external-instructions.md
```diff
@@ -454,6 +459,136 @@ The cell type label file is a TSV file with the following required columns:
...
+To support the first two steps of this procedure, you will need to provide two metadata TSV files:
+
+* [The diagnosis groups metadata file](#diagnosis-groups-metadata-file), which maps individual sample diagnoses to broad diagnosis groups
+* [The diagnosis cell types metadata file](#diagnosis-cell-types-metadata-file), maps broad diagnosis groups to consensus cell type validation groups
+
+The Data Lab has prepared these metadata files for all sample diagnoses present in the ScPCA Portal, which `scpca-nf` uses by default:
+
+* The diagnosis groups metadata file is provided in [`references/broad-diagnosis-map.tsv`](references/broad-diagnosis-map.tsv)
+  * Broad diagnoses are listed in the `diagnosis_group` column, and sample-level diagnoses are listed in the `submitted_diagnosis` column
+* The diagnosis cell types metadata file is provided in [`references/diagnosis-celltype-groups.tsv`](references/diagnosis-celltype-groups.tsv)
```
allyhawkins, line 515: If we keep the instructions for using a custom diagnoses, we need to provide a list of all possible validation groups. I think we have a table that matches consensus to validation so we should direct them to that table.

allyhawkins (commented):
I mostly left some overview comments that I think we should address and flush out before continuing. 

In general, I think we should first focus the instructions on running with existing defaults, including diagnoses we already have and the assigned normal cell type groups for each diagnoses. Then you should have a subsection that talks about how to run with a diagnoses that we don't yet support and/or modify the list of cell types used as normal for a given diagnoses. This second section is where you can expand on the provided metadata files that you need. 

Additionally, I think we should make the broad diagnoses mapping file optional, and if that file is not provided, check that the diagnosis matches the diagnosis column in the normal cell group to broad diagnosis map. That way external users only need to create one additional metadata file. That will probably require code changes though. 

Because of that, I think this first PR should just be "how to run with our existing metadata" and then in another PR we add in running with custom metadata.

external-instructions.md
sjspielman: > A thought I had while reading this is that we could make this file optional and if that's the case, then they just need to make sure the diagnosis in the sample metadata table matches the diagnosis in the column for broad diagnosis in diagnosis_celltypes_file. I feel like the broad groups for diagnoses are really just helpful for us and not for external users that are running samples independently.

In a scenario where this file is optional, it sounds like you still envision _us_ using the file?

sjspielman:
@allyhawkins I just wanted to give you an update here: First, I am testing `scpca-nf` with only one of these files being optional. Having now thought more about it, I agree this move seems reasonable to not overly burden external users, **but** this approach does make me feel more strongly about this comment I left:

> Overall, I am leaning towards renaming these files in this repo to more closely match both these example file names and their workflow parameter names, as well as stripping out those extra columns, and then documenting those manual steps in references/README.md. What do you think of this?

This is because, with this shift to only 1 required metadata file, I will want to add to external instructions "you might want to double check that your sample diagnosis is represented in the broad diagnosis file" (maybe only saying this as a as a potential troubleshooting step). If others might have to poke around in these files, I'd prefer to keep things more consistent and without extra columns we don't use, since this seems like a better user experience. 

After testing with `scpca-nf`, I will proceed with:
- first, a PR to update `scpca-nf`
- then, back to these instructions; I'll open fresh PRs to get this back going in 2 stages.

allyhawkins:
> After testing with `scpca-nf`, I will proceed with:
> 
> * first, a PR to update `scpca-nf`
> * then, back to these instructions; I'll open fresh PRs to get this back going in 2 stages.

This plan sounds good to me.

sjspielman:
Closing this PR in favor of updates PRs as discussed!

PR #1037, opened 2025-09-22
account for missing pool file and update cavatica exclude file

### LINKED ISSUE #1021: [BUG] "development" version does not run on Cavatica
**Describe the bug**

I was trying to use the `development` version on Cavatica as it includes the 10Xv3_5prime chemistry per my request in https://github.com/AlexsLemonade/scpca-nf/issues/958. I have no issue running it on my local ec2 instance.
But I encountered an error on Cavatica. I followed the [instruction](https://github.com/AlexsLemonade/scpca-nf/blob/development/cavatica-instructions.md) to push the app to Cavatica without issue. And I used the [example run metadata](https://cavatica.sbgenomics.com/u/jashapiro/scpca-nf-test/files/682ca828f8492c6e343921d6) and [example sample metadata](https://cavatica.sbgenomics.com/u/jashapiro/scpca-nf-test/files/681b7a812b97f154cf6e477b) that you shared previous in the [scpca-nf-test](https://cavatica.sbgenomics.com/u/jashapiro/scpca-nf-test) project. I used `cavatica` profile in the setting. This is the error from the `stdout.log`.

> ERROR ~ Argument of `file` function cannot be empty
> 
>  -- Check script 'main.nf' at line: 234 or see '.nextflow.log' file for more details

Can you help on checking this? or do you have a different version with the 10Xv3_5prime chemistry on Cavatica? Thank you! @jashapiro 

**To reproduce**
If possible, please provide example code to reproduce the problem, including any information about the dataset where the problem occurred, if possible.

What command did you use to invoke the `scpca-nf` workflow?

What type of system are you using to run the workflow (local, slurm, AWS, etc.)?
If you are using a custom profile or config file, please include the file and settings you used.

**Expected behavior**

**Version information**
What version of `scpca-nf` are you using, and what version of Nextflow?

**Additional context**

jashapiro:
This branch was started to track any changes for #1021, but I ended up deciding that the only change I wanted to make at this point was to slightly modify the exclude file to add excluding the `docker` directory, and sorting the file. 

However, I also needed to test the updated workflow on Cavatica, and in doing so I found that we are not properly allowing for the cellhash pool file to be missing, so I made a small change to accommodate that. 

I did not add support for providing new params when running, which I think is fine for now, but we may want to add support for files required for InferCNV, but I think that can wait until we have completed docs and any potential redesign of the input files for that. 

In the meantime, I have now successfully tested this on Cavatica with these changes, so this should close #1021

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -251,7 +251,7 @@ workflow {
   spaceranger_quant(runs_ch.spatial)
 
   // **** Process 10x flex RNA-seq data ***
-  flex_quant(runs_ch.flex, flex_probesets, file(params.cellhash_pool_file))
+  flex_quant(runs_ch.flex, flex_probesets, file(params.cellhash_pool_file ?: empty_file))
   flex_sce_ch = generate_sce_cellranger(flex_quant.out, file(params.sample_metafile))
     .branch{
       continue_processing: it[2].size() > 0 || it[2].name.startsWith("STUBL")


allyhawkins (approved):
LGTM

PR #1038, opened 2025-09-22
Replace list of columns to include to be columns to exclude 

### LINKED ISSUE #942: Update merge column list to be columns to exclude
Opening this as a placeholder issue to circle back to soon, we should update how we handle columns to include in the merged object:

https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/bin/merge_sces.R#L201-L202

Rather than a vector of columns to include, we should consider updating this to columns to exclude for future-proofing.

allyhawkins:
Closes #942 

As noted in the issue, rather than having a long list of columns to include in the merged objects to maintain, this PR changes to providing a list of columns to exclude from the merged objects. To get the exclude list I grabbed all columns from a recent processed SCE and then ran `setdiff()` with the current list of columns to keep. The only columns that were missing were `sizeFactor` and `cluster`, which makes sense since those aren't applicable anymore once the objects are merged. The `sizeFactor` column was the only one to exclude for ADT `colData`, so I just made one list and use it for subsetting columns in both the main and alt Exps. 

Just a note that the current [`merge_sce_list()` function from `scpcaTools`](https://github.com/AlexsLemonade/scpcaTools/blob/main/R/merge_sce_list.R) actually asks for a list of columns to retain. We could change this to be the columns to exclude instead, but I think I prefer being explicit when calling that function about what columns we want in the merged object. Because of that, I still make the list of columns to retain, but instead of hard coding a long list, I grab all columns in the objects and then remove any of the "exclude" columns. Let me know if you disagree with this approach and would prefer to see changes to the underlying function instead.


bin/merge_sces.R
```diff
@@ -198,57 +198,17 @@ for (annotation in names(annotation_list)) {
...
+# these apply to both the main and altExps
+exclude_columns <- c(
+  "sizeFactor",
+  "cluster"
 )
 
-# Define colData columns to retain based on intersection with present columns
-retain_coldata_columns <- intersect(possible_columns, present_columns)
+# Define colData columns to retain by removing columns to exclude
+retain_coldata_columns <- setdiff(possible_columns, exclude_columns)
```
jashapiro, line 209: I think you would have run into an error here!
```suggestion
retain_coldata_columns <- setdiff(present_columns, exclude_columns)
```

bin/merge_sces.R
```diff
@@ -198,57 +198,17 @@ for (annotation in names(annotation_list)) {
...
-  "altexps_adt_detected",
-  "altexps_adt_percent",
-  # cellhash statistics & demux results
-  "altexps_cellhash_sum",
-  "altexps_cellhash_detected",
-  "altexps_cellhash_percent",
-  "hashedDrops_sampleid",
-  "HTODemux_sampleid",
-  "vireo_sampleid"
+# Define columns we do not want to include in the merged object
```
jashapiro, line 201:
```suggestion
# Define colData columns we do not want to include in the merged object
```

jashapiro (approved):
I agree with your approach here, so just one (pretty important!) bug fix and small comment update

PR #1042, opened 2025-09-24
Make the diagnosis map file optional

sjspielman:
This is the first of three PRs now associated with #1005, based on this discussion: https://github.com/AlexsLemonade/scpca-nf/pull/1034#issuecomment-3328741593 

---

This PR does two main things:

1. Updates the code in `add_celltypes_to_sce.R` to no longer require the diagnosis map file. I have tested this with different metadata conditions. The code only uses the diagnosis map file if a) it exists, and b) it actually contains the sample diagnosis. This covers us for a scenario where users provide the group <-> celltype map file _but don't remove_ the default diagnosis map file, so we'll still have a diagnosis to fall back on. Implementing this didn't involve any nextflow changes, mostly - I just removed the check for the diagnosis map file.

2. Implements some diagnosis file cleanups, as I laid out in #1034:
  - I renamed the diagnosis map file to match its param name, removed unused columns, and changed the submitted_diagnosis -> sample_diagnosis. This level of diff resulted in GitHub declaring this a new file, not modified
  - I documented these changes in `internal-instructions.md`
  - I updated config files and the schema file (a couple things were missing/outdated there also, so a couple more tiny changes) accordingly

Let me know what you think!


bin/add_celltypes_to_sce.R
```diff
@@ -373,21 +372,36 @@ if (has_singler && has_cellassign) {
...
     consensus_validation_df <- readr::read_tsv(opt$consensus_validation_ref)
     diagnosis_celltype_df <- readr::read_tsv(opt$diagnosis_celltype_ref)
-    diagnosis_map_df <- readr::read_tsv(opt$diagnosis_groups_ref)
 
-    sample_diagnosis <- metadata(sce)$sample_metadata$diagnosis
+    diagnosis <- metadata(sce)$sample_metadata$diagnosis
+
+    # Get the broad_diagnosis from the map file if possible, or use the sample diagnosis otherwise
+    broad_diagnosis <- diagnosis
+    if (tidyr::replace_na(file.size(opt$diagnosis_groups_ref), 0) > 0) {
```
allyhawkins, line 380: I think you could change this check to just `if(file.exists())` and then make a few changes in Nextflow. 
Mainly https://github.com/AlexsLemonade/scpca-nf/blob/f33b3eb10b70d004d8aff4233003fee9fde7479b/modules/classify-celltypes.nf#L305 should be changed to `params.diagnosis_groups_file ? file(params.diagnosis_groups_file, checkIfExists: true) : []`

Then the `add_celltypes_to_sce` process will need to be changed to just pass in the path of the file like you normally would. Using `[]` will cause the path to be passed in as `""` which will cause `file.exists()` to be false.

allyhawkins (commented):
This mostly looks good, I just had one suggestion to clean up the logic a little bit. I think you could generally switch to using `[]` instead of empty file for these files.

bin/add_celltypes_to_sce.R
sjspielman: I will look into switching to `[]` for overall workflow consistency, but I do need to say for posterity that I'm sad to lose my `name:` approach which was a fun journey to test out: https://github.com/AlexsLemonade/scpca-nf/pull/987#issuecomment-3253776069

sjspielman: We may need to detour here: I have encountered a bug when running this script in the consensus cell type section of code here: https://github.com/AlexsLemonade/scpca-nf/blob/f33b3eb10b70d004d8aff4233003fee9fde7479b/bin/add_celltypes_to_sce.R#L344-L351

The joining with `many-to-many` is now giving us multiple rows per barcode with the updated reference file, so consensus cell types are not being assigned:
```
> dplyr::count(celltype_df, barcodes) |> head()
          barcodes n
1 AAACCCAAGTGGACGT 1
2 AAACCCAGTCTTCAAG 3
3 AAACGAAAGAGTCGAC 3
4 AAACGAAAGCAAGCCA 8
5 AAACGAATCCATACAG 1
6 AAACGCTTCACGACTA 3
```

In the interest of time since you're the expert on the consensus code, I may leave this for you to have a look at when you're back from AACR peds and catch you asynchronously later or next sprint!

sjspielman: Back today! I will develop/test code for this PR this specifying the previous references file, since I'm realizing it's probably not working since SCimilarity isn't yet in `scpca-nf`, so it should presumably sort itself out as we develop that.

modules/classify-celltypes.nf
```diff
@@ -135,8 +138,8 @@ process add_celltypes_to_sce {
       ${cellassign_present ? "--panglao_ontology_ref ${panglao_ref_file}" : ''} \
       --consensus_celltype_ref ${consensus_ref_file} \
       --consensus_validation_ref ${validation_ref_file} \
-      --diagnosis_celltype_ref "diagnosis_celltypes.txt" \
-      --diagnosis_groups_ref "diagnosis_groups.txt" \
+      ${has_diagnosis_celltypes ? "--diagnosis_celltype_ref ${diagnosis_celltypes_file}" : ''} \
+      ${has_diagnosis_groups ? "--diagnosis_groups_ref ${diagnosis_groups_file}" : ''} \
```
allyhawkins, line 142: Do you need the `has_diagnosis_celltypes` variable? I think you should just be able to pass in the argument as is. `[]` evaluates to `""`, which is the default for the files and if they don't exist then that part of the code won't be run, which is what we want. 

```suggestion
      --diagnosis_celltype_ref ${diagnosis_celltypes_file} \
      --diagnosis_groups_ref ${diagnosis_groups_file}
```

modules/classify-celltypes.nf
```diff
@@ -123,6 +123,9 @@ process add_celltypes_to_sce {
     singler_results = "${singler_dir}/singler_results.rds"
     cellassign_present = "${cellassign_dir.name}" != "NO_FILE"
     cellassign_predictions = "${cellassign_dir}/cellassign_predictions.tsv"
+
+    has_diagnosis_celltypes = diagnosis_celltypes_file && diagnosis_celltypes_file.isFile()
+    has_diagnosis_groups = diagnosis_groups_file && diagnosis_groups_file.isFile()
```
allyhawkins, line 128: I think you can delete these lines. 
```suggestion

```

allyhawkins (commented):
Thanks for changing to `[]`, I think it's much cleaner. I had one more suggestion to simplify but otherwise this should be good. 

The consensus assignment code is going to be updates as part of continuing to implement scimilarity. I plan to address that in my PRs, so if you are seeing failures there I'm not totally surprised.

allyhawkins (approved):
LGTM

PR #1044, opened 2025-09-30
Add SCimilarity to cell typing processes

### LINKED ISSUE #1029: Process for running `SCimilarity`
We will need to add a new process for running `SCimilarity` to the `classify-celltypes.nf` module. This process should mirror the process we use for `CellAssign` where the processed RDS files are read in, converted to AnnData, and then annotated with `SCimilarity`. The output of the process should be the TSV file with annotations and any metrics we want to add into the processed SCE objects. 

When we do this we will want to add a new parameter for the path to the `SCimilarity` reference. The name of the reference/ model version should be tracked in the `scpca-meta.json` that lives in the cell type checkpoints folder so that we can skip `SCimilarity` if the versions are the same and the files already exist. This mirrors how we handle skipping `SingleR` and `CellAssign`.

allyhawkins:
Closes #1029 

Here I am adding a process for classifying cells using `SCimilarity` to the `classify-celltypes.nf` module. The actual process mirrors the CellAssign process where first the objects are converted to AnnData and then `SCimilarity` is run using the `.h5ad` file as input. The output is a TSV file with cell type predictions that ultimately will be added to the processed SCE object. The next PR will handle that addition, so here I stop just at running `SCimilarity`. 

- I added two new parameters, one for the `scimilarity_model_dir` and one for `scimilarity_ontology_map_file`. I made both of these required if the `--perform_celltyping` flag is true. Unlike with `SingleR` and `CellAssign`, there's no option for the model directory to be missing since the same model is used on all samples, but let me know if you disagree with this and we should have an option to not provide a model directory? 
- As with the other cell typing methods, I created a `checkpoints/celltype/scimilarity/<library id>/scpca-meta.json` file that gets exported. I added the path to the model directory and the ontology map file to the `meta` object and then if SCimilarity results exist, I check that the model directory is the same as the previous one to confirm that we should skip it. Are there other things we should check here? 
- While here, I made some other small changes to move away from using the `empty_file` option and instead just use `[]`. We are doing this in #1042 also. 
- I also made a temporary fix to the consensus cell type code to get around that failing without the addition of scimilarity. This will be modified as part of the next PR.


modules/classify-celltypes.nf
```diff
@@ -237,23 +302,28 @@ workflow annotate_celltypes {
...
+
+    /////////////////////////////////////////////////////
+    //                  CellAssign                     //
+    /////////////////////////////////////////////////////
+
     // create cellassign input channel: [meta, processed sce, cellassign reference file]
     cellassign_input_ch = celltype_input_ch
       // add in cellassign reference
-      .map{it.toList() + [file(it[0].cellassign_reference_file ?: empty_file, checkIfExists: true)]}
+      .map{it.toList() + [it[0].cellassign_reference_file ? file(it[0].cellassign_reference_file, checkIfExists: true) : [] ]}
```
jashapiro, line 318: making the map explicit
```suggestion
      .map{meta, processed_sce ->
        def cellassign_ref = meta.cellassign_reference_file ? file(meta.cellassign_reference_file, checkIfExists: true) : []
        [meta, processed_sce, cellassign_ref]
```

modules/classify-celltypes.nf
```diff
@@ -268,11 +338,53 @@ workflow annotate_celltypes {
...
+    // run SCimilarity and export results
+    classify_scimilarity(scimilarity_input_ch.do_scimilarity)
+
+    // scimilarity output channel: [unique id, scimilarity_dir]
+    scimilarity_output_ch = scimilarity_input_ch.skip_scimilarity
+      // provide existing scimilarity predictions dir for those we skipped
+      .map{ meta, _processed_sce, _scimilarity_model_dir, _scimilarity_ontology_map_file -> tuple(
+        meta.unique_id,
+        file(meta.scimilarity_dir, type: 'dir', checkIfExists: true)
+      )}
```
jashapiro, line 380: Trying to think about keeping style consistent for the future. (This is a bigger conversation about style guides, but Stephanie & I seemed to agree on trailing `->` for multiline lambda functions like this
```suggestion
      .map{ meta, _processed_sce, _scimilarity_model_dir, _scimilarity_ontology_map_file ->
         tuple(meta.unique_id, file(meta.scimilarity_dir, type: 'dir', checkIfExists: true))
      }
```

jashapiro (commented):
This looks good. The only think I really want to see added is the stub file for the scimilarity model directory. I would probably include there at least one file name that you expect to have in the directory, and also maybe check for the presence of that file when checking that the directory was given as an input?

Most of the other comments are about style and future things, some of which are for some future discussion (like whether we want to keep channel contents as tuples or use lists for convenience; right now we kind of do both, with no particular logic). Feel free to incorporate what you like there; I have a feeling we will sweep through to make these more consistent at some point before the next release.

allyhawkins:
> This looks good. The only think I really want to see added is the stub file for the scimilarity model directory. I would probably include there at least one file name that you expect to have in the directory, and also maybe check for the presence of that file when checking that the directory was given as an input?

I included a stub for the model directory already, but can add one of the files. Do you mean to check for the presence of that file in Nextflow or in the python script? I'm assuming you mean Nextflow, since that will fail first, but just wanted to be sure.

modules/classify-celltypes.nf
```diff
@@ -102,6 +102,64 @@ process classify_cellassign {
     """
 }
 
+process classify_scimilarity {
+  container params.SCPCATOOLS_SCIMILARITY_CONTAINER
+    publishDir (
+      path: "${meta.celltype_checkpoints_dir}",
+      mode: 'copy',
+      pattern: "${scimilarity_dir}"
```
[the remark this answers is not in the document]
jashapiro: Sorry, I meant one "file" object, which is actually a directory in this case.

jashapiro:
> I included a stub for the model directory already, but can add one of the files. Do you mean to check for the presence of that file in Nextflow or in the python script? I'm assuming you mean Nextflow, since that will fail first, but just wanted to be sure.

Yes, I meant checking in Nextflow during the parameter checks. And you might have a stub directory there, but if there are no files in it, it isn't in the git repo. So you need at least something there, and it might as well be an expected filename rather than just `.gitkeep`

bin/add_celltypes_to_sce.R
```diff
@@ -331,7 +331,9 @@ if (has_singler && has_cellassign) {
   # read in consensus table
   consensus_ref_df <- readr::read_tsv(opt$consensus_celltype_ref) |>
     # select unique combinations of consensus refs based on ontology columns
-    dplyr::select(blueprint_ontology, panglao_ontology, consensus_ontology, consensus_annotation) |>
+    # TODO: Update when incorporating scimilarity
+    dplyr::select(blueprint_ontology, panglao_ontology, consensus_ontology = cellassign_singler_pair_ontology, consensus_annotation = cellassign_singler_pair_annotation) |>
```
jashapiro, line 335: just noting the long line, but since this will be updated, not a big deal
```suggestion
    dplyr::select(
      blueprint_ontology,
      panglao_ontology,
      consensus_ontology = cellassign_singler_pair_ontology,
      consensus_annotation = cellassign_singler_pair_annotation
    ) |>
```

modules/classify-celltypes.nf
```diff
@@ -102,6 +100,61 @@ process classify_cellassign {
...
+    )
+  label 'mem_96'
+  label 'cpus_4'
+  tag "${meta.library_id}"
+  input:
+    tuple val(meta), path(processed_rds), path(scimilarity_model_dir), path(scimilarity_ontology_map_file)
+  output:
+    tuple val(meta.unique_id), path(scimilarity_dir)
+script:
+    scimilarity_dir = file(meta.scimilarity_dir).name
```
jashapiro, line 117: Okay, it is odd that you can't use `def` here. I wonder if it is "defined" when first referenced in the output?

jashapiro (approved):
LGTM!

PR #1051, opened 2025-10-08
Test support for 10xv2.1

### LINKED ISSUE #1041: [BUG] Error handling 10x v2.1 files
**Describe the bug**

A user reports that when running a sample created with the apparently briefly-available 10x v2.1 kit, reporting it as  10Xv2 to the workflow results in only a few hundred cells, whereas if reported as 10Xv3, it works as expected, giving results similar to Cell Ranger.

**To reproduce**

I am hoping to get some small example files so I can start to reproduce a bit of this (the mapping phase specifically)

**Expected behavior**

Results for v2.1 should mirror Cell Ranger

**Version information**

v0.8.8

**Additional context**

I can not find information about what actually changed for v2 vs v2.1. I know that both the barcode list and the UMI length changed from v2 to v3; it is possible that one or the other changed in v3, which could affect the running of salmon/alevin or the barcode whitelist.  (If the problem is that v2.1 updated the barcodes but left the UMI length unchanged, then I expect the runs with v3 will be perfectly valid: the inferred UMI may be a bit too long, but if the sequence following the UMI is constant, any inference from the UMIs will still be correct)

I have requested a bit of additional information to debug, and will update here as I do.

jashapiro:
From the best we can see, the 10x v2.1 uses  v2 barcodes, but perhaps with the longer UMI used for v3. Under this assumption, this PR changes all steps to allow for appropriate mapping with alevin-fry.

I also added checks in case the technology changes that force remapping, as I wondered if that might have also been part of the initial error that was reported. 

While I was in there, I also expanded support for the v4 file, adding the barcode lists for citeseq and cellhash modalities, in case those come up.

hopefully closes #1041


jashapiro:
These changes were not needed, https://github.com/AlexsLemonade/scpca-nf/issues/1041 was closed separately.

PR #1058, opened 2025-10-13
Scimilarity metrics section to supplemental report

### LINKED ISSUE #1031: Include `SCimilarity` in the report
We will want to update the report to include the `scimilarity_celltype_annotation` column as one of the automated cell typing methods to display. This should include: 

- A description of `SCimilarity`
- The same tables, UMAPs, and heatmaps that we show for the other automated methods 
- A new section looking at metrics for plotting the `min_dist`. As mentioned in the [tutorial](https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html#Annotation-QC), this metric can be used to help assess confidence in the assignments. I think we want to make a similar plot to the ridge plot we show for `CellAssign` across all cell types.

allyhawkins:
Closes #1031 

Here I'm adding a new section to the automated annotations QC for `SCimilarity` results. As noted in the [cell type annotation vignette](https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html#Annotation-QC), we can use the `min_dist` metric as a confidence measure. 

I ended up using a similar plot to the `SingleR` plot, where the annotation is on the y-axis and the x-axis is the minimum distance metric. This illustrates the distribution of the minimum distance across all cells for each predicted cell type. I played around with UMAPs too, but ultimately the sina plot does a better job of displaying the distribution and gives the information we are looking for. 

As noted in the vignette, they use a value of 0.05 where any cell with a minimum distance from the query cells > 0.05 has less confidence in the prediction. I mentioned this in the description and then also added a dashed line for this value, which I think is helpful when we have lots of cell types and the plot gets really long. 

Are there other things that we should be sure to include in the text or point out about these results? 

While here, I also moved the `lump_wrap_celltypes()` function back to the `celltype-functions.rmd` file. I noticed in #1022, where it was removed, there was a comment that the reports were generated without the celltype sections, which is probably why it needed to be moved. In reality, `inferCNV` does not run without cell typing, so it did not need to be moved and addresses the concern brought up in https://github.com/AlexsLemonade/scpca-nf/pull/1049#pullrequestreview-3325710636.

Here's a copy of the supplemental report for three different libraries:
[reports.zip](https://github.com/user-attachments/files/22892167/reports.zip)

### DIFF
--- a/templates/qc_report/celltypes_supplemental_report.rmd
+++ b/templates/qc_report/celltypes_supplemental_report.rmd
@@ -543,7 +543,7 @@ You can interpret this plot as follows:
   In other sections of this report, these cells are labeled as `Unknown cell types`.
   - For more information on how `SingleR` calculates annotation quality, please refer to [this `SingleR` documentation](https://rdrr.io/bioc/SingleR/man/pruneScores.html).
 - Diamonds represent the median of the delta median statistic specifically among high-quality annotations for the given cell type annotation.
-- Cell types with fewer than 3 labeled cells are labeled as 'Other' in the plot.
+- Cell types with fewer than 3 labeled cells are combined and labeled as 'Other' in the plot.
 ")
 ```
 
@@ -680,7 +680,7 @@ The plot below shows the distribution of `CellAssign`-calculated probabilities f
 Line segments represent individual values that comprise each distribution.
 
 Line segments are taller for any cell type label with 5 or fewer cells.
-Cell types with fewer than 3 labeled cells are labeled as 'Other' in the plot.
+Cell types with fewer than 3 labeled cells are combined and labeled as 'Other' in the plot.
 ")
 
 # Determine height in inches for plot area
@@ -781,7 +781,7 @@ The plot below shows the distribution of the minimum distance metric reported by
 Diamonds represent the median of the minimum distance metric among all cells assigned a given cell type annotation.
 Cells to the right of the dashed line are those for which the model is lesonfident in the prediction.
 
-Cell types with fewer than 3 labeled cells are labeled as 'Other' in the plot.
+Cell types with fewer than 3 labeled cells are combined and labeled as 'Other' in the plot.
 ")
 ```
 


templates/qc_report/celltypes_supplemental_report.rmd
```diff
@@ -85,12 +85,13 @@ has_clusters <- "cluster" %in% names(colData(processed_sce))
 
 # what celltypes are available?
 available_celltypes <- c(
+  # keep consensus as the first method so it shows up first in comparison plots
+  ifelse(has_consensus, "Consensus", NA),
   ifelse(has_submitter, "Submitter", NA),
   ifelse(has_openscpca, "OpenScPCA", NA),
```
jashapiro: I was thinking just having the consensus before the three automated, so keeping the Submitter and OpenScPCA first: 
```suggestion
  ifelse(has_submitter, "Submitter", NA),
  ifelse(has_openscpca, "OpenScPCA", NA),
  ifelse(has_consensus, "Consensus", NA),
```

jashapiro: I realized that this whole section can actually be simplified a bit by not using `ifelse`, but just regular `if` statements: 
```
# what celltypes are available?
available_celltypes <- c(
  if(has_submitter) "Submitter",
  if(has_openscpca) "OpenScPCA",
  if(has_consensus) "Consensus",
  if(has_singler) "SingleR",
  if(has_cellassign) "CellAssign",
  if(has_scimilarity) "SCimilarity"
)

# use an empty character vector if no celltypes present
if(is.null(available_celltypes){
  available_celltypes <- character()
}

```

This should save the `NA` removal and conversion, though you need to add a check that you don't end up with just `NULL`

allyhawkins:
Alright, I added back "other" and made sure they appear in the plots at the bottom of the list but before unknowns, which only applies to CellAssign. 

I also simplified the available celltypes chunk that you suggested.

PR #1060, opened 2025-10-16
Add inferCNV meta json and update checkpointing

### LINKED ISSUE #1039: Use a json file to support checkpoints in the inferCNV module
Related to #1025 

We are currently not exporting any `scpca-meta.json` file in the `inferCNV` module, which could be used to assist in checkpointing as we do for cell type assignment. This issue tracks adding this file, although still in progress is finalizing the information to include in it. Regardless of what information this contains, we would need to update the `run_infercnv` process to export it to the checkpoint directory.

----

Here's what I am now thinking for information to store: The main circumstance when we would want to ensure a repeat is when the set of reference cells has changed. There are several conditions under which they might change, including:
  - A change which cells are included the reference
  - Any kind of mapping file change which could include:
    - the diagnosis group/cell type mapping
    - the validation group/consensus cell types mapping

It has been proposed to track the _number_ of cells in the reference, but (it's unlikely but possible!) this number could in theory stay the same just with a different set of cells. But, if we track the literal reference cell barcodes as a json array, I'm pretty sure this would cover our bases for any circumstance. For this, we'd want to capture that array in the meta json at the same time of (instead of?) the reference cell count to use for comparison later; "instead of" because the length of that array could be checked instead of the reference cell count. Implementing this would entail:
- The script `add_celltypes_to_sce.R` would export a file listing barcodes `barcode1,barcode2,barcode3` and stash this cat'd content as an environment variable, as we currently do for the reference cells: https://github.com/AlexsLemonade/scpca-nf/blob/fb5ba02c988848f3d0b17d6699adc2add56e952b/modules/classify-celltypes.nf#L142-L143
- Convert this to a list when saving to `meta` 
- To determine whether to run inferCNV in the first place, we could check the length of this array rather than the reference cell count
- When the time comes to branch for inferCNV, compare this meta list to the array stored in the checkpoint json, which as noted will need to be created within the `run_infercnv` process, likely from R (not nextflow) since we'll be exporting barcodes as seen in the processed SCE.

 
We may also want to repeat if the data itself has changed, but this might be unnecessary. For example, in the cell type module we do allow for "old" singler/cellassign results run on a previous data version to make it through the pipeline. In these cases, we include a message about in the QC to explain why some cells aren't cell typed: https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/templates/qc_report/celltypes_qc.rmd#L342-L344 We could adopt a similar strategy for inferCNV; note that this is the scenario where we'd get `NA` values in the `infercnv_total_cnv` column.

That said, a way to do "both" would be to export both sets of barcodes, for cells in the reference and cells in the query. If there are any mismatches, we could trigger a repeat.

sjspielman:
Closes #1039 

This PR updates the inferCNV bits as follows:
- Save a hash for the sorted/collapsed barcodes to meta for comparing to the checkpoint. This is pretty analogous to the existing code that saves the reference cell count
- When branching for skip vs. run inferCNV, include the hash comparison in the logic
- Save meta as a json in the `run_infercnv` process
  - Note that I have to output this file to ensure it's actually published, but I drop it immediately. Let me know if you prefer a different file management strategy here
- One recreational change, renaming `infercnv_dir` to `infercnv_checkpoints_dir`; it's longer (which is why I originally didn't use it), but I think it's better to be explicit especially if we're writing checkpoint jsons now. 

I've tested this and it seems to behave as expected! Doesn't repeat when the hash matches, repeats when the hash has changed.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -111,6 +111,13 @@ option_list <- list(
       This calculation is only performed if `diagnosis_celltype_ref`, `diagnosis_groups_ref`, and `consensus_validation_ref` are provided.
       If not calculated, this file will be created with the value NA instead of a count",
     default = "reference_cell_count.txt"
+  ),
+  make_option(
+    opt_str = c("--reference_cell_hash_file"),
+    type = "character",
+    help = "Path to write out unique hash for all concatenated barcodes in reference cell set; used for checkpointing.
+      If not calculated, this file will be created with the value NA instead of a hash",
+    default = "reference_cell_hash.txt"
   )
 )
 
@@ -161,6 +168,7 @@ sce <- readr::read_rds(opt$input_sce_file)
 
 # We'll count inferCNV reference cells when assigning consensus cell types
 reference_cell_count <- NA_integer_
+reference_cell_hash <- NA_character_
 
 # set automated methods list
 # we'll use this to assign consensus and add cell type methods to the metadata
@@ -514,11 +522,19 @@ if (assign_consensus) {
 
     # get the full count
     reference_cell_count <- sum(sce$is_infercnv_reference)
+
+    # get hash for concatenated barcodes
+    concat_barcodes <- paste(
+      sce$barcodes[sce$is_infercnv_reference],
+      collapse = ""
+    )
+    reference_cell_hash <- digest::digest(concat_barcodes, algo = "sha256")
   }
 }
 
 # export annotated object with cell type assignments
 readr::write_rds(sce, opt$output_sce_file, compress = "bz2")
 
-# export normal cell count
+# export normal cell count and hash
 readr::write_lines(reference_cell_count, opt$reference_cell_count_file)
+readr::write_lines(reference_cell_hash, opt$reference_cell_hash_file)
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -109,14 +109,14 @@ option_list <- list(
     type = "character",
     help = "Path to write number of calculated inferCNV reference cells to.
       This calculation is only performed if `diagnosis_celltype_ref`, `diagnosis_groups_ref`, and `consensus_validation_ref` are provided.
-      If not calculated, this file will be created with the value NA instead of a count",
+      If not calculated, this file will be empty",
     default = "reference_cell_count.txt"
   ),
   make_option(
     opt_str = c("--reference_cell_hash_file"),
     type = "character",
     help = "Path to write out unique hash for all concatenated barcodes in reference cell set; used for checkpointing.
-      If not calculated, this file will be created with the value NA instead of a hash",
+      If not calculated, this file will be empty",
     default = "reference_cell_hash.txt"
   )
 )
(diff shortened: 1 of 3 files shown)


bin/add_celltypes_to_sce.R
```diff
@@ -111,6 +111,13 @@ option_list <- list(
       This calculation is only performed if `diagnosis_celltype_ref`, `diagnosis_groups_ref`, and `consensus_validation_ref` are provided.
       If not calculated, this file will be created with the value NA instead of a count",
     default = "reference_cell_count.txt"
+  ),
+  make_option(
+    opt_str = c("--reference_cell_hash_file"),
+    type = "character",
+    help = "Path to write out unique hash for all concatenated barcodes in reference cell set; used for checkpointing.
+      If not calculated, this file will be created with the value NA instead of a hash",
```
jashapiro:
```suggestion
    help = "Path to write out unique hash for all concatenated barcodes in reference cell set; used for checkpointing.
      If not calculated, this file will be empty",
```

bin/add_celltypes_to_sce.R
```diff
@@ -519,11 +527,19 @@ if (assign_consensus) {
...
   }
 }
 
 # export annotated object with cell type assignments
 readr::write_rds(sce, opt$output_sce_file, compress = "bz2")
 
-# export normal cell count
+# export normal cell count and hash
 readr::write_lines(reference_cell_count, opt$reference_cell_count_file)
+readr::write_lines(reference_cell_hash, opt$reference_cell_hash_file)
```
jashapiro: Prevent trailing `\n` so we get empty files if empty. 
```suggestion
readr::write_file(reference_cell_count, opt$reference_cell_count_file)
readr::write_file(reference_cell_hash, opt$reference_cell_hash_file)
```

jashapiro (commented):
This looks good overall.

I added a few suggestions to simplify a bit. The first one is that we can have multiple named outputs, so that removes the need to do extra processing after output, but still send the scpca-meta.json files for publication.

The other main one is to change the script output a bit to simplify some of the later logic: just write an empty file if the calculations were not done.

jashapiro:
I realized that our stub runs are not checking the full workflow very well, as they do not include cell typing or CNV inference. We should add the appropriate flags set to true in `config/profile_stub.config`, and then make sure we have any of the needed stub files that might be missing.

sjspielman:
> I realized that our stub runs are not checking the full workflow very well, as they do not include cell typing or CNV inference. We should add the appropriate flags set to true in config/profile_stub.config, and then make sure we have any of the needed stub files that might be missing.

I thought I had filed an issue for this at one point but I guess I didn't! For a while I was adding it into stub locally/manually and I intended at least to follow up more formally.... Maybe it's best to put this PR on hold for a moment and pivot to get that in.

jashapiro:
> Because last time I put too much in one PR I got called out ;)

Fair, but in this case it is related to testing this PR, not a totally separate thing.

jashapiro (approved):
LGTM
(the stub check stalled for some reason, but I kicked it and it ran the next time)

PR #1064, opened 2025-10-20
Make sure run01/librar01 don't have an extra CITE-seq sample

### LINKED ISSUE #879: Update example instructions and data to remove CITE-seq from run01
> Just as a side note, the only run besides these two that can actually be used in real life for processing is `run01`. We don't have the FASTQs or any other data available for the other samples. Mainly because we instruct people to download their own example data rather than running the workflow with the data that's on S3. We could update the example directory as a whole to document being able to use the example profile to process example data and ensure that any samples in the example metadata file actually have corresponding data, but that is something we should do separately from the 10X flex support if we want to do it.

That makes some sense. I realized that was probably the case when I looked at the example profile and saw that it had that run specified. I was mostly confused I think by `run01` having a CITE-seq run associated with it. It might make sense to have that be a separate sample.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/issues/877#issuecomment-2895322793_
            
We should remove the CITE-seq library from library01 (run01 and run02) to have an example that is just single-cell and one that is single-cell with CITE in the example metadata file. We could also update instructions as mentioned in the comment above, but that is a nice to have item.

allyhawkins:
Closes #879 

Here I'm updating the example run metadata to make sure that the first library listed is separate from the example CITE-seq library. We don't actually run anything but `run01` when running example data, so it was confusing to also have CITE-seq with that library that doesn't get used. 

I just added a new library that includes the CITE-seq one and updated the run/library numbering to match that. I did not change the sample numbering since we can have more than one library per sample and this helps illustrate that. 

I also set the default for running the example profile to include running `inferCNV`. I think we want to have that so we always generate the full example output. Once AWS is back I'll actually regenerate the files, but this takes care of any of the changes needed. 

I also noticed an error when describing the organization of the bulk output files that I fixed.

### DIFF
--- a/external-instructions.md
+++ b/external-instructions.md
@@ -575,7 +575,7 @@ results
         └── {library_id}_qc.html
 ```
 
-If bulk libraries were processed, the following three files will also be present in the `project_id` directory:
+If bulk libraries were processed, the following three files will also be present in the `project_id/bulk` directory:
 - `{project_id}_bulk_metadata.tsv`: a summary table of metadata across all project bulk libraries
 - `{project_id}_bulk_quant.tsv`: a matrix of gene expression count values for all bulk libraries
 - `{project_id}_bulk_tpm.tsv`: a matrix of gene expression TPM values for all bulk libraries


sjspielman (approved):
LGTM! Can't help with the queued precommit though, looks like both of your jobs are queued https://results.pre-commit.ci/repo/github/386739699.... let's blame aws? 🤷‍♀️

sjspielman:
>  looks like both of your jobs are queued https://results.pre-commit.ci/repo/github/386739699.... let's blame aws? 🤷‍♀️

Clearly I un-queued one by complaining about it*! 🎉 Maybe this one will make it through soon then

Issue #1066, opened 2025-10-21
Ensure merge.nf gets fully run in stub


sjspielman:
I realized while working on/testing #1035 that the `merge.nf` workflow doesn't entirely get run in stub. This is because the workflow checks that the processed files have a size greater than 0, which they won't for the test files that are being input. So, no process ends up being triggered. 

We should come up with a workaround for this so `merge.nf` actually gets tested all the way through. Probably something to consider as part of #1045, too.

sjspielman:
Just noting it's here! https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/main.nf#L212-L215

Issue #1035, opened 2025-09-22
Update workflow style to be more consistent


sjspielman:
> The style questions are mostly about consistency that we have not to this point been very good about, so they really don't matter too much, but we might want to start thinking about some standards. Specifically things like: 
> - Do we care about specifying channel entries as tuples or lists? (Personally, I think lists are nicer for reading, but technically channel elements come out of processes as tuples, so there is an argument to be made there about consistency.)
> - Should we favor one line return statements for our lambdas?
> - Should tuple arrows be the end of the line?

> These are all trivial things, but it would be good to be consistent. Maybe we should make an issue to try to clean all this up before the next release.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/1026#pullrequestreview-3246840305_

---

This issue tracks making some style decisions and implementing them in `scpca-nf` before we cut the next release. We may also want to include a "style guide" type section in internal instructions or contributing.md to support future workflow additions.

jashapiro:
> Do we also want to take this opportunity to pivot all the `NO_FILE` usage to `[]`? I'm on the fence for it...

I'd say yes, but probably in a separate PR

PR #1067, opened 2025-10-21
Workflow style: top-level nextflow files

sjspielman:
Towards #1035 
This is the first of several PRs to tackle style updates. I decided to split these PRs up by groups of files, rather than by specific type of change, since that means reviewers ultimately need to look at fewer files per PR.

For this first one, I have updated style for all the `.nf` files in the top-level of the repo (aka, everything that is not in `modules/` - that will come next in 2 more PRs for your reviewing sanity). In addition to updating explicit lambda variables in `.map{}`, I also did this for the other closure operations like filter, branch, and more, since all of this seems to be needed for future proofing. Stub is running (locally at least; edit and here too!) for me for all workflows, except I realized stub isn't really running `merge.nf` (see #1066), so that's something to bear in mind...

Happy checking and let me know if/where I missed any spots, or if anything needs to be further stylized. I'm going to wait to file remaining PRs until this has been reviewed since I imagine insights from this review will inform completing the rest.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -157,8 +157,8 @@ workflow {
   ref_ch = Channel.fromPath(params.ref_metadata)
     .splitCsv(header: true, sep: '\t')
     .map{ it ->
-      def reference_name = "${it.organism}.${it.assembly}.${it.version}";
-      def ref_name_paths = ref_paths[reference_name];
+      def reference_name = "${it.organism}.${it.assembly}.${it.version}"
+      def ref_name_paths = ref_paths[reference_name]
       // return reference name & reference file paths for each organism
       // return this is as a map (dictionary) so we can refer to items by name
       [
@@ -200,7 +200,7 @@ workflow {
   infercnv_ref_ch = ref_ch
     .filter{ it.include_infercnv }
     .map{ it ->
-      def cytoband_path = file("${params.ref_rootdir}/${it.ref_paths["cytoband"]}");
+      def cytoband_path = file("${params.ref_rootdir}/${it.ref_paths["cytoband"]}")
       [it.ref_name, it.ref_paths, it.gtf_path, cytoband_path]
     }
 
--- a/main.nf
+++ b/main.nf
@@ -195,7 +195,7 @@ workflow {
   // convert row data to a metadata map, keeping columns we will need (& some renaming) and reference paths
   unfiltered_runs_ch = all_runs_ch.valid
     .map{ it ->
-      def sample_refs = ref_paths[it.sample_reference];
+      def sample_refs = ref_paths[it.sample_reference]
       [
         run_id: it.scpca_run_id,
         library_id: it.scpca_library_id,
--- a/merge.nf
+++ b/merge.nf
@@ -203,8 +203,8 @@ workflow {
   // print out warning message for any libraries not included in merging
   filtered_libraries_ch.single_sample
     .map{ it ->
-      def processed = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds");
-      def meta_json = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json");
+      def processed = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds")
+      def meta_json = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
       [it.library_id, processed, meta_json ]
     }
     .subscribe{ library_id, processed, meta_json ->
@@ -221,8 +221,8 @@ workflow {
 
   grouped_libraries_ch = filtered_libraries_ch.single_sample
     .map{ it ->
-      def processed = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds");
-      def meta_json = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json");
+      def processed = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds")
+      def meta_json = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
       [it.project_id, it.library_id, processed, meta_json]
     }
     // only include libraries that have been processed through scpca-nf and have at least 3 cells


build-cellbrowser.nf
```diff
@@ -52,26 +52,26 @@ workflow {
...
+    .filter{ it ->
       run_all
       || (it.run_id in run_ids)
       || (it.library_id in run_ids)
       || (it.sample_id in run_ids)
       || (it.project_id in project_ids)
     }
-    .unique{ it.library_id }
-    .map{it -> [
+    .unique{it -> it.library_id }
```
jashapiro: I think for single values like this, we can skip the `it ->`
```suggestion
    .unique{ it.library_id }
```

merge.nf
```diff
@@ -158,100 +158,96 @@ workflow {
...
+    .map{ it ->
+      [
+        it.library_id,
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds"),
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
+      ]
     }
-    else if (Utils.getMetaVal(it[2], "processed_cells") < 3){
-      log.warn("Library ${it[0]} has fewer than 3 cells. This library will not be included in the merged object.")
+    .subscribe{ it ->
```
jashapiro: For something where we are using multiple values (especially all of them as here), I think unpacking is worthwhile for clarity
```suggestion
    .subscribe{ library_id, processed_rds, meta_json ->
```

merge.nf
```diff
@@ -158,100 +158,96 @@ workflow {
...
+        it.library_id,
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds"),
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
+      ]
+    }
     // only include libraries that have been processed through scpca-nf and have at least 3 cells
-    .filter{it[2].exists() && it[2].size() > 0 && Utils.getMetaVal(it[3], "processed_cells") >= 3}
-    .map{it[0..2]} // remove metadata file from tuple
+    .filter{it -> it[2].exists() && it[2].size() > 0 && Utils.getMetaVal(it[3], "processed_cells") >= 3}
+    .map{it -> it[0..2]} // remove metadata file
```
jashapiro: unpacking might be worth doing here for clarity too

main.nf
```diff
@@ -243,24 +243,24 @@ workflow {
...
-    .collect{it.sample_id}
+    .collect{it -> it.sample_id}
 
   // get genetic multiplex libs with all bulk samples present
   genetic_multiplex_libs = runs_ch.rna
-    .filter{!params.skip_genetic_demux} // empty channel if skipping genetic demux
-    .filter{it.sample_id.contains(",")}
-    .filter{it.sample_id.tokenize(",").every{it in bulk_samples.getVal()}}
-    .collect{it.library_id}
+    .filter{it -> !params.skip_genetic_demux} // empty channel if skipping genetic demux
```
jashapiro: This one doesn't even use `it`! 
```suggestion
    .filter{ !params.skip_genetic_demux } // empty channel if skipping genetic demux
```

build-cellbrowser.nf
```diff
@@ -52,26 +52,26 @@ workflow {
   libraries_ch = Channel.fromPath(params.run_metafile)
     .splitCsv(header: true, sep: '\t')
     // filter to run all ids or just specified ones
-    .map{it -> [
+    .map{ it -> [
```
allyhawkins: So my understanding was that we were going to have any multi line statements have the lambda at the end so this would be: 
```suggestion
    .map{ it -> 
      [
```

jashapiro (commented):
Overall I think this looks pretty good, but I actually think we can back off just a bit on declaring _every_ lambda. `filter { it.thing }` seems fine to me, and I don't mind keeping those as they are. 

`filter { it[3] }` is a bit less readable, so I suggested we change the early step in the build-index workflow to build a dictionary so we can use that more easily for later filters. In some other cases I think we are probably better off unpacking for clarity. 

One much more minor comment is that we should try to be consistent with how we space around `{}`. I just looked at how the Nextflow formatter handles it, and it uses a space before and after `{` and before `}`. I think we probably want to just go with their standard formatting. (`()` and `[]` do not have spaces.) I'm still resisting just enabling the formatter globally, because I like more indentation in process definitions, but I might cave.

build-index.nf
```diff
@@ -172,30 +172,30 @@ workflow {
       ]
     }
     // filter to only regenerate specified references
-    .filter{ build_all || it[0] in params.build_refs.tokenize(",") }
+    .filter{ it -> build_all || it[0] in params.build_refs.tokenize(",") }
 
   // filter to relevant references and drop the boolean flags
   salmon_ref_ch = ref_ch
-    .filter{ it[2] }
+    .filter{ it -> it[2] }
```
allyhawkins: Same with all these filtering statements. I thought the requirement was just for `.map` calls?

main.nf
sjspielman: 😵‍💫

build-celltype-ref.nf
```diff
@@ -170,17 +170,17 @@ workflow build_celltype_ref {
...
+
   // singler refs to download and train
   singler_refs_ch = celltype_refs_ch.singler
-    .map{[
-      ref_name: it.celltype_ref_name,
-      ref_source: it.celltype_ref_source
-    ]}
+    .map{ ref ->
+      [ref.celltype_ref_name, ref.celltype_ref_source]
+    }
```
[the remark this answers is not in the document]
sjspielman: Yeah, I think this change is fine, and maybe even good b/c it contributes to consistency? For each such change I made the result is still compatible with the workflow logic and relevant process inputs.

merge.nf
```diff
@@ -140,118 +140,121 @@ workflow {
...
 
   oversized_projects = libraries_ch
-    .filter{it.technology.startsWith("10x")} // only count single-cell or single-nuclei libraries, no cell hash, ADT, bulk or spatial
-    .map{[
-      it.project_id, // pull out project id for grouping
-      it
-    ]}
+    .filter{ it.technology.startsWith("10x") } // only count single-cell or single-nuclei libraries, no cell hash, ADT, bulk or spatial
+    // pull out project id for grouping
+    .map{ it -> [it.project_id, it] }
```
jashapiro: I think this helps clarify the `it[1]` below
```suggestion
    // pull out project id and library id for grouping and counting
    .map{ it -> [it.project_id, it.library_id] }
```

merge.nf
```diff
@@ -140,118 +140,121 @@ workflow {
...
-      log.warn("Processed files do not exist for ${it[0]}. This library will not be included in the merged object.")
-    }
-    else if(!(it[2].exists() && it[2].size() > 0)){
-      log.warn("Metadata file does not exist for ${it[0]}. This library will not be included in the merged object.")
+    .map{ it ->
+      [
+        it.library_id,
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds"),
+        file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
+      ]
```
jashapiro: I think this might be clearer?
```suggestion
      def processed = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds")
      def meta_json = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
      [it.library_id, processed, meta_json ]
```

build-celltype-ref.nf
```diff
@@ -74,9 +72,8 @@ process train_singler_models_flex {
   output:
     path celltype_model
   script:
-    ref_file_basename = file("${ref_file}").baseName
     date_str = java.time.LocalDate.now().toString() // get current date in ISO8601 format
```
allyhawkins: Same comments here about `def`?

allyhawkins (commented):
This looks good, but I did have one question about if we want to be using `def` in the `script` blocks or not? I feel like in more recent PRs, we've been using that.

jashapiro:
> This looks good, but I did have one question about if we want to be using `def` in the `script` blocks or not? I feel like in more recent PRs, we've been using that.

I feel like I remember that sometimes this causes trouble, especially if the variable is used in tags or outputs. I think if it is in an output it gets implicitly defined, so then the `def` fails.

sjspielman:
Also want to note - I caught a bad thing I did! https://github.com/AlexsLemonade/scpca-nf/pull/1067/commits/6758e5a4daff0f815a35cb0cd2b72a1399852d46

So I'm going to briefly pause and tackle #1066 to protect me from me before merging this so we don't merge a bad merge.nf (obligatory i heard you like merge so...)

jashapiro:
I'd actually vote for no `def` in `script:` statements, for consistency. I find the "if it is used in an output" condition to be a bit too fiddly, personally, and likely to cause errors.

sjspielman:
Confirming merge is running in stub (#1066): https://github.com/AlexsLemonade/scpca-nf/actions/runs/18753024002/job/53497570164?pr=1067#step:8:12

allyhawkins (approved):
New changes look good to me!

merge.nf
```diff
@@ -140,118 +140,121 @@ workflow {
...
-    ]}
-  .subscribe{
-    if(!(it[1].exists() && it[1].size() > 0)){
-      log.warn("Processed files do not exist for ${it[0]}. This library will not be included in the merged object.")
-    }
-    else if(!(it[2].exists() && it[2].size() > 0)){
-      log.warn("Metadata file does not exist for ${it[0]}. This library will not be included in the merged object.")
+    .map{ it ->
+      def processed = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds");
+      def meta_json = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json");
```
jashapiro: Semicolons should not be required for groovy code. We seem to be using them inconsistently, but I would say we should not include them if they are not needed.
```suggestion
      def processed = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_processed.rds")
      def meta_json = file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
```

merge.nf
```diff
@@ -140,118 +140,121 @@ workflow {
...
-      file("${params.results_dir}/${it[0]}/merged/${it[0]}_merged.rds"),
-      it[0],
-      it[1]
-    ]}
+    .map{ project_id, has_adt, _library_id_list, _sce_file_list ->
+      [
+        file("${params.results_dir}/${project_id}/merged/${project_id}_merged.rds"),
+        project_id,
+        has_adt
+      ]
```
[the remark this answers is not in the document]
sjspielman: Ok, thanks for clarifying. For the general comment, I would really prefer to make that type of general change in a fresh PR since this is one already extremely long and getting harder to follow.

But this I will do here:
>  Can we make the project_id first here (and in the merge_sce output) to match the general pattern of an id/metadata before data that we are using generally?

PR #1071, opened 2025-10-23
Merge stub libraries in stub

### LINKED ISSUE #1066: Ensure merge.nf gets fully run in stub

sjspielman:
Closes #1066 

In this PR, I add a check that allows the `merge.nf` to proceed in stub as discussed in the linked issue, though here we're able to directly check the library ID and don't need to grab it from a file name.
 
I did not add this logic into the subscribe code that is printing logs - should I?

### DIFF
--- a/merge.nf
+++ b/merge.nf
@@ -228,7 +228,10 @@ workflow {
       file("${params.results_dir}/${it.project_id}/${it.sample_id}/${it.library_id}_metadata.json")
     ]}
     // only include libraries that have been processed through scpca-nf and have at least 3 cells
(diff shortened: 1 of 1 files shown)


jashapiro:
The code here would have to be heavily modified in #1067; can you just add this there instead in that format?

PR #1072, opened 2025-10-23
Add AWS ECR push for vireo-snp

jashapiro:
Following #1069, Here I am adding the ECR push of images for vireo-snp. Unfortunately, I didn't manage to fully test the push with that PR, so this is a second attempt to do so. I added the AWS images, as well as authentication. There is an update to the base image as well, mostly to trigger the push, but that also closes some security holes.

### DIFF
--- a/build-cellbrowser.nf
+++ b/build-cellbrowser.nf
@@ -52,13 +52,15 @@ workflow {
   libraries_ch = Channel.fromPath(params.run_metafile)
     .splitCsv(header: true, sep: '\t')
     // filter to run all ids or just specified ones
-    .map{it -> [
+    .map{ it ->
+      [
         project_id: it.scpca_project_id,
         library_id: it.scpca_library_id,
         sample_id: it.scpca_sample_id.split(";").sort().join(","),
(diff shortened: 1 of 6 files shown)


docker/vireo-snp/Dockerfile
```diff
@@ -7,11 +7,6 @@ LABEL org.opencontainers.image.description="Docker image to run the Vireo SNP in
 LABEL org.opencontainers.image.authors="Childhood Cancer Data Lab, Alex's Lemonade Stand Foundation <scpca@ccdatalab.org>"
 LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/scpca-nf/tree/main/docker/vireo-snp"
 
-## Install ps for nextflow
```
sjspielman: Just confirming this is no longer needed? I don't see it in the dockerfile for the updated python version which I am 99% sure is this - https://github.com/docker-library/python/blob/54a65f4ff531391810946ee17b0806accbda0fae/3.10/slim-trixie/Dockerfile

sjspielman (approved):
👍

PR #1079, opened 2025-10-24
Fix merge bugs I introduced and address remaining def patterns

### LINKED ISSUE #1074: Use separate def patterns
I think I'd like to make the separate `def` a pattern if it can facilitate single line outputs. 
```suggestion
    .map{ project_id, has_adt, _library_id_list, _sce_file_list ->
      def merged_rds = file("${params.results_dir}/${project_id}/merged/${project_id}_merged.rds")
      [merged_rds, project_id, has_adt]
```

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/1067#discussion_r2455725911_

sjspielman:
Closes #1074 

This gets the remaining `def`s we should probably add in the top-level `.nf` files. Other spots should be taken care of in the open PRs. Please let me know if you see other spots in the top-level files to catch!

I also did some silly things with `merge.nf` that didn't get flagged but I am fixing them!
1. `starts_with` is R, not groovy which is `startsWith`
2. an `||` should have been `&&`
3. referenced wrong variable in one of the subscribes

### DIFF
--- a/merge.nf
+++ b/merge.nf
@@ -208,7 +208,7 @@ workflow {
       [it.library_id, processed, meta_json ]
     }
     .subscribe{ library_id, processed, meta_json ->
-      if(!(processed.exists() && processed.size() > 0) && !(processed.exists() && library_id.startsWith("STUBL"))){
+      if(!processed.exists() || !(processed.size() > 0 || library_id.startsWith("STUBL")) )
         log.warn("Processed files do not exist for ${library_id}. This library will not be included in the merged object.")
       }
       else if(!(meta_json.exists() && meta_json.size() > 0)){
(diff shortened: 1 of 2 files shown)


merge.nf
```diff
@@ -208,13 +208,13 @@ workflow {
       [it.library_id, processed, meta_json ]
     }
     .subscribe{ library_id, processed, meta_json ->
-      if(!(processed.exists() && processed.size() > 0) || !(processed.exists() && library_id.starts_with("STUBL"))){
+      if(!(processed.exists() && processed.size() > 0) && !(processed.exists() && library_id.startsWith("STUBL"))){
```
jashapiro: Simplifying a bit, I think. 
```suggestion
      if(!(processed.exists() && (processed.size() > 0 || library_id.startsWith("STUBL"))){
```

jashapiro (approved):
LGTM, with a few suggestions about maybe making the logic simpler, but obviously simpler is a matter of opinion!

PR #1083, opened 2025-10-28
A few more Nextflow linting updates

jashapiro:
The main change here is to replace `Channel` with `channel` to remove a deprecation warning. I also caught one missing `_variable` and simplified a file search.

Annoyingly, the warnings about implicit `{ it }` closures are still there even for the simple filters, etc. but I decided not to follow that as per our previous discussions. If we are lucky, maybe https://github.com/nextflow-io/nextflow/issues/5934 will be picked up again and we will be able to configure this to hide the warnings.


sjspielman (approved):
👍 onwards!

Issue #1065, opened 2025-10-21
Use an empty list instead of NO_FILE


sjspielman:
> Do we also want to take this opportunity to pivot all the `NO_FILE` usage to `[]`? I'm on the fence for it...
> 
> I'd say yes, but probably in a separate PR 

 _Originally posted by @jashapiro in [#1035](https://github.com/AlexsLemonade/scpca-nf/issues/1035#issuecomment-3428191836)_

---

This issue tracks what it says! We often use `NO_FILE` via a variable `empty_file` to handle optional files. We have started transitioning to using an empty list `[]` in these cases instead which has the added benefit of allowing multiple optional files. This issue tracks making that swap across the board.

PR #1087, opened 2025-10-29
scpca-nf bug fixes

### LINKED ISSUE #1085: Update diagnosis handling for multiplexed libraries
While debugging post file styling, I encountered an error running a multiplexed sample through because of how we handle diagnoses leading into CNV inference; multiplexed will have multiple diagnosis values which the code does not account for: https://github.com/AlexsLemonade/scpca-nf/blob/47a39ad5a7f31a3ccd0b71eeacf23c2d189c754f/bin/add_celltypes_to_sce.R#L488-L508

This bug is addressable enough, but it does raise some questions about how we want to proceed with inferCNV and multiplexed. Specifically, we can only run multiplexed libraries through CNV inference if _all_ their associated samples have the same overall diagnosis group. This is currently the case for all data in the portal, but may not always be the case in the future. This is a bridge to cross later, but worth pointing out!

For now, this issue tracks updating this diagnosis parsing code to account for the fact that there may be multiple sample diagnoses. If there are multiple _broad_ diagnoses mapped across the samples, we'll error out (there is a world where we could also just not proceed with CNV inference instead of _erroring_, if someone wants to advocate for that!).

sjspielman:
Closes #1085

This PR catches and fixes a couple bugs....

1. Fix bug introduced during styling in `modules/call-cnvs.nf`
2. Caught and fixed a bug in `main.nf` where citeseq libraries weren't being flagged (at some point they got changed from CITEseq->citeseq, but the findAll did not!)
3. Fix a couple `inferCNV` bugs I caught while testing different library types...
    - I fixed the bug noted in #1085 where diagnoses weren't being properly dealt with for multiplexed
    - I found/caught another inferCNV bug in the logic checks in `sce_qc_report.R`, now confirmed fixed and works for all kinds of inferCNV success's

Noting that now that these have been addressed and the workflow is running across library types, I can proceed to #1065

### DIFF
--- a/bin/sce_qc_report.R
+++ b/bin/sce_qc_report.R
@@ -322,7 +322,7 @@ if (has_consensus) {
 # if TRUE, we need infercnv_min_reference_cells and a heatmap file
 infercnv_success <- metadata(processed_sce)$infercnv_success
 heatmap_path <- NULL
-if (!isFALSE(infercnv_success)) {
+if (is.na(infercnv_success) | infercnv_success) {
   stopifnot("infercnv_min_reference_cells parameter value was not provided" = !is.na(opt$infercnv_min_reference_cells))
   if (isTRUE(infercnv_success)) {
     stopifnot("inferCNV heatmap file does not exist" = file.exists(opt$infercnv_heatmap_file))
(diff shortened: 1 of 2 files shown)


jashapiro (commented):
This looks good to me, with one suggested change to avoid double negatives. 
I noted that the stub check seems to have stalled, which seems to be a bug we are hitting periodically. Not sure if it is due to the 25.10 version of nextflow, or what, but cancelling and rerunning usually fixes it, so I will do that here. If it becomes too much of a problem, we can specify the Nextflow version for a bit and see if that helps. (v25.10 has a lot of big changes, so I'm not too surprised it might be a bit buggy)

PR #1088, opened 2025-10-29
Use empty list instead of empty_file

### LINKED ISSUE #1065: Use an empty list instead of NO_FILE

sjspielman:
Closes #1065

This PR swaps all the `empty_file` -> `[]`, and also removes `assets/NO_FILE` from the repo. I've successfully tested this with a "regular" library, a CITEseq library, and multiplexed library 🎉

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -266,7 +266,11 @@ workflow {
   spaceranger_quant(runs_ch.spatial)
 
   // **** Process 10x flex RNA-seq data ***
-  flex_quant(runs_ch.flex, flex_probesets, file(params.cellhash_pool_file ?: [])) // if it's empty but needed, this would have failed already
+  flex_quant(
+    runs_ch.flex, 
+    flex_probesets, 
+    params.cellhash_pool_file ? file(params.cellhash_pool_file) : []
+  )
   flex_sce_ch = generate_sce_cellranger(flex_quant.out, file(params.sample_metafile))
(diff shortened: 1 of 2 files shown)


main.nf
```diff
@@ -335,7 +332,7 @@ workflow {
     }
 
   // apply cellhash demultiplexing
-  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file ?: empty_file))
+  cellhash_demux_ch = cellhash_demux_sce(feature_sce_ch.cellhash, file(params.cellhash_pool_file ?: []))
```
jashapiro:
```suggestion
  cellhash_demux_ch = cellhash_demux_sce(
    feature_sce_ch.cellhash, 
    params.cellhash_pool_file ? file(params.cellhash_pool_file): []
  )
```

PR #1089, opened 2025-10-30
Update contributing guide

### LINKED ISSUE #1073: Keep track of scpca-nf style choices
I'm opening this issue to keep a living record of the decisions we are making for #1035, starting off with the following. Additional items will be aded in comments. This information can potentially be used to put together in internal style guide (in lieu of/addition to? tbd! )a formal styler.

- We keep `->` at the end of the line when it's for a multi-line transformation
- We return lists from channel transformations, not tuples
- Spacing style:
  - For closures, `{ braces need spaces }`
  - But, `(parentheses dont)`
  - `if (thing) {`
- We only use `"${}"` for variables when there is string interpolation. Otherwise, just use the variable name
- Explicit lambda variables:
  - Use in all `map`s
  - Do not use in other small one-line closures, but do use for multi-line ones!
  - When a variable isn't being used in the transformation, preface it with an underscore
- We do not use `def` in script blocks
- For processes that optionally take a file, use `[]` (an empty list) rather than the `NO_FILE` approach

sjspielman:
Closes #1073

This PR updates `CONTRIBUTING.md` with style recommendations tracked in the linked issue. I tried to make examples somewhat consistent as well, ordered in what you should do and then what you shouldn't do. Let me know if anything is unclear or could be differently (more/less) emphasized, or if you prefer a different organization.


CONTRIBUTING.md
```diff
@@ -85,25 +89,146 @@ This can be set in [Visual Studio Code](https://code.visualstudio.com) with the
...
+
+#### Declaring groovy variables 
 
 Variables in Groovy are set as global by default, which can have some unexpected consequences.
-To avoid this, any variables declared in functions or closures (such as `.map{}` statements) should be prefixed with `def`, which defines them as locally-scoped variables.
+To avoid this, any variables declared in functions or closures (such as `.map{ }` statements) should be prefixed with `def`, which defines them as locally-scoped variables.
 
 ```groovy
+// do this:
+good.map{
```
jashapiro:
```suggestion
good.map{ it ->
```

sjspielman:
I just want to note that I have heavily referred to this PR while working on #1205 and still think it would be a useful thing to merge in! It's here for reference either way, but it's really useful for maintaining consistency in our code.

sjspielman:
I'm going to re-request review again, since I think it's important to actually get this in for (at least) these two reasons:

- As a team member who develops in nextflow, I have found it useful for development
- I realized that using `[]` for optional files had not been added here yet, but since this style choice we made was brought up in the spaceranger PR, it would be appropriate to add it here for future reference

Having this PR does not preclude automatic styling, which was the initial reason this did not move forward. But, as that has not yet happened, it is important to have this reference for keeping track of our style until it does, at least.

jashapiro (approved):
LGTM

PR #1090, opened 2025-10-30
Bump version to v0.9.0 and update containers

allyhawkins:
Towards #1061 

Here I'm bumping the containers to use the most recent release of `scpcaTools`. I did a test run of all library types with those containers and things looked good. 

I also bumped the cellbrowser and vireo containers to be the new tag. So once this goes in the `development` branch will not work for those specific processes until we actually make the release.


jashapiro (approved):
LGTM

PR #1093, opened 2025-10-31
Convert Utils class to Nextflow functions and add pull-through support

jashapiro:
This PR started as just wanting to add the pull through cache support for the new infrastructure in a way similar to what we have in openscpca-nf, but when I went to test it I started running into a strange nextflow bug that seems to be related to the future evolution of the Nextflow language. 

Specifically, using a class with functions defined within it is currently being deprecated, and while it still works with the `development` branch of this repo, something in this branch was causing some trouble. Rather than trying to fix it, I decided to embrace the future, and moved the functions to be standalone, and included explicitly. 

I also added the `pullthroughContainer` function and related params. Since I was modifying all the container directives, I also updated the container params to be lower case. 

One funny thing I found was that I did have to quote the calls to `pullthroughContainer()` when using it in the container directive. A bit annoying, but it is what it is.

### DIFF
--- a/config/profile_awsbatch.config
+++ b/config/profile_awsbatch.config
@@ -13,7 +13,7 @@ aws {
   region = 'us-east-1'
 }
 
-params.pullthrough_registry = '997241705947.dkr.ecr.us-east-2.amazonaws.com'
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
LGTM

PR #1105, opened 2025-11-11
Fix bugs in rendering metrics report

allyhawkins:
I was working on rendering the metrics reports for a few projects and ran into a few errors that needed to be fixed before the report could render successfully. 

* When reading in the JSON files, the `x == "NA"` check was causing a failure for any real `NA` values that were not quoted since `NA == "NA"` evaluates to `NA`. 
* `n_added` was being used throughout when it wasn't defined in those blocks. I changed this to `n_libraries` where appropriate. 
* The pivoting when counting cell type changes was throwing an error because sometimes it was a list of cell types and other times there was no change so it was just `NA`. To combat that, I first turn everything into a character vector, then pivot, and then turn back into a list for counting. I'm not sure this is the most elegant way of doing that, but this fixed the problem. 

I tested this on a few projects and the report is rendering correctly.


jashapiro (approved):
LGTM

Issue #1101, opened 2025-11-11
Refactor script for adding cell types to SCEs and counting reference cells


allyhawkins:
But `bin/add_celltypes_to_sce.R` could really use a refactoring: the level of if/else nesting is hard to follow especially given the length of some of the blocks. If some of those blocks could be pulled out as functions, that would probably help a lot.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/1099#pullrequestreview-3445344335_

sjspielman:
We function-ified this whole thing a little while ago, so this is set.

PR #1107, opened 2025-11-12
Handle unknown diagnoses and add a check for infercnv columns in report

### LINKED ISSUE #1106: [BUG] Errors rendering QC report
I was doing some testing with new infrastructure, and ran into a couple of bugs when using the development branch.

The main error came from `infercnv_qc.rmd`, which seems to be an error getting a table with the specified columns, which are `c("is_infercnv_reference", "consensus_celltype_annotation", "infercnv_total_cnv")`

```
  Error in `.use_names_to_integer_indices()`:
  ! 'use.coldata' contains invalid values
  Backtrace:
   1. dplyr::mutate(...)
   6. scuttle::makePerCellDF(...)
   7. scuttle:::.use_names_to_integer_indices(...)
  Quitting from lines 140-218 [unnamed-chunk-142] (/tmp/RtmpMw0VPl/infercnv_qc.rmd)
  Quitting from lines 757-757 [unnamed-chunk-19] (main_qc_report.rmd)
``` 

I thought this might be related to having to rerun infercnv with updated scripts, so I tried to use `--perform_cnv_inference and --repeat_cnv_inference`, but despite those flags, the steps did not run. 

I'm not super-familiar with this part of the workflow, so I am not sure the best way to test this, and I am not sure whether some of this might just be due to mismatches in versions, but it seems like it may be worth investigating and resolving before the next release. 

The full error report is here: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/4HBUiqYp3kUcCC/v2/logs

jashapiro:
This does the main part of #1106, though it does not yet address my final question in https://github.com/AlexsLemonade/scpca-nf/issues/1106#issuecomment-3522338904, so I am not closing that yet.

Update, I am now confident that the `--repeat_cnv_inference` question was the expected behavior, so:
closes #1106

First an apology; I did a chunk of #1101, so this looks like a lot of change, but most of it is just moving things. In particular, adding each of the automated method annotations is now in a separate function, as is creating the consensus.

I did not include moving the reference cell counting at this stage, so I would not consider #1101 done. I could do more of that here, but I really didn't want to quite yet.

The main _important_ change is the one in 7dd3a280bab14d56f0e935a04b8135d226db1835. I changed from filtering to joining the `diagnosis` vector, so that we get a `diagnosis_group` value for every diagnosis, regardless of whether it is in the diagnosis groups file. If the value is `NA`, we replace that with the sample diagnosis.  To me this seems like the most logical thing to do, as it allows us to handle any new diagnosis, at least this far.

I also made a small update to avoid having an error when making the inferCNV plots in the case that sce files were made with an older version of the workflow. I'm not sure it is the absolute best approach, but it is a pretty simple addition, and that seems worth something. This included also moving a bit of the logic out of that subnotebook and into the main notebook, mostly because I was very confused looking at it in isolation about what `has_infercnv_attempt` meant. So I also updated that variable to be more immediately informative. 

I am also leaving this as a draft for a bit, as I have not yet run it through a test, but I thought you might want to have a look early, given that it changes more than expected. And if you would prefer that I make a separate PR with just the important changes, I should be able to cherry pick those without much trouble.

### DIFF
--- a/lib/Utils.groovy
+++ b/lib/Utils.groovy
@@ -23,6 +23,10 @@ class Utils {
       meta.unique_id = meta.technology.contains("_multi") ? "${meta.library_id}-${meta.sample_id}" : meta.library_id
     }
 
+    if(meta.technology) {
+      meta.technology = meta.technology.toLowerCase()
+    }
+
     return(meta)
   }
 
--- a/main.nf
+++ b/main.nf
@@ -125,9 +125,11 @@ workflow {
     'citeseq_10xv2': '737K-august-2016.txt',
     'citeseq_10xv3': '3M-february-2018.txt',
     'citeseq_10xv3.1': '3M-february-2018.txt',
+    'citeseq_10xv4': '3M-3pgex-may-2023_TRU.txt.gz',
     'cellhash_10xv2': '737K-august-2016.txt',
     'cellhash_10xv3': '3M-february-2018.txt',
-    'cellhash_10xv3.1': '3M-february-2018.txt'
+    'cellhash_10xv3.1': '3M-february-2018.txt',
+    'cellhash_10xv4': '3M-3pgex-may-2023_TRU.txt.gz',
   ]
 
   // 10x flex probe set files
--- a/merge.nf
+++ b/merge.nf
@@ -154,7 +154,7 @@ workflow {
         library_id: it.scpca_library_id,
         sample_id: it.scpca_sample_id.split(";").sort().join(","),
         seq_unit: it.seq_unit,
-        technology: it.technology
+        technology: it.technology.toLowerCase()
       ]
     }
 
--- a/modules/af-features.nf
+++ b/modules/af-features.nf
@@ -165,6 +165,7 @@ workflow map_quant_feature {
       // branch based on whether mapping should be run (make_rad) or skipped (has_rad)
       // if neither fastq or rad dir are present, run goes into missing_inputs branch
       .branch{ it ->
+        def stored_tech = Utils.getMetaVal(file("${it.feature_rad_dir}/scpca-meta.json"), "technology") ?: ""
         make_rad: (
           // input files exist
           it.files_directory && file(it.files_directory, type: "dir").exists() && (
@@ -172,6 +173,8 @@ workflow map_quant_feature {
             params.repeat_mapping
             // or the feature rad file directory does not exist
             || !file(it.feature_rad_dir).exists()
+            // or the technology has changed (to ensure re-mapping if tech was updated)
+            || it.technology.toLowerCase() != stored_tech.toLowerCase()
           )
         )
         has_rad: file(it.feature_rad_dir).exists()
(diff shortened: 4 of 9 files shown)


templates/qc_report/infercnv_qc.rmd
```diff
@@ -4,33 +4,6 @@
 This file is meant to be run as a child report from `main_qc_report.rmd`
 -->
 
-```{r results = 'asis', eval = has_infercnv_attempt}
```
allyhawkins: I originally had this moved into this report so that all the inferCNV code was together and the messages get printed under the `inferCNV results summary` header. By moving it into the main report, then it's just a box with no header, which is fine, but not my favorite.

allyhawkins (commented):
I'm on board with most of these changes and don't have a lot of comments, but I did have a preference for where the inferCNV message boxes should live. I didn't go line by line in the functions you added yet and focused on just the error handling changes for this round.

allyhawkins:
I also don't know what's going on with the `--repeat_cnv_inference` situation, but was looking into that. There must be something with how the logic is set up that's not working.

bin/add_celltypes_to_sce.R
```diff
@@ -13,7 +13,337 @@ suppressPackageStartupMessages({
...
+  return(sce)
+}
+
+# check that input file exists and output file ends in rds
+stopifnot(
+  "Missing input SCE file" = file.exists(opt$input_sce_file),
+  "output sce file name must end in .rds" = stringr::str_ends(opt$output_sce_file, ".rds"),
+  "output file to store counted normal cells was not provided" =
+    !is.null(opt$reference_cell_count_file)
+)
```
allyhawkins: Was doing some testing here to try and figure out what's going on with inferCNV flags and this is in the wrong place. It should be after the options are read in.

jashapiro: I think the inferCNV flags are actually fine; it was probably skipping because there were no reference cells.

jashapiro: Not sure how that got moved!

templates/qc_report/infercnv_qc.rmd
```diff
@@ -173,8 +183,13 @@ total_cnv_lumped_df <- total_cnv_df |>
...
   tibble::rownames_to_column("celltype_lumped") |>
-  dplyr::filter(celltype_lumped %in% unique(total_cnv_lumped_df$celltype_lumped))
-hex_df$celltype_lumped <- factor(hex_df$celltype_lumped, levels = levels(total_cnv_lumped_df$celltype_lumped))
+  dplyr::mutate(
+    celltype_lumped = factor(
+      celltype_lumped,
+      levels = levels(total_cnv_lumped_df$celltype_lumped)
+    )
+  ) |>
+  dplyr::filter(!is.na(celltype_lumped))
```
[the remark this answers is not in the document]
jashapiro: yeah, I reverted that locally and am testing it.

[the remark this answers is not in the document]
jashapiro: Okay, I finally fixed it. That took way too long to find. The whole problem was me using `colnames(sce)` rather than `colnames(colData(sce))` So it was quitting early, just as it was told to do. 

During my many wrong turns, I ended up making a few more changes in the infercnv subnotebook, trying to simplify the code to make it easier for me to trace, even though all of that was for naught. But I left it.

Should be ready for rereview.

PR #1109, opened 2025-11-12
Ensure techs are lowercase when reading in existing meta files

### LINKED ISSUE #1108: [BUG] Missing ADT section in QC report
It appears that something isn't working correctly for CITE-seq samples, but I'm not entirely sure what the issue is. I noticed that SCPCP000003, which has CITE-seq, has `_adt.h5ad` files but no ADT section of the QC report. This appears to affect more than the report since the `_metrics.json` file that's getting exported is reporting `NA` for the ADT related keys and the ADT keys, `adt_filtering_method` and `adt_normalization_method` is missing from the `_metadata.json` file that gets output. 

We should definitely fix this before v0.9.1

allyhawkins:
Closes #1108 

It looks like when we changed the techs to be lower case, we didn't account for the fact that the metadata being read in from existing `scpca-meta.json` files would still be upper case. This meant when we were creating the initial unfiltered SCE with `adt` as the `altExp`, this section of the code wasn't working as anticipated and was instead providing `CITEseq` as the name of the `altExp`. 

https://github.com/AlexsLemonade/scpca-nf/blob/79fa07462cab458bd664a67138f901ba07345c94/modules/sce-processing.nf#L79-L81

This meant all the code that check for cite seq by using `adt` as the `altExpName` wasn't working as expected. 

To fix this, I'm just updating the `readMeta` function to convert `technology` to lower case if present. This should hopefully make sure that any older meta files we read in match the expected lowercase formatting we specify in the main workflow. I'm still waiting for the full test run to complete, but the `--feature_name` argument is now correct when running the process to generate the unfiltered SCE, so I expect it to resolve the problem.

### DIFF
--- a/lib/Utils.groovy
+++ b/lib/Utils.groovy
@@ -23,7 +23,7 @@ class Utils {
       meta.unique_id = meta.technology.contains("_multi") ? "${meta.library_id}-${meta.sample_id}" : meta.library_id
     }
 
-    if( meta.technology) {
+    if(meta.technology) {
       meta.technology = meta.technology.toLowerCase()
     }
 


allyhawkins:
> Just to be more certain, I'd also update:
> 
> ```
>  if (meta.feature_type.toLowerCase() == "citeseq") { 
>    meta.feature_type = "adt" 
>  } 
> ```

Done in https://github.com/AlexsLemonade/scpca-nf/pull/1109/commits/bfc5e5c6f979d13e108b2b165c6e8c56a80ef306. I also did a search for any other places where we check the technology is `citeseq` or `cellhash` and found a spot in the merged file where we were not ensuring lower case technology, so I changed that in  https://github.com/AlexsLemonade/scpca-nf/pull/1109/commits/ad6b4fab9ab060c9a3a2b452f2bbff84a6bb2696.

PR #1124, opened 2025-11-25
Account for incompatible HVGs and cell line projects in metrics report

allyhawkins:
I was working on generating the metrics report for all projects and found a few issues that need to be fixed. 

There were a few occasions where the number of HVGs were less than 2000 and the length of the total HVGS did not match between reference and comparison. This seems to happen for low quality libraries that have very few cells. This caused a failure to calculate the correlation, so I added a check for the length of the HVGs and then only compute correlation if they are equal. Otherwise, I report `NA` since we can't get a correlation. 

There was also an issue with the cell line projects where the list of celltypes were all just empty lists in the data frame. I added a check for cell types being present and then skip any cell type summaries if they are missing. 

I decided to generate one notebook per project since looking at all projects together seemed like a lot. With the changes here I'm able to generate a report for each project.

### DIFF
--- a/scripts/compare-metrics/compare-metrics-template.rmd
+++ b/scripts/compare-metrics/compare-metrics-template.rmd
@@ -1093,8 +1093,8 @@ cluster_changes |>
 # use this to make sure cell types are actually present
 all_celltypes <- metrics_df |>
   dplyr::select(contains("celltypes")) |>
-  tidyr::pivot_longer(everything(), names_to = "celltype_method", values_to = "celltype") |>
-  dplyr::pull(celltype) |>
+  tidyr::pivot_longer(everything(), names_to = "celltype_method", values_to = "celltype_list") |>
+  dplyr::pull(celltype_list) |>
   unlist() |>
   na.omit()
 


scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -1083,8 +1088,31 @@ cluster_changes |>
...
 
-
 ```{r}
+# get a list of all celltypes found
+# use this to make sure cell types are actually present
+all_celltypes <- metrics_df |>
+  dplyr::select(contains("celltypes")) |>
+  tidyr::pivot_longer(everything(), names_to = "celltype_method", values_to = "celltype") |>
+  dplyr::pull(celltype) |>
+  unlist()
```
jashapiro: did you mean `unique()` here?

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -1083,8 +1088,31 @@ cluster_changes |>
 
 ### Cell type assignments {.tabset}
 
-
 ```{r}
+# get a list of all celltypes found
+# use this to make sure cell types are actually present
+all_celltypes <- metrics_df |>
+  dplyr::select(contains("celltypes")) |>
```
jashapiro: If _all_ we care about is the presence of cell type columns (which there will not be in cell lines), could you just do:

```suggestion
has_celltypes <- colnames(metrics_df) |>
  stringr::str_detect("celltypes") |>
  any()
```

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: Ah, I get it. The entries in the data frame themselves are lists. I'd probably rename the column from `celltype` to `celltype_list` just for clarity

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -1083,8 +1088,31 @@ cluster_changes |>
...
+  tidyr::pivot_longer(everything(), names_to = "celltype_method", values_to = "celltype") |>
+  dplyr::pull(celltype) |>
+  unlist()
+
+has_celltypes <- !is.null(all_celltypes)
```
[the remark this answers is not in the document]
jashapiro: Because of the way `na.omit()` works, you will still need to check the length here:
```suggestion
has_celltypes <- !is.null(all_celltypes) && length(all_celltypes) > 0
```

jashapiro (approved):
Thank you for the explanations. I think this is mostly good now, but I think you need to change the `NA` filtering since the lists won't be `NA` even if all of their elements are.

Issue #1095, opened 2025-11-05
Update scpca-nf on Cavatica to 0.9.1


jashapiro:
With the release of 0.9.0, we should update the version of the workflow on Cavatica to the release version.

This will follow the directions in the `cavatica/README.md` document to upload the new version to Cavatica.

PR #1128, opened 2025-11-26
Updates to Cavatica description and more careful checks

jashapiro:
While working on #1095, I updated the description for the tool, so I wanted to get those changes in here (I had updated one doc, but not the yaml file).

I also ran into a small error where `file(foo).exists()` fails if `foo` is empty, so I added an extra bit to all of those checks so that we get the nice error message.

Finally, I ran into a strange error where the sb_pack scripts couldn't find a dependency, so I made it install more explicitly.


jashapiro:
Updating the yaml version number was actually already in the release notes (added when I added it to the header), but I did not have the same thing for the md file. Other changes made.

PR #1132, opened 2025-12-03
remove use of existing output directory for Cell Browser

jashapiro:
Reusing the output directory seems to have been causing some trouble, so we will try not doing that anymore. 

This will slow down minor rebuilds, so it might be nice to add part of this back in if we want to be able to run a single project at a time and have it added to or updated within the existing site. I think to do this we would need to add a staging step that removed any projects or libraries from the existing site directory before building, but that seems like it will require a lot more testing, so I went with the simpler solution for now.

### DIFF
--- a/modules/cellbrowser.nf
+++ b/modules/cellbrowser.nf
@@ -44,7 +44,7 @@ process cellbrowser_library {
 }
 
 process cellbrowser_site {
-  container "${pullthroughContainer(params.cellbrowser_container, params.pullthrough_registry)}"
+  container "${params.CELLBROWSER_CONTAINER}"
   publishDir "${params.outdir}"
   input:
     tuple val(project_ids), path(library_dirs)


modules/cellbrowser.nf
```diff
@@ -72,7 +71,6 @@ process cellbrowser_site {
 
     # build the site
     CBDATAROOT=cb_data cbBuild -r -i cb_data/cellbrowser.conf \
-      --redo ${params.cellbrowser_rebuild ? "matrix" : "meta"} \
```
allyhawkins: If we aren't using this parameter anymore than we should also remove it from the config files.

jashapiro: https://github.com/AlexsLemonade/scpca-nf/commit/1801286ed000c304cb351f9feb3c892097fefc24 if you are interested to look...

PR #1141, opened 2025-12-22
Don't move input files

jashapiro:
While testing the new infrastructure, I discovered that when using wave containers and Fusion file system, input files that are moved from an input directory are actually _moved_, which is to say they are removed from their original location. In essence this is because the Fusion drive is treating the directory as a mounted file system. Presumably this same error could also occur with other true shared or local file systems. 

So what I have done here is converted any place where we were moving an input file within a script to copy it instead, leaving the original intact.

I only looked for places where we actually used `mv`, so it is possible that there may be some places where we overwrite files, but I think we had done a pretty good job of removing those earlier. Nonetheless, we should probably check this with a more formal process for all actions, especially those that use checkpoint directories, as modifying the checkpoints is _bad_.  (I'll file a separate issue for this).

### DIFF
--- a/modules/af-rna.nf
+++ b/modules/af-rna.nf
@@ -79,7 +79,7 @@ process fry_quant_rna {
     if [ $barcode_file == *.gz ]; then
       gunzip -c ${barcode_file} > permitted_barcodes.txt
     else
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
LGTM

PR #1149, opened 2026-01-16
Update metrics comparison script with new buckets

jashapiro:
Part of https://github.com/AlexsLemonade/ScPCA-admin/issues/1195

Here I am updating the default locations for the `compare-metrics` script and template, as well as adding an earlier check that the buckets are available and creds are correct. I also made the default output file include a date, because that seems useful. If you don't like it, let me know and I will return to the previous default.

I also made a readme in the script directory, preparing for the impending cuts to the internal instructions. 

Finally, I realized that the default reference output should be a local folder rather than something that a general user would not be able to write to, so I changed that and added the actual location to the appropriate profiles.

### DIFF
--- a/config/ccdl_profiles.config
+++ b/config/ccdl_profiles.config
@@ -35,6 +35,7 @@ profiles {
 
       // output base directory
       outdir = "s3://ccdl-scpca-results-staging-997241705947-us-east-1"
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
LGTM

Issue #1127, opened 2025-11-26
Workflow should not fail if infercnv reference cells can not be determined


jashapiro:
Right now we have a number of `stop`  (or `stopifnot`) statements when counting reference cell types in `add_celltypes_to_sce.R`, which can cause the whole workflow to fail. It would be better if we simply returned `NA`  for the consensus, (or didn't add the column, but I think `NA` is better?)

I think the best way to handle this may be to factor out the reference counting code to a function, and replace the `stop` statements with early returns that return an unmodified SCE object if reference counting can not be done.

allyhawkins:
> I might also change the name of the field to infercnv_status, but I'm not sure how much breaks in that case.

I agree about this name change, but before we do this we should check where else it's used that it might cause issue. I _think_ it's just in the report and the docs so we theoretically could change it. 

> I agreed up to here, but I think a shorter status strings make more sense, which can be converted to appropriate messages in the report. While I understand not liking string comparisons, I like mixed types even less.

+1 to this comment and to the values that Josh suggested.

sjspielman:
> I agree about this name change, but before we do this we should check where else it's used that it might cause issue. I think it's just in the report and the docs so we theoretically could change it.

Just to confirm, the variable (currently named) `infercnv_success` only appears in:
- `bin/add_infercnv_to_sce.R`
- `bin/sce_qc_report.R`
- `templates/qc_report/main_qc_report.rmd`
- `templates/qc_report/infercnv_qc.rmd`

PR #1150, opened 2026-01-21
Refactor add_celltypes_to_sce inferCNV code

sjspielman:
Towards #1127 

This PR takes the first steps of the linked issue to nail down inferCNV behavior:

1. I refactored the script, including getting inferCNV reference cell code into a function
2. I added a bunch of statuses for tracking conditions (keep reading!)
3. I added the associated environment variable to the nextflow code and got it into `meta`, but I'm not using it yet - that'll be in the next PR!

These are the `infercnv_status` values I currently have in the code - feedback appreciated! Please let me know if/where you think these can be tweaked.

| string                                  | meaning                                                                                                                                                                             |
| --------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| ""                                      | no problems were flagged; infercnv can proceed                                                                                                                                      |
| "no_diagnosis_celltype_reference"       | the broad diagnosis <-> cell type reference wasn't provided                                                                                                                         |
| "no_consensus"                          | no consensus cell types were generated                                                                                                                                              |
| "unknown_diagnosis"                     | no diagnosis was provided in the sample metadata                                                                                                                                    |
| "multiple_diagnoses_multiplexed"        | we have multiple diagnoses for multiplexed, but the diagnosis group file wasn't provided so we can't determine which group to get cell types for                                    |
| "multiple_diagnosis_groups_multiplexed" | we have multiple diagnosis _groups_ for multiplexed. this is sort of similar to "no_diagnosis_multiplexed", except a group mapping was provided and we still ended up with multiple |
| "unknown_diagnosis_group"               | the diagnosis in the sample metadata couldn't be mapped to a diagnosis group                                                                                                        |
| "skipped_non_cancerous"                 | the sample is normal, won't proceed.                                                                                                                                                |
| "unknown_reference_celltypes"           | the diagnosis group doesn't have associated cell types                                                                                                                              |

A couple notes on these:

- There is an edge that I do not currently account for. I went back and forth on this and ended up feeling it's more of an internal implementation detail and not worth reporting as a standalone situation. Please let me know if you disagree!
  - If the diagnosis mapping file (sample diagnosis <-> broad diagnosis) was not provided, this means we'll just assume the sample diagnosis is the broad diagnosis
  - If there are no cell types associated with that broad diagnosis, the code will give the status `"unknown_reference_celltypes"`. But, it's technically a different circumstance from the _broad diagnosis obtained via mapping the sample diagnosis_ not having associated cell types. Do you feel this reaches the level of needing its own flag?
- In addition, I have a TODO in there about the behavior for "non-cancerous." The code was a bit inconsistent on this previously, so now's a good time to lock it down (and in the docs next!)....
  - Currently, if it's not multiplexed and diagnosis is non-cancerous, don't run infercnv. Status is  `"skipped_non_cancerous"`
  - Currently (aka in this PR), if it IS multiplexed and e.g. we have 2 diagnosis: a cancer and a "non-cancerous," do NOT run infercnv. Status would be `"multiple_diagnosis_groups_multiplexed"`. This is the more conservative approach.
  - But, a different way to think about this is - if we have 2 diagnoses and one is "non-cancerous", then we could drop the non-cancerous and favor the other diagnosis and then indeed proceed to process that sample with inferCNV. This would be a more permissive approach. Let me know if you prefer the permissive approach here!

Finally, I've locally run and tested this in a bunch of ways and seems ok 🤞

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -710,7 +710,7 @@ if (length(automated_methods) > 1) {
     opt$diagnosis_celltype_ref
   )
   # calculate if passing status
-  if (metadata(sce)$infercnv_status == "") {
+  if (metadata(sce)$infercnv_status == "feasible") {
     sce <- add_infercnv_reference_cells(
       sce,
       opt$consensus_validation_ref,


bin/add_celltypes_to_sce.R
```diff
@@ -336,6 +336,90 @@ assign_consensus_celltypes <- function(
...
+) {
+  # read in tsvs we know exist
+  diagnosis_celltype_df <- readr::read_tsv(opt$diagnosis_celltype_ref)
+  consensus_validation_df <- readr::read_tsv(opt$consensus_validation_ref)
+
+  # unique in case of multiplexed
+  diagnosis <- unique(metadata(sce)$sample_metadata$diagnosis)
+
+  if (is.null(diagnosis)) {
+    infercnv_status <- "unknown_diagnosis"
```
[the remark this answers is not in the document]
sjspielman: This turns out to be somewhat less elegant than I might have hoped.

What's nice about having a function is early `return` statements to bypass a lot of the nesting, and of course it's more modular. 

But, we can't really factor out reading in reference files, because checking if they exist is part of assigning the status. Also, we'd have to define `broad_diagnosis` in both functions since it's used both for assigning status and the reference calculations. To make it work, we'd have to return a list with a buncha stuff from function setting the status, and I'm unconvinced that's an improvement on having it all in one function. Feel free to try to convince me though!

sjspielman:
@allyhawkins should be ready for another look. The status is now stored in the SCE metadata, and I fixed the non-cancerous/multiplexed situation. I attempted a grander refactor into 2 functions, but I did not find this to be as pleasant as I hoped since we'd necessarily end up with either a bunch of duplicate code, or passing a ton of info around - https://github.com/AlexsLemonade/scpca-nf/pull/1150#discussion_r2714456610

allyhawkins (commented):
This looks mostly fine to me, I just had some small comments about moving around some of the if statements. I did leave a suggestion on how I would separate the cell counting into a separate function, but I don't think its absolutely required. It would help add modularity though.

bin/add_celltypes_to_sce.R
```diff
@@ -336,6 +336,127 @@ assign_consensus_celltypes <- function(
...
+  } else {
+    diagnosis_celltype_df <- readr::read_tsv(diagnosis_celltype_ref)
+    if (!(broad_diagnosis %in% diagnosis_celltype_df$diagnosis_group)) {
+      metadata(sce)$infercnv_status <- "unknown_reference_celltypes"
+      return(sce)
+    }
+  }
+
+  # If we made it here, no edge case was found
+  metadata(sce)$infercnv_status <- ""
```
jashapiro: I'd actually prefer if/else in this case. Early returns make sense to me at the start of a function, but not when most of the logic in the function has already happened. 

I'd pull the file check out from the main if/else:

Something like:
```
if(file.exists(diagnosis_celltype_ref)){
  diagnosis_celltype_df <- readr::read_tsv(diagnosis_celltype_ref)
} else {
  diagnosis_celltype_df <- data.frame()
}
```
Then you replace the file in your status setting code with `if(length(diagnosis_celltype_df) > 0)` and just go through all the cases in an unnested set of if/elses.

sjspielman: I did this here https://github.com/AlexsLemonade/scpca-nf/pull/1150/commits/e93531521eae3565fbd2ccea93f87fe97f426f73

But note, we need an `else if` for this one still, since with a defined data frame (even an empty one) it will still check that column, and we need to confirm the value is present.

```
  if (nrow(diagnosis_celltype_df) == 0) {
    metadata(sce)$infercnv_status <- "no_diagnosis_celltype_reference"
  } else if (!(broad_diagnosis %in% diagnosis_celltype_df$diagnosis_group)) {
    metadata(sce)$infercnv_status <- "unknown_reference_celltypes"
  }
```

or I could have written as this, but I liked without the `&` more because line length 🤷‍♀️ 
```
  if (nrow(diagnosis_celltype_df) > 0 & broad_diagnosis %in% diagnosis_celltype_df$diagnosis_group) {
    metadata(sce)$infercnv_status <- "unknown_reference_celltypes"
  } else {
    metadata(sce)$infercnv_status <- "unknown_reference_celltypes"
  }
```

bin/add_celltypes_to_sce.R
```diff
@@ -336,6 +336,129 @@ assign_consensus_celltypes <- function(
...
+  if (nrow(diagnosis_celltype_df) == 0) {
+    metadata(sce)$infercnv_status <- "no_diagnosis_celltype_reference"
+  } else if (!(broad_diagnosis %in% diagnosis_celltype_df$diagnosis_group)) {
+    metadata(sce)$infercnv_status <- "unknown_reference_celltypes"
+  }
+
+  # Empty string if no edge case was found
+  if (is.null(metadata(sce)$infercnv_status)) {
+    metadata(sce)$infercnv_status <- ""
+  }
```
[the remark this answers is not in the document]
sjspielman: > A note on the final case though.... is that a pass/success? Should we make that explicit?

We _will_ make that explicit but later once we actually get to that process. it could be pass/success, fail, or insufficient_reference_cells (this gets checked as part of branching for infercnv process, unless we want moar refactoring..!) 

But that doesn't mean it needs to be _empty_ now, if you have a suggestion for a value you think might be safer?

bin/add_celltypes_to_sce.R
```diff
@@ -581,108 +699,30 @@ if (length(automated_methods) > 1) {
...
-        # in case of multiplexed
-        unique()
-    }
-
-    # Check that we have at least one broad diagnosis
-    stopifnot(
-      "Could not determine broad diagnosis from sample diagnosis." =
-        length(broad_diagnosis) >= 1
+  # calculate if passing status
+  if (metadata(sce)$infercnv_status == "") {
```
allyhawkins:
```suggestion
  if (metadata(sce)$infercnv_status == "feasible") {
```

Issue #1146, opened 2026-01-14
Probably update report dotplot sizing


sjspielman:
I noticed that for SCPCL000001, the dotplot in the report is a bit squished such that some circles are overlapping. This probably happens across several reports!

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/938cf2de-b1f9-408e-8818-935a40acaf1a" />

We might just make the figure frame a bit wider in the R chunk.

allyhawkins:
> > Yup, we definitely need to update this. See this incidental one I just came across - SCPCS000527/SCPCL000859
> 
> Wooooah now noticing we've got 2 `NA`s down there on the y-axis? Seems bad! Looking into it....

My very initial inkling is that we are not accounting for an observed consensus cell types with a validation group, so they are getting classified as `NA` when joining.

sjspielman:
Closing in favor of focusing future efforts on #1157

PR #1156, opened 2026-01-28
QC report updates

### LINKED ISSUE #1151: Update UMAP report message
While working on #1127, I noticed a spot in the QC report that appears to be outdated.

https://github.com/AlexsLemonade/scpca-nf/blob/f8ee2a65887eaeb3ad137b64291984c3342e27f0/templates/qc_report/main_qc_report.rmd#L524-L537

The post-processing R script doesn't contain this specific check (the value `50` isn't used anywhere), but we do use a `try()`:
https://github.com/AlexsLemonade/scpca-nf/blob/f8ee2a65887eaeb3ad137b64291984c3342e27f0/bin/post_process_sce.R#L243-L270

We should update the report here to use the `has_umap` instead to communicate if there's a reduced dimension section.

sjspielman:
Closes #1151 
Towards #1146 ? 

This PR takes care of some miscellaneous updates to the QC report (although we're not done!). Changes include:

- For #1151, I updated the messaging in the main report to still print out a warning about 50 cells, but not mention UMAP as part of that warning. At the same time, I added a message box to the UMAP child report to print out if there's no UMAP in the object, and skip the remaining code. Because of this change, we now always run the UMAP child report, but bail after the first chunk if there's no actual UMAP. This is conceptually similar (but not identical) to how we treat the infercnv child report. Tested and working as expected! 
- For #1146, a handful of changes...
  - I updated the palette file as discussed in the issue
  - For future-proofing, I added a filter statement to make sure that we're only including validation groups that have marker genes (aka that have colors). I also added a sentence to print in the report saying that we do this. I think this is probably the move for now, but we can always change the approach to include all the groups, even those without marker genes. Since that would involve much bigger code updates, I suggest this route for now and we can open an issue to circle back to that approach later. 
  - I attempted some more dynamic dotplot sizing, and I think I did _ok_.... It's certainly an improvement on before, but I'm not convinced it's done, so I'm hesitant to close out #1146. 

Here are some reports with resized dotplots (ignore all the warnings about package versions at the top... that's a my-computer problem!). Does this like an ok spot to at least temporarily land for dotplot sizing?
[SCPCL00001.html](https://github.com/user-attachments/files/24919774/SCPCL00001.html)
[SCPCL000859.html](https://github.com/user-attachments/files/24919775/SCPCL000859.html)
[SCPCL000864.html](https://github.com/user-attachments/files/24919776/SCPCL000864.html)

While I was testing all this, I also caught/fixed a bug where `include_infercnv_report` was able to sneak by undefined in certain circumstances.

### DIFF
--- a/templates/palettes/validation-group-palette.tsv
+++ b/templates/palettes/validation-group-palette.tsv
@@ -6,7 +6,7 @@ innate lymphoid cell	#325A9B
 dendritic cell	#16FF32
 macrophage	#1C8356
 monocyte	#1CBE4F
-myeloid	#90AD1C
+myeloid cell	#90AD1C
 natural killer cell	#FBE426
 hematopoietic precursor cell	#FEAF16
 stem cell	#F7E1A0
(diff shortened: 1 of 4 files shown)


allyhawkins (commented):
Thanks for doing this. The biggest change that should be made is to keep the unknown cells in the plot. I had a suggestion to only show the top 10 marker genes for each cell type like we do in the paper figures, but I'm not sure that's 100 percent necessary.

allyhawkins (approved):
LGTM

PR #1163, opened 2026-02-02
Update CCDL profile buckets

jashapiro:
Here I am updating the default buckets for the CCDL profiles. 

I also updated the actions to use github-held info that is only in the `prod` environment.

I removed the default pull through repository, as we don't really want to have that in this repository.


allyhawkins (approved):
LGTM

PR #1170, opened 2026-02-03
Add metrics for new features (and catch teeny scimilarity "bug")

### LINKED ISSUE #1154: Add additional workflow metrics for new features
We have updated `scpca-nf` with some new features since originally compiling this set of metrics to check: https://github.com/AlexsLemonade/scpca-nf/blob/main/workflow-testing.md

In particular, we now run `SCimilarity`, `scDblFinder` and `inferCNV` which we could use some checks for. Some ideas:

| Metric | Description | Calculation | Required files | 
|--------|--------------------------------------|-------------|---------------|
| `scimilarity_model` | Name of the foundation model  |`metadata(processed_sce)$scimilarity_model` (or `NA` if it's `NULL`)   | `{library_id}_processed.rds` |
| `scimilarity_celltypes` | Named list of cell types  |`as.list(table(sce$scimilarity_celltype_annotation))`   | `{library_id}_processed.rds` |
| `scdblfinder_total_doublets` | Total number of cells designated as `doublet` by `scDblFinder`  |`sum(sce$scdblfinder_class == "doublet")`   | `{library_id}_filtered.rds` |
| `infercnv_reference_celltypes` | Vector of reference cell types used for defining normal reference | `metadata(sce)$infercnv_reference_celltypes` | `{library_id}_processed.rds` |
| `infercnv_total_reference_cells` | Total number of cells used in the inferCNV normal reference | `metadata(sce)$infercnv_num_reference_cells` | `{library_id}_processed.rds` |
| `infercnv_sum_totalcnv` | Sum of the total CNV events in the processed library | `sum(sce$infercnv_total_cnv)`  | `{library_id}_processed.rds` |

- noting the scimilarity ideas are inspired by existing cell type checks https://github.com/AlexsLemonade/scpca-nf/blob/f8ee2a65887eaeb3ad137b64291984c3342e27f0/bin/sce_metrics.R#L145-L158
- we don't need all the inferCNV checks I listed, but this is a pool to choose from

sjspielman:
Closes #1154 

This PR adds some new metrics to capture new features we've implemented. This doesn't have to go in as part of #1160, but I may have caught something we want to include in that release, so read on!

I updated the metrics script to add:
- `scdblfinder_total_doublets`: Total number of cells that == "doublet", calculated from the filtered sce (rest from processed)
- `infercnv_total_cnv`: Sum of `infercnv_total_cnv` colData column
- `infercnv_reference_cells`: Named list of the reference cells (gives us both counts and cell types)
- `scimilarity_model`:  The model name
- `scimilarity_celltypes`: Named list of cell types

I ran the script with `SCPCS000001` as an example and it produced this `metrics.json` with these additions _mostly_ looking as expected: [metrics.json](https://github.com/user-attachments/files/25046341/metrics.json)

The one that doesn't look how I'd expect is the `scimilarity_model` - looks like we saved the full S3 path, which probably wasn't our goal! To address this, we'd want to change this line to include `basename()` probably:
https://github.com/AlexsLemonade/scpca-nf/blob/f61e6767096339fcaa0c2f4aaf65ec5ffc142196/bin/add_celltypes_to_sce.R#L256-L257

Please let me know if you'd like me to make that (or some other?) change here or in a standalone PR to make it into `v0.9.3` without bringing the new metric calculations along.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -444,7 +444,7 @@ add_infercnv_reference_cells <- function(
     dplyr::filter(validation_group_annotation %in% reference_validation_groups) |>
     dplyr::select(consensus_ontology, consensus_annotation) |>
     dplyr::distinct() |>
-    # keep only the cell types that appear in the SCE
+    # keep only the cell types that actually appear in the SCE
     dplyr::filter(consensus_annotation %in% unique(sce$consensus_celltype_annotation))
 
   # Add reference cell information to SCE


allyhawkins:
> As I'm typing this, would we actually want to use the infercnv_diagnosis_groups instead of all these cells? Slightly smaller!

I think it makes more sense to show the actual cell type labels that were used for the given library as normal reference cells, so not the diagnosis groups. 

A related comment is that it looks like `infercnv_reference_celltypes` in the objects is a list of all possible consensus cell types that can be used to create a reference for a given diagnosis group. I actually wonder if we should update this function to instead only add the cell types that are reference cell types AND found in the actual object? I don't think it makes sense to include cell types that aren't even present in the first place. If you agree we should update that function to filter based on what's present in the `sce$consensus_celltype_annotation` column before adding to the object. 

https://github.com/AlexsLemonade/scpca-nf/blob/f61e6767096339fcaa0c2f4aaf65ec5ffc142196/bin/add_celltypes_to_sce.R#L426-L456

allyhawkins (approved):
LGTM

PR #1172, opened 2026-02-04
Merge `development` into `main` for v0.9.3

allyhawkins:
Towards #1160 

This merges in the changes that are in `development` to `main`, which includes: 

- Refactoring of inferCNV to use `infercnv_status`
- Updates to the QC report to account for missing validation groups and update dot plot sizing
- A new copy of the consensus cell type reference from `OpenScPCA-analysis` v0.2.4
- An update to the validation palette to include all possible validation groups 

I ran this through with `-profile ccdl,batch --perform_celltyping --perform_cnv_inference` and everything looked as expected.

PR #1173, opened 2026-02-04
Merge changes from `development` to `main`

allyhawkins:
Re-opened version of #1172 but using a feature branch instead. 

Towards #1160 

This merges in the changes that are in `development` to `main`, which includes: 

- Refactoring of inferCNV to use `infercnv_status`
- Updates to the QC report to account for missing validation groups and update dot plot sizing
- A new copy of the consensus cell type reference from `OpenScPCA-analysis` v0.2.4
- An update to the validation palette to include all possible validation groups 

I ran this through with `-profile ccdl,batch --perform_celltyping --perform_cnv_inference` and everything looked as expected.

### DIFF
--- a/.github/workflows/run-batch.yaml
+++ b/.github/workflows/run-batch.yaml
@@ -1,0 +1,80 @@
+name: Run Workflow on AWS Batch
+
+on:
+  workflow_dispatch:
+    inputs:
+      revision:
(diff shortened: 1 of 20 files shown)


jashapiro (approved):
🚀

PR #1177, opened 2026-02-10
remove example containers

jashapiro:
As part of the new infra revisions, we don't want to have these containers set, even for the example repo. This PR removes them from the last place we had them.


sjspielman (approved):
lgtm!

Noting to review this, I ran in the `scpca-prod-account` branch ` grep -r .  `, and there were no other remaining spots besides here. Nothing appears to be in docs about maintaining these either, so nothing else to rip out.

PR #1178, opened 2026-02-10
add a bit more memory for infercnv

jashapiro:
I noticed during the last set of runs runs that infercnv has a strong chance of failure in its first attempt with the current memory setup, so I added a bit more for it. 

This meant adding one more tag, but otherwise a pretty minimal change!


sjspielman (approved):
Yeah, seems for the best. 👍

Issue #1164, opened 2026-02-02
Create a formatting reference file


allyhawkins:
As discussed in https://github.com/AlexsLemonade/scpca-nf/issues/1148#issuecomment-3781463940, we will need to first create a reference file with all the expected formatting items to be checked in SCEs and AnnData objects. We want this to be a JSON file that is created programmatically and not manually. 

I was originally thinking we could have something like this, with keys for each type of object: 

```
[
"unfiltered_sce": {
	"assays": ["counts", "spliced"},
	"colData": {
		"column1" : "column_type",
		"column2" : "column_type"
	},
	"rowData": {
		 "column1" : "column_type", 
		"column2" : "column_type" 
	},
	…
},
"filtered_sce": { … } ,
"processed_sce": { … } , 
"processed_sce_celltyping": {…}, 
"processed_sce_infercnv": {…}
} 
```

allyhawkins:
We have initiated this and made many of the necessary design decisions so I'm going to close this. I filed follow up issues to track completing this: 
#1193 
#1194 
#1195

PR #1187, opened 2026-02-16
Initiate scpca-nf output formatting reference

allyhawkins:
Towards #1164 

This PR includes the initial creation of a JSON file that can serve as a reference to checking the output format of objects from scpca-nf. I'm filing this first as a draft because I want to see what other people think about this direction. 

The overall structure of the file that I'm going for is to have two main dictionaries, one for SCE and one for AnnData. Then within each of those we can have a dictionary with the expected contents of the unfiltered, filtered, and processed objects (and eventually merged). I spent a lot of time trying to figure out the best approach here and even played around with using a TSV, but I think I landed on something reasonable.

Where this gets a little tricky is with the special conditions like adt and cellhashing. Here we have both the `altExp` objects that contain their own associated sets of assays and col/rowData. Then we also have additional columns that are present in the main `colData`. To account for this, I'm using two separate dictionaries: 
- Within each unfiltered/filtered/processed object is the `altExp` dictionary which holds the contents of the altExp for either adt or cellhash. This includes anything that's found in the `altExp`, not the main experiment. 
- The `has_adt` dictionary is used to specify any additional columns that should be present in the main experiment. 

I am only showing adt as an example here, but I would do the same for cell hash. We can build on this concept for cell typing and inferCNV and include a `has_celltyping` and `has_infercnv` entry within the dictionary for processed objects. 

The contents of each of the dictionaries are defined at the top of the script that's used to build the file. So if there are new columns that we want to add or new things we want to check we will just need to add them to the appropriate dictionary at the top of the script. 

I went through example SCE objects to get everything that we should check for and I noted a few todos of items that are present in most objects, but not all in the script. I think if the column is not a gaurantee, we should not include a check for it. This applies to `sizeFactor` which is only present if normalization succeeds and `cluster`. It also applies to some of the special adt normalization columns that are only present with certain types of normalization. 

I'm going to request review to get some initial feedback on the structure and setup. Once we finalize that with just the SCEs, I'll start documenting these decisions and continue building the rest of these sections in subsequent PRs.


sjspielman:
I think the approach you are taking to store & build this all up this is fine. In terms of the items that might be missing due to failures, I actually think we might want to take a step back and track those failures so that (perhaps in the future) we can more easily check for those columns? Something akin to what we do for `infercnv_status` (speaking of which, I'm assuming the `metadata` slot is coming in future PRs, which makes sense since there's a lot here!).

A somewhat related thought I have too is, I think it might be worth establishing a checklist (e.g., issue template?) for adding new content to SCEs, with tasks (which can be "open issues for accomplishing this task") like:
- add it to the formatting checks
- add to metrics checks
- add it to scpca-docs
- update example outputs
- (are there more things?)

Not something we need to do _now_, but something like this wouldn't hurt in the not-terribly-distant future.

allyhawkins:
> I think the approach you are taking to store & build this all up this is fine. In terms of the items that might be missing due to failures, I actually think we might want to take a step back and track those failures so that (perhaps in the future) we can more easily check for those columns? Something akin to what we do for infercnv_status (speaking of which, I'm assuming the metadata slot is coming in future PRs, which makes sense since there's a lot here!).

So we do have items in the metadata that help us indicate which columns are present for these scenarios, but I'm struggling with how to encode that information in this specific reference, or at least for some of them we do. For example, with `sizeFactor`, it's not there if normalization fails, which is indicated in `metadata(sce)$normalization`. Maybe for things like this, we could use the metadata dictionary to define different columns that should be present? 

```
sce{
  filtered {
    metadata {
      normalization {
        deconvolution{
           colData { sizeFactor: numeric }  
           }
        }
      }   
   } 
}
```

> A somewhat related thought I have too is, I think it might be worth establishing a checklist (e.g., issue template?) for adding new content to SCEs, with tasks (which can be "open issues for accomplishing this task") like:

> - add it to the formatting checks
> - add to metrics checks
> - add it to scpca-docs
> - update example outputs
> - (are there more things?)

> Not something we need to do now, but something like this wouldn't hurt in the not-terribly-distant future.

I think this is a reasonable future idea, but should probably live in `ScPCA-admin`. Also, the updated example output should reflect the latest release of the workflow, so we only want to update that during an `scpca-nf` release, which is currently part of the checklist.

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,115 @@
...
+
+# TODO: do we include the colData that's present in filtered projects?
+# See https://scpca.readthedocs.io/en/stable/sce_file_contents.html#additional-singlecellexperiment-components-for-cite-seq-libraries-with-adt-tags
+# A lot of these are only present in certain conditions, so it doesn't seem worth the check here?
+
+# build base SCE -----------------
+
+unfiltered_sce = {
+    "assayNames": assays,
+    "colData": {**cell_metadata},
```
jashapiro: Do you need to expand this? I would expect just using the dictionary directly to work unless you were going to later modify it. 
```suggestion
    "colData": cell_metadata,
```

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,115 @@
...
+        "assayNames": adt_assays,
+        "colData": {**adt_unfiltered_altexp_cell_metadata},
+        "rowData": {**adt_unfiltered_altexp_gene_metadata},
+    }
+}
+
+# build filtered SCE using unfiltered as a base ------
+filtered_sce = deepcopy(unfiltered_sce)
+
+filtered_sce["colData"].update({**filtered_cell_metadata})
```
jashapiro: Oh, you **are** updating it here.

I think it might be clearer to make `filtered_cell_metadata` at the top a copy that is updated with additional fields rather than doing this piece by piece. So you would replace all of `colData` with the `filtered_cell_metadata`

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,115 @@
...
+    "total": "numeric",
+}
+
+filtered_cell_metadata = {
+    "prob_compromised": "numeric",
+    "miQC_pass": "logical",
+    "scpca_filter": "character",
+    "scDblFinder_class": "character",
+    "scDblFinder_score": "numeric",
+}
```
jashapiro, line 45: I think it might be nice if this were the complete set (see some comments below)
```suggestion
filtered_cell_metadata = deepcopy(cell_metadata).update({
    "prob_compromised": "numeric",
    "miQC_pass": "logical",
    "scpca_filter": "character",
    "scDblFinder_class": "character",
    "scDblFinder_score": "numeric",
})
```

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,115 @@
...
+
+# TODO: don't include these actually because if either of these processes fail then they won't be there
+processed_cell_metadata = {
+    "sizeFactor": "numeric",
+    "cluster": "factor",
+}
```
jashapiro, line 80: I think you do want to include these in the document, but maybe with a type prefix, like `optional_*`?
```suggestion
# TODO: don't include these actually because if either of these processes fail then they won't be there
processed_cell_metadata = {
    "sizeFactor": "optional_numeric",
    "cluster": "optional_factor",
```

references/formatting-reference.json
```diff
@@ -0,0 +1,157 @@
...
+            "rowData": {
+                "gene_ids": "character",
+                "gene_symbol": "character",
+                "mean": "numeric",
+                "detected": "numeric"
+            },
+            "has_submitter": {
+                "colData": {
+                    "submitter_celltype_annotation": "character"
+                }
```
jashapiro: I think I would reverse the nesting for these to make the organization a bit more uniform at the top level. If you use something like `colData_conditional`, then it could look like this:

```suggestion
            "colData_conditional": {
                "has_submitter": {
                    "submitter_celltype_annotation": "character"
                },
                "has_adt": {
                    "adt_scpca_filter": "character"
                }
                
```

jashapiro (commented):
Overall I think this is on the right track. My first thought is that you may want to have separate files for SCE and AnnData; they will likely be handled by separate programs and it takes the nesting down one level.

The other big thought was to group conditional fields first by where they go, rather than the conditions; I made a suggestion to that point but the overall look would be something like this:

```
"colData_conditional":{
    "has_submitter": {"submitter_celltype_annotation": "character"},
    "has_adt": {"adt_scpca_filter": "character"}
}
```

My other thoughts are a little about how you build things up in the Python code, which are really minor and I don't necessarily hate what you have, just that it might be a bit hard to debug later.

allyhawkins:
> Overall I think this is on the right track. My first thought is that you may want to have separate files for SCE and AnnData; they will likely be handled by separate programs and it takes the nesting down one level.

This is a great idea! 

> The other big thought was to group conditional fields first by where they go, rather than the conditions; I made a suggestion to that point but the overall look would be something like this:
> 
> ```
> "colData_conditional":{
>     "has_submitter": {"submitter_celltype_annotation": "character"},
>     "has_adt": {"adt_scpca_filter": "character"}
> }
> ```

I also think this will work and be a bit nicer. I like your comment about using `optional` too. I will revisit this with these suggestions and get it ready for a real review.

allyhawkins:
Okay I made a few updates to this based on initial feedback and am marking this ready for review. 

- I am now exporting a separate SCE and AnnData reference file. 
- For any of the cell metadata items that are conditional, I've grouped them together as `colData_conditional` and then include `has_adt`, `has_clusters`, etc. within that. I had originally started with optional and conditional, but I think we just need conditional. 
  - I did leave one TODO here about handling the cell typing columns. I think we might want to nest that further with `has_singler`, `has_scimilarity` etc.
- I built each SCE with the full structure rather than copying and updating specific parts. What do we think of this approach? 

I have not yet included the object metadata (since that list is very long), the cellhash information, merged objects, or anything for AnnData. Those things will have to come later.

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+
+# processed cell metadata ------------
+
+processed_conditional_cell_metadata = {
+    **filtered_conditional_cell_metadata,
+    "has_normalization": {"sizeFactor": "numeric"},
+    "has_clusters": {"cluster": "factor"},
+    # TODO: should add another level of nesting with has_singler, has_cellassign, etc?
+    "has_celltyping": {
```
sjspielman: I think you probably need to split this more into `has_<method>`. Presumably we'll use vector we save to the SCE metadata slot to determine which of these columns needs to be checked, so the nesting seems important to go hand-in-hand with that.

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+    "assayNames": assays,
+    "colData": cell_metadata,
+    "rowData": gene_metadata,
+    "colData_conditional": conditional_cell_metadata,
+    "altExp": {
+        "adt": {
+            "assayNames": adt_assays,
+            "colData": adt_unfiltered_altexp_cell_metadata,
+            "rowData": adt_unfiltered_altexp_gene_metadata,
+        }
```
sjspielman:
```suggestion
        },
```

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+    "colData": filtered_cell_metadata,
+    "rowData": gene_metadata,
+    "colData_conditional": filtered_conditional_cell_metadata,
+    "altExp": {
+        "adt": {
+            "assayNames": adt_assays,
+            "colData": adt_filtered_altexp_cell_metadata,
+            "rowData": adt_unfiltered_altexp_gene_metadata,
+            "colData_conditional": adt_filtered_conditional_altexp_cell_metadata,
+        }
```
sjspielman:
```suggestion
        },
```

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+    "rowData": gene_metadata,
+    "colData_conditional": processed_conditional_cell_metadata,
+    "reducedDimNames": processed_embeddings,
+    "altExp": {
+        "adt": {
+            "assayNames": adt_assays,
+            "colData": adt_filtered_altexp_cell_metadata,
+            "rowData": adt_unfiltered_altexp_gene_metadata,
+            "colData_conditional": adt_processed_conditional_altexp_cell_metadata,
+        }
```
sjspielman:
```suggestion
        },
```

sjspielman (commented):
> I built each SCE with the full structure rather than copying and updating specific parts. What do we think of this approach?

I like the "verbosity" of this actually, being able to quite explicitly see what's going into each dictionary. But, I do worry that as we continue building this script up, having the full structure might make the script hard to follow/maintain just because of sheer length. To that point, I wonder if storing (some of?) the core dictionaries that are used to build up the final one we export in standalone jsons that this script would read in. I'm really just considering this from an ease-of-maintenance perspective. 

Personally, I'm of the opinion that this structure is a fine place to start and mior refinements might come later when we actually start _using_ this json, since (for me at least) would make it easier to know the most helpful organizational tweaks.

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+filtered_cell_metadata = {
+    **cell_metadata,
+    "prob_compromised": "numeric",
+    "miQC_pass": "logical",
+    "scpca_filter": "character",
+    "scDblFinder_class": "character",
+    "scDblFinder_score": "numeric",
+}
+
+filtered_conditional_cell_metadata = {
```
jashapiro: This is quite of minor, but I don't like the `conditional` hiding in the middle of the this value. I think I might change the original to `cell_metadata_conditional` and then use `filtered_cell_metadata_conditional` or not change the main value and use `conditional_filtered_cell_metadata`. The advantage of the former is that it better matches the nesting order.

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+# row metadata ----------
+gene_metadata = {
+    "gene_ids": "character",
+    "gene_symbol": "character",
+    "mean": "numeric",
+    "detected": "numeric",
+}
+
+# reduced dimensionality ----------
+processed_embeddings = ["PCA", "UMAP"]
```
jashapiro: I feel like maybe just `reduced_dims` as the name here?

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+
+# alt exps gell/gene metadata ------------
+
+adt_unfiltered_altexp_cell_metadata = {
+    "altexps_adt_sum": "numeric",
+    "altexps_adt_detected": "integer",
+    "altexps_adt_percent": "numeric",
+}
+
```
jashapiro: I think these should be included in the main conditional cell metadata with `has_adt`? I see a quite reasonable argument for keeping all the ADT code together, so I might just update here: 
```suggestion
conditional_cell_metadata.update({
    "has_adt": {
         "altexps_adt_sum": "numeric",
         "altexps_adt_detected": "integer",
        "altexps_adt_percent": "numeric",
    }
})

```

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,188 @@
...
+adt_unfiltered_altexp_gene_metadata = {
```
jashapiro: By contrast, this is in a separate table, so makes sense to be separate. 

But maybe `altexp_adt_feature_metadata` to match the nesting order?

(add that I left off `unfiltered` because it wasn't in the original `gene_metadata`; I can go either way on whether that should be included, but I do think it should be consistent.)

references/scripts/create-formatting-reference.py
jashapiro: It we are going to worry about things like this, we should add a precommit hook and be done with it.

jashapiro (commented):
Overall this looks quite good to me. 

The most important suggestion here is the placement of adt fields that should be in the main colData. (And follow the same pattern for cellhash data as you add it.

The other "big"  suggestion is to make your variable names match the nesting order in the final JSON file, which I think will make everything a bit easier to track. I also have a small preference for `feature_metadata` over `gene_metadata` but that is quite weak.

allyhawkins:
@sjspielman and @jashapiro this should be ready for another round of review. 

- I broke up the cell typing methods further, since that should make it easier to track those when we write the script to do the actual checking. 
- I fixed the error in the altExp colData that is part of the main colData that Josh pointed out in https://github.com/AlexsLemonade/scpca-nf/pull/1187#discussion_r2828367254. 
- I switched the variable naming as suggested to be more uniform and consistent throughout. 

> I like the "verbosity" of this actually, being able to quite explicitly see what's going into each dictionary. But, I do worry that as we continue building this script up, having the full structure might make the script hard to follow/maintain just because of sheer length. To that point, I wonder if storing (some of?) the core dictionaries that are used to build up the final one we export in standalone jsons that this script would read in. I'm really just considering this from an ease-of-maintenance perspective.

I think this could be something we consider, but I also worry about generating too many reference files and making things overcomplicated. I also like the verbosity of having the full SCE structure so I'm going to leave it and then if we find this gets too long or we need to split things up more in the future we can.

sjspielman (approved):
Still looking good to me!

references/scripts/create-formatting-reference.py
```diff
@@ -0,0 +1,198 @@
...
+        "has_scimilarity": {
+            "scimilarity_celltype_annotation": "character",
+            "scimilarity_celltype_ontology": "character",
+            "scimilarity_min_distance": "numeric",
+        },
+        "has_consensus": {
+            "consensus_celltype_annotation": "character",
+            "consensus_celltype_ontology": "character",
+        },
+    },
```
jashapiro: I'm not sure you will want this nested; I think the extra complexity will make it harder to use when you go to access these. 


```suggestion
    "has_singler": {
        "singler_celltype_annotation": "character",
        "singler_celltype_ontology": "character",
    },
    "has_cellassign": {
        "cellassign_celltype_annotation": "character",
        "cellassign_celltype_ontology": "character",
        "cellassign_max_prediction": "numeric",
    },
    "has_scimilarity": {
        "scimilarity_celltype_annotation": "character",
        "scimilarity_celltype_ontology": "character",
        "scimilarity_min_distance": "numeric",
    },
    "has_consensus": {
        "consensus_celltype_annotation": "character",
        "consensus_celltype_ontology": "character",
    },
```

jashapiro (approved):
Looks good, but I think you might come to regret nesting the cell typing methods, so I would not do that.

PR #1189, opened 2026-02-18
Fix metrics script and report

### LINKED ISSUE #1175: Fix script comment
Noticed this wrong script comment which is definitely left over from something. 
https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/sce_metrics.R#L3

We should update this!

### LINKED ISSUE #1176: Fix cell type counting bug in metrics report
While familiarizing myself with the metrics report template, I believe I have caught a bug in how we tabulate the number of changed cell types.

To see this in action, I'm attaching two json files that I've been using to test things out with - I engineered them to have a couple different values relevant to what I'm working on for #1162, including changing some of the counts for cellassign cell types (but preserving overall counts). I did _not_ change any cell type labels however, so there should not be a difference in cell types (I just swapped b-cell and astrocyte counts).
[SCPCL000001_metrics_v1.json](https://github.com/user-attachments/files/25215131/SCPCL000001_metrics_v1.json)
[SCPCL000001_metrics_v2.json](https://github.com/user-attachments/files/25215132/SCPCL000001_metrics_v2.json)

Here's a full diff between the files for context:
```
<   "scpca_filter_count": [2500],
---
>   "scpca_filter_count": [2629],
29c29
<   "scdblfinder_total_doublets": [100],
---
>   "scdblfinder_total_doublets": [248],
36,37c36,37
<   "infercnv_total_cnv": [10],
<   "infercnv_reference_cells": ["B cell", "naive B cell", "T cell", "CD4-positive, alpha-beta T cell", "mature alpha-beta T cell", "mature T cell", "memory T cell", "CD8-positive, alpha-beta T cell", "CD8-positive, alpha-beta memory T cell", "CD4-positive, alpha-beta memory T cell", "endothelial cell", "macrophage", "microglial cell", "monocyte", "stephanie cell"],
---
>   "infercnv_total_cnv": [90163],
>   "infercnv_reference_cells": ["B cell", "naive B cell", "T cell", "CD4-positive, alpha-beta T cell", "mature alpha-beta T cell", "mature T cell", "memory T cell", "CD8-positive, alpha-beta T cell", "CD8-positive, alpha-beta memory T cell", "CD4-positive, alpha-beta memory T cell", "endothelial cell", "macrophage", "microglial cell", "monocyte", "myeloid leukocyte", "mononuclear phagocyte", "natural killer cell"],
69,70c69,70
<     "Astrocytes": [12],
<     "B cells": [3],
---
>     "Astrocytes": [3],
>     "B cells": [12],
```

## The bug

When I run this through the report, I get a result that there is a 1 cell type added and 1 cell type removed. This seems wrong! It seems to be coming from here:
https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/scripts/compare-metrics/compare-metrics-template.rmd#L1155-L1175

Specifically...
- First in `dplyr::mutate()`, we ensure all values are character. In the case when there are no changes, this vale would be `NA`. So, here, `NA` values here become `"NA"`.
- Then in the next `dplyr::mutate()` after pivoting, we unlist to create `types_added` and `types_removed`. But, this will leave us with a string value `"NA"` - which has a length of 1. As such, we end up with a value of 1 when we really should end up with a value of 0.

Locally, I updated the code to as follows and it now reports 0 as expected; see lines following the `# UPDATED: ...` comments.
```
  dplyr::mutate(
    dplyr::across(
      dplyr::ends_with(c("types_added", "types_removed")),
      \(x) purrr::map_chr(x, \(y) {
	     # UPDATED: use NA_character_ if it's actually NA
        ifelse(is.na(y), NA_character_, paste0(y, collapse = ","))
      })
    )
  ) |>
  # pivot by celltype method
  tidyr::pivot_longer(
    cols = starts_with(c("singler", "cellassign", "consensus")),
    names_to = c("celltype_method", ".value"),
    names_pattern = "(singler|cellassign|consensus)_(.*)"
  ) |>
  dplyr::mutate(
    # convert characters back to lists for types added and removed prior to counting
    types_added = purrr::map(types_added, \(x) unlist(stringr::str_split(x, ","))),
    types_removed = purrr::map(types_removed, \(x) unlist(stringr::str_split(x, ","))),
    # UPDATED: record NA instead of length if the value is NA
    types_added_count = dplyr::if_else(is.na(types_added), NA_integer_, purrr::map_int(types_added, length)),
    types_removed_count = dplyr::if_else(is.na(types_removed), NA_integer_, purrr::map_int(types_removed, length))
```

I tested this with actual cell type name differences as well (I popped 1 new cell type into added to each json file which corresponds to 1 removed and 1 added), and it works as expected now. See reports:

- current version: [current-report-no-changes-expected.html](https://github.com/user-attachments/files/25215215/current-report-no-changes-expected.html)
- updated code without cell type changes: [updated-report-no-changes-expected.html](https://github.com/user-attachments/files/25215219/updated-report-no-changes-expected.html)
- updated code with cell type changes: [updated-report-changes-expected.html](https://github.com/user-attachments/files/25215216/updated-report-changes-expected.html)

sjspielman:
Closes #1175: updates the script comment
Closes #1176: updates the report

For this PR, I updated the metrics report to fix a bug in cell type counting I caught the other week. I tested this with a comparison JSON that had an additional cell type (aptly named `v2 additional celltype`): 
[compare-metrics-template.html](https://github.com/user-attachments/files/25400504/compare-metrics-template.html)
More details about testing are in #1176, but I didn't repeat the full set of testing I did the other week. Please let me know if you would like to see some refreshed examples and I can certainly generate reports for some more cases!

There are also some comments that say temporary for the cache rds - perhaps this began as temporary but ended up being a fixture? I cleaned those up since it doesn't seem temporary anymore - let me know if it should be though and I can take things the other direction.

### DIFF
--- a/scripts/compare-metrics/compare-metrics-template.rmd
+++ b/scripts/compare-metrics/compare-metrics-template.rmd
@@ -292,13 +292,12 @@ metrics_df <- dplyr::full_join(
   by = c("project_id", "sample_id", "library_id"),
   suffix = c(".ref", ".comp")
 )
-# temporary write for testing
+# write for caching
 readr::write_rds(metrics_df, "metrics.rds")
 ```
 
 ```{r, eval=use_cache}
 #| include: false
-# temporary chunk for testing
 metrics_df <- readr::read_rds("metrics.rds")
 ```
 
@@ -1156,7 +1155,10 @@ types_stats <- metrics_df |>
   dplyr::mutate(
     dplyr::across(
       dplyr::ends_with(c("types_added", "types_removed")),
-      \(x) purrr::map_chr(x, \(y) paste0(y, collapse = ","))
+      \(x) purrr::map_chr(x, \(y) {
+	      # use NA_character_ if it's actually NA
+        ifelse(is.na(y), NA_character_, paste0(y, collapse = ","))
+      })
     )
   ) |>
   # pivot by celltype method
@@ -1169,8 +1171,9 @@ types_stats <- metrics_df |>
     # convert characters back to lists for types added and removed prior to counting
     types_added = purrr::map(types_added, \(x) unlist(stringr::str_split(x, ","))),
     types_removed = purrr::map(types_removed, \(x) unlist(stringr::str_split(x, ","))),
-    types_added_count = purrr::map_int(types_added, length),
-    types_removed_count = purrr::map_int(types_removed, length)
+    # record NA instead of length if the value is actually NA
+    types_added_count = dplyr::if_else(is.na(types_added), NA_integer_, purrr::map_int(types_added, length)),
+    types_removed_count = dplyr::if_else(is.na(types_removed), NA_integer_, purrr::map_int(types_removed, length))
   ) |>
   dplyr::relocate(types_added_count, types_removed_count, .before = types_added)
 
@@ -1279,8 +1282,8 @@ changed_types |>
     colnames = c(
       "Project", "Sample", "Library",
       "Cell type method", "Minimum changed cell types", "Scaled distance",
+      "Added cell types", "Removed cell types",
       "Added cell type count", "Removed cell type count",
-      "Added cell types", "Removed cell types"
     )
   )
 ```
--- a/scripts/compare-metrics/compare-metrics-template.rmd
+++ b/scripts/compare-metrics/compare-metrics-template.rmd
@@ -1156,7 +1156,7 @@ types_stats <- metrics_df |>
     dplyr::across(
       dplyr::ends_with(c("types_added", "types_removed")),
       \(x) purrr::map_chr(x, \(y) {
-	      # use NA_character_ if it's actually NA
+        # use NA_character_ if it's actually NA
         ifelse(is.na(y), NA_character_, paste0(y, collapse = ","))
       })
     )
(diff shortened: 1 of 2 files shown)


scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -1279,8 +1282,8 @@ changed_types |>
     colnames = c(
       "Project", "Sample", "Library",
       "Cell type method", "Minimum changed cell types", "Scaled distance",
-      "Added cell type count", "Removed cell type count",
-      "Added cell types", "Removed cell types"
+      "Added cell types", "Removed cell types",
+      "Added cell type count", "Removed cell type count"
```
allyhawkins: Looking at the example report you have the first column is the cell type count and the second one is the cell types themselves, so this change should be reverted.

allyhawkins (commented):
This looks good with one small comment about a change you made that I don't think should have been made.

scripts/compare-metrics/compare-metrics-template.rmd
jashapiro: Just noting that we reorder these columns on line 1178; we have to compute the changed types first and then count them, but we want to present them in the opposite order.

allyhawkins (approved):
LGTM

Issue #1196, opened 2026-02-27
Update Cavatica version to v0.9.3


allyhawkins:
We need to update the `scpca-nf` app on Cavatica to reflect the new version v0.9.3

allyhawkins:
This was updated and saved to a new app, `scpca-nf-development`. The app is within the `scpca-nf-test` project.

PR #1199, opened 2026-03-03
Add command line for app name

allyhawkins:
As part of #1196, we updated the `scpca-nf-test` project to include a new app `scpca-nf-development`.  I'm just adding a command line argument to specify the app name and setting the default to be the new development one so we don't accidentally update the other one while testing.

### DIFF
--- a/cavatica/cavatica-build.sh
+++ b/cavatica/cavatica-build.sh
@@ -13,7 +13,17 @@ exclude_file="sb_exclude.txt"
 # use "build" or "push": default is build
 ACTION=${1:-"build"}
 # cavatica app id: <username>/<project>/<app>
-APP_ID=${2:-"jashapiro/scpca-nf-test/scpca-nf-development"}
+APP_TYPE=${2:-development}
+
+project_id="jashapiro/scpca-nf-test"
+
+if [ $APP_TYPE == "development" ] ; then
+  app_id="${project_id}/scpca-nf-development"
+elif [ $APP_TYPE != "main" ]; then
+  app_id="${project_id}/scpca-nf"
+else
+  echo "The second argument must be 'development' or 'main'."
+fi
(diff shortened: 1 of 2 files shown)


cavatica/cavatica-build.sh
```diff
@@ -5,13 +5,15 @@ set -euo pipefail
...
+# default files for docs and exclusions for upload
 doc_file="sb_doc.md"
 exclude_file="sb_exclude.txt"
 
-# get the action from the command line, default is build
+# get the action and app ID from the command line
+# use "build" or "push": default is build
 ACTION=${1:-"build"}
+# cavatica app id: <username>/<project>/<app>
+APP_ID=${2:-"jashapiro/scpca-nf-test/scpca-nf-development"}
```
jashapiro: I'd keep the prefix as an internal variable and just pull in the final element from the command line. Or make an argument that is rather "dev" or "main" and then set app_id based on that. 

Just trying to save typing/typos on the command line.

A simple version of this might be: 

```suggestion
APP_TYPE=${2:-development}

project_id="jashapiro/scpca-nf-test"

if [ $APP_TYPE == "development" ] ; then
  app_id="${project_id}/scpca-nf-development"
elif [ $APP_TYPE != "main" ]; then
  app_id="${project_id}/scpca-nf"
else
  echo "The second argument must be 'development' or 'main'."
fi
```

allyhawkins:
While working on running this in Cavatica, I noticed that there were a few outdated reference buckets, so I also fixed that here. I was also able to update the app to reflect this.

jashapiro (approved):
Overall this seems good, with just a suggestion to be a bit more restrictive (and concise) about the argument.

Thanks for updating the reference path. Not sure why the references there are not up to date though!

PR #1200, opened 2026-03-04
Add cellhash section of the reference file

### LINKED ISSUE #1193: Formatting reference file: Cellhash section
This issue tracks filling in the expected formatting for libraries with cell hashing to the reference file created in https://github.com/AlexsLemonade/scpca-nf/pull/1187. We should follow the same design choices that we made for adt.

allyhawkins:
Closes #1193 

Here I'm adding the information for libraries with cell hashing. This includes what's present in both the main experiment and altExp. I think the trickiest thing here is that there are columns present for each of the demultiplexing methods, but depending on how things were run not every method will be present. So I followed the same format we use for cell typing methods where I just use "has_hashedDrops" and "has_vireo" and include the columns that are present for each of those.

### DIFF
--- a/references/scripts/create-formatting-reference.py
+++ b/references/scripts/create-formatting-reference.py
@@ -161,6 +161,7 @@ filtered_altexp_cellhash_feature_metadata = {
 filtered_altexp_cellhash_cell_metadata_conditional = {
     # indicate columns that are conditionally present based on demultiplexing methods used
     # this is tracked based on demux columns in colData
+    # this is the same as what's in the processed object
     "has_hashedDrops": {
         "hashedDrops_Total": "numeric",
         "hashedDrops_Best": "integer",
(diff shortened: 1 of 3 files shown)


references/scripts/create-formatting-reference.py
```diff
@@ -132,6 +146,44 @@
...
+    "detected": "numeric",
+}
+
+filtered_altexp_cellhash_feature_metadata = {
+    **unfiltered_altexp_cellhash_feature_metadata,
+    "barcode_id": "character",
+    "sample_id": "character",
+}
+
+filtered_altexp_cellhash_cell_metadata_conditional = {
```
sjspielman, line 161: Let's add a comment that this is also the same content as the processed altExp, so future us remembers assigning this to processed isn't a bug.

sjspielman (approved):
LGTM! 

Noting that I reviewed this using a set of files from a SCPCP000009 library to confirm both the fields and their types. Using the `has_demuxflavor` makes definitely makes sense here. The code here also appears consistent with how we treat adts, updates for which are reasonable now that we have more stuff in the conditionals.

PR #1202, opened 2026-03-04
account for metadata in SCE

### LINKED ISSUE #1201: Add metadata for SCE objects to reference files
Before we move to AnnData, we need to add in the metadata section to the formatting reference for SCEs.

allyhawkins:
Stacked on #1200 
Closes #1201 

This PR adds in the contents of the `metadata` slot in the SCE objects. I chose not to track the type for each of the metadata slots since depending on what happens in the workflow the actual content of some of these is slightly different and I think it's more important that we just check they are present. To do this I loaded in different versions of the objects and listed out the names and cross-checked that with what we have in the docs: https://scpca.readthedocs.io/en/stable/sce_file_contents.html#singlecellexperiment-experiment-metadata

There are some minor differences that will ultimately be present for samples processed with cellranger-multi vs alevin-fry so I accounted for that now in the `metadata_conditional` slot. We also have one random metadata value that's only present if the filtering slot is `UMI cutoff` so I included that as a conditional too. The other conditionals mirror the ones we have for other slots (colData, etc).

### DIFF
--- a/references/scripts/create-formatting-reference.py
+++ b/references/scripts/create-formatting-reference.py
@@ -126,7 +126,7 @@ unfiltered_experiment_metadata = {
     "assay_ontology_term_id": "character",
     "seq_unit": "character",
     "transcript_type": "character",
-    "sample_metadata": "character",
+    "sample_metadata": "data.frame",
     "sample_type": "character",
 }
 
@@ -195,7 +195,7 @@ processed_experiment_metadata_conditional = {
     },
     "has_singler": {
         "celltype_methods": "character",
-        "singler_results": "S4",
+        "singler_results": "DFrame",
         "singler_reference": "character",
         "singler_reference_label": "character",
         "singler_reference_source": "character",
@@ -203,7 +203,7 @@ processed_experiment_metadata_conditional = {
     },
     "has_cellassign": {
         "celltype_methods": "character",
-        "cellassign_predictions": "character",
+        "cellassign_predictions": "data.frame",
         "cellassign_reference": "character",
         "cellassign_reference_organs": "character",
         "cellassign_reference_source": "character",
@@ -219,7 +219,7 @@ processed_experiment_metadata_conditional = {
         "infercnv_num_reference_cells": "integer",
         "infercnv_diagnosis_groups": "character",
         "infercnv_status": "character",
-        "infercnv_options": "character",
+        "infercnv_options": "list",
         "infercnv_table": "character",
     },
     "has_adt": {
--- a/references/scripts/create-formatting-reference.py
+++ b/references/scripts/create-formatting-reference.py
@@ -155,7 +155,7 @@ filtered_experiment_metadata = {
     **unfiltered_experiment_metadata,
     "filtering_method": "character",
     # this is NA if miQC failed
-    "prob_compromised_cutoff": "NA",
+    "prob_compromised_cutoff": "logical",
     "scpca_filter_method": "character",
     "min_gene_cutoff": "integer",
 }
@@ -179,7 +179,7 @@ processed_experiment_metadata_conditional = {
     # if empty drops filtering_method is UMI cutoff
     "umi_filtering": {"umi_cutoff": "numeric"},
     # if miQC is present, these should have specific contents
-    "has_miQC": {"miQC_model": None, "prob_compromised_cutoff": "numeric"},
+    "has_miQC": {"prob_compromised_cutoff": "numeric"},
     "has_clusters": {
         "cluster_algorithm": "character",
         "cluster_weighting": "character",
@@ -200,6 +200,10 @@ processed_experiment_metadata_conditional = {
         "singler_reference_label": "character",
         "singler_reference_source": "character",
         "singler_reference_version": "character",
+        # TODO: these are NA right now because we are using the old SingleR refs
+        # only projects that use the new refs will have these set correctly...
+        "singler_gene_set_version": "logical",
+        "singler_date": "logical",
     },
     "has_cellassign": {
         "celltype_methods": "character",
@@ -260,7 +264,10 @@ processed_altexp_adt_cell_metadata_conditional = {
     "sizeFactor": "numeric",
 }
 
-filtered_altexp_adt_experiment_metadata = {"ambient_profile": "character"}
+filtered_altexp_adt_experiment_metadata = {
+    **unfiltered_experiment_metadata,
+    "ambient_profile": "character",
+}
 
 # cellhash specific items
 unfiltered_altexp_cellhash_feature_metadata = {
@@ -314,6 +321,8 @@ unfiltered_sce = {
         "adt": {
             "assayNames": adt_assays,
             "rowData": altexp_adt_feature_metadata,
+            "metadata": unfiltered_experiment_metadata,
+            "metadata_conditional": unfiltered_experiment_metadata_conditional,
         },
         "cellhash": {
             "assayNames": cellhash_assays,
@@ -339,6 +348,7 @@ filtered_sce = {
             "rowData": altexp_adt_feature_metadata,
             "colData_conditional": filtered_altexp_adt_cell_metadata_conditional,
             "metadata": filtered_altexp_adt_experiment_metadata,
+            "metadata_conditional": unfiltered_experiment_metadata_conditional,
         },
         "cellhash": {
             "assayNames": cellhash_assays,
@@ -365,6 +375,7 @@ processed_sce = {
             "rowData": altexp_adt_feature_metadata,
             "colData_conditional": processed_altexp_adt_cell_metadata_conditional,
             "metadata": filtered_altexp_adt_experiment_metadata,
+            "metadata_conditional": unfiltered_experiment_metadata_conditional,
         },
         "cellhash": {
             "assayNames": cellhash_assays,
--- a/references/scripts/create-formatting-reference.py
+++ b/references/scripts/create-formatting-reference.py
@@ -224,7 +224,7 @@ processed_experiment_metadata_conditional = {
         "infercnv_diagnosis_groups": "character",
         "infercnv_status": "character",
         "infercnv_options": "list",
-        "infercnv_table": "character",
+        "infercnv_table": "data.frame",
     },
     "has_adt": {
         "adt_scpca_filter_method": "character",
--- a/references/scripts/create-formatting-reference.py
+++ b/references/scripts/create-formatting-reference.py
@@ -115,101 +115,117 @@ feature_metadata = {
 reduced_dims = ["PCA", "UMAP"]
 
 # experiment metadata --------------
-unfiltered_experiment_metadata = [
-    "library_id",
-    "sample_id",
-    "project_id",
-    "reference_index",
-    "total_reads",
-    "mapping_tool",
-    "tech_version",
-    "assay_ontology_term_id",
-    "seq_unit",
-    "transcript_type",
-    "sample_metadata",
-    "sample_type",
-]
+unfiltered_experiment_metadata = {
+    "library_id": "character",
+    "sample_id": "character",
+    "project_id": "character",
+    "reference_index": "character",
+    "total_reads": "integer",
+    "mapping_tool": "character",
+    "tech_version": "character",
+    "assay_ontology_term_id": "character",
+    "seq_unit": "character",
+    "transcript_type": "character",
+    "sample_metadata": "character",
+    "sample_type": "character",
+}
 
 unfiltered_experiment_metadata_conditional = {
     # experiment metadata that's based on mapping tool
     # stored in metadata(sce)$mapping_tool
-    "alevin-fry": [
-        "salmon_version",
-        "mapped_reads",
-        "alevinfry_version",
-        "af_permit_type",
-        "af_resolution",
-        "usa_mode",
-        "af_num_cells",
-        "include_unspliced",
-    ],
-    "cellranger-multi": [
-        "cellranger_version",
-        "reference_probeset",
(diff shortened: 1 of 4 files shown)


sjspielman (commented):
> I chose not to track the type for each of the metadata slots since depending on what happens in the workflow the actual content of some of these is slightly different and I think it's more important that we just check they are present. 

I'm not sure I'm sold on not checking types here, since I think most of them are going to be consistently the same type. Can you say a bit more about which fields seemed tricky to track here that led to this decision, and are they..
- situations where it could be eg one of two types?
- situations where we could know the expected type based on other information in the sce?

Also, do we want to track the `metadata(sce)$sample_metadata` contents themselves? I can certainly see that being a separate PR though!

allyhawkins:
> I'm not sure I'm sold on not checking types here, since I think most of them are going to be consistently the same type.

The main reason I wasn't tracking this is because things get kind of tricky with the miQC model and we have some S4 objects in there. The `miQC_model` is only present if miQC was successful and it gets removed in the processed object to be `NULL`. If `miQC` works then we have a `prob_compromised_cutoff` thats a number, otherwise its NA. I tried to handle this with adding it as something that's always present, but only numeric if miQC has worked. The other option is we just don't track what that value is but just that it's present. 

I think for the other ones it's the same type throughout or not present based on conditionals that we already account for. 

> Also, do we want to track the metadata(sce)$sample_metadata contents themselves? I can certainly see that being a separate PR though!

I'm going to push back on this because we don't require any columns in the sample metadata to run the workflow other than `scpca_sample_id`. If this isn't there the script is going to throw an error and error out the workflow. Because of this, I think it's presence is enough to check personally.

sjspielman:
> If miQC works then we have a prob_compromised_cutoff thats a number, otherwise its NA. I tried to handle this with adding it as something that's always present, but only numeric if miQC has worked. The other option is we just don't track what that value is but just that it's present.

I will come back and give this a full review Monday (way too much Friday brain today), but wanted to reply that I think it would be reasonable for this case just to confirm it's present. But if we do want to check types, note that NA counts as logical
```
> class(NA)
[1] "logical"
```

allyhawkins:
> But if we do want to check types, note that NA counts as logical

Yes, I used `logical` for `prob_compromised_cutoff` already, but the issue is with the NULL value that we assign to the miQC_model. Fine to revisit next week!

references/scripts/create-formatting-reference.py
```diff
@@ -114,6 +114,120 @@
...
+    "sample_id": "character",
+    "project_id": "character",
+    "reference_index": "character",
+    "total_reads": "integer",
+    "mapping_tool": "character",
+    "tech_version": "character",
+    "assay_ontology_term_id": "character",
+    "seq_unit": "character",
+    "transcript_type": "character",
+    "sample_metadata": "data.frame",
```
sjspielman, line 129: This one might be fiddly because running `class()` on it gives a vector since we constructed it as a tibble:

```
> class(metadata(sce)$sample_metadata)
[1] "tbl_df"     "tbl"        "data.frame"
```

If we are going to use `class()` for checking, this may need all three and then use `all.equal()`. But, if we run `is.data.frame()` it _will_ be TRUE. So, whether this needs to change depends on how we'll ultimately perform the checking.

references/scripts/create-formatting-reference.py
```diff
@@ -114,6 +114,120 @@
...
+        "pct_mapped_reads": "numeric",
+        "cellranger_num_cells": "integer",
+    },
+}
+
+filtered_experiment_metadata = {
+    **unfiltered_experiment_metadata,
+    "filtering_method": "character",
+    # this is NA if miQC failed
+    "prob_compromised_cutoff": "NA",
```
sjspielman: Sorry, but I'm still confused why this says `NA` and not `logical` if the goal here is to record the _type_?

references/scripts/create-formatting-reference.py
```diff
@@ -114,6 +114,120 @@
...
+    },
+    # note that celltype_methods exists if either one of these methods is present
+    "has_submitter": {"celltype_methods": "character"},
+    "has_openscpca": {
+        "celltype_methods": "character",
+        "openscpca_celltype_module_name": "character",
+        "openscpca_celltype_nf_version": "character",
+        "openscpca_celltype_release_date": "character",
+    },
+    "has_singler": {
```
sjspielman, line 196: We recently added some more fields here: `singler_gene_set_version` and `singler_date`:

https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/add_celltypes_to_sce.R#L130-L133

references/scripts/create-formatting-reference.py
```diff
@@ -114,6 +114,120 @@
...
+        "celltype_methods": "character",
+        "singler_results": "DFrame",
+        "singler_reference": "character",
+        "singler_reference_label": "character",
+        "singler_reference_source": "character",
+        "singler_reference_version": "character",
+    },
+    "has_cellassign": {
+        "celltype_methods": "character",
+        "cellassign_predictions": "data.frame",
```
sjspielman, line 210: Another case where we have a tibble
```
class(metadata(sce)$cellassign_predictions)
[1] "spec_tbl_df" "tbl_df"      "tbl"        
[4] "data.frame" 
```

references/scripts/create-formatting-reference.py
```diff
@@ -114,6 +114,120 @@
...
+        "scimilarity_model": "character",
+    },
+    "has_consensus": {"consensus_celltype_methods": "character"},
+    "has_infercnv": {
+        "infercnv_reference_celltypes": "character",
+        "infercnv_num_reference_cells": "integer",
+        "infercnv_diagnosis_groups": "character",
+        "infercnv_status": "character",
+        "infercnv_options": "list",
+        "infercnv_table": "character",
```
sjspielman: This one is definitely not a tibble at least!

```suggestion
        "infercnv_table": "data.frame",
```

references/scripts/create-formatting-reference.py
```diff
@@ -114,6 +114,120 @@
...
+    "normalization": "character",
+    "highly_variable_genes": "character",
+}
+
+processed_experiment_metadata_conditional = {
+    ## reuse filtered conditional since miQC model gets removed for processed
+    # if empty drops filtering_method is UMI cutoff
+    "umi_filtering": {"umi_cutoff": "numeric"},
+    # if miQC is present, these should have specific contents
+    "has_miQC": {"miQC_model": None, "prob_compromised_cutoff": "numeric"},
```
sjspielman: Can you explain a bit more why this needs a `None` vs just not including `miQC_model` as a field here at all for processed object metadata?

references/scripts/create-formatting-reference.py
```diff
@@ -146,6 +260,8 @@
     "sizeFactor": "numeric",
 }
 
+filtered_altexp_adt_experiment_metadata = {"ambient_profile": "character"}
```
sjspielman: Should this not also include the other fields that are here? It seems that our altExp metadata is inheriting the main exp metadata fields (can't immediately find where in the code that is happening though? possibly related, i still need lunch.)
I say this, but also, I don't think we _need_ re-check those fields!!

sjspielman:
Noting I was mid-review when I saw that it wanted to me refresh so I quickly clicked the submit review button..! But, this was all my comments anyways :) 

Edit - Ah, I see that it just canceled out the reviews I left to make all those things data.frames that you were doing in that commit :)

references/scripts/create-formatting-reference.py
allyhawkins: So I was thinking about this and we get a vector of `FALSE FALSE TRUE` if we use `class()` here. I think we just want to make sure at least one of the fields match up and can use an `any()` statement, but I'm also not opposed to recording all three?

references/scripts/create-formatting-reference.py
allyhawkins: Sorry that's my mistake! I had thought I had changed this.

references/scripts/create-formatting-reference.py
allyhawkins: I think this is because the original metadata fields that are present in the unfiltered object are added when we use the function from `scpcaTool` to create the SCE object. We use that same function to create the ADT or cellhash objects and then save them as an `altExp`. I updated this so that the metadata is the same as the unfiltered objects plus any additional metadata that may be present.

references/scripts/create-formatting-reference.py
allyhawkins: No reason at all, this should just be empty so I updated that!

references/scripts/create-formatting-reference.py
allyhawkins: I added these in but note that they were added in because we had to make updated versions of the reference files when we were setting up things for flex. When we did that, we updated the naming for the SingleR ref files, but we kept the old versions of the files and still use those old files for references for all the ScPCA projects. So because of that, these new fields are just `NA` in the objects. I left a TODO note about that.

allyhawkins:
Thanks for the review, and sorry to cause a refresh! I updated things based on your comments and I think I replied to everything in your inline comments. Let me know if anything is unclear.

references/scripts/create-formatting-reference.py
sjspielman: > I think this is because the original metadata fields that are present in the unfiltered object are added when we use the function from scpcaTool to create the SCE object.

Ah this would make sense because I didn't look in that repo yet when looking for why all the metadata was there!

references/scripts/create-formatting-reference.py
sjspielman: I confess I was hoping to learn that this was some neat Python trick 😄

references/scripts/create-formatting-reference.py
sjspielman: Ahhh this makes sense. TODO makes sense too!

references/scripts/create-formatting-reference.py
sjspielman: I would kind of rather just make it a real data frame? Having to treat this field in the format check script differently makes me feel like too many weird things could potentially happen....

We define it here: https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/format_unfiltered_sce.R#L74-L85
And presumably it's a tibble because of `readr`, but if we pop an `as.data.frame()` into the piping it should let us just use `"data.frame"` as the type here.

And similarly, here we just make it a data frame:
https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/add_celltypes_to_sce.R#L192

What do you think?

allyhawkins: I'm good with that! Done in https://github.com/AlexsLemonade/scpca-nf/pull/1202/commits/d5aafade062a5e6eb236c0fbdfbd6e049e3ef9c2

references/scripts/create-formatting-reference.py
sjspielman: giving this comment a bump!

references/scripts/create-formatting-reference.py
```diff
@@ -212,12 +339,16 @@
...
+    "metadata": filtered_experiment_metadata,
+    "metadata_conditional": filtered_experiment_metadata_conditional,
     "altExp": {
         "adt": {
             "assayNames": adt_assays,
             "colData": filtered_altexp_adt_cell_metadata,
             "rowData": altexp_adt_feature_metadata,
             "colData_conditional": filtered_altexp_adt_cell_metadata_conditional,
+            "metadata": filtered_altexp_adt_experiment_metadata,
+            "metadata_conditional": unfiltered_experiment_metadata_conditional,
```
sjspielman, line 351: Generally, I just really like tracking when we use a dictionary associated with different sce processing level (unfiltered, filtered, processed) since it seems confusing (or like a potential bug when it's not!) on its face.

We have quite a bit of this throughout, and I like that we added this comment in the last PR:

https://github.com/AlexsLemonade/scpca-nf/blob/d5aafade062a5e6eb236c0fbdfbd6e049e3ef9c2/references/scripts/create-formatting-reference.py#L287

I do think it would be a helpful move to add comments when dictionaries are defined that are used beyond their processing level and we'll appreciate this bookkeeping in the long run. For example when `unfiltered_experiment_metadata_conditional` is defined, we could have a comment like "also used to check altexp metadata at all processing levels".

**That said** I would do this for the existing dictionaries as a separate/standalone PR - easier to tackle and review this on its own, but when future ones are defined we can add comments as we go and they start to get used.

----

I also noticed this comment which at first I thought was in a similar vein, but is it outdated?
https://github.com/AlexsLemonade/scpca-nf/blob/d5aafade062a5e6eb236c0fbdfbd6e049e3ef9c2/references/scripts/create-formatting-reference.py#L178

sjspielman (approved):
LGTM! I left 2 comments but don't need to see again:
- bumping a comment that got lost in the data frame shuffle
- I would like to see a separate PR that adds more comments to dictionaries that are used at different levels, e.g. if a defined dictionary `unfiltered_blah_blah` is used for checking filtered/processed, we should comment to that effect to make maintaining this when we all forget it in a year more straightforward!

references/scripts/create-formatting-reference.py
allyhawkins: > I also noticed this comment which at first I thought was in a similar vein, but is it outdated?

This is still accurate since I "reuse" the contents from filtering and just don't include miQC model rather than do the dictionary expansion that I use elsewhere.

Issue #1204, opened 2026-03-04
Accommodate visium HD technologies


sjspielman:
This issue tracks expanding the spaceranger functionality to be able to use visium hd and visium hd 3'. Some of the groundwork will likely be laid for this as part of #1192, but we'll need to make these updates:

- Ensure probesets are going into hd 3' but not hd
- Update the spaceranger publishing process
  - These runs will not have the standard `raw` and `filtered` directories, but instead `binned_outputs` and `segmented_outputs` (rest is the same). As such, we might want a separate publishing process for this, or we'd need to have some logic with these different strings 
- (edit) We'll also need to figure out which files we prefer to read in for barcode counting among the binned outputs (2, 8, 16 um)

sjspielman:
> (edit) We'll also need to figure out which files we prefer to read in for barcode counting among the binned outputs (2, 8, 16 um)

segmented, because those are as close as we get to cells

sjspielman:
> > (edit) We'll also need to figure out which files we prefer to read in for barcode counting among the binned outputs (2, 8, 16 um)
> 
> segmented, because those are as close as we get to cells

Now that I'm in the thick of testing, I think we want to revise this choice! It turns out segmentation has non-zero potential to fail. Currently it's failed for me because I ensmallened fastqs for testing, but reading about this step more in depth suggests failure could not unreasonably occur in other cases e.g. due to image quality.

I would suggest we instead use the 2 um or 8um binning which will always have the output we need. 2 because it's smallest, 8 because I've seen this level commonly used in tutorials (although 16 is also not uncommon...). But as long as we document it I don't think it will be a huge deal which one we pick.
cc @jashapiro @allyhawkins if anyone has strong opinions

(I'll note also while I'm here, I indeed need a second `spaceranger_hd` process for more memory...)

jashapiro:
> (I'll note also while I'm here, I indeed need a second `spaceranger_hd` process for more memory...)

We can maybe make this dynamic within a process something like: 
```
label { meta.technology.endsWith("_hd") ? 'mem_32': 'mem_24' }
```

As to the binning, I don't have a strong opinion, as long as we clearly describe it in the docs/on the portal.

sjspielman:
> label { meta.technology.endsWith("_hd") ? 'mem_32': 'mem_24' }

Sure, we can try this. (Noting, will need to be contains not ends and mem_48/32 - i added the 48 for hd, regular spaceranger is now at 32)

[...]

Issue #1203, opened 2026-03-04
Update visium technology to be visium1


sjspielman:
This is a placeholder issue for renaming our current technology `visium` to something that denotes v1. This will need to happen in nextflow code and example files here, and over in `scpca-docs` as well.

sjspielman:
Noting this additional spot (pun! no apologies) to update in `scpca-admin`: https://github.com/AlexsLemonade/ScPCA-admin/blob/4a0727ec7710b4fd3f7ef992f1ae02dd79a0dd9b/sample_info/ontologies/technology-ontologies.tsv?plain=1#L16

sjspielman:
We don't actually need to do this - we are using `visium1` moving forward, and `visium` has been retained in the code for legacy reasons. The place to update in `scpca-admin` has its own issue tracking it. I also opened a docs placeholder issue which we are very far from getting to, since that depends on actually having more portal spatial data which we don't currently.

PR #1205, opened 2026-03-05
Update spaceranger invocation and accommodate visium2

### LINKED ISSUE #1192: Update spaceranger invocation for updated version
As part of adding support for new Visium types, we need to test out the workflow with the updated spaceranger container. A Visium2 dataset is available in the testing bucket for use with this.

The goal of this issue is update/test the spaceranger module for this new version including updating the spaceranger invocation. 

Resources:
- commands used for testing on EC2: https://github.com/AlexsLemonade/ScPCA-admin/issues/1235#issuecomment-3935185793
- notes on general usage of the new spaceranger version: https://github.com/AlexsLemonade/scpca-nf/issues/1169#issuecomment-3841723018

sjspielman:
Closes #1192 

This PR establishes the initial spaceranger update to be able to accommodate visium v1 (the only technology we previously accommodated) and visium v2 (cytassist technology). Results are in `s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/scpca_out` (see section below on testing notes for more info). 

(Edit: note that #1206 is stacked on this, but in a draft state since it's subject to changes based on review here).

## Implementation notes (this list is longer than the code changes are I think..!)
- In `main.nf`, I defined several new lists:
	- Following how 10x flex probes were managed, I defined a list with species specific cytassist probes which can be indexed out using the reference name (after some regex) 
		- Currently, the probes can't be directly uploaded to the references bucket due to permissions. They are temporarily in the testing bucket (more on this below), and I have a parameter pointing to that bucket for now with a TODO that something will need to be updated once they make it to their final destinations.
		- A future note is we probably want to rename the probe_dir param, since we have multiple kinds of probes now! That variable might have a flex indicator in it to further distinguish from cytassist
	- I expanded the list of spatial technologies, including placeholders for hd technologies (forthcoming in https://github.com/AlexsLemonade/scpca-nf/issues/1204). Throughout, I have TODOs that we should update the visium name to be something like visium1 (see https://github.com/AlexsLemonade/scpca-nf/issues/1203)
	- I added a list of which technologies require a probeset to be specified to handle various logic pieces
- Also in `main.nf`, I added some additional checks to warn if required spatial metadata is missing, to ensure those runs don't make it into the `valid` channel
	- All technologies except visium1 require a cytaimage
	- The visium1 technology requires an image
	- All technologies require slide serial number and slide section
- Finally also in `main.nf`, I added lines to parse the various kind of image files we might accept (see here that we would like to accommodate several image types: https://github.com/AlexsLemonade/ScPCA-admin/issues/1234#issuecomment-3967930264) 
	- Here are the official [spaceranger count docs](https://www.10xgenomics.com/support/software/space-ranger/latest/analysis/running-pipelines/command-line-arguments) If there are other arguments we'd like to accommodate, we can add those in a subsequent PR. Nothing else jumped out at me, but let me know if any look good to you to include.
- The spaceranger module changes:
	- The module now takes some of the additional lists I defined to handle providing probes
	- The spaceranger invocation has been updated, and the process can now accept multiple image files as well as probes
	- Line to remove bams is gone, since we donm't create them! _But_, it does seem like `SPATIAL_RNA_COUNTER_CS` isn't getting properly removed from the checkpoints dir and I'm not sure why yet? Thoughts?
- Spaceranger publishing script:
	- I had to update the R script that creates the output json, because some field names differ based on the technology. As part of this I laid some groundwork for accommodating hd technologies since I was writing some `if`s anyways
	- But, a question I have is: Why don't we just return the metrics summary csv to users directly, or similarly just put all the fields into the json? Less to maintain moving forward with that approach, and also we get to avoid the last bullet here altogether:  https://github.com/AlexsLemonade/scpca-nf/issues/1204

## Testing notes
Files to run this are available here:
```
➤ aws s3 ls s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/                 
                           PRE cytassist-probes/
                           PRE runs/
                           PRE scpca_out/
2026-03-05 09:12:15       1191 run-metadata.tsv
2026-03-05 09:12:08        126 sample-metadata.tsv


➤ aws s3 ls s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/runs/ 
                           PRE visium1-mouse-brain/
                           PRE visium2-human-lung-ffpe/
```
I set this up so that the `visium1` mouse dataset is `SCPCR100000` , and I set this up so that the `visium2` human dataset is `SCPCR200000`, so this is how `scpca_out/{checkpoints & results}` are organized as well. (Fun fact: we set up a pretty solid regex in code to submit jobs to batch, so I needed run ids in scpca format here.)

To re-run, you can use this line. If you are feeling masochistic, you can toss in a `--repeat_mapping` into the custom options string, but note that I was already masochistic enough to do that so I don't recommend it again ;) 
```
gh workflow run run-scpca-nf.yaml \
  -f scpca_run_mode=testing \
  -f scpca_revision=sjspielman/1192-accommodate-visium2 \
  -f scpca_run_ids="SCPCR100000,SCPCR200000" 
  -f scpca_nf_options="--run_metafile s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/run-metadata.tsv --sample_metafile s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/sample-metadata.tsv --outdir s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/scpca_out"
```


sjspielman:
Noting that I requested everyone for review here on the basis of:
- Ally is in the office and makes sense to review
- But, I strongly suspect Josh wants to review nextflow changes!

bin/generate_spaceranger_metadata.R
```diff
@@ -162,7 +180,14 @@ metadata_list <- list(
...
   workflow_commit = opt$workflow_commit
 ) |>
-  purrr::map(~ if (is.null(.)) NA else .) # convert any NULLS to NA
+  # convert any NULLS to NA
+  purrr::map(\(x){
+    ifelse(is.null(x), NA, x)
+  })
+
+# update the tissue column name
+names(metadata_list)[names(metadata_list) == "temp_tissue_colname"] <- tissue_colname
```
jashapiro: Since this is kind of ugly, might we want to just add this element separately?
```suggestion
metadata_list[[tissue_colname]] <- metrics_summary[[tissue_spots_column]]
```

modules/spaceranger.nf
```diff
@@ -6,10 +6,12 @@ process spaceranger {
...
   tag "${meta.run_id}-spatial"
   label 'cpus_12'
-  label 'mem_24'
+  label 'mem_32'
   label 'disk_big'
   input:
-    tuple val(meta), path(fastq_dir), path(image_file), path(index)
+    tuple val(meta), path(fastq_dir), path(index), 
+    path(cytaimage_file), path(cytassist_probe_file), 
+    path(image_file), path(darkimage_file), path(colorizedimage_file)
```
jashapiro: Clarifying these are not separate inputs
```suggestion
      path(cytaimage_file), path(cytassist_probe_file), 
      path(image_file), path(darkimage_file), path(colorizedimage_file)
```

bin/generate_spaceranger_metadata.R
```diff
@@ -143,6 +143,24 @@ metrics_summary <- readr::read_csv(opt$metrics_summary_file)
 # read in versions file
 spaceranger_versions <- jsonlite::read_json(opt$spaceranger_versions_file)
 
+# determine column names for exporting metrics, which depends on technology
+# TODO: visium -> visium1
+if (opt$technology %in% c("visium", "visium_hd_3prime")) {
```
jashapiro: Can you add a check that `opt$technology` is a valid value before getting too far?

config/reference_paths.config
```diff
@@ -14,6 +14,11 @@ params {
   // probe sets for 10x flex
   probes_dir = "${params.ref_rootdir}/flex-probe-refs"
 
+  // probe sets for CytAssist
+  // TODO: update link once these make it to S3
+  // Not unlikely this is no longer a parameter though, since directories may be species-specific
```
jashapiro: I'm not sure what you mean by this?

modules/spaceranger.nf
```diff
@@ -151,14 +161,29 @@ workflow spaceranger_quant{
         log.error("The expected input files or Space Ranger results files for ${it.run_id} are missing.")
       }
 
-    // create tuple of [metadata, fastq dir, and path to image file]
+    // create tuple of [metadata, fastq dir, cytaimage file, index, probe set, paths, to, other, images]
```
jashapiro: be specific here

bin/generate_spaceranger_metadata.R
```diff
@@ -143,6 +143,24 @@ metrics_summary <- readr::read_csv(opt$metrics_summary_file)
...
+  reads_mapped_column <- "Reads Mapped Confidently to Genome"
+} else {
+  # visium2 and visium_hd
+  reads_mapped_column <- "Reads Mapped Confidently to Probe Set"
+}
+
+if (opt$technology %in% c("visium_hd", "visium_hd_3prime")) {
+  tissue_spots_column <- "Number of Squares Under Tissue 2 µm"
+  tissue_colname <- "tissue_squares"
```
allyhawkins, line 164: Before we change this we need to double check with how this is accounted for on the engineering side since this is a column that's included in metadata.

config/reference_paths.config
allyhawkins: What do you mean directories are species-specific?

modules/spaceranger.nf
```diff
@@ -116,8 +124,10 @@ def getCRsamples(files_dir) {
 
 
 workflow spaceranger_quant{
-  take: spatial_channel
-  // a channel with a map of metadata for each spatial library to process
+  take:
+    spatial_channel // a channel with a map of metadata for each spatial library to process
+    cytassist_probesets // map of CytAssist probe set files for each species
+    cytassist_probe_techs // list of which technologies require a probe set
```
jashapiro: I feel like this can be defined within this workflow and does not need to be a separate list?

modules/spaceranger.nf
```diff
@@ -24,18 +26,22 @@ process spaceranger {
...
+      ${cytassist_probe_file ? "--probe-set ${cytassist_probe_file}" : ""} \
+      ${cytaimage_file ? "--cytaimage ${cytaimage_file}" : ""} \
+      ${image_file ? "--image ${image_file}" : ""} \
+      ${darkimage_file ? "--darkimage ${darkimage_file}" : ""} \
+      ${colorizedimage_file ? "--colorizedimage ${colorizedimage_file}" : ""}
 
     # write metadata
     echo '${meta_json}' > ${out_id}/scpca-meta.json
 
+    # TODO: Why is this still present in the checkpoints directory???
```
allyhawkins, line 39: I think we removed it after we actually ran everything so it's in the checkpoints for all the runs that we processed previously.

modules/spaceranger.nf
```diff
@@ -151,14 +161,29 @@ workflow spaceranger_quant{
         log.error("The expected input files or Space Ranger results files for ${it.run_id} are missing.")
       }
 
-    // create tuple of [metadata, fastq dir, and path to image file]
+    // create tuple of [metadata, fastq dir, cytaimage file, index, probe set, paths, to, other, images]
     spaceranger_reads = spatial_channel.make_spatial
       .map{ meta ->
+        // probeset logic
+        def species = meta.ref_assembly.split('\\.')[0]
```
allyhawkins: I think it might make sense to directly track the organism in the `meta` object in the main workflow. As we expand to processing multiple species, this seems like the move rather than pulling it out only in the spaceranger workflow.

modules/spaceranger.nf
```diff
@@ -151,14 +161,29 @@ workflow spaceranger_quant{
...
+        def use_probeset = meta.technology in cytassist_probe_techs
+        def probeset_file = use_probeset ? file("${params.cytassist_probes_dir}/${cytassist_probesets[species]}", checkIfExists: true) : []
```
jashapiro: I don't love the empty array here if it is not always going to be an array? I would either use `null` or use `files()` so you always get an array.

main.nf
```diff
@@ -142,11 +142,18 @@ workflow {
     '10xflex_v1.1_multi': 'Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv'
   ]
 
+  // 10x cytassist probe files
+  def cytassist_probesets = [
+    'Homo_sapiens': 'Visium_Human_Transcriptome_Probe_Set_v2.1.0_GRCh38-2024-A.csv',
```
allyhawkins: Is the probe set the same for both visium2 and visum HD 3'? As visium grows I imagine we will have different probe sets for species and tech version so we might want to think about how best to define that together rather than just using the organism? Should this instead be in the `scpca-refs.json` file?

main.nf
```diff
@@ -216,6 +244,10 @@ workflow {
         files_directory: parseNA(it.files_directory),
         slide_serial_number: parseNA(it.slide_serial_number),
         slide_section: parseNA(it.slide_section),
+        cytaimage_file: parseNA(it.cytaimage_file),
+        image_file: parseNA(it.image_file),
+        darkimage_file: parseNA(it.darkimage_file),
+        colorizedimage_file: parseNA(it.colorizedimage_file),
```
allyhawkins: Just confirming that this means we are extending the library metadata file to have each of these columns? I think it would be helpful to make updates to the example metadata files either ahead of this PR or alongside these changes in this PR so we have a row for each of the types we are adding in. Similar to how I added in rows for the two flex data techs we now support.

main.nf
```diff
@@ -180,12 +187,20 @@ workflow {
...
     .branch{ it ->
       unknown_tech: !(it.technology.toLowerCase() in all_techs)
       unknown_ref: !(it.sample_reference in ref_paths)
+      unknown_spatial_slide_serial: it.technology.toLowerCase() in spatial_techs && !parseNA(it.slide_serial_number)
+      unknown_spatial_slide_section: it.technology.toLowerCase() in spatial_techs && !parseNA(it.slide_section)
+      // TODO: this should say visium1
+      unknown_spatial_image: it.technology.toLowerCase() == "visium" && !parseNA(it.image_file)
+      unknown_spatial_cytaimage: it.technology.toLowerCase() in spatial_techs 
+        && it.technology.toLowerCase() != "visium" 
+        && !parseNA(it.cytaimage_file)
```
jashapiro: I would move most of this to the spatial subworkflow.

config/reference_paths.config
allyhawkins: Oh you mean for probe sets from Human vs. Mouse. I think we need to decide how we are going to store and add in references here as noted in my comment about the references.

main.nf
jashapiro: I was expecting all of these to be parsed out of the files directory structure, not as separate columns. This will make it much easier to accommodate multiple image files, for example.

modules/spaceranger.nf
```diff
@@ -6,10 +6,12 @@ process spaceranger {
   publishDir "${meta.spaceranger_publish_dir}", mode: 'copy'
   tag "${meta.run_id}-spatial"
   label 'cpus_12'
-  label 'mem_24'
+  label 'mem_32'
   label 'disk_big'
   input:
-    tuple val(meta), path(fastq_dir), path(image_file), path(index)
```
jashapiro, line 12: Rather than adding inputs, I might actually reduce this to a generic files_dir and then find the image paths within this rule.

modules/spaceranger.nf
jashapiro: Maybe just reduce this to the parent directory and find the right image files within here?
```suggestion
    tuple val(meta), path(files_dir), path(index), path(cytassist_probe_file)
```

e.g. 
```
--fastqs=${files_dir}/fastq
```

modules/spaceranger.nf
allyhawkins: Hmm I see it's still there in your more recent runs, so this doesn't seem like the reason... I'm not sure what's going on.

allyhawkins (commented):
Overall I think this is mostly the right idea, but I do have some questions that I think will spark some discussion. 

> Why don't we just return the metrics summary csv to users directly, or similarly just put all the fields into the json? Less to maintain moving forward with that approach, and also we get to avoid the last bullet here altogether: https://github.com/AlexsLemonade/scpca-nf/issues/1204

The json file is used by devs to construct the metadata file so any changes there need to be checked with them, but I am okay with making this change for these libraries specifically. This is the only technology type that we provide the json file to users anyways. I would probably stay away from adding another file since that means more updates for the Portal and updates to docs. You could also ask engineering what they would prefer. 

I also think it's worth talking about how we are going to track references here. We have both version and species now, which means I might favor adding the probeset paths to the `scpca-refs.json` file instead of as the dictionary that you added in.

jashapiro (commented):
I had an initial look at this, and my main comment is that that I was hoping not to have the image files be provided as arguments/columns, but rather by reading the directory structure. 

I would also consolidate as much of the spaceranger-specific logic to the spaceranger subworkflow, rather than the main workflow.

modules/spaceranger.nf
sjspielman: Yep, definitely on board with this.

main.nf
sjspielman: > Should this instead be in the scpca-refs.json file?

Yes, I imagined that might be where all of this ends up because it is species-specific even though the references workflow isn't really needed to create the files. The probe file is indeed the same for visium2 and hd-3', so at this time it's really just a per-species files.

config/reference_paths.config
sjspielman: I thought we may want it in scpca-refs.json as part of species-specific information; some discussion to this effect in response to Ally's comments here

modules/spaceranger.nf
sjspielman: The reason I put it at the top of main was because it's consistent with where we define lists in general and helps keep things discoverable, but I don't mind moving it into this file instead. But, I really prefer to keep the list for future maintenance purposes.

modules/spaceranger.nf
sjspielman: I'm confused - we changed the whole workflow to use `[]` for optional files and this seems like effectively the same thing. The file of course exists, we just don't always want to provide it. This syntax was brought to you by copy/paste from half of the other modules...

sjspielman:
> I had an initial look at this, and my main comment is that that I was hoping not to have the image files be provided as arguments/columns, but rather by reading the directory structure.

This implies to me that there would have to be either 1 directory per image type (named accordingly) with the requirement that there is only 1 file per directory, or 1 directory with all images where the files are required to follow a certain naming format so we can figure out which is which. Can you say a bit more about what you are thinking here?

modules/spaceranger.nf
jashapiro: If that is what we always use, keep it. I'm on limited access here!

modules/spaceranger.nf
jashapiro: I meant a separate input, not a separate list; I would define it as a list within this workflow, just not as part of the input.

jashapiro:
> > I had an initial look at this, and my main comment is that that I was hoping not to have the image files be provided as arguments/columns, but rather by reading the directory structure.
> 
> This implies to me that there would have to be either 1 directory per image type (named accordingly) with the requirement that there is only 1 file per directory, or 1 directory with all images where the files are required to follow a certain naming format so we can figure out which is which. Can you say a bit more about what you are thinking here?

I would probably do the former, where we create defined directories for each file type. I really want to minimize the changes to the library file.

modules/spaceranger.nf
jashapiro: My memory was bad. I remember now that nextflow won't allow `null` as an input to a process, which is why we did this.

jashapiro: Another thought: can we just do this on the basis of whether there is a cytassist file? Do we need a list of the cytassist technologies or is the presence of the file a sufficient indicator? (we could check earlier for the presence of a cytassist file based on technologies)

sjspielman: > Another thought: can we just do this on the basis of whether there is a cytassist file?

If you mean the cytassist image file, no. Visium HD requires a cytassist image but it doesn't take the probe file. I'll note that I do have a check that a value for the cytassist image file was provided; this is currently in main which as you commented I'll move to the module file.

sjspielman: > I remember now that nextflow won't allow null as an input to a process, which is why we did this.

Other framing - lucky you then, we would all love to forget this "feature"

main.nf
allyhawkins: > The probe file is indeed the same for visium2 and hd-3', so at this time it's really just a per-species files.

In this case I would definitely just add it to the `scpca-refs.json` file and not add it here. The only reason the flex ones are here is because they are linked to the technology.

modules/spaceranger.nf
sjspielman: Starting to look towards implementing, and I don't think this is what we want since we will need to check the image files before entering the process. This suggests to me that in the workflow we'd want to uncover those file names for checking, which further suggests to go ahead and individually stage at that point.

A thought I have instead for how to handle images is, write a helper function that takes a directory argument and checks if it contains a single file with an allowed extension. If yes, return that file to stage for the process. If not, return `[]`. We could use this helper function 4x, for each of the image files, and stage them individually for the process. 

In terms of checks - what I had previously to check that a cytaimage/image were provided won't work out of the box anymore if they aren't specified in the library file. But, if I run that helper function to define add these items into `meta` early in the spaceranger workflow where we define `spaceranger_publish_dir`  & `spaceranger_results_dir`, then I could confirm those files exist with a few more branches where we split off for  `make_spatial`/`has_spatial`/`missing_inputs`.

Thoughts welcome!

allyhawkins: Can we use something similar to how we deal with the fastq files for regular libraries where we use a regex to find the fastq files within a specific directory? I think if we ensure that there's a specific organization followed for spaceranger files then we can use that to pull out any image files that may exist and check if they exist before entering the process. 

So the expected organization within the files directory would be: 
```
SCPCRXXXXXX
├── R1.fastq.gz
├── R2.fastq.gz
└── cytaimage
    └── image.jpg

```

So you would look for presence of the `image`, `cytaimage`, `darkimage` or `colorizedimage` directory and find the image file within that.

jashapiro: I think that seems reasonable enough. I don't think you need to have the files in `meta` though; I would just expect them to be passed along in the channel. I also don't expect to need branching other than to report the missing files and stop processing those samples, much like what I think you had.

sjspielman: > Can we use something similar to how we deal with the fastq files for regular libraries where we use a regex to find the fastq files within a specific directory? 

Yes, this is essentially what I was getting at, but I don't think regex specifically.  For example, we previously used a glob and assumed a jpg https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/modules/spaceranger.nf#L160 

But now, we have to be more permissive because it could be one of several extensions, so we'd want to check:
- is there one file?
- does it have an acceptable extension?
  -  For this point, I looked a bit more closely at [spaceranger image requirements](https://www.10xgenomics.com/support/software/space-ranger/latest/analysis/inputs/image-image-recommendation) as well: 
    -  First, glad I did this since there's more going on with `darkimage` that I had not realized, namely that it can be up to 6 files...! These can be specified as `--darkimage file1,file2,file3...`
    - Second, valid file types are apparently jpg, tiff, bigtiff, and qptiff. This corresponds to a lot of possible extensions, and I'm not convinced it's worth checking?

**BUT!! most importantly**, I am no longer convinced we want/need to support all these images? I don't think that darkimage or colorizedimage impact mapping/quant here and but are just additional items that can get added to the loupe file. Part of the issue here is I haven't been able to (in spite of much trying...) locate a 10x dataset which actually provides multiple tissue images so I wasn't able to actually test combos of these arguments and their affect. While trying to find more docs here, I came across [cumulus docs](https://cumulus.readthedocs.io/en/latest/spaceranger.html) (note that the [cumulus code](https://github.com/lilab-bcb/cumulus/tree/master/workflows/spaceranger) seems to be a very straightforward wrapper around spaceranger (v4) I don't think there's other considerations besides spaceranger requirements). These docs suggest that image, darkimage, and colorizedimage are _mutually exclusive_. Again, I would love to test this, but I don't immediately have files to use for that! **tl;dr** - what if we just take cytaimage and image for now and pause on including the other two image types until we _know_ how to support them (if at all)?

(edit): also see the diagrams here which drive the point home more: https://www.10xgenomics.com/support/software/space-ranger/latest/analysis/running-pipelines/probe-based-assay-count-cytassist-gex

jashapiro: I'm fine with only supporting brightfield images, but I don't think it would be hard to add the logic for full support. I'd probably do it by adding two fields to the current process input: `cytaimage` and `image_type`. The current `image` input would contain a cell ranger supported image file or files, and the `image_type` would describe what kind of image is there. 

It we set that up, then we can start with just brightfield support and add support for more image types later without modifying the input tuple.

sjspielman: > I'm fine with only supporting brightfield images, but I don't think it would be hard to add the logic for full support. 

I agree that implementing it isn't hard, but what is hard at this time is that I don't have example/test data to be confident it's implemented properly! On one hand I think I have a handle now, but on the other hand I'd really like to not "think" but "know" 😬 

> The current image input would contain a cell ranger supported image file or files, and the image_type would describe what kind of image is there.

Sounds like `image_type` would have to be something provided in the library file, no? Let me know if I'm reading you right here, I'm envisioning:

```
runs
├── cytaimage
│   └── agnostically-named-file.tif
├── fastq
│   ├── blah_R1.fastq.gz
│   ├── blah_R2.fastq.gz
│   ├── blahblah_R1.fastq.gz
│   └── blahblah_R2.fastq.gz
└── image    <------- has associated `image_type` field with values `image` (to match the spaceranger arg, or `brightfield`), `darkimage`, or `colorizedimage` except for now it's `image` all the way down until we can 100% confirm how darkimage & colorizedimage interact? 
    └── agnostically-named-file.jpg
```

jashapiro: > Sounds like image_type would have to be something provided in the library file, no?

I was thinking the file structure would be the same, and we would just pull the correct files and fill in the image type as part of our code. But I could be convinced to put it in the library file

allyhawkins: > I was thinking the file structure would be the same, and we would just pull the correct files and fill in the image type as part of our code. But I could be convinced to put it in the library file

I was just typing out that I would favor this approach where we keep the file structure where the name of the folder corresponds to the image type. I think that might make it easier to expand on image types and have multiple image types per library. If we add a column to the library metadata we would need to account for having multiple image types.

bin/generate_spaceranger_metadata.R
sjspielman: I thought about this, but decided against it because I had a stronger preference to keep all the spaceranger items together in the json. While it's ugly, it's legible. I can be overruled here though.

modules/spaceranger.nf
```diff
@@ -116,21 +129,27 @@ def getCRsamples(files_dir) {
...
         meta.spaceranger_results_dir = "${meta.spaceranger_publish_dir}/${meta.run_id}-spatial"
+
         meta // return modified meta object
       }
       .branch{ it ->
         def stored_ref_assembly = getMetaVal(file("${it.spaceranger_results_dir}/scpca-meta.json"), "ref_assembly")
         def stored_tech = getMetaVal(file("${it.spaceranger_results_dir}/scpca-meta.json"), "technology") ?: ""
+        // branch for invalid cases
+        missing_slide_serial: !parseNA(it.slide_serial_number)
+        missing_slide_section: !parseNA(it.slide_section)
```
jashapiro: These were parsed on read, so they should not need to be parsed again.
```suggestion
        missing_slide_serial: !it.slide_serial_number
        missing_slide_section: !it.slide_section
```

lib/utils.nf
```diff
@@ -118,3 +118,24 @@ def pullthroughContainer(image_url, pullthrough_url = ""){
...
+def getImageFiles(dir, max_files = 1) {
+  def f = files("${dir}/*")
+  // empty or too many
+  if (f.size() == 0 || f.size() > max_files) {
+    return ([])
+  }
+  // return the 1+ files if they exist
+  if (max_files == 1) {
+    return (f[0]) // we want to return the single file here, so index it out
+  }
```
jashapiro: Make your life easier later and always return an array (as implied by the function name)
```suggestion
```

modules/spaceranger.nf
sjspielman: resolving this comment for PR real estate purposes since it's covered here https://github.com/AlexsLemonade/scpca-nf/pull/1205#discussion_r2891746178

lib/utils.nf
```diff
@@ -118,3 +118,24 @@ def pullthroughContainer(image_url, pullthrough_url = ""){
...
+
+/**
+  * For a given directory, return its files as long as there are <= max_files present
+  * This is used to get images files for space ranger
+  *
+  * @param dir Directory to search for a single file in
+  * @return File path(s) or [] if no or too many files 
+  */
+def getImageFiles(dir, max_files = 1) {
+  def f = files("${dir}/*")
```
jashapiro, line 137: I'd still filter on extensions (Add others as needed)
```suggestion
  def f = files("${dir}/*")
    .findAll{it =~ /(?i)\.(jpg|jpeg|tif|tiff)$/}
```

lib/utils.nf
```diff
@@ -118,3 +118,24 @@ def pullthroughContainer(image_url, pullthrough_url = ""){
...
+
+
+/**
+  * For a given directory, return its files as long as there are <= max_files present
+  * This is used to get images files for space ranger
+  *
+  * @param dir Directory to search for a single file in
+  * @return File path(s) or [] if no or too many files 
+  */
+def getImageFiles(dir, max_files = 1) {
```
jashapiro: I think I would keep the max files out of this function. I find it confusing that if there are too many files you return nothing. Return all the files and check after if needed.

sjspielman:
Update: many commits because testing/debugging requires pushing! But I hit a wall. 

I was testing out approaches for the image files and a run was hanging (after erroring? something was clearing wrong) so I killed it: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/57DZHHHkZCVUkM/tasks

But, something seems to have gone wrong with killing that run because it thinks a run I can't submit fresh: https://github.com/AlexsLemonade/ScPCA-admin/actions/runs/22868395134/job/66341315026
Note that `tmux ls` on the runner comes up empty as well.

So, I guess I'm leaving this as a WIP but folks can review at this stage! For handling image files, I have two competing versions of code at the moment because WIP...
- Currently commented out are pieces of code to stage image files individually. There's also a helper function I added for this, currently still in the utils file but its use is commented out.
- Currently not commented out but probably still buggy is code to pass in the files directory and pull out files within the process. I think I now care less to check that there is _only one file_, so I was trying to pivot in that direction when the runs got borked.

Not so WIP though, I updated the reference json to include the probe set files rather than defining them in the workflow.

bin/generate_spaceranger_metadata.R
```diff
@@ -116,6 +116,15 @@ if (is.null(opt$spaceranger_versions_file) || !file.exists(opt$spaceranger_versi
   stop("Versions file missing or `spaceranger_versions_file` not specified.")
 }
 
+# check that technology exists
+# TODO: visium -> visium1
+allowed_techs <- c("visium", "visium2", "visium_hd", "visium_hd_3prime")
+if (!opt$technology %in% allowed_techs) {
+  stop(
+    glue::glue("The `technology` must be one of: {paste(allowed_techs, collapse = ', ').}")
```
jashapiro:
```suggestion
    glue::glue("The `technology` must be one of: {paste(allowed_techs, collapse = ', ')}.")
```

bin/generate_spaceranger_metadata.R
jashapiro: Technically, JS objects are unordered, so I would not want to rely on ordering set here (the list is ordered, so the initial output will be, but after that who knows). But it is also fine to leave as is.

modules/spaceranger.nf
```diff
@@ -1,41 +1,69 @@
...
     out_id = file(meta.spaceranger_results_dir).name
     meta += getVersions()
     meta_json = makeJson(meta)
+
+    // TODO: We are currently assuming that only one of the non-cytassist images should exist, 
+    //  and therefore we only want to provide a single argument
+   // image_arg = image_type == "image"           ? "--image ${image_file}" :
+   //             image_type == "colorizedimage"  ? "--colorizedimage ${image_file}" :
+    //            image_type == "darkimage"       ? "--darkimage ${[image_file].flatten().join(',')}" :
+    //            ""
```
jashapiro: I think my preference is to use something between the two version you have here. 

Use `path(fastq_dir), path(cytaimage_file), path(image_file), val(image_type)` as the inputs, then within the code you can do something a bit simpler than what you have here:


```suggestion
  image_arg = image_type ? "--{image_type} ${image_file.join(',')}" : ""
```

Note that this assumes that `path(image_file)` is an array, but that can be done by using the `files()` function to create it and not converting back to a single value.

modules/spaceranger.nf
```diff
@@ -1,41 +1,69 @@
...
     """
+    ###### TODO: In this approach I'm not counting how many files there are but assuming any presence is ok
+    # First, set up image arguments
+    # Cytaimage
+    cytaimage_arg=""
+    if [ -n "\$(ls ${files_directory}/cytaimage/ 2>/dev/null)" ]; then
+      cytaimage_arg="--cytaimage \$(ls ${files_directory}/cytaimage/*)"
```
jashapiro: While I am recommending against the bash approach in general, I want to note that you should approximately never use `ls` in a bash script. Its output is too variable and can depend on the calling environment.

modules/spaceranger.nf
```diff
@@ -145,20 +181,47 @@ workflow spaceranger_quant{
         missing_inputs: true
       }
 
-    // send run ids in spatial_channel.missing_inputs to log
-    spatial_channel.missing_inputs
-      .subscribe{ it ->
-        log.error("The expected input files or Space Ranger results files for ${it.run_id} are missing.")
-      }
+    // TODO: Should these be warn or error?
```
jashapiro: As long as these are not actually killing the workflow, I think either is fine. But I might lean toward error since the result is not processing a sample and we want to call that out somewhat louder (we may or may not do that universally, but we should aim consistency)

jashapiro (commented):
I think this should be pretty close! 

I lean toward doing more of the file processing in Nextflow if possible, rather than the bash approach. One reason for this is that you should be able to check most of the logic with stub runs, which will be much faster than submission to batch. The bash script will _only_ be checked on a real run, which is not so great.

I would expect the `getImageFiles()` function you have to mostly work, but I would modify/simplify it to always return an array, and remove the check for the number of elements. You can always check that separately. If you know you are getting an array, you can simplify some of your later logic as well.

modules/spaceranger.nf
sjspielman: I'm going with error (it doesn't kill the workflow), but opened a followup issue to inventory this more generally for overall consistency.

bin/generate_spaceranger_metadata.R
```diff
@@ -116,6 +116,15 @@ if (is.null(opt$spaceranger_versions_file) || !file.exists(opt$spaceranger_versi
   stop("Versions file missing or `spaceranger_versions_file` not specified.")
 }
 
+# check that technology exists
+# TODO: visium -> visium1
+allowed_techs <- c("visium", "visium2", "visium_hd", "visium_hd_3prime")
```
allyhawkins: Why not just include `visium1` now? 

```suggestion
allowed_techs <- c("visium", "visium1", "visium2", "visium_hd", "visium_hd_3prime")
```

modules/spaceranger.nf
```diff
@@ -1,21 +1,23 @@
...
+  input: 
+    tuple val(meta), path(index), path(probeset_file),
+      path(fastq_dir), path(cytaimage_file), path(image_file), val(image_type) 
   output:
     tuple val(meta), path(out_id)
   script:
     out_id = file(meta.spaceranger_results_dir).name
     meta += getVersions()
     meta_json = makeJson(meta)
+    image_arg = image_type ? "--{image_type} ${image_file.join(',')}" : "" // join in case of multiple darkimages
```
allyhawkins: Can you have multiple image types in addition to cytaimage or is it always just one image type here?

lib/utils.nf
```diff
@@ -118,3 +118,24 @@ def pullthroughContainer(image_url, pullthrough_url = ""){
...
+  * For a given directory, return its space ranger image files 
+  *
+  * @param dir Directory to search image files(s) in
+  * @param cytaimage Boolean if the image file(s) to grab are the cytaimage
+  * @return Array of file paths
+  */
+def getImageFiles(dir, cytaimage = false) {
+  
+  // https://www.10xgenomics.com/support/software/space-ranger/latest/analysis/inputs/image-image-recommendation
+  def image_extensions = cytaimage
```
jashapiro, line 133: Named argument are strange in groovy. If you want to call this with `getImageFiles(my_directory, cytaimage: true), you would need the following:
```suggestion
def getImageFiles(Map args, dir) {
  
  // https://www.10xgenomics.com/support/software/space-ranger/latest/analysis/inputs/image-image-recommendation
  def image_extensions = args.cytaimage
```

main.nf
```diff
@@ -146,7 +146,7 @@ workflow {
   def single_cell_techs = cell_barcodes.keySet()
   def flex_techs = flex_probesets.keySet()
   def bulk_techs = ['single_end', 'paired_end']
-  def spatial_techs = ['visium']
+  def spatial_techs = ['visium', 'visium2', 'visium_hd', 'visium_hd_3prime'] // TODO: Make the original visium "visium1" or similar
```
allyhawkins: I might say the same thing here and then we can remove `visium` later or keep it depending on when/how we update tech for the existing data. 
```suggestion
  def spatial_techs = ['visium', 'visium1', 'visium2', 'visium_hd', 'visium_hd_3prime'] // TODO: Make the original visium "visium1" or similar
```

main.nf
```diff
@@ -180,12 +180,13 @@ workflow {
       || (it.submitter in project_ids)
       || (it.scpca_project_id in project_ids)
     }
-    // make sure technology and references are known
+    // make sure technology and references are known, and all spatial information is provided
```
allyhawkins:
```suggestion
    // make sure technology and references are known
```
I assume this is a leftover comment?

main.nf
```diff
@@ -209,6 +210,7 @@ workflow {
         technology: it.technology.toLowerCase(),
         assay_ontology_term_id: parseNA(it.assay_ontology_term_id),
         seq_unit: it.seq_unit,
+        organism: it.sample_reference.split('\\.')[0],
```
allyhawkins:
```suggestion
        organism: it.organism
```
This is a direct field in our metadata, so let's just use it.

allyhawkins (commented):
These new updates look pretty good to me. I left some very minor comments.

modules/spaceranger.nf
sjspielman: Cytaimage is always one. I have a check for this file so far, but only this file. It's in the workflow code prepping for this process. Note that I am currently testing though!

main.nf
sjspielman: Ah, fwiw, I am no longer using this. I can either remove it entirely, or add it in and it will just sit in meta but not be used.

modules/spaceranger.nf
allyhawkins: No, I mean can you have multiple image types in general? For example, can you have both `image` and ` brightimage`? I think the answer to this is yes and in that case this would break because you would have two image types.

bin/generate_spaceranger_metadata.R
sjspielman: After much back and forth with myself, yeah ok I agree now let's add it afterwards.

modules/spaceranger.nf
sjspielman: Noting that we tested and the answer is no - you get cytaimage and _one other_ image.

sjspielman:
Here's where we are:
- This version of code is currently running https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/AgteEnHylCKgu/tasks. Sadly though the first try of the processes ran out of memory with `143` so they're gonna be running for a little while longer. I wish I understood why these memory issues seem so stochastic...
-  Currently, I am only checking that there is one cytaimage file but I have opened a separate issue to check the other image files since we want to move this particular PR along https://github.com/AlexsLemonade/scpca-nf/issues/1212
- I have added visium1 as a technology (but didn't remove visium because legacy) 
- The probe files have been integrated into our refs setup, with TODOs to update the paths once the probe files land in the right bucket

Once the runs complete and I can confirm they look correct, I will re-request for final.final review

bin/generate_spaceranger_metadata.R
sjspielman: We have David's blessing

sjspielman:
We made it!

- Full run through completed successfully: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/AgteEnHylCKgu
- But after I submitted that I made changes to the generate metadata script so I wanted to run that through again, _without repeating mapping_: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/1ayeyme2czYg4H/tasks

Results are in `s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/scpca_out/results/`.

lib/utils.nf
sjspielman: Frankly, I didn't do this because I didn't want to repeat mapping again........

jashapiro: It is fine as is! Just thought you might appreciate the strangeness.

modules/spaceranger.nf
jashapiro: Just noting we should try to debug this before release.

jashapiro (approved):
LGTM, I just want to make sure we get an issue filed for the directory not being deleted. 

In prep for that, you might want to generate a cut down set of input fastqs for faster testing?

bin/generate_spaceranger_metadata.R
```diff
@@ -152,8 +177,7 @@ metadata_list <- list(
   filtered_spots = nrow(filtered_barcodes),
   unfiltered_spots = nrow(unfiltered_barcodes),
```
allyhawkins, line 178: This doesn't need to happen in this PR, but should we account for the "squares" change in these columns too?

main.nf
```diff
@@ -209,6 +210,7 @@ workflow {
         technology: it.technology.toLowerCase(),
         assay_ontology_term_id: parseNA(it.assay_ontology_term_id),
         seq_unit: it.seq_unit,
+        organism: it.organism,
```
allyhawkins: If you aren't using this maybe we just remove it until we need it? Because when we process scpca data we don't actually use the meta here but instead use the meta stored in `scpca-meta.json`, I'm worried we might go to use this later and be confused why we see failures about organism not existing.

allyhawkins (approved):
LGTM, but I might remove the organism addition.

main.nf
sjspielman: No disagreement here, will delete it!

Issue #1197, opened 2026-02-27
Test run of scpca-nf with public dataset on Cavatica


allyhawkins:
After we update scpca-nf on Cavatica in #1196 we want to do a full test run starting from FASTQ files. We will do that with the data from [PRJNA952244](https://www.ncbi.nlm.nih.gov/bioproject/PRJNA952244).

allyhawkins:
I was able to get a test run started, but not fully processed yet and wanted to post some initial thoughts and comments. 

I started by grabbing the FASTQ files for SCPCR001279 (one of the libraries that we have added from PRJNA952244). I used the ftp/http import option and copied the ftp links from the `run-metadata.tsv` for that library. This was a pretty quick and easy process and even though I did it through GUI, I was looking at the docs and we should be able to do something similar through the command line and handle that programmatically. 

I then created metadata files and saved them to `public_data_example_metadata` in the project files for now. This just includes the metadata for that single run ID and contains the appropriate paths on Cavatica to where input files live. 

Then I ran the `scpca-nf-development` app using the specified run and sample metadata files for the public data. At first this failed because the files don't have `_R1` or `_R2` (see https://github.com/AlexsLemonade/ScPCA-admin/issues/1232#issuecomment-3993435107). I updated the file names to have `_R1` and `_R2` and then re-ran. 

After running for 2 hours, I did hit this error message with the `fry_quant_rna` step that I'm going to look into. It looks like the mapping completed successfully but then hit a failure on the quantification step. It appears that even though the mapping step completed, there is no output files being published so no checkpoints folder with the resulting output. The only place I can see the output is by going into the actual task and searching through the logs.  

```
2026-03-03 23:34:07 INFO number of unfiltered bcs read = 6,794,880
2026-03-03 23:34:07 INFO paired : false, ref_count : 424,237, num_chunks : 88
2026-03-03 23:34:07 INFO read 2 file-level tags
2026-03-03 23:34:07 INFO read 2 read-level tags
2026-03-03 23:34:07 INFO read 1 alignemnt-level tags
2026-03-03 23:34:07 INFO File-level tag values FileTags { bclen: 16, umilen: 12 }
2026-03-03 23:34:07 INFO observed 412,364 reads (250,363 orientation consistent) in 88 chunks --- max ambiguity read occurs in 851 refs
2026-03-03 23:34:07 INFO minimum num reads for barcode pass = 10
2026-03-03 23:34:07 INFO num_passing = 3,533
[...]

PR #1208, opened 2026-03-05
Fix mismatching happening for index building

allyhawkins:
While investigating https://github.com/AlexsLemonade/scpca-nf/issues/1197#issuecomment-3998720962, I found that rebuilding the index didn't fix the problem I was having and I was getting the same error with the index files not matching correctly. In looking into this, I realized we weren't matching the fasta output with the correct information in the salmon channel, so sometimes when we ran it, Nextflow would mix the human fasta with the mouse salmon index information. You can see an example of this here: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/3Bo7wPnJiOwiSx/tasks

This was the command being executed for the salmon_index process: 
```
salmon index       -t Homo_sapiens.GRCh38.104.spliced_intron.txome.fa.gz       -i Mus_musculus.GRCm39.104.spliced_intron.txome       -k 31       -p 8
gunzip -c Mus_musculus.GRCm39.dna.primary_assembly.fa.gz       | grep "^>" | cut -d " " -f 1       | sed -e 's/>//g' > decoys.txt
cat Homo_sapiens.GRCh38.104.spliced_cdna.txome.fa.gz Mus_musculus.GRCm39.dna.primary_assembly.fa.gz > gentrome.fa.gz

salmon index       -t gentrome.fa.gz       -d decoys.txt       -i Mus_musculus.GRCm39.104.spliced_cdna.txome       -k 31       -p 8
```

I fixed that by joining the channels using the `ref_name` and things now appear to be building correctly (see https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/447IzDaeaw9z1e/tasks). While here I also increased the memory since I was noticing memory failures in the previous runs and added some tags to easily see which index is being built.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -37,7 +37,7 @@ process salmon_index {
   container "${pullthroughContainer(params.salmon_container, params.pullthrough_registry)}"
   publishDir "${params.ref_outdir}/${meta.ref_dir}/salmon_index", mode: 'copy'
   label 'cpus_8'
-  label 'mem_16'
+  label 'mem_24'
   tag "${ref_name}"
   input:
     tuple val(ref_name), path(splici_fasta), path(spliced_cdna_fasta), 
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
Thank you for catching this! That explains how things got messed up the last time, and makes me feel much better on many levels. 

Enthusiastic approval!

PR #1209, opened 2026-03-09
Build AnnData formatting reference

### LINKED ISSUE #1194: AnnData formatting reference file
We will want to create a separate formatting reference file for AnnData objects following the same design choices we made in https://github.com/AlexsLemonade/scpca-nf/pull/1187. This should be created within the same script but output a separate file.

allyhawkins:
Closes #1194 
Stacked on #1202 

This PR adds in the formatting reference file for AnnData objects. I wanted to utilize as much information from the SCE reference dictionaries as possible, so I ended up writing two functions that map the object type in R to the object type in Python depending on if it's coming from the cell/row metadata which is stored in a pandas DataFrame or the `uns` slot which is a dictionary. 

As reference, this is the script we use to reformat and convert SCE objects -https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/sce_to_anndata.R

For the most part I used these two functions to do the conversions for all the R/SCE dictionaries, with some exceptions: 

- There were a few columns that are in the `obs` that are not in the `colData` for SCE objects. Specifically, we add a `library_id` column, the columns from the sample metadata and a few other metadata columns based on what's required by CZI. That being said, the only column that we require to be present in the `sample_metadata` slot is the `sample_id`, so that's the only one I'm explicitly checking right now. 
- If a library is multiplexed, we don't include a `sample_id` in the main `obs`. I added a conditional statement to account for that. 
- We add two new booleans to the `var` metadata and two strings to `uns`. 
- The `uns` contains the `pca` dictionary only if the object is processed. 
- These objects just have a `layers` slot which stores just the `spliced` object. The `X` slot contains the `counts` object, but that name is tracked within the `uns`. For processed objects only, we have a `.raw.X` and an `X`, so I added a check for `has_raw.X` boolean to track that. 

When building the schema itself, I chose to include `rna` and `adt` dictionaries for each object and removed the `has_adt` and `has_cellhash`. We output separate `adt` objects with the contents of the AltExp from the adt SCE and we dont' do the equivalent for cell hash. 

As an added reference, I printed out the types for a processed object with ADT so you can see what they are and why I added some of the exceptions that I did. 

Example of types for `obs` from the processed object with ADT: 
```
rna_adata.obs.dtypes
barcodes                                      object
sum                                          float64
detected                                       int32
subsets_mito_sum                             float64
subsets_mito_detected                          int32
subsets_mito_percent                         float64
altexps_adt_sum                              float64
altexps_adt_detected                           int32
altexps_adt_percent                          float64
total                                        float64
prob_compromised                             float64
miQC_pass                                       bool
scDblFinder_class                           category
scDblFinder_score                            float64
scpca_filter                                category
adt_scpca_filter                            category
sizeFactor                                   float64
cluster                                     category
singler_celltype_ontology                   category
singler_celltype_annotation                 category
cellassign_celltype_annotation              category
cellassign_celltype_ontology                category
cellassign_max_prediction                    float64
scimilarity_celltype_annotation             category
scimilarity_celltype_ontology               category
scimilarity_min_distance                     float64
consensus_celltype_annotation               category
consensus_celltype_ontology                 category
is_infercnv_reference                           bool
library_id                                  category
sample_id                                   category
scpca_project_id                            category
submitter_id                                category
participant_id                              category
submitter                                   category
age                                         category
age_timing                                  category
sex                                             bool
diagnosis                                   category
subdiagnosis                                    bool
tissue_location                             category
disease_timing                              category
organism                                    category
is_xenograft                                    bool
is_cell_line                                    bool
development_stage_ontology_term_id          category
sex_ontology_term_id                        category
organism_ontology_id                        category
self_reported_ethnicity_ontology_term_id    category
disease_ontology_term_id                    category
tissue_ontology_term_id                     category
assay_ontology_term_id                      category
suspension_type                             category
is_primary_data                                 bool
```

Example of the `uns` contents and their associated type: 
```
for key,value in rna_adata.uns.items():
...   print(key, type(value))
X_name <class 'str'>
adt_normalization <class 'str'>
adt_scpca_filter_method <class 'str'>
af_num_cells <class 'numpy.int64'>
af_permit_type <class 'str'>
af_resolution <class 'str'>
alevinfry_version <class 'str'>
assay_ontology_term_id <class 'str'>
cellassign_predictions <class 'pandas.core.frame.DataFrame'>
cellassign_reference <class 'str'>
cellassign_reference_organs <class 'str'>
cellassign_reference_source <class 'str'>
cellassign_reference_version <class 'str'>
celltype_methods <class 'numpy.ndarray'>
cluster_algorithm <class 'str'>
cluster_nn <class 'numpy.int64'>
cluster_weighting <class 'str'>
consensus_celltype_methods <class 'numpy.ndarray'>
filtering_method <class 'str'>
highly_variable_genes <class 'numpy.ndarray'>
include_unspliced <class 'numpy.bool_'>
infercnv_diagnosis_groups <class 'str'>
infercnv_num_reference_cells <class 'numpy.int64'>
infercnv_reference_celltypes <class 'numpy.ndarray'>
infercnv_status <class 'str'>
library_id <class 'str'>
mapped_reads <class 'numpy.int64'>
mapping_tool <class 'str'>
min_gene_cutoff <class 'numpy.int64'>
normalization <class 'str'>
pca <class 'dict'>
prob_compromised_cutoff <class 'numpy.float64'>
project_id <class 'str'>
reference_index <class 'str'>
salmon_version <class 'str'>
sample_id <class 'str'>
sample_type <class 'str'>
schema_version <class 'str'>
scimilarity_model <class 'str'>
scpca_filter_method <class 'str'>
seq_unit <class 'str'>
singler_date <class 'str'>
singler_gene_set_version <class 'str'>
singler_reference <class 'str'>
singler_reference_label <class 'str'>
singler_reference_source <class 'str'>
singler_reference_version <class 'str'>
singler_results <class 'pandas.core.frame.DataFrame'>
tech_version <class 'str'>
total_reads <class 'numpy.int64'>
transcript_type <class 'numpy.ndarray'>
usa_mode <class 'numpy.bool_'>
```


references/scripts/create-formatting-reference.py
```diff
@@ -385,7 +385,7 @@
     },
 }
 
-# build and expoort sce schema --------------
+# build and export sce schema --------------
```
sjspielman, line 389: haha I almost want to tell you not to fix this, it's very charming

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,245 @@
 with open(sce_ref_file, "w") as f:
     json.dump(sce_schema, f, indent=4)
     f.write("\n")  # Explicitly add a newline character to appease GitHub
+
+# --------------------- AnnData ---------------------------
+
+# cell and row metadata are dtypes
+cell_row_metadata_map = {
```
sjspielman: I would do one of two things differently with these map dictionaries since they are being used in functions but not being explicitly passed in to them. Either:
- make the variables name all caps for a better indicator that they are being used as constants. I tend to prefer this one, given that contents of these dictionaries make sense to be constants!
- or actually pass it in as function arguments

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,245 @@
...
+experiment_metadata_map = {
+    "character": "str",
+    "integer": "numpy.int64",
+    "numeric": "numpy.float64",
+    "logical": "numpy.bool_",
+    "data.frame": "pandas.DataFrame",
+    "DFrame": "pandas.DataFrame",
+}
+
+# these all get stored as numpy.ndarray and not str even though they are considered character in R
```
sjspielman: super fun!

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,245 @@
...
+# layer/assays ---------
+layers = ["spliced"]
+
+# unfiltered cell metadata ------------
+anndata_specific_obs_metadata = {
+    # sample metadata is present in obs for anndata
+    # this minimally includes library id, sample id and assay ontology term id
+    "library_id": "category",
+    "sample_id": "category",
+    "assay_ontology_term_id": "category",
```
sjspielman: we also explicitly add this one in sce_to_annndata https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/sce_to_anndata.R#L114

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,245 @@
...
+
+filtered_uns_metadata = convert_experiment_metadata_types(filtered_experiment_metadata)
+
+filtered_uns_metadata_conditional = convert_experiment_metadata_types(
+    filtered_experiment_metadata_conditional
+)
+
+processed_uns_metadata = {
+    **convert_experiment_metadata_types(processed_experiment_metadata),
+    "pca": "dict",
```
sjspielman: It seems to me we might also want to check these fields? Perhaps you already thought about this and decided against it? This feels different to me somewhat than e.g. checking the df columns for `metadata(sce)$sample_metadata`; since it's for CZI compliance, these fields do need to be there

https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/reformat_anndata.py#L106

sjspielman (commented):
> That being said, the only column that we require to be present in the `sample_metadata` slot is the `sample_id`, so that's the only one I'm explicitly checking right now.

I want to talk through this comment a bit.
- I assumed by sample_metadata slot mean metadata(sce)$sample_metadata - slot means too many things and it's not our fault
 - Anwyays, I am assuming it's based on this check: https://github.com/AlexsLemonade/scpcaTools/blob/5f8dcddc6ae852868f097f22f0ff51ff530bf8a2/R/add_sample_metadata.R#L23
 - Further, it makes sense why we would not also consider values like `is_xenograft` or `is_cell_line` since those are added _after_ we add the sample_metadata : https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/format_unfiltered_sce.R#L103
 - Similarly, it seems to make sense not to include `diagnosis` since we handle cases where it's missing: https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/bin/add_celltypes_to_sce.R#L359-L360
 - _But_, these three columns really don't feel to me like they would ever be missing for data we process for the portal, and we might like to know if they are? 


I asked Claude for some help to find any discrepancies between the actual json we create and the example of types you pasted in and it made me a nice table... Some of this is intentional since it's the sample_metadata columns, but I wanted to put everything it flagged here to be sure!
But, it does seem we have cases like `detected` and `cluster_nn` for example that R has as numeric, but since they don't have decimals get converted to ints not floats. We'll need to handle these exceptions somehow.

Slot | Variable | Type in JSON | Type in Printed Output
-- | -- | -- | --
obs | detected | float64 | int32
obs | infercnv_total_cnv | int32 | (not present in output)
obs | scpca_project_id | (not in JSON) | category
obs | submitter_id | (not in JSON) | category
obs | participant_id | (not in JSON) | category
obs | submitter | (not in JSON) | category
obs | age | (not in JSON) | category
obs | age_timing | (not in JSON) | category
obs | sex | (not in JSON) | bool
obs | diagnosis | (not in JSON) | category
obs | subdiagnosis | (not in JSON) | bool
obs | tissue_location | (not in JSON) | category
obs | disease_timing | (not in JSON) | category
obs | organism | (not in JSON) | category
obs | is_xenograft | (not in JSON) | bool
obs | is_cell_line | (not in JSON) | bool
obs | development_stage_ontology_term_id | (not in JSON) | category
obs | sex_ontology_term_id | (not in JSON) | category
obs | organism_ontology_id | (not in JSON) | category
obs | self_reported_ethnicity_ontology_term_id | (not in JSON) | category
obs | disease_ontology_term_id | (not in JSON) | category
obs | tissue_ontology_term_id | (not in JSON) | category
uns | cluster_nn | numpy.float64 | numpy.int64
uns | prob_compromised_cutoff | NA | numpy.float64
uns | transcript_type | str | numpy.ndarray
uns | schema_version | (not in processed JSON) | str
uns | singler_date | (not in JSON) | str
uns | singler_gene_set_version | (not in JSON) | str

edit - noting the `prob_compromised_cutoff` is a false flag, if you will, as well!

allyhawkins:
> But, these three columns really don't feel to me like they would ever be missing for data we process for the portal, and we might like to know if they are?

So the plan is to add this into `scpca-nf` as a process within the workflow, not something that will be run separately from the workflow. Because of that we should accommodate any data that could be run through this and not just Portal data. So in general, I'm keeping the formatting checks here general enough to apply to all data and not just data we process. 

I also think I got all of the mismatches and missing items that were relevant that you noted. Thanks for making that table.

In addition to those, I added fields for the pca dictionary that's present in objects. This should be ready for another look.

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,259 @@
...
+
+def convert_cell_row_metadata_types(metadata):
+    for key, value in metadata.items():
+        # if the value is a dict, this is conditional metadata and we need to convert the inner values
+        # use recursion to do this
+        if isinstance(value, dict):
+            convert_cell_row_metadata_types(value)
+        # if detected, this is an int
+        elif key == "detected":
+            metadata[key] = "int32"
```
sjspielman: just confirming (given my cluster_nn comment) that I see int32:
```
>>> type(adata.obs["detected"][0])
<class 'numpy.int32'>
```

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,259 @@
...
+]
+
+
+def convert_experiment_metadata_types(metadata):
+    for key, value in metadata.items():
+        # account for nested ditionaries in the conditional experiment metadata
+        if isinstance(value, dict):
+            convert_experiment_metadata_types(value)
+        elif key == "cluster_nn":
+            metadata[key] = "numpy.int64"
```
sjspielman: Hm maybe this was wrong actually? 

I just read in a processed anndata file to be sure, and I'm seeing:
```
>>> type(adata.uns["cluster_nn"])
<class 'int'>
>>> isinstance(adata.uns["cluster_nn"], np.int64)
False
```
But you clearly printed this and saw: `cluster_nn <class 'numpy.int64'>`. Do you know which library you printed out there? I am printing `SCPCL000001`

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,259 @@
...
+layers = ["spliced"]
+
+# unfiltered cell metadata ------------
+anndata_specific_obs_metadata = {
+    # sample metadata is present in obs for anndata
+    # this minimally includes library id, sample id and assay ontology term id
+    "library_id": "category",
+    "sample_id": "category",
+    "assay_ontology_term_id": "category",
+    # columns that we explicitly add in sce_to_annndata
```
sjspielman:
```suggestion
    # columns that we explicitly add in sce_to_annndata
    "assay_ontology_term_id": "category",
```

sjspielman (commented):
> So the plan is to add this into scpca-nf as a process within the workflow, not something that will be run separately from the workflow. Because of that we should accommodate any data that could be run through this and not just Portal data. So in general, I'm keeping the formatting checks here general enough to apply to all data and not just data we process.

Ah! Ok, this makes total sense then. Thanks for clarifying this!

references/scripts/create-formatting-reference.py
sjspielman: Either way, I think we want to have a dictionary of exceptions and then we can make these `elif`s --> `elif key in that dictionary's keys`, to track them better.

allyhawkins: Are you sure you are reading in the correct data? I was using SCPCL000698 but I just grabbed SCPCL000001 from the Portal bucket and read it in: 
```
>>> adata = anndata.read_h5ad("SCPCL000001_processed_rna.h5ad")
>>> type(adata.uns["cluster_nn"])
<class 'numpy.int64'>
>>> isinstance(adata.uns["cluster_nn"], numpy.int64)
True
```

allyhawkins:
I updated this to use a dictionary for the exceptions for both the row/cell metadata and experiment metadata. I'm not sure why you are seeing the discrepancy in type in `cluster_nn`, but I tried two libraries that I freshly downloaded and it looks correct.

references/scripts/create-formatting-reference.py
sjspielman: This has got to be a numpy versioning thing? What version are you on?

We're going to need to run this in the pixi environment. I didn't use it here, but our pixi environment here has version 2.4.2. For this though, we'll need anndata in the toml. (This may mean we have more changes to make here)

```
➤ aws s3 cp s3://ccdl-scpca-nf-results-prod/results/SCPCP000001/SCPCS000001/SCPCL000001_processed_rna.h5ad .
download: s3://ccdl-scpca-nf-results-prod/results/SCPCP000001/SCPCS000001/SCPCL000001_processed_rna.h5ad to ./SCPCL000001_processed_rna.h5ad
➤ python 
Python 3.11.14 | packaged by conda-forge | (main, Oct 22 2025, 22:56:31) [Clang 19.1.7 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import anndata
>>> adata = anndata.read_h5ad("SCPCL000001_processed_rna.h5ad")
>>> type(adata.uns["cluster_nn"])
<class 'int'>
>>> numpy.__version__
'2.4.1'
```

references/scripts/create-formatting-reference.py
```diff
@@ -395,3 +395,264 @@
...
+
+# if the library is multiplexed, the sample ID doesn't get added
+anndata_specific_obs_metadata_conditional = {
+    "is_multiplexed": {
+        "sample_id": None,
+    }
+}
+
+obs_metadata = {
+    **convert_cell_row_metadata_types(cell_metadata),
```
jashapiro: Since the function you have is working directly on the `metadata` object and modifying in place (which makes sense for the recursion), you probably want the objects that go into these calls to be copies of the originals.
```suggestion
    **convert_cell_row_metadata_types(copy.deepcopy(cell_metadata)),
```

references/scripts/create-formatting-reference.py
allyhawkins: > We're going to need to run this in the pixi environment.

We actually are going to want to match the version that's in `scpcaTools` since we are going to be using the `anndata` container for the process that does the checking. It looks like that version is 2.1.2 (see https://github.com/AlexsLemonade/scpcaTools/blob/main/docker/requirements_anndata.txt). I will double check using that version that the types are right before re-requesting review.

sjspielman: Ugh yes right it's part of the workflow!!! 🤦‍♀️

allyhawkins:
Okay so I used the version of `numpy` that's in `scpcaTools` and there were a few changes to how the metadata types are stored which is why Stephanie was seeing `int` when I wasn't. The types should now be `int` `float` and `bool` instead of `numpy.int64` etc. 

I also implemented Josh's suggestion to use `deepcopy`.

jashapiro:
> Okay so I used the version of `numpy` that's in `scpcaTools` and there were a few changes to how the metadata types are stored which is why Stephanie was seeing `int` when I wasn't. The types should now be `int` `float` and `bool` instead of `numpy.int64` etc.

It is worth noting that you may want to be flexible here when you get to checking. I think it is correct to store here as `int` but you may want to use a more flexible check in your code like `isinstance(x, (int, np.integer))` that can handle both native python and numpy integer types (without regard to how many bits). Same story for floats: `isinstance(x, (float, np.floating))`.

references/anndata-formatting-reference.json
```diff
@@ -0,0 +1,460 @@
...
+                "umi_filtering": {
+                    "umi_cutoff": "float"
+                },
+                "has_miQC": {
+                    "prob_compromised_cutoff": "float"
+                },
+                "has_clusters": {
+                    "cluster_algorithm": "str",
+                    "cluster_weighting": "str",
+                    "cluster_nn": "float"
```
sjspielman: we lost the int here

references/anndata-formatting-reference.json
```diff
@@ -0,0 +1,460 @@
...
+                "min_gene_cutoff": "int",
+                "X_name": "str",
+                "schema_version": "str"
+            },
+            "uns_conditional": {
+                "umi_filtering": {
+                    "umi_cutoff": "float"
+                },
+                "has_miQC": {
+                    "miQC_model": "S4",
```
sjspielman: Sorry, I just noticed this - the miqc model itself will be dropped before anndata conversion, so we need to handle this also.

references/anndata-formatting-reference.json
```diff
@@ -0,0 +1,460 @@
...
+                    "high.ambient": "bool"
+                },
+                "sizeFactor": "float64"
+            },
+            "uns": {
+                "library_id": "category",
+                "sample_id": "category",
+                "project_id": "category",
+                "reference_index": "category",
+                "total_reads": "int32",
```
sjspielman: This should be just `int`, as we have for the rna section. The types only seem to have the bits when it's been numpy-ified by a data structure like an array or pandas df. For spots like this where it's just a value, we don't have the bits.

sjspielman (commented):
Homestretch!! 

We lost the cluster_nn fix, and I realized we need to not be checking the miqc model in python. Maybe drop uns items that don't have a correspond key in `EXPERIMENT_METADATA_MAP`?

references/anndata-formatting-reference.json
allyhawkins: So it looks like this is because I had a typo in how I was converting the `uns` for the ADT experiment metadata and was accidentally using the row/cell conversion function instead of the experiment conversion function. This should fix this and all the other incorrect entries here!

allyhawkins:
Okay I updated this to remove the miQC model all together, fixed the cluster_nn situation and then made a fix to the conversion for ADT experiment metadata.

sjspielman (approved):
LGTM! 🐈‍⬛

PR #1214, opened 2026-03-11
Add stub for new visium additions

### LINKED ISSUE #1185: Prepare test files for updated space ranger version
Probably as an earlier step in developing updated space ranger functionality, we should update test data files as needed to ensure we can test through `stub` during development.

Opening this as a placeholder for now; details to be filled in.

sjspielman:
Closes #1185

This PR adds/updates stub files for visium. Here are the changes I made:

- I kept the existing `visium` row for legacy reasons, but rearranged its run files to match our current file organization. 
- I added two new runs: `visium1` and `visium2` with associated run files.
  - Let me know if you would prefer that I just swap the `visium` -> `visium1`. Currently I am retaining `visium` for legacy reasons, but I see a fair argument that we don't need to and can just keep `visium1` and `visium2` moving forward.
- Caught and fixed a small bug - the way I am currently checking for a cytaimage file was logging and error even when we don't need the cytaimage file (eg visium/visium1 tech). It runs fine, but it just logs the unnecessary error. I fixed this so an error is no longer logged unnecessarily  
- Miscellaneous other changes
  - I finally fixed the files directory bug associated with `STUBR07`, hurray!
  - I noticed we had an extra/unused infercnv gene order file hanging around, so I removed it. The one that is actually used as reference file is still there!


test/references/stub/annotation/stub_cytassist_probes.csv
sjspielman, line 1: In which github decided that removing the extra gene order stub file and adding the probe stub file was renaming a file. They are empty so this is fair!

test/stub-run-metadata.tsv
```diff
@@ -5,7 +5,7 @@ STUBR03	STUBL02	STUBS02	STUBP02	CITEseq_10Xv3	cell	stub	test/runs/STUBR03	test/r
 STUBR04	STUBL04	STUBS04	STUBP02	paired_end	bulk	stub	test/runs/STUBR04	NA	NA	NA	NA	NA	NA
 STUBR05	STUBL05	STUBS05	STUBP03	10Xv3	nucleus	stub	test/runs/STUBR05	NA	NA	NA	NA	NA	NA
 STUBR06	STUBL06	STUBS06	STUBP04	10Xv2_5prime	cell	stub	test/runs/STUBR06	NA	NA	NA	NA	NA	NA
-STUBR07	STUBL07	STUBS07	STUBP05	visium	spot	stub	tests/runs/STUBR07	NA	NA	A1	V10L14-093	NA	NA
+STUBR07	STUBL07	STUBS07	STUBP05	visium	spot	stub	test/runs/STUBR07	NA	NA	A1	V10L14-093	NA	NA
```
allyhawkins, line 8: Just noting that I agree with keeping this right now.

allyhawkins (approved):
LGTM

test/runs/STUBR21/image/STUBR21-tissue.jpg
jashapiro, line 1: I'd go ahead and give this a different image extension, just for giggles.

modules/spaceranger.nf
jashapiro: The logic here doesn't seem quite right, so I would separate the cases, which also allows more informative errors.

I also made the two errors first, just to keep them together
```suggestion
        if (meta.technology in non_cytassist_techs && cytaimage_file){
          log.error("Did not expect a cytaimage file for ${meta.techology} in ${meta.files_directory} but found ${cytaimage_file.size()} files.")
          cytaimage_file = []
        } else if (cytaimage_file.size() > 1) {
          log.error("Expected exactly 1 cytaimage file in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
          cytaimage_file = [] 
        } else {
          // we have 1 cytaimage file as we should
          cytaimage_file = cytaimage_file[0]
       }
```

jashapiro (commented):
I'm suggesting a small logic/error update, but I should not need to see this again.

test/runs/STUBR21/image/STUBR21-tissue.jpg
sjspielman: I will do this, especially because you graciously let me keep my clean up comment over in `scpca-admin`

modules/spaceranger.nf
sjspielman: So something about this was off, and while debugging it, my brain decided to assume that `[]` was truthy. This is obviously incorrect, but once I reminded my brain that that was incorrect, I landed here - it was missing a check to confirm that it _is_ size 1, otherwise we're indexing out something that doesn't exist and getting `null` files. 

```
        if (meta.technology in non_cytassist_techs && cytaimage_file) {
          log.error("Did not expect a cytaimage file for ${meta.technology} in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
          cytaimage_file = []
        } else if (!(meta.technology in non_cytassist_techs)) {
          if (cytaimage_file.size() != 1) {
            log.error("Expected exactly 1 cytaimage file in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
            cytaimage_file = []
          } else {
              // we correctly have one cytaimage file, so index it out
              cytaimage_file = cytaimage_file[0]
          }
        } // if no condition was hit, that means it is already []
``` 

Confirming it runs through before merging: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/5wM3D9L2L3GUOa/tasks
Note that there are now fewer fastq files for these test datasets, so we should move quicker! (I confess it was faster to remove files than subset them, but if this is still annoyingly slow I will upload ensmallened fastqs to S3)

modules/spaceranger.nf
```diff
@@ -176,18 +179,25 @@ workflow spaceranger_quant{
...
-        }
+        } else if (!(meta.technology in non_cytassist_techs)) {
+          if (cytaimage_file.size() != 1) {
+            log.error("Expected exactly 1 cytaimage file in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
+            cytaimage_file = []
+          } else {
+              // we correctly have one cytaimage file, so index it out
+              cytaimage_file = cytaimage_file[0]
+          }
+        } // if no condition was hit, that means it is already []
```
jashapiro: I don't like recurrent testing for things and their not, and you are already nesting, so...
```suggestion
        if (meta.technology in non_cytassist_techs) {
          if (cytaimage_file) log.error("Did not expect a cytaimage file for ${meta.technology} in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
          cytaimage_file = []
        } else { // we have a cytassist tech
          if (cytaimage_file.size() != 1) {
            log.error("Expected exactly 1 cytaimage file in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
            cytaimage_file = []
          } else {
              // we correctly have one cytaimage file, so index it out
              cytaimage_file = cytaimage_file[0]
          }
        } 
```

sjspielman:
I do want to note something I am seeing locally in the stub log though, so it's a bit curious to me that it didn't fail since I would think inputting `path("")` into a process is something nextflow would complain about!

```
DEBUG nextflow.util.HashBuilder - Unable to get file attributes file: /Users/sjspielman/ALSF/scpca/scpca-nf/stub/annotation/stub_cytassist_probes.csv -- Cause: java.nio.file.NoSuchFileException: /Users/sjspielman/ALSF/scpca/scpca-nf/stub/annotation/stub_cytassist_probes.csv
```

Indeed the workflow doesn't currently parse the probe file path into the meta map using the correct ref_dir, since it's not there yet! These files are temporarily still in example data, but the way the stub is coded up expects the probe file to be in ref_dir. So we probably want to get this into the references bucket sooner rather than later if we can, which involves some elevated permissions (cc @jashapiro).

modules/spaceranger.nf
sjspielman: done in https://github.com/AlexsLemonade/scpca-nf/pull/1214/commits/194d25a01a71511280b65d811df086ab89eb8f04

sjspielman:
🎉 https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/aPtz9H2JV1q4L/tasks
run at https://github.com/AlexsLemonade/scpca-nf/pull/1214/commits/194d25a01a71511280b65d811df086ab89eb8f04

PR #1216, opened 2026-03-12
Update spaceranger image handling

### LINKED ISSUE #1212: Add checks for other image files
In #1205, we included code to check that there is only one cytassist image. We should probably do this for image, colorizedimage, and darkimage (but this one is <=6 files).

sjspielman:
Closes #1212 

This PR adds (and fixes on occasion!) some image checks and allows for us to use any kind of image. I have tested this under a variety of conditions: https://github.com/AlexsLemonade/scpca-nf/issues/1212#issuecomment-4047954125 

Here's how it now works (edited for how it final.final actually works):
- if cytassist, _error entirely_ if there is not 1 cytaimage
- for the other images, log an error (but don't error out!) if there are too many images (>1 image and colorized image, >6 darkimage)
- error out if:
  -  there's >1 image type. there can only be one
  - it's not cytassist and there is _no_ image at all
- then, we pick final image as what should now be the only one that exists

### DIFF
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -185,7 +185,8 @@ workflow spaceranger_quant{
           cytaimage_file = []
         } else { // we have a cytassist tech
           if (cytaimage_file.size() != 1) {
-            log.error("Expected exactly 1 cytaimage file in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
+            // fully error if no cytaimage, since it is required
+            error("Expected exactly 1 cytaimage file in ${meta.files_directory}/cytaimage but found ${cytaimage_file.size()} files.")
             cytaimage_file = []
           } else {
             // we correctly have one cytaimage file, so index it out
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -36,6 +36,10 @@ process spaceranger {
     # write metadata
     echo '${meta_json}' > ${out_id}/scpca-meta.json
 
+    # DEBUGGING - why still intermediates????
+    ls ${out_id} > ${out_id}/contents.txt
+
+
     # TODO: Why is this still present in the checkpoints directory???
     # TODO: Is there more we'd like to remove?
     # remove Space Ranger intermediates directory
--- a/modules/spaceranger.nf
+++ b/modules/spaceranger.nf
@@ -193,16 +193,37 @@ workflow spaceranger_quant{
           }
         } 
 
-
+        // look for any additional images and check that there are not too many
         def image_file = getImageFiles("${meta.files_directory}/image")
+        if (image_file.size() > 1) {
+          log.error("Expected no more than 1 brightfield image file for ${meta.technology} in ${meta.files_directory}/image but found ${image_file.size()} files.")
+          image_file = []
+        } 
         def colorizedimage_file = getImageFiles("${meta.files_directory}/colorizedimage")
-        def darkimage_files = getImageFiles("${meta.files_directory}/darkimage") 
+        if (colorizedimage_file.size() > 1) {
+          log.error("Expected no more than 1 colorized image file for ${meta.technology} in ${meta.files_directory}/colorizedimage but found ${colorizedimage_file.size()} files.")
+          colorizedimage_file = []
+        } 
+        def darkimage_file = getImageFiles("${meta.files_directory}/darkimage")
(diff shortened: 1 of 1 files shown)


modules/spaceranger.nf
```diff
@@ -185,32 +185,54 @@ workflow spaceranger_quant{
...
+          colorizedimage_file = []
+        } 
+        def darkimage_file = getImageFiles("${meta.files_directory}/darkimage")
+        if (darkimage_file.size() > 6) {
+          log.error("Expected between 1-6 dark image files for ${meta.technology} in ${meta.files_directory}/darkimage but found ${darkimage_file.size()} files.")
+          darkimage_file = []
+        } 
+
+        // Determine which image file to pass in to spaceranger - there can only be one
+        // Here we simultaneously define the image_type (the spaceranger flag), and the image file itself
```
sjspielman:
```suggestion
        // Here we simultaneously define the the image file itself and image_type (the spaceranger flag)
```

modules/spaceranger.nf
```diff
@@ -211,7 +211,13 @@ workflow spaceranger_quant{
           darkimage_file = []
         } 
 
-        // Determine which image file to pass in to spaceranger - there can only be one
+        // there can only be one of image, colorizedimage, or darkimage
+        def found_images = [image_file, colorizedimage_file, darkimage_file].count { it.size() > 0 }
+        if (found_images > 1) {
+            error("Expected no more than 1 type of provided image in ${meta.files_directory} but found these: ${found_images}")
```
jashapiro: I feel like this logic should be combined with the next block; the repetitive logic is going to be a sources of errors. 
I would do something like the following:

```
def found_images = [
            [image_file, "image"],
            [colorizedimage_file, "colorizedimage"],
            [darkimage_file, "darkimage"]
].findAll {it[0].size() > 0}

if (found_images.size() > 1) {
            error("Expected no more than 1 type of provided image in ${meta.files_directory} but found these: ${found_images}")
} 

// one image found, get values
def (selected_image_file, image_type) = found_images ?: [[], ""]

Comment (sjspielman) [reply modules/spaceranger.nf]: Wait. Were we doing this at the same time?

Comment (sjspielman) [reply modules/spaceranger.nf]: b5849cd9f2ac99330c34088a8384f467cf86b221

File: modules/spaceranger.nf [hunk 3]
```diff
@@ -185,32 +185,61 @@ workflow spaceranger_quant{
...
+          .findAll{ it[0].size() > 0 } 
+
+        // confirm we have the right number of images:
+        // no more than 1 image for any technology, but require 1 for non-cytassist
+        if (present_images.size() > 1) {
+          // ensure informative print statement with info about the detected files
+          error("Expected no more than 1 type of provided image in ${meta.files_directory} but found: ${present_images.collect { "${it[1]}: ${it[0]}" }.join(", ")}")
+        } else if (meta.technology in non_cytassist_techs && present_images.size() == 0) {
+          error("At least one image file is required for ${meta.technology} but none were found in ${meta.files_directory}.")
+        }
```
allyhawkins, line 230: This is minor, but I might move this check to be before the actual number of image checks. We probably want this to fail first since it's an actual error and then do the logging?

sjspielman: Sorry, this comment is a little hard to read which checks you're referring to because of github formatting. Let me confirm! Are you asking to move `else if (meta.technology in non_cytassist_techs && present_images.size() == 0)` before `(present_images.size() > 1)`?

allyhawkins: I was suggesting to move this whole chunk to be before the log.error checks where you check the total number of images for each type. But I see you reset `image = []` if there's too many found so then this throws an error, so you can ignore this.

allyhawkins (approved):
LGTM

modules/spaceranger.nf
sjspielman: > But I see you reset image = [] if there's too many found so then this throws an error, so you can ignore this.

Wait, maybe not. I wonder if I can actually simplify this, let me look closely. Please hold!

sjspielman: I was thinking, could I get rid of resetting them to `[]`? But, that means the following code wouldn't work because we assume only 0 or 1 files is present by that point. I think this could be condensed overall, but the logging would not be as thorough, so _now_ I will ignore your comment ;)

PR #1221, opened 2026-03-14
Add support for visium hd & hd 3'

### LINKED ISSUE #1204: Accommodate visium HD technologies

### LINKED ISSUE #1213: Figure out why intermediate spaceranger directory isn't being deleted
Just noting we should try to debug this before release.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/1205#discussion_r2914400425_

sjspielman:
Closes #1204 
Closes #1213 

This PR adds support for Visium HD/HD 3'. Note this comment about files in outputs - it's a lot!!: https://github.com/AlexsLemonade/scpca-nf/issues/1204#issuecomment-4060593867

Changes include:
- added stub files for hd/hd3' (one intentionally has no image/)
- Bug fix, I set up before for the wrong hd tech to use probes. Fixed that list!
- Fixed the bug where the intermediate directory was being copied to checkpoints against our wishes
- Added a new spaceranger_hd process to grant more memory/cpu; we can't do a dynamic `label`, unfortunately, so we do need both processes
- Updated output file handling
  - The full results (you can see in checkpoints e.g. `s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/scpca_out/checkpoints/spaceranger/SCPCL300000/SCPCR300000-spatial/`) contain lots of nested `analysis/` directories across binning levels, which we do not want. There are also a lot of fusion symlinks that got carried forward into the nested `spatial` subdirectories, so I had to use a bit more logic to ensure we are publishing the result the files we really need and not copy over the symlinks. The core `spatial` files shared by all binnings are all at the top-level of `outs`.
  - Note that there are parquet files for tissue positions in HD land, not csvs! I renamed these according to our current convention with the web summary. 
- Segmentation is only performed if a brightfield image is specified, so I had to handle this logic too. Here's what I did:
  - First, I added a check in the workflow to log an error if segmentation results were expected (i.e. a brightfield image was specified) but there is no segmented output. (edit) I also ensured this error doesn't get logged for stub libraries, since stub processes don't create any checkpoints. But, I confirmed that this error would get logged for the stub library I set up with the image file!
  - Then, in the publish process, I only copy the directory if it exists
- A couple more updates to the metadata R script to take care of different column names between regular visium and visium hd.

I will say, the logic in the publishing process is necessarily many lines, since we need finer-grained control over which files per directory get copied. I am not too worried about this since I think the bash code is clear (there's just a lot of this), but if we prefer a separate process for publishing hd because of future-proofing/maintenance, I think that's not unreasonable.


config/process_base.config
```diff
@@ -21,6 +21,9 @@ process {
   withLabel: mem_32 {
     memory = { 32.GB * task.attempt }
   }
+  withLabel: mem_48 {
+    memory = { 48.GB * task.attempt }
```
jashapiro: Do you think doubling is really needed here? Maybe
```suggestion
    memory = { 24.GB + 24.GB * task.attempt }
```

modules/spaceranger.nf
```diff
@@ -65,21 +114,60 @@ process spaceranger_publish {
...
 
     # copy over needed files to outs directory
-    cp -r ${spatial_out}/outs/filtered_feature_bc_matrix ${spatial_publish_dir}
-    cp -r ${spatial_out}/outs/raw_feature_bc_matrix ${spatial_publish_dir}
     cp -r ${spatial_out}/outs/spatial ${spatial_publish_dir}
     cp ${spatial_out}/outs/web_summary.html ${spatial_publish_dir}/${meta.library_id}_spaceranger-summary.html
 
+    # copy over files to outs directory that depend on the technology, and define inputs to the R script
+    if [[ ${is_hd} == "true" ]]; then
+      # ensure we are only copying over the count & spatial results, not nested analysis dirs, and that we aren't copying symlinks
```
jashapiro: I would expect fusion to resolve symlinks on publication, is this not happening? According to nextflow-io/nextflow#4725  and nextflow-io/nextflow#5004 it should work. But if it does not, we can probably identify a better pattern than having to manually copy files.

Update: Do you mean the `.fusion.symlinks` files? I'm surprised those aren't being cleaned up too, but I wouldn't be concerned about them, especially if it allows simplifying the logic.

modules/spaceranger.nf
```diff
@@ -28,18 +28,68 @@ process spaceranger {
...
-    # TODO: Is there more we'd like to remove?
-    # remove Space Ranger intermediates directory
+
+// this process is identical to spaceranger, but with more memory for HD & HD 3' datasets
+process spaceranger_hd {
+  container "${pullthroughContainer(params.spaceranger_container, params.pullthrough_registry)}"
+  publishDir "${meta.spaceranger_publish_dir}", mode: 'copy'
+  tag "${meta.run_id}-spatial"
+  label 'cpus_24'
+  label 'mem_128'
```
jashapiro: Just for curiosity, when you were running your local tests, were you using 24 cpus or fewer? Memory requirements scale with CPUs, so that could explain increased memory requirements. 

Did you see a large benefit from increasing the CPU count? Often the scaling benefit falls off a lot sooner than that, especially given instance cost and availability. Note that while we have that cpu label, we don't actually use it anywhere else.

modules/spaceranger.nf
```diff
@@ -65,21 +114,60 @@ process spaceranger_publish {
...
-    cp -r ${spatial_out}/outs/filtered_feature_bc_matrix ${spatial_publish_dir}
-    cp -r ${spatial_out}/outs/raw_feature_bc_matrix ${spatial_publish_dir}
     cp -r ${spatial_out}/outs/spatial ${spatial_publish_dir}
     cp ${spatial_out}/outs/web_summary.html ${spatial_publish_dir}/${meta.library_id}_spaceranger-summary.html
 
+    # copy over files to outs directory that depend on the technology, and define inputs to the R script
+    if [[ ${is_hd} == "true" ]]; then
+      # ensure we are only copying over the count & spatial results, not nested analysis dirs, and that we aren't copying symlinks
+
+      cp -r ${spatial_out}/outs/spatial ${spatial_publish_dir}/
```
jashapiro: duplicate copy
```suggestion
```

modules/spaceranger.nf
```diff
@@ -65,21 +114,60 @@ process spaceranger_publish {
...
     mkdir ${spatial_publish_dir}
 
     # copy over needed files to outs directory
-    cp -r ${spatial_out}/outs/filtered_feature_bc_matrix ${spatial_publish_dir}
-    cp -r ${spatial_out}/outs/raw_feature_bc_matrix ${spatial_publish_dir}
     cp -r ${spatial_out}/outs/spatial ${spatial_publish_dir}
     cp ${spatial_out}/outs/web_summary.html ${spatial_publish_dir}/${meta.library_id}_spaceranger-summary.html
 
+    # copy over files to outs directory that depend on the technology, and define inputs to the R script
+    if [[ ${is_hd} == "true" ]]; then
```
jashapiro: I always like to put the short version first to make the end of the block easier to find, so I would switch this test to `!= "true"`

modules/spaceranger.nf
```diff
@@ -65,21 +114,60 @@ process spaceranger_publish {
...
+      cp ${spatial_out}/outs/barcode_mappings.parquet ${spatial_publish_dir}/${meta.library_id}_barcode_mappings.parquet
+
+      for square_dir in square_002um square_008um square_016um; do
+        mkdir -p ${spatial_publish_dir}/binned_outputs/\${square_dir}/spatial
+
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/spatial/scalefactors_json.json ${spatial_publish_dir}/binned_outputs/\${square_dir}/spatial/
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/spatial/tissue_positions.parquet ${spatial_publish_dir}/binned_outputs/\${square_dir}/spatial/
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/raw_feature_bc_matrix ${spatial_publish_dir}/binned_outputs/\${square_dir}/
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/filtered_feature_bc_matrix ${spatial_publish_dir}/binned_outputs/\${square_dir}/
+      done
```
jashapiro: A bit less repetition?
```suggestion
      for square_src in ${spatial_out}/outs/binned_outputs/square_*; do
        square_dest=${spatial_publish_dir}/binned_outputs/\$(basename \$square_src)
        mkdir -p \${square_dest}

        cp  \${square_src}/spatial/scalefactors_json.json \${square_dest}/spatial/
        cp  \${square_src}/spatial/tissue_positions.parquet \${square_dest}/spatial/
        cp -r \${square_src}/raw_feature_bc_matrix \${square_dest}/
        cp -r \${square_src}/filtered_feature_bc_matrix \${square_dest}/
      done
```

modules/spaceranger.nf
```diff
@@ -232,7 +322,10 @@ workflow spaceranger_quant{
         // Simultaneously define the the image file itself and image_type (the spaceranger flag)
         def (selected_image_file, image_type) = present_images[0] ?: [[], ""]
 
-        // input to spaceranger process
+        // used for checking for segmented_results for hd technologies
+        meta.has_brightfield_image = image_type == "image"
```
jashapiro: Why don't we add the visium image type directly to the metadata so we can access it more easily in any downstream application? Then we might also be able to reduce the passed tuple to remove `image_type`.

The check for `has_brightfield_image` can just be where we need it. 
```suggestion
        // store image type for downstream reference
        meta.visium_image_type = image_type
```

modules/spaceranger.nf
```diff
@@ -257,7 +355,21 @@ workflow spaceranger_quant{
...
+      }
+
+
+    grouped_spaceranger_ch = spaceranger.out
+      .mix(spaceranger_hd.out)
+      .mix(spaceranger_quants_ch)
+
 
     // generate metadata.json
     spaceranger_publish(grouped_spaceranger_ch)
```
jashapiro: I don't see what is grouped here, so renaming
```suggestion
    spaceranger_output_ch = spaceranger.out
      .mix(spaceranger_hd.out)
      .mix(spaceranger_quants_ch)


// generate metadata.json
spaceranger_publish(spaceranger_output_ch)
```

modules/spaceranger.nf
```diff
@@ -257,7 +355,21 @@ workflow spaceranger_quant{
...
 
-    grouped_spaceranger_ch = spaceranger.out.mix(spaceranger_quants_ch)
+    // log error about segmented_outputs as needed
+    spaceranger_hd.out
+      .map{ meta, out_id ->
+        def has_segmented_outputs_dir = file("${meta.spaceranger_results_dir}/segmented_outputs").exists()
+        if (meta.has_brightfield_image && !has_segmented_outputs_dir) {
+          log.error("Expected segmented_outputs directory for ${meta.library_id} with brightfield image, but Space Ranger did not create it.")
+        }
+        [meta, out_id]
```
jashapiro: The output file might not exist yet in the results directory, so we have to check the output directly. Note that I also updated to use `meta.visium_image_type` assuming that will be created instead of `has_brightfield_image` 

We can also use `subscribe` here as we are not doing anything with the output here.
```suggestion
      .subscribe{ meta, output_path ->
        if (meta.visium_image_type == "image" && !(${output_path} / "segmented_outputs").exists()) {
          log.error("Expected segmented_outputs directory for ${meta.library_id} with brightfield image, but Space Ranger did not create it.")
        }
```

jashapiro (commented):
This looks good overall, but I had a number of style suggestions. The most consequential one is probably the suggestion to store the image type in meta rather than just a boolean for one specific type.

modules/spaceranger.nf
sjspielman: I mean both! There are `.fusion.symlinks` files, and the symlink files themselves.

modules/spaceranger.nf
sjspielman: For local tests, I was running with `--localcores 14 --localmem 32`. 
I had repeated failures with memory below 128 here, but actually I'm not sure I ever tried 128/12! I can do a repeat mapping with those resources and see if it makes it through.

modules/spaceranger.nf
jashapiro: What do the symlink files look like in the output? I would not expect any that were not resolved. (They all seem resolved in the checkpoint directory)

sjspielman:
I just want to bring this comment left on a commit to the main PR, since I think it requires more discussion given that we often check for STUB libraries in `main.nf`. 

https://github.com/AlexsLemonade/scpca-nf/commit/eef70a51c4c1be401ee407d5bfe32e2385bc144d#r179684889

modules/spaceranger.nf
jashapiro: From your answer, I can't tell if you every tried with 12/64, so I would try that first. The other thing you can do is change `--localmem` to be a small undershoot of the available memory. Something like `--localmem=${task.memory.toGiga() * 0.9}`

config/process_base.config
sjspielman: Ah, this was left over since I had to bump _way up_ to 128.

modules/spaceranger.nf
sjspielman: I didn't, because it was repeatedly failing after running with 48 even after 48 got doubled to 96 on a retry. So I skipped it altogether.

jashapiro:
> I just want to bring this comment left on a commit to the main PR, since I think it requires more discussion given that we often check for STUB libraries in `main.nf`.

I would not call twice often!

modules/spaceranger.nf
sjspielman: The issue is these files aren't expected in the spaceranger output structure. The images are only expected in the top-level spatial directory, so my concern is having (even resolved) symlinks in the nested spatial directories could break expectations for reading this data in. The only files that expected in the nested spatial directories are the json and parquet files.

modules/spaceranger.nf
jashapiro: Again, I am unclear if you were running with 12 or 24 cpus at that point

modules/spaceranger.nf
jashapiro: > The issue is these files aren't expected in the spaceranger output structure. The images are only expected in the top-level spatial directory, so my concern is having (even resolved) symlinks in the nested spatial directories could break expectations for reading this data in. The only files that expected in the nested spatial directories are the json and parquet files.

I don't understand what you mean here. Either the files are there (symlinks or full) or not at all? I don't understand how symlinks to files you don't expect to be created would be there.

sjspielman: There are symlinks present in nested spatial directories, and I don't expect them to be present at all. 
But I also don't know why.

sjspielman:
```
➤ aws s3 ls s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/scpca_out/checkpoints/spaceranger/SCPCL300000/SCPCR300000-spatial/outs/binned_outputs/square_008um/spatial/
2026-03-13 23:51:01          0 
2026-03-13 23:51:01        140 .fusion.symlinks
2026-03-13 23:51:01         38 aligned_fiducials.jpg
2026-03-13 23:51:01         41 aligned_tissue_image.jpg
2026-03-13 23:51:02         37 cytassist_image.tiff
2026-03-13 23:51:02         42 detected_tissue_image.jpg
2026-03-13 23:51:02        296 scalefactors_json.json
2026-03-13 23:51:02         39 tissue_hires_image.png
2026-03-13 23:51:02         40 tissue_lowres_image.png
2026-03-13 23:51:02   11301911 tissue_positions.parquet
```

I only expect to see `scalefactors_json.json` and `tissue_positions.parquet` here. I don't expect anything else to be present.

modules/spaceranger.nf
sjspielman: Alright let's back up. I previously tested:

- 12 cpu/24 gb
- 12 cpu/48 gb
- 12 cpu/96 gb
- 24 cpu/128 gb

I do not roundly test 24 cpu. These were not tested:
- 24 cpu/48 gb
- 24 cpu/96 gb
- 12 cpu/128 gb
- any cpu/64 gb

I will proceed to test 12/64.

sjspielman:
> I would not call twice often!

I do not understand this spirit of this comment and the practical advice it gives. I am trying to keep our code consistent and follow existing precedent, so when I receive reviews asking me to deviate from that precedent, more discussion is merited so we can decide as a team to either:
- change existing precedent
- continue using existing precedent

jashapiro:
> > I would not call twice often!
> 
> I do not understand this spirit of this comment and the practical advice it gives. I am trying to keep our code consistent and follow existing precedent, so when I receive reviews asking me to deviate from that precedent, more discussion is merited so we can decide as a team to either:
> 
>     * change existing precedent
> 
>     * continue using existing precedent

Inconsistent code is common as development of a large project progresses. Sometimes an undesired practice can sneak in a couple of times, but I would be very careful about assuming that makes it a precedent. 

The goal of stub runs is to test the logic of the workflow, independent of the code that needs to run within each process. As such, the workflow logic should not include special cases for stub runs. I have submitted #1222 to correct the places where we had done this.

modules/spaceranger.nf
```diff
@@ -65,21 +114,59 @@ process spaceranger_publish {
...
+      for square_dir in square_002um square_008um square_016um; do
+        mkdir -p ${spatial_publish_dir}/binned_outputs/\${square_dir}/spatial
+
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/spatial/scalefactors_json.json ${spatial_publish_dir}/binned_outputs/\${square_dir}/spatial/
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/spatial/tissue_positions.parquet ${spatial_publish_dir}/binned_outputs/\${square_dir}/spatial/
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/raw_feature_bc_matrix ${spatial_publish_dir}/binned_outputs/\${square_dir}/
+        cp -r ${spatial_out}/outs/binned_outputs/\${square_dir}/filtered_feature_bc_matrix ${spatial_publish_dir}/binned_outputs/\${square_dir}/
+      done
+
+      # only present if a brightfield image was supplied, so copy segmented_outputs conditionally
```
allyhawkins: If we aren't planning on making the segmented outputs available, why are we doing this? I thought we had decided to just pick one folder and provide that, which should be the 8 um folder based on your other comment? https://github.com/AlexsLemonade/scpca-nf/issues/1204#issuecomment-4060593867

sjspielman: Sorry for confusion - that comment was about which binning to use for calculating statistics for the json file. The 2, 8, and 16 binned_outputs should be provided (and segmented if doable) since this is standard spaceranger output.

modules/spaceranger.nf
jashapiro: How are those files getting into the checkpoints directory in the first place? Does `spaceranger` create them to have a "complete" `spatial` directory in each binning output that looks like the ones at `/outs/spatial`? 

If so, we may want to remove them during the spaceranger step itself before publishing?  Unfortunately, there is no good way to make S3 "relink" them on download. The other option is to copy the originals to those locations, if we think there is a reason users (or a later processing step) would want them in that arrangement.

jashapiro: If we think the symlinks are worth preserving, we could also `tar` up the `spaceranger` output? This will make it harder to look within the directory from nextflow, but we could perhaps add a `val` output to indicate whether segmented output was created. I think we would do this later if needed; we'd want to change all the spaceranger outputs to be the same format, and that goes beyond the scope here.

modules/spaceranger.nf
allyhawkins: Hmm okay. I thought when we chatted about implementation last week that we were only going to provide one type of output for people to use. I'm fine with providing all of it if that's what we want to do, but is there a reason to not just include the smallest output for people (2 um) and simplify the overall structure of the files we provide?

allyhawkins: Wait, I think I'm definitely misremembering this, and I now remember we said we would just give it all to them. Sorry for the confusion!

modules/spaceranger.nf
sjspielman: > How are those files getting into the checkpoints directory in the first place? Does spaceranger create them to have a "complete" spatial directory in each binning output that looks like the ones at /outs/spatial?

Ok, I have figured it out - `rclone` has been obfuscating things here. That's what I used to copy result files off of EC2 from my local runs, but it seems that `rclone` by default doesn't copy symlinks. So, in my own record of outputs from Space Ranger, these files aren't present. I can confirm this with a fresh E2 instance, but I'm pretty sure that's what has been going on and leading to confusion between my local runs and runs with scpca-nf.

> The other option is to copy the originals to those locations, if we think there is a reason users (or a later processing step) would want them in that arrangement.

I've tested this out with VisiumIO, and indeed the nested folders do need symlinks or actual files. I think what we do is copy over the files into spatial for now. This will increase the overall size of the output by somewhere in the 100-200 mb range. I will port your comment about tar'ing to preserve symlinks instead to keep in our back pockets.

sjspielman:
Thanks for feedback, everyone! Status: reviews implemented, much testing incoming in which I expect to catch syntax things and pin down resource requirements.

config/process_base.config
```diff
@@ -21,6 +21,9 @@ process {
   withLabel: mem_32 {
     memory = { 32.GB * task.attempt }
   }
+  withLabel: mem_64 {
+    memory = { 64.GB + 32.GB * task.attempt }
```
jashapiro:
```suggestion
    memory = { 32.GB + 32.GB * task.attempt }
```

sjspielman: woops, thanks!

modules/spaceranger.nf
```diff
@@ -28,18 +28,76 @@ process spaceranger {
...
+    meta_json = makeJson(meta)
+    image_arg = meta.visium_image_type ? "--${meta.visium_image_type} ${image_file.join(',')}" : "" // join in case of multiple darkimages
+    """
+    spaceranger count \
+      --id=${out_id} \
+      --transcriptome=${index} \
+      --fastqs=${fastq_dir} \
+      --sample=${meta.cr_samples} \
+      --localcores=${task.cpus} \
+      --localmem=${task.memory.toGiga()} \
```
jashapiro: This mirrors the default that `spaceranger` would use according to https://www.10xgenomics.com/support/software/space-ranger/2.1/analysis/running-pipelines/computing-options#single-server

If it prevents OOM errors, that would be nice to add everywhere (probably for `cellranger` too)
```suggestion
      --localmem=${task.memory.toGiga() * 0.9} \
```

modules/spaceranger.nf
sjspielman: I am now marinating on whether we need to return the top-level `spatial/` directory. It's technically expected space ranger output, but only if symlinks are provided since _the actual IO_ needs to have them within binned/segmented directories. Right now it just makes things bigger and wouldn't be used in IO code.

I'd appreciate feedback here to break my tie with myself - do you agree it's reasonable to skip the overall spatial directory until if/when we swap to tarball w/ symlinks?

sjspielman: Noting we decided to return the top-level spatial directory at this time.

sjspielman:
Update from resource testing: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/2FlwSZoHBraK9/tasks
- Run sans image (and therefore sans segmentation) finished fine with 12/64 in 2h58m
- Run _with_ image (and therefore with segmentation) failed until at 64 and 96 (with error 143) but succeeded at 128 in 4h16m, which is consistent with what I previously observed

So, it doesn't seem that the 0.9 memory scaling prevented OOM errors. 

We can compare this back to a run with 24/128: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/3wMWE5COfCADkL/tasks
- Run sans image completed in 1h47m, vs 2h58m with 12 cpu. This is not exactly but essentially a 2x speedup.
- With _with_ image completed in 2h22m, vs 4h16m with 24 cpu. Again, pretty darn close to 2x. 

---> I'm liking 24/64 sans image (not yet directly tested but seems like it would be on point) and 24/128 with image. But, it also seems to me we don't want to request 128 unless it's _definitely needed_

It was previously floated to use a dynamic memory label: https://github.com/AlexsLemonade/scpca-nf/issues/1204#issuecomment-4056114712, but nextflow doesn't allow this for labels. We could, however, still have dynamic memory and just not use a _label_. This would streamline code to allow for a single spaceranger process. For example (still pending testing, but conceptually) 32 for visium1/2, 64 for hd sans brightfield image, and 128 hd with brightfield image. https://www.nextflow.io/docs/latest/process.html#dynamic-task-resources

1. Would we be ok with not using a label for this specific case?
2. Taking a stab at how this could look.
```
memory {
    def mem = !meta.technology.contains("_hd") ? 16.GB 
             : meta.image_type != "image"      ? 32.GB
             :                                   64.GB
    mem + mem * task.attempt
}
```
--> visium1/2 start at 32 and scale up by 16 each attempt (_different from_ 32 label but seems fine for this case)
--> hd sans image starts at 64 and scales up by 32 each attempt (same as 64 label)
--> hd with image starts at 128 and scales up by 64 each attempt (same as 128 label)

jashapiro:
> * Run sans image completed in 1h47m, vs 2h58m with 12 cpu. This is not exactly but essentially a 2x speedup.
> 
> * With _with_ image completed in 2h22m, vs 4h16m with 24 cpu. Again, pretty darn close to 2x.

I do not think that is close enough to 2x to be worth it, given pricing. That is a 60-80% speedup for double the price.

That said, what we probably want to be more aware of is matching memory to cpus: the highest memory instances we use have 8GB/cpu, so 16cpus and 128GB memory will probably give us the best fit (other instance types can handle the lower memory runs when convenient).

sjspielman:
Confirming that dynamic memory is being allocated based on image presence as intended: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/XT6XU9TmvXyEk/tasks
I also did this for cpu, so 12 without image and 16 with image. Therefore, it's 12/64 & 16/128. 
Even though the 0.9 memory scaling didn't clearly help, it also didn't clearly hurt, so I think it's worth keeping just in case.

At this point, I think this PR is set and we can continue onto #1226 for next steps. We can do another round of review to implement any small changes, but larger changes if requested should be in a subsequent PR since this one is big enough!

Edit: argh, I need to fix something because the stub log still has
``` 2026-03-17T15:42:15.1132125Z [7d/8daeaa] Submitted process > spaceranger_quant:spaceranger_hd (STUBR23-spatial)
2026-03-17T15:42:15.1782405Z ERROR ~ Expected segmented_outputs directory for STUBL23 with brightfield image, but Space Ranger did not create it.
```

will re-request review when I nail this down!

modules/spaceranger.nf
```diff
@@ -257,10 +381,23 @@ workflow spaceranger_quant{
         ]
       }
 
-    grouped_spaceranger_ch = spaceranger.out.mix(spaceranger_quants_ch)
+    // log error about segmented_outputs as needed
+    // don't log for stub libraries
+    spaceranger_hd.out
+      .subscribe{ meta, out_id ->
+        def segmented_dir_exists = file("${meta.spaceranger_results_dir}/outs/segmented_outputs", type: 'dir').exists()
```
jashapiro: I think this suggestion got lost along the way. You should not use the meta value here because it refers to the publish directory, which is not guaranteed to exist yet. We want to check the process output here. 
```suggestion
    spaceranger_hd.out
      .subscribe{ meta, out_dir ->
        def segmented_dir_exists = (out_dir / "outs/segmented_outputs").isDirectory()
```

jashapiro (approved):
Looks good, with one comment that I think may have gotten missed along the way. It's an important one though.

modules/spaceranger.nf
sjspielman: I had seen this comment and said "no, this is the checkpoints, this is right".

But looking closer I see we just have confusing as heck variable names that got me. `spaceranger_publish_dir` is not a good name for a checkpoints directory. 
```
        meta.spaceranger_publish_dir =  "${params.checkpoints_dir}/spaceranger/${meta.library_id}"
        meta.spaceranger_results_dir = "${meta.spaceranger_publish_dir}/${meta.run_id}-spatial"
```

jashapiro: The main point is that anything from a `publishDir` statement is not guaranteed to exist while the pipeline is running; Nextflow makes those copies asynchronously.

sjspielman: >  Nextflow makes those copies asynchronously.

Perhaps an explanation for why removing that intermediate directory was fiddly

sjspielman:
confirming stub is happy: https://productionresultssa1.blob.core.windows.net/actions-results/6c6ad8a1-ec52-494c-bf91-84f233cdd545/workflow-job-run-dad9fe7f-9588-57d4-bdb1-6d065c1525b3/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-03-17T16%3A42%3A05Z&sig=tKKK7c7W2%2B7frdpFdUFLi4v8tutVDOW7jLkHLKNcgEE%3D&ske=2026-03-17T18%3A01%3A59Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-03-17T14%3A01%3A59Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-03-17T16%3A32%3A00Z&sv=2025-11-05

PR #1222, opened 2026-03-16
Remove special logic for stub processing

jashapiro:
To have more complete testing, we should not have any special logic for stub tests outside of the stub blocks themselves. Here I am removing all the places I could find where that had snuck into our workflow. All of these were places that we were checking file sizes and expecting non-zero sizes, so to remove them I simply made the stub block insert some text into the file rather than creating an empty file. 

I also did some indentation cleanup, so you might want to turn off whitespace!

### DIFF
--- a/lib/utils.nf
+++ b/lib/utils.nf
@@ -121,21 +121,21 @@ def pullthroughContainer(image_url, pullthrough_url = ""){
 
 
 /**
-  * For a given directory, return its space ranger image files 
+  * For a given directory, return its space ranger image files
   *
   * @param dir Directory to search image files(s) in
   * @param cytaimage Boolean if the image file(s) to grab are the cytaimage
   * @return Array of file paths
   */
 def getImageFiles(dir, cytaimage = false) {
-  
+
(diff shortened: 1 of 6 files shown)


sjspielman (commented):
There are actually 2 more* spots in main.nf that appear to have originated here https://github.com/AlexsLemonade/scpca-nf/pull/738/

https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/main.nf#L321-L323

https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/main.nf#L377-L382

jashapiro:
Thank you for pointing out the places I had missed!

For some reason my nextflow linter didn't like the regex (it seemed to be working, so I don't know) so I changed it to something that the linter didn't complain about. It might be a bit more readable too? regex: fun to write, impossible to read.

sjspielman (approved):
👍 

> regex: fun to write, impossible to read.

confirmed!

Issue #1226, opened 2026-03-17
Update spaceranger probe handling and add associated references


sjspielman:
The visium probe situation is more nuanced than* originally appreciated. I've made a table of how technologies pair up with probes and genomes...
| Technology                         | Sample Type | Probe Version | Species         | Probe Set CSV Filename (abbreviated) | Genome Ref      | Ensembl         |
| :--------------------------------- | :---------- | :------------ | :-------------- | :----------------------------------- | :-------------- | :-------------- |
| Visium v1 (Direct), poly-A capture | FF          | NA            | dealer's choice | NA                                   | dealer's choice | dealer's choice |
| Visium v1 (Direct), probe capture  | FFPE        | v1.0          | Human           | `v1.0_GRCh38-2020-A.csv`             | GRCh38          | 98              |
| Visium v1 (Direct), probe capture  | FFPE        | v1.0          | Mouse           | `v1.0_mm10-2020-A.csv`               | mm10            | 98              |
| Visium CytAssist and HD            | FF / FFPE   | v2.1          | Human           | `v2.1_GRCh38-2024-A.csv`             | GRCh38          | 110             |
| Visium CytAssist and HD            | FF / FFPE   | v2.1          | Mouse           | `v2.1_GRCm39-2024-A.csv`             | GRCm39          | 110             |
| Visium CytAssist and HD            | FF / FFPE   | v2.0          | Human           | `v2.0_GRCh38-2020-A.csv`             | GRCh38          | 98              |
| Visium CytAssist and HD            | FF / FFPE   | v2.0          | Mouse           | `v2.0_mm10-2020-A.csv`               | mm10            | 98              |
| Visium HD (poly-A capture)         | FF          | NA            | dealer's choice | NA                                   | dealer's choice | dealer's choice |

- The references are:
  - 2020-A: 
    - Human GRCh38 / ensembl 98
    - Mouse mm10 / ensembl 98
  - 2024-A is:
    - Human GRCh38 / ensembl 110
    - mouse GRCm39 / ensembl 110
- Noting that v2.0/v2.1 JUST refer to probe sets (which map to reference versions), not chemistry versions. In other words, there really is only visium2 and no such thing as visium2.1.

**Proposed actions to take**:
- Our current tracking of `visium` and `visium1` is insufficient. The v1 dataset I have been testing with is a fresh-frozen mouse brain, so it didn't need probes, but ffpe v1's will need probes. Therefore:
[...]

PR #1227, opened 2026-03-17
Rearrange flex and cytassist probe sets

allyhawkins:
Related to https://github.com/AlexsLemonade/ScPCA-admin/issues/1281#issuecomment-4074994650, here I'm making the changes needed to reorganize the probes so that we have them nested by organism and ensembl id. I'm filing this as a draft PR mostly to help convey the idea I had about how we can organize these files. 

This would mean the Visium viles would be here: 
```
s3://scpca-nf-references/homo_sapiens/ensembl-110/cytassist-probe-refs/Visium_Human_Transcriptome_Probe_Set_v2.1.0_GRCh38-2024-A.csv
s3://scpca-nf-references/mus_musculus/ensembl-104/cytassist-probe-refs/Visium_Mouse_Transcriptome_Probe_Set_v2.1.0_GRCm39-2024-A.csv
```

And the flex probes would be here: 
```
s3://scpca-nf-references/homo_sapiens/ensembl-110/flex-probe-refs/Chromium_Human_Transcriptome_Probe_Set_v2.1.0_GRCh38-2024-A.csv
s3://scpca-nf-references/mus_musculus/ensembl-104/flex-probe-refs/Chromium_Mouse_Transcriptome_Probe_Set_v2.1.0_GRCm39-2024-A.csv
```

When we store these file locations in the `scpca-refs.json` file we would first track organism and Ensembl ID and then track technology version. So instead of just adding a single file entry to the json, I added a list with the name of the technology so that we can first map the organism and assembly and then map the technology when creating the `meta` channel in the workflow. 

I added some stub files in the organization that I would expect these to be in as well.


references/scpca-refs.json
```diff
@@ -45,6 +53,10 @@
     "star_index": "mus_musculus/ensembl-104/star_index/Mus_musculus.GRCm39.104.star_idx",
     "infercnv_gene_order": "mus_musculus/ensembl-104/infercnv/Mus_musculus.GRCm39.104_gene_order_arms.txt.gz",
     "cytoband": "mus_musculus/ensembl-104/annotation/Mus_musculus.GRCm39.104_cytoband.txt.gz",
-    "cytassist_probe": "s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/cytassist-probes/Mus_musculus.GRCm39.104_cytassist-v2.1.0.csv"
+    "flex_probe_files": "",
+    "cytassist_probe_files": {
+      "visium2": "mus_musculus/ensembl-104/flex-probe-refs/Visium_Mouse_Transcriptome_Probe_Set_v2.1.0_GRCm39-2024-A.csv",
+      "visium_hd_3prime": "mus_musculus/ensembl-104/flex-probe-refs/Visium_Mouse_Transcriptome_Probe_Set_v2.1.0_GRCm39-2024-A.csv"
```
jashapiro, line 59: I would expect these to all move to a new `Mus_musculus.GRCm39.110`

references/scpca-refs.json
```diff
@@ -13,7 +13,8 @@
     "star_index": "homo_sapiens/ensembl-104/star_index/Homo_sapiens.GRCh38.104.star_idx",
     "infercnv_gene_order": "homo_sapiens/ensembl-104/infercnv/Homo_sapiens.GRCh38.104_gene_order_arms.txt.gz",
     "cytoband": "homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104_cytoband.txt.gz",
-    "cytassist_probe": ""
+    "flex_probe_files": "",
+    "cytassist_probe_files": ""
```
jashapiro, line 17: For type consistency
```suggestion
    "flex_probe_files": {},
    "cytassist_probe_files": {}
```

references/scpca-refs.json
```diff
@@ -45,6 +53,10 @@
     "star_index": "mus_musculus/ensembl-104/star_index/Mus_musculus.GRCm39.104.star_idx",
     "infercnv_gene_order": "mus_musculus/ensembl-104/infercnv/Mus_musculus.GRCm39.104_gene_order_arms.txt.gz",
     "cytoband": "mus_musculus/ensembl-104/annotation/Mus_musculus.GRCm39.104_cytoband.txt.gz",
-    "cytassist_probe": "s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/cytassist-probes/Mus_musculus.GRCm39.104_cytassist-v2.1.0.csv"
+    "flex_probe_files": "",
```
jashapiro, line 56:
```suggestion
    "flex_probe_files":  {},
```

main.nf
```diff
@@ -230,8 +230,9 @@ workflow {
         cellranger_index: sample_refs.cellranger_index ? "${params.ref_rootdir}/${sample_refs.cellranger_index}" : '',
         star_index: sample_refs.star_index ? "${params.ref_rootdir}/${sample_refs.star_index}" : '',
         infercnv_gene_order: sample_refs.infercnv_gene_order ? "${params.ref_rootdir}/${sample_refs.infercnv_gene_order}" : '',
+        flex_probeset: sample_refs.flex_probe_files ? "${params.ref_rootdir}/${sample_refs.flex_probe_files[it.technology.toLowerCase()]}" : '',
```
jashapiro, line 233: we probably need another check here that the probe for the technology exists.

jashapiro (commented):
Overall this looks reasonable to me. A few comments added here, and an overall note that I would prefer to call them visium probes (vs cytassist) throughout.

jashapiro:
But I forgot: this does not address the question of what to do when 10x releases version 2.1.1 or 2.2 with the same genome build.

references/scripts/create-reference-json.R
```diff
@@ -132,15 +142,31 @@ create_ref_entry <- function(
...
-    json_entry$cytassist_probe <- file.path(
-      cytassist_probe_dir,
-      glue::glue("{reference_name}_cytassist-v2.1.0.csv")
+  # add directory for cytassist probes
+  if (include_cytassist) {
+    json_entry$cytassist_probe_files <- list(
+      "visium2" = file.path(
+        flex_probe_dir,
+        glue::glue("Visium_{cellranger_organism}_Transcriptome_Probe_Set_v2.1.0_{assembly}-2024-A.csv")
+      ),
```
sjspielman, line 165: I think this is the likely outcome here for technology naming:
https://github.com/AlexsLemonade/scpca-nf/issues/1226#issuecomment-4076665240


```suggestion
      "visium2_v2.1" = file.path(
        flex_probe_dir,
        glue::glue("Visium_{cellranger_organism}_Transcriptome_Probe_Set_v2.1.0_{assembly}-2024-A.csv")
      ),
```

We'll need more of these, but we should add them alongside the new references so not now!

sjspielman (commented):
This is the organization I was imagining as well, so we're on the same page!

allyhawkins:
> But I forgot: this does not address the question of what to do when 10x releases version 2.1.1 or 2.2 with the same genome build.

It does address this. The version used for the probe set for flex is encoded in the `technology` string. That's why I'm using the technology to map to individual files within the organism and assembly. As noted in https://github.com/AlexsLemonade/scpca-nf/pull/1227#discussion_r2948444601 and the discussion in https://github.com/AlexsLemonade/scpca-nf/issues/1226#issuecomment-4076665240, we can do the same thing for spatial so that we have the probe version in the technology and use that to grab the appropriate file.

sjspielman (commented):
I'd expect the `flex_probesets` list to be removed as well.
https://github.com/AlexsLemonade/scpca-nf/blob/71f9e95f9a345b92b6249510b1a1dca3c823327e/main.nf#L139-L143
https://github.com/AlexsLemonade/scpca-nf/blob/71f9e95f9a345b92b6249510b1a1dca3c823327e/main.nf#L281
https://github.com/AlexsLemonade/scpca-nf/blob/71f9e95f9a345b92b6249510b1a1dca3c823327e/modules/cellranger-flex.nf#L129

It's currently only used to define the flex_techs, but I would just define this with the keys itself given how we now handle probes with `def flex_techs =  ['10xflex_v1.1_single', '10xflex_v1.1_multi']`
https://github.com/AlexsLemonade/scpca-nf/blob/71f9e95f9a345b92b6249510b1a1dca3c823327e/main.nf#L147

references/scripts/create-reference-json.R
sjspielman: FYI you can skip this comment if you want! I took care of the visium stuff in my stacked PR, using a slightly modified approach to ensure we only list the visium probes for relevant techs

allyhawkins:
@sjspielman it looks like you are making additional changes on top of what I made here to account for the additional visium organisms in #1228. I think it might be easier to just remove this PR and then rebase #1228 onto `development`? That would make it easier to review all these changes related to the probe organization together. We'll just need to double check that all the flex changes are still present when you re-base. Thoughts?

sjspielman:
> That would make it easier to review all these changes related to the probe organization together. We'll just need to double check that all the flex changes are still present when you re-base. Thoughts?

That's fine with me if it's fine with you!

allyhawkins:
That works for me! I just filed this to better explain the idea I had for handling probes. I'll go ahead and close this.

PR #1228, opened 2026-03-17
Update visium probe handling

### LINKED ISSUE #1220: Update cytassist probe directory
Once the cytassist probes are in S3, we need to update the reference paths that point to them. This will require updating and re-running the references script to update the json, and update how we parse this directory in main.nf

### LINKED ISSUE #1226: Update spaceranger probe handling and add associated references

sjspielman:
Closes #1220 
Closes #1226 

This PR is stacked on #1227 and fills in the remaining visium probe aspects. (Note there's a big diff when I'm filing because #1227 is behind development. This will sort itself out shortly.)

Changes include:
- Updated `references/` to handle the new references that are needed for the cytassist probes. The associated annotation files have been successfully uploaded to S3, so after this goes in we can trigger the GHA in `scpca-admin` to run that workflow 
- Updated the workflow probe names and handling. As part of this, I changed naming from cytassist -> visium as well.

Some of the changes I made here may end up conflicting with the final state of #1227, so I'm opening this as a draft until eventual potential conflicts can be resolved. 
Edit - it is also good that I opened as a draft because more changes are needed, but I am working in diminishing returns territory. Incoming tomorrow. (edit again! tomorrow arrived, and it's better now)


references/scripts/create-reference-json.R
```diff
@@ -156,18 +167,54 @@ create_ref_entry <- function(
...
       )
-    )
+    ) |>
+      purrr::keep_at(include_techs)
+
+    # add to json
+    if (length(visium_probe_files) > 0) {
+      json_entry$visium_probe_files <- visium_probe_files
+    } else {
+      json_entry$visium_probe_files <- ""
```
jashapiro: I really want these to be empty objects, (`{}`),  but I don't know how to make one in R that won't be stored as `NULL` in the list. I think I would still prefer an empty vector to a string though:
```suggestion
      json_entry$visium_probe_files <- c()
```

Update: I found it! you have to refer to the list element with a single bracket and fill it with a null list:
```suggestion
      json_entry["visium_probe_files"] <- list(NULL)
```

This will then result in an empty object as the value in the JSON output.

jashapiro (commented):
I really wanted to figure out how to get proper empty objects from R into JSON, and I did. It's a bit weird, but satisfying.

references/scripts/create-reference-json.R
```diff
@@ -179,7 +207,7 @@ if (!file.exists(opt$ref_metadata)) {
   stop("ref_metadata file does not exist.")
 }
 
-ref_metadata <- readr::read_tsv(opt$ref_metadata, col_types = "c") |>
+ref_metadata <- readr::read_tsv(opt$ref_metadata) |>
```
jashapiro: I'd rather fix and keep the col_types for the quiet & ensure no numeric weirdness
```suggestion
ref_metadata <- readr::read_tsv(opt$ref_metadata, col_types = list(.default = "c")) |>
```

sjspielman:
Status: This now contains flex changes that were previously in https://github.com/AlexsLemonade/scpca-nf/pull/1227

I'm going to do some spaceranger test runs with the probe set up outlined here, and then will request review. When we do get around to review, I want to draw attention to this comment: https://github.com/AlexsLemonade/scpca-nf/issues/1226#issuecomment-4082074346 If we want to support additional versions of flex probes, we should probably set up here in preparation for when we add that sooner rather than later.

references/scripts/create-reference-json.R
```diff
@@ -132,17 +155,48 @@ create_ref_entry <- function(
...
+        ))
+      )
+
+    if (nrow(probes_df) == 0) {
+      visium_probe_files <- list(NULL)
+    } else {
+      visium_probe_files <- as.list(probes_df$probe_file) |>
+        purrr::set_names(probes_df$technology)
+    }
+    json_entry$visium_probe_files <- visium_probe_files
```
jashapiro: For the `list(NULL)` trick to work, this has to be filled in with single bracket notation. 
```suggestion
    json_entry["visium_probe_files"] <- visium_probe_files
```

allyhawkins:
> Status: This now contains flex changes that were previously in https://github.com/AlexsLemonade/scpca-nf/pull/1227
> 
> 
> 
> I'm going to do some spaceranger test runs with the probe set up outlined here, and then will request review. When we do get around to review, I want to draw attention to this comment: https://github.com/AlexsLemonade/scpca-nf/issues/1226#issuecomment-4082074346 If we want to support additional versions of flex probes, we should probably set up here in preparation for when we add that sooner rather than later.

We are already set up to support additional versions of flex probes. We encoded the probe version in the technology string for this purpose.

references/scripts/create-reference-json.R
```diff
@@ -132,30 +154,43 @@ create_ref_entry <- function(
...
-  if (include_cytassist) {
-    # TODO: UPDATE THIS DIRECTORY to: file.path(ref_dir, "cytassist")
-    cytassist_probe_dir <- "s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/cytassist-probes"
+  # add directory for flex probes
+  if (include_flex) {
+    json_entry$flex_probe_files <- flex_probe_map |>
+      purrr::pluck(reference_name) |>
+      purrr::map(
+        \(f) file.path(flex_probe_dir, f)
+      )
```
jashapiro: I think this may work? (will need to be applied to visium too)
```suggestion
      )
    # be sure we end up with an empty object in the JSON
    if(length(json_entry$flex_probe_files) == 0) { json_entry['flex_probe_files'] <- list(NULL) }
```

sjspielman: I'm not clear on why this is needed? We already end up with empty fields where there* is no probe with the code as is, e.g. https://github.com/AlexsLemonade/scpca-nf/blob/95eb7c93db10764ace668d5c13924ec4e93c0d0c/references/scpca-refs.json#L16-L17

sjspielman:
This should be ready for real review! Visium probes have all tested with all but the new references, which I will need to come back to once we actually generate the cellranger references from build-index.nf.

For handling probe files, I ended up adding two json files that we can maintain moving forward when we add new probes. These are parsed by the `create-references-json.R` script, and it will make our lives easier moving forward to edit these jsons instead of the script when we add new probes in the future. We also avoid duplicating probe files with this approach. If we agree on using the jsons, I will go ahead and add to `internal_instructions.md` as part of review.

edit - requesting Josh since he's already been looking, Ally also since this started off with her PR and so we can all be on same page for probe handling moving forward.

references/scripts/create-reference-json.R
jashapiro: Those are empty strings, not empty objects. I want them to look like this:
```
 "flex_probe_files": {}, 
 "visium_probe_files": {} 
```

sjspielman: Ah! I got this backwards in my head. Incoming.

sjspielman: This approach didn't end up working because the logic skips `if (include_<visium/flex>)` if there are no probes for the reference, but got it where we want it to be here https://github.com/AlexsLemonade/scpca-nf/pull/1228/commits/559fb313f8b4e94c3ce51742b8ea93fff0098690

jashapiro: Oh, I was assuming these would not exist if `include_*` was false. But it looks like we could just update the defaults here and removed all the `if`s and`else`s   https://github.com/AlexsLemonade/scpca-nf/blob/056e9b435dd8b18b63a72d5456b3f6fb05fe5358/references/scripts/create-reference-json.R#L100-L101

jashapiro: I take it back. Not quite. We still have to use `json_entry['flex_probe_files'] <- list(NULL)`, but we can do it up there after creating the initial list object. Just not in the list creation call.

modules/spaceranger.nf
```diff
@@ -363,7 +364,7 @@ workflow spaceranger_quant{
         ]
     }
     .branch {
-      hd: it[0].technology.contains("_hd")
+      hd: it[0].technology.contains("-hd")
```
jashapiro: for consistency with above
```suggestion
      hd: it[0].technology.contains("visium-hd")
```

references/scripts/create-reference-json.R
```diff
@@ -5,7 +5,15 @@
...
-# `organism`, `assembly`, `version`, `include_salmon`, `include_cellranger`, `include_star`
+# - `organism`
+# - `assembly`
+# - `version`
+# - `include_salmon`
+# - `include_cellranger`
+# - `include_star`
+# - `include_infercnv`
+# - `include_flex`
+# - `include_visium`
```
jashapiro: We probably don't need these if we have the metadata file we can look at to see whether the probe files exist (and I would generally rather have a single source of truth), but I don't think this is worth changing at this time.

references/scripts/create-reference-json.R
```diff
@@ -76,7 +97,8 @@ create_ref_entry <- function(
     star_index = "",
     infercnv_gene_order = "",
     cytoband = "",
-    cytassist_probe = ""
+    flex_probe_files = "",
+    visium_probe_files = ""
   )
```
jashapiro:
```suggestion
)

# set default empty objects for probe files
json_entry["flex_probe_files"] <- list(NULL)
json_entry["visium_probe_files"] <- list(NULL)
```

references/scripts/create-reference-json.R
```diff
@@ -76,7 +97,8 @@ create_ref_entry <- function(
     star_index = "",
     infercnv_gene_order = "",
     cytoband = "",
```
jashapiro: Just removing the comma to prevent errors with my other suggestion.
```suggestion
    cytoband = ""
```

references/scripts/create-reference-json.R
```diff
@@ -132,30 +154,52 @@ create_ref_entry <- function(
...
+      purrr::pluck(reference_name) |>
+      purrr::map(
+        \(f) file.path(flex_probe_dir, f)
+      )
+    if (length(json_entry$flex_probe_files) == 0) {
+      json_entry["flex_probe_files"] <- list(NULL)
+    }
+  } else {
+    json_entry["flex_probe_files"] <- list(NULL)
+  }
```
jashapiro: With defaults already set we shouldn't need any of this.
```suggestion
  }
```

references/scripts/create-reference-json.R
```diff
@@ -132,30 +154,52 @@ create_ref_entry <- function(
...
+  # add directory for visium probes
+  # need the else to ensure it's {} if empty
+  if (include_visium) {
+    json_entry$visium_probe_files <- visium_probe_map |>
+      purrr::pluck(reference_name) |>
+      purrr::map(
+        \(f) file.path(visium_probe_dir, f)
+      )
+  } else {
+    json_entry["visium_probe_files"] <- list(NULL)
```
jashapiro: Should not be needed.
```suggestion
```

references/scripts/create-reference-json.R
```diff
@@ -132,30 +154,52 @@ create_ref_entry <- function(
...
-      glue::glue("{reference_name}_cytassist-v2.1.0.csv")
-    )
+  # add directory for visium probes
+  # need the else to ensure it's {} if empty
+  if (include_visium) {
+    json_entry$visium_probe_files <- visium_probe_map |>
+      purrr::pluck(reference_name) |>
+      purrr::map(
+        \(f) file.path(visium_probe_dir, f)
+      )
```
jashapiro: On second thought, I kind of want to keep a check that we are not adding an empty vector.
```suggestion
    probe_files <- visium_probe_map |>
      purrr::pluck(reference_name) |>
      purrr::map(\(f) file.path(visium_probe_dir, f))
    # only add if non-empty
    if (length(probe_files) > 0) json_entry$visium_probe_files <- probe_files
```

jashapiro (commented):
This looks good overall. I just want to be a bit more sure about my type obsession. I _think_ all my suggestions will make it work, but I can't be sure there isn't a lingering syntax error in there or an introduced bug.

references/scripts/create-reference-json.R
```diff
@@ -132,30 +154,52 @@ create_ref_entry <- function(
...
-    # TODO: UPDATE THIS DIRECTORY to: file.path(ref_dir, "cytassist")
-    cytassist_probe_dir <- "s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/cytassist-probes"
+  # add directory for flex probes
+  # need the else to ensure it's {} if empty
+  if (include_flex) {
+    json_entry$flex_probe_files <- flex_probe_map |>
+      purrr::pluck(reference_name) |>
+      purrr::map(
+        \(f) file.path(flex_probe_dir, f)
+      )
```
jashapiro: Meant to suggest this too for completeness, but never clicked to add the comment.
```suggestion
    probe_files <- flex_probe_map |>
      purrr::pluck(reference_name) |>
      purrr::map(\(f) file.path(flex_probe_dir, f))
    if (length(probe_files) > 0) json_entry$flex_probe_files <- probe_files

Comment (jashapiro) [reply references/scripts/create-reference-json.R]: And of course github didn't put it in the right place, but you get the idea.

File: references/scripts/create-reference-json.R (hunk 6 above)
Comment (sjspielman) [reply references/scripts/create-reference-json.R]: https://github.com/AlexsLemonade/scpca-nf/issues/1231

File: references/scripts/create-reference-json.R (hunk 12 above)
Comment (sjspielman) [reply references/scripts/create-reference-json.R]: saw this comment after I did the thing.
also, thanks for letting me have the pluck ;)

Comment (jashapiro) [reply references/scripts/create-reference-json.R]: It's a good use, since it behaves well when the item may not exist!

File: internal-instructions.md [hunk 13]
```diff
@@ -87,11 +87,14 @@ Inside the `references` folder are files and scripts related to maintaining the
    The columns included specify the `organism` (e.g., `Homo_sapiens`), `assembly`(e.g.,`GRCh38`), and `version`(e.g., `104`) of the `fasta` obtained from [Ensembl](https://www.ensembl.org/index.html) that was used to build the reference files.
    This file is used as input to the `build-index.nf` workflow, which will create all required index files for `scpca-nf` for the listed organisms in the metadata file, provided the `fasta`, `gtf`, and `cytoband` (used to build `inferCNV` gene order files) files are stored in the proper location on S3.
 
-2. `scpca-refs.json`: Each entry of this file contains a supported reference for mapping with `scpca-nf` and the name used to refer to that supported reference, e.g., `Homo_sapiens.GRCh38.104`.
+2. `flex-probes.json` and `visium-probes.json` track probe set files used by 10X Flex and 10X Visium technologies, respectively.
```
jashapiro: I wouldn't have noticed, but you called attention to it.
```suggestion
2. `flex-probes.json` and `visium-probes.json` track probe set files used by 10x Flex and 10x Visium technologies, respectively.
```

sjspielman: I accept responsibility.

references/scripts/create-reference-json.R
sjspielman: This turned out to be a very small change. I went ahead and implemented it here, but can revert if preferred at this time!
https://github.com/AlexsLemonade/scpca-nf/pull/1228/commits/fb60e4263d377cf36e26053c647f45dd4346bf2a

sjspielman:
Just don't want this comment to buried: https://github.com/AlexsLemonade/scpca-nf/pull/1228#discussion_r2965473811

modules/spaceranger.nf
```diff
@@ -382,11 +384,10 @@ workflow spaceranger_quant{
       }
 
     // log error about segmented_outputs as needed
-    // don't log for stub libraries
     spaceranger_hd.out
       .subscribe{ meta, out_dir ->
-        def segmented_dir_exists = file("${out_dir}/outs/segmented_outputs").isDirectory()
-        if (meta.visium_image_type == "image" && !segmented_dir_exists) { 
+       def segmented_exists = out_dir.resolve("outs/segmented_outputs/cell_segmentations.geojson").exists()
```
jashapiro:
```suggestion
        def segmented_exists = out_dir.resolve("outs/segmented_outputs/cell_segmentations.geojson").exists()
```

references/flex-probes.json
```diff
@@ -0,0 +1,6 @@
+{
+  "Homo_sapiens.GRCh38.110": {
+    "10xflex_v1.1_single":   "Chromimum_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv",
```
jashapiro:
```suggestion
    "10xflex_v1.1_single": "Chromimum_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv",
```

jashapiro (approved):
LGTM with only a few spacing suggestions

The one other thing I might suggest is to add in all of the flex probe sets (mouse and human). I don't see why we would not want to support those, and adding the files to our references should be straightforward now (You can do this in a separate PR.)

allyhawkins (approved):
Apologies for the late review, I must have missed this while we are at C8 but LGTM

PR #1233, opened 2026-03-20
Check reference file

### LINKED ISSUE #1232: Add GHA to check that all references in scpca-refs.json exist
When reviewing https://github.com/AlexsLemonade/scpca-nf/pull/1228, I realized we should probably have a check that all of the reference files we include exist at our default reference location. 

We should write a small shell (jq + awscli) or python script to check that all of the reference paths exist. This will trigger on any changes to `scpca-refs.json`

jashapiro:
Closes #1232 

Here I am adding a little action to check that the reference files we have in the default references file exist. 

I did this all with a shell script because I wanted it fast and to work with pre-installed tools.

### DIFF
--- a/.github/workflows/refs-check.yaml
+++ b/.github/workflows/refs-check.yaml
@@ -1,0 +1,106 @@
+name: Check reference paths
+
+on:
+  pull_request:
+    branches:
+      - development
+      - main
+    paths:
+      - ".github/workflows/refs-check.yaml"
+      - "references/scpca-refs.json"
+
+concurrency:
+  # only one run per branch at a time
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  refs-check:
+    runs-on: ubuntu-latest
+    name: Check reference paths
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Check that reference paths exist on S3
+        env:
+          REF_JSON: references/scpca-refs.json
+          REF_CONFIG: config/reference_paths.config
+        run: |
+          # Get the default reference location from the config file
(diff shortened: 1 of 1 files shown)


jashapiro:
This is failing tests, but it seems to be correctly doing so at the moment, so I am going to request review. We can either merge it and fix the files as #1228 goes in or wait for #1228 and fix any remaining refs as part of this.

.github/workflows/refs-check.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+on:
+  pull_request:
+    branches:
+      - development
+      - main
+    paths:
+      - ".github/workflows/refs-check.yaml"
+      - "references/scpca-refs.json"
+      - "config/reference_paths.config"
```
sjspielman, line 11: I think we have to run this manually or run this as part of https://github.com/AlexsLemonade/ScPCA-admin/blob/main/.github/workflows/build-references.yaml, or else it's doomed to fail?

In our current setup, the `references/scpca-refs.json` will always get updated with new reference files before they get added to S3, since we have set up to merge `scpca-refs.json` before running `build-index.nf` for prod (staging here).

https://github.com/AlexsLemonade/ScPCA-admin/blob/af2380a803aac93203496eff4b671048f8cb3428/data-processing-and-handoff/README.md?plain=1#L539-L540

jashapiro: Maybe we should make it fire only on merge to main?

sjspielman: > Maybe we should make it fire only on merge to main?

Yeah, that makes sense. I saw [this template](https://github.com/peter-evans/repository-dispatch?tab=readme-ov-file#dispatching-an-event-to-a-remote-repository) which seems interesting to use too but it wouldn't work for us since this would get run before the nextflow workflow actually complete. 

So I'd vote for manual (which I can definitely see being helpful for convenient checking after running the build-index workflow), and to `main`. Or instead of manual, we pull out the bash script to run locally, but manual dispatch seems like much less work!

sjspielman (approved):
👍

PR #1234, opened 2026-03-20
Add a space in cellranger multi

allyhawkins:
When working on the docs in https://github.com/AlexsLemonade/scpca-docs/pull/494, I noticed that we use `cellranger-multi` as the `mapping_tool` for flex samples, but the function is actually `cellranger multi`. This PR just adds in the space.


sjspielman (approved):
🐈

PR #1238, opened 2026-03-24
Update example visium data and associated internal instructions

### LINKED ISSUE #1218: Update internal instructions with spatial testing data
We will need to update internal instructions to add a section about testing the spatial workflow, analogous to the section about 10x flex.
We can do this once those are formally uploaded to S3.

sjspielman:
Closes #1218 

This PR updates our metadata for incoming visium examples. I wanted this to be reviewed _before_ I upload to S3, since I want to make sure we're in agreement first what with permissions.

For this PR, I updated the example run metadata TSV with proper information for the 5 visium libraries I have been using to test.  I also updated the internal instructions to match. 

While I was doing this, I cleaned up some numbering as well, which had gotten a bit off with the flex example datasets. On S3, the flex example fastqs live at `run08` and `run09` but the metadata file recorded them as `run09` and `run10`. In addition, the multiplex example tsv includes `library07` for the flex multi, but the run metadata file recorded this library as `library08`. The sample ids all looked correct though.

Rather than move files on S3 (since move is actually copy/delete*), I rearranged ids here a bit, mostly by moving our original visium example run down to the bottom with the other visiums and re-numbering from there. The run ids for flex now match what's on S3, and the multi library id now matches what is in the multiplex tsv. Note also that only other run actually in our example data is `run01`, so this should all be safe. 

*Ok, I did try to move though, since I was naively hopeful that the DataManagement role would stop me running `aws s3 mv` altogether if it didn't like that. I found out the hard way that it won't stop you because it's glad to do the first part and copy, but then fails at deleting. In retrospect this makes sense! Anyways the punchline is that after attempting to move files on S3, I duplicated some fastqs that _now DO need to be removed_ from S3, cc @jashapiro 🙏🙏🙏  . 
```
# remove a couple duplicates of what lives in run09
aws s3 rm s3://scpca-nf-references/example-data/example_fastqs/run10/ --recursive
```


examples/example_run_metadata.tsv
```diff
@@ -1,14 +1,15 @@
...
+run03	library02	sample01	project01	CITEseq_10Xv3	EFO:0030011	cell	Homo_sapiens.GRCh38.104	example_fastqs/run03	example_barcode_files/cite_barcodes.tsv	2[1-15]	NA	NA	NA
+run04	library03	sample03;sample04	project01	10Xv3.1	EFO:0009922	cell	Mus_musculus.GRCm39.104	example_fastqs/run04	NA	NA	NA	NA	NA
+run05	library03	sample03;sample04	project01	cellhash_10Xv3.1	EFO:0030012	cell	Mus_musculus.GRCm39.104	example_fastqs/run05	example_barcode_files/cellhash_barcodes.tsv	2[1-10]	NA	NA	NA
+run06	library04	sample05	project01	paired_end	EFO:0003747	bulk	Mus_musculus.GRCm39.104	example_fastqs/run06	NA	NA	NA	NA	NA
+run07	library05	sample06	project01	10Xv3.1	EFO:0009922	cell	Homo_sapiens.GRCh38.104	example_fastqs/run07	NA	NA	NA	NA	example_metadata_files/submitter_celltypes.tsv
+run08	library06	sample07	project02	10Xflex_v1.1_single	EFO:0022606	nucleus	Homo_sapiens.GRCh38.110	example_fastqs/run08	NA	NA	NA	NA	NA
+run09	library07	sample08;sample09;sample10;sample11	project02	10Xflex_v1.1_multi	EFO:0022606	cell	Homo_sapiens.GRCh38.110	example_fastqs/run09	NA	NA	NA	NA	NA
+run10	library08	sample02	project03	visium1	EFO:0022857	spot	Mus_musculus.GRCm39.104	example_files/run10	NA	NA	C1	V19L01-041	NA
```
jashapiro: According to the listing in the instructions below, this is human?

jashapiro (commented):
I think my main question is whether we need all of these examples? 

The examples we provide are not really meant for full testing, but just to show data formats, so I don't think we necessarily need all of these, especially given the size of the data and that the output is mostly the same as Space Ranger. So maybe one "regular" and one HD?

sjspielman:
> I think my main question is whether we need all of these examples?
>
> The examples we provide are not really meant for full testing, but just to show data formats, so I don't think we necessarily need all of these, especially given the size of the data and that the output is mostly the same as Space Ranger. So maybe one "regular" and one HD?

100% I agree. This was a crux reason I didn't want to be uploading yet, since this seems like a lot of samples! I'm very good with one "regular" (visium2) and one HD. 
Edit - noting I chose an hd 3' since this technology doesnt use a probe file, which gives us a bit more code coverage.

examples/example_run_metadata.tsv
sjspielman: Ah, it was mouse but my instructions were misordered. Either way, we're dropping this example dataset because it's too much.

allyhawkins:
> I think my main question is whether we need all of these examples?
> 
> The examples we provide are not really meant for full testing, but just to show data formats, so I don't think we necessarily need all of these, especially given the size of the data and that the output is mostly the same as Space Ranger. So maybe one "regular" and one HD?

I would agree with this and also take it one step further. The example metadata files themselves are meant to show how to input data of different formats into the run and sample metadata files. We then provide example output for just a regular single cell library for people to explore. That is the only run that gets actively used in our example profile (https://github.com/AlexsLemonade/scpca-nf/blob/80e7bbb78202ba6cb618feab59af433c750fcdc7/config/profile_example.config#L17). 

The reason we have 10x flex example data was for development and have a way to test the workflow. So I think we should have enough example datasets for spatial that allow us to properly test the workflow if we make future changes. I might argue that we should have a regular visium, visium + cytassist, and visium hd then? We should be sure we cover the different configurations that might be possible without being overkill on the amount of data. 

Another question I have is if it's worth adding the links to the outputs for any of the spatial data as an example for users to see what they should expect? I think it could be helpful but also could cause confusion if there are different outputs based on different configurations.

sjspielman:
> I might argue that we should have a regular visium, visium + cytassist, and visium hd then? We should be sure we cover the different configurations that might be possible without being overkill on the amount of data.

Maybe... Let me break it down for clarity and to talk through it.
- With visium2, we hit the main spaceranger process and probe usage
- With hd 3', we hit the spaceranger hd process and _no_ probe usage
- With "regular visium" (the old visium v1 technology), we have an option with and without a probe file, but the code this touches would already be covered with the other examples. 
- The main code piece we're not covering with these potential examples is providing an optional image file, so what we _might_ want is:
  - visium1 sample. The image has to be provided here, since there's only 1! I don't think it matters probe or not, so I'd choose the one with smaller size files.
  - visium2 but _don't_ provide the extra image, only provide the cytassist image. This will have two effects, one I think is helpful but one I worry is not: a) helpful: coverage to test code, but b) maybe unhelpful: it might be misleading as an example
  - visium hd 3'

> Another question I have is if it's worth adding the links to the outputs for any of the spatial data as an example for users to see what they should expect? I think it could be helpful but also could cause confusion if there are different outputs based on different configurations.

This strikes me as something to maybe do later if/when we get more spatial onto the portal. At this time, the people most likely to be running spatial libraries through scpca-nf would be doing so in the context of contributing to the portal, and they have this information in the contribution guidelines.

allyhawkins:
> With hd 3', we hit the spaceranger hd process and no probe usage

I thought hd 3' was a cytassist technology so has probes? Or are you saying to use an example dataset without cytassist? 

I was just looking through the code more and I think it's going to be hard to capture every single configuration without storing a lot of data, so I think I'm fine with making sure we are at least just testing the `spaceranger` and `spaceranger_hd` process. So what you have here is probably sufficient for now and we can always add more datasets if we find we want to test more configurations in the future.

sjspielman:
> I thought hd 3' was a cytassist technology so has probes? Or are you saying to use an example dataset without cytassist?

Sadly no. You'd think so, but it's a polyA protocol, so no probes are ~needed~ used for 3' hd. But _yes_ probes for regular hd which does the hybridization. This is why we need both these lists to track technologies in the code 🫠 
https://github.com/AlexsLemonade/scpca-nf/blob/da00b35aecd1d8e0c5a2d9f1b50de0bfd2f03933/modules/spaceranger.nf#L246-L250

allyhawkins:
> > I thought hd 3' was a cytassist technology so has probes? Or are you saying to use an example dataset without cytassist?
> 
> Sadly no. You'd think so, but it's a polyA protocol, so no probes are ~needed~ used for 3' hd. But _yes_ probes for regular hd which does the hybridization. This is why we need both these lists to track technologies in the code 🫠
> 
> https://github.com/AlexsLemonade/scpca-nf/blob/da00b35aecd1d8e0c5a2d9f1b50de0bfd2f03933/modules/spaceranger.nf#L246-L250

Sorry I was thinking regular hd not hd 3' (since obviously 3' is in fact 3' and not probe based...). There's too many versions to keep track of...

sjspielman:
> There's too many versions to keep track of...

Can absolutely confirm.

allyhawkins (approved):
LGTM

Issue #1195, opened 2026-02-27
Formatting reference file: Merged objects


allyhawkins:
Once we have the individual object formatting reference for AnnData and SCE objects, we should add the information for the merged objects.

allyhawkins:
When we do this, we should also be sure to update the documentation to include any relevant info.

allyhawkins:
Closed by #1304

PR #1240, opened 2026-03-25
Document formatting references

### LINKED ISSUE #1210: Document formatting reference decisions
We should include some documentation around the decisions made for tracking formatting of files. When we do this we should also address this comment: 

Generally, I just really like tracking when we use a dictionary associated with different sce processing level (unfiltered, filtered, processed) since it seems confusing (or like a potential bug when it's not!) on its face.

We have quite a bit of this throughout, and I like that we added this comment in the last PR:

https://github.com/AlexsLemonade/scpca-nf/blob/d5aafade062a5e6eb236c0fbdfbd6e049e3ef9c2/references/scripts/create-formatting-reference.py#L287

I do think it would be a helpful move to add comments when dictionaries are defined that are used beyond their processing level and we'll appreciate this bookkeeping in the long run. For example when `unfiltered_experiment_metadata_conditional` is defined, we could have a comment like "also used to check altexp metadata at all processing levels".

**That said** I would do this for the existing dictionaries as a separate/standalone PR - easier to tackle and review this on its own, but when future ones are defined we can add comments as we go and they start to get used.

----

I also noticed this comment which at first I thought was in a similar vein, but is it outdated?
https://github.com/AlexsLemonade/scpca-nf/blob/d5aafade062a5e6eb236c0fbdfbd6e049e3ef9c2/references/scripts/create-formatting-reference.py#L178

_Originally posted by @sjspielman in https://github.com/AlexsLemonade/scpca-nf/pull/1202#discussion_r2911635995_

allyhawkins:
Closes #1210 

Here I'm just adding a little bit of docs about the reference files for checking formatting. I included a brief description of the files and then some instructions for how to update them. Let me know if anything is unclear or if I should add more information. 

I also added some comments throughout about dictionaries that are used in more than one object type.


references/README.md
```diff
@@ -115,3 +115,23 @@ The table includes the following columns:
 | --- | --- |
 | `diagnosis_group` | Broad diagnosis group |
 | `celltype_groups` | Consensus cell type validation groups to include in an `inferCNV` normal reference for samples of the given broad diagnosis group  |
+
+### Formatting reference files 
+<!--TODO: Update these when we include merged reference-->
```
sjspielman, line 120: There's an existing issue for adding info about merged, but not one specifically about docs, if you wanted to add one for yourself. Your call.

references/README.md
```diff
@@ -115,3 +115,23 @@ The table includes the following columns:
...
+
+* `sce-formatting-reference.json` specifies formatting for all `SingleCellExperiment` objects
+* `anndata-formatting-reference.json` specifies formatting for all `AnnData` objects
+
+#### Updating the formatting reference
+
+If a new item is to be added to the objects output from `scpca-nf`, such as adding a new column in the `colData`, the formatting files will need to be updated. 
+
+1. Add the appropriate information to the `SCE` or `AnnData` dictionary at the top of `create-formatting-reference.py`. 
+  * Note that any item that will only be present under certain conditions, should be included in the `_conditional` dictionary, e.g., `colData_conditional`.
```
sjspielman, line 133: this suggestion will shock you! ;)

```suggestion
  * Note that any item that will only be present under certain conditions, should be included in the `_conditional` dictionary, e.g., `colData_conditional`. 
  * If an item will be used in multiple dictionaries, indicate this in a comment, e.g., "used in both filtered and processed" 
```

sjspielman (commented):
Thanks for doing this! LGTM, just a couple small comments but I don't need to see this again.

I found one more spot we might want a comment - this one is also used for filtered & processed _adt_
https://github.com/AlexsLemonade/scpca-nf/blob/0facb6e091672d078e6148cf5f744480f48c21af/references/scripts/create-formatting-reference.py#L137

references/README.md
allyhawkins: I made a comment on https://github.com/AlexsLemonade/scpca-nf/issues/1195#issuecomment-4135502776 to do it at the same time.

allyhawkins:
@sjspielman I added the extra comment you noted. Based on your original comment though I think you meant to approve here? Either way, re-requesting review.

sjspielman (approved):
sure did mean to approve! but still looks good!

PR #1241, opened 2026-03-26
switch some to log.warn

### LINKED ISSUE #1211: Ensure log warn/error is consistent across the workflow
As long as these are not actually killing the workflow, I think either is fine. But I might lean toward error since the result is not processing a sample and we want to call that out somewhat louder (we may or may not do that universally, but we should aim consistency)

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/1205#discussion_r2911952718_

----

The goal of this issue is to have a specific look at how we use log.warn and log.error throughout the workflow: https://github.com/search?q=repo%3AAlexsLemonade%2Fscpca-nf+log.&type=code

Then, pick a universal strategy and apply it as needed. We should also our decision to CONTRIBUTING.

sjspielman:
Closes #1211 

This PR changes a couple `log.error`s --> `log.warn`s in the updated spatial code to make it more consistent with the overall workflow. 

Let me know if you disagree with any of these!

edit - also includes a bonus change I just realized we should make to deprecate zenhub.


jashapiro (approved):
LGTM

PR #1242, opened 2026-03-26
Change CellBrowser display to normalized expression

### LINKED ISSUE #1239: Check Cell Browser expression display values
While looking at the Cell Browser with Max form UCSC, the question of whether we were displaying raw counts or lognorm counts came up. We should confirm which we are displaying (and which we want to display) and update the cell browser build accordingly.

jashapiro:
closes #1239
First, thank you to @sjspielman for doing the check to confirm we are showing raw counts!

With a bit of digging, I was able to find the correct flag for import to get the processed data. I also added a config option to display the correct units.

I have not yet had a chance to do a test run, but I will get to that soon!


sjspielman (commented):
Well that certainly was a non-zero amount of digging!!
Confirming I found it it as well:
https://github.com/maximilianh/cellBrowser/blob/ca964f1583f751228b2ccce442e76e02846f051e/src/cbPyLib/cellbrowser/convert.py#L588-L589

And I confirmed that attribute
```
>>> adata=anndata.read_h5ad("SCPCL000001_processed_rna.h5ad")
>>> adata.raw
<anndata._core.raw.Raw object at 0x1371e6b10>
```

sjspielman (approved):

jaclyn-taroni:
Subtle!

jashapiro:
Confirming that the new browser displays logcounts, complete with updated units
Screenshot 2026-03-26 at 10 18 41 AM

PR #1243, opened 2026-03-26
Update where example data lives

sjspielman:
As briefly discussed on Slack, this PR rearranges some S3 prefix bits so to use:

- `s3://scpca-nf-references/examples/` as top-level for example things
- `s3://scpca-nf-references/examples/example-data/` to store the data files themselves

Edit - noting also I added a TODO about uploading one remaining metadata file to S3. Just looking for someone to agree before I do elevated permissions things!


allyhawkins (approved):
LGTM

PR #1244, opened 2026-03-26
Remove example todo

sjspielman:
Wrapping up something from #1243 - I removed that TODO because I did the thing. You can see it at both:

- `s3://scpca-nf-references/examples/example_diagnosis_celltypes.tsv`
- `s3://scpca-nf-references/example-data/example_diagnosis_celltypes.tsv`


sjspielman:
Hm. Looks like we've got some other problems. Glad I filed this I guess!

sjspielman:
I'm a little confused about some of these files that got flagged (edit: but unconfused myself while writing the comment!):
```
Reference file s3://scpca-nf-references/homo_sapiens/ensembl-110/annotation/Homo_sapiens.GRCh38.110.mitogenes.txt does not exist
Reference file s3://scpca-nf-references/homo_sapiens/ensembl-110/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.fai does not exist
Reference file s3://scpca-nf-references/homo_sapiens/ensembl-110/flex-probe-refs/Chromimum_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv does not exist
Reference file s3://scpca-nf-references/homo_sapiens/ensembl-98/annotation/Homo_sapiens.GRCh38.98.mitogenes.txt does not exist
Reference file s3://scpca-nf-references/mus_musculus/ensembl-110/annotation/Mus_musculus.GRCm39.110.mitogenes.txt does not exist
Reference file s3://scpca-nf-references/mus_musculus/ensembl-98/annotation/Mus_musculus.GRCm38.98.mitogenes.txt does not exist
```

The mitogenes files I would have expected to be there, e.g. homo sapiens ensembl 110 is not a new reference but it was one presumably added for flex. The ones for ensembl=98 (both species) are new, and I would have expected these files to have been created - https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/5wlesRmvb5qPnh/tasks. But indeed like `generate_reference` wasn't run there, which looking at `build-index.nf` makes sense since don't have salmon turned on for these references. We either I guess need to turn on salmon for references we added for flex & visium, or we need to change the index workflow. The latter is maybe what we actually want, but the former is going to get us to the finish line faster. tagging @jashapiro @allyhawkins 

The probe set probably needed to be directly placed but wasn't yet.

sjspielman:
I guess we can also update the check to not check for mitogenes if salmon is off too.

jashapiro:
> I guess we can also update the check to not check for mitogenes if salmon is off too.

I think we want to have mitochondrial gene lists for everything, so I would just add the salmon indices if that gets the job done.

sjspielman:
Noting I got the probes in place and I added the fasta index (although I'm not sure these are actually used, so we may not need to be checking them in the first place? separate situation, though) that was missing.

mitofiles will land probably tomorrow.

allyhawkins:
> Noting I got the probes in place and I added the fasta index (although I'm not sure these are actually used, so we may not need to be checking them in the first place? separate situation, though) that was missing.

Those are only used when creating the star index, so they aren't needed for the references that don't have that. That's why I didn't add them in for the 110 reference.

jashapiro:
> Those are only used when creating the star index, so they aren't needed for the references that don't have that. That's why I didn't add them in for the 110 reference.

Still not a bad thing to have. Pretty tiny.

jashapiro:
> You can see it at both:
> 
>     * `s3://scpca-nf-references/examples/example_diagnosis_celltypes.tsv`
> 
>     * `s3://scpca-nf-references/example-data/example_diagnosis_celltypes.tsv`

Why both?

sjspielman:
> Why both?

Hm, ok fair. I was thinking to mirror files we need in both `examples/` and `example-data/` to guarantee things work before and after we merge into main. But, main isn't currently looking for this file on S3, only development would be.

sjspielman:
Ah, I realize the only reason the ref-check GHA ran at first was because when I filed this I accidentally set the base to `main`! I fixed it right away, but the GHA had still been kicked off. 
But, I can confirm the files that weren't there are now there:

I also addressed the chromium probe spelling, which notably is not spelling chromimum. 

```
 ➤ aws s3 ls s3://scpca-nf-references/homo_sapiens/ensembl-110/annotation/Homo_sapiens.GRCh38.110.mitogenes.txt
aws s3 ls s3://scpca-nf-references/homo_sapiens/ensembl-110/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.fai
aws s3 ls s3://scpca-nf-references/homo_sapiens/ensembl-110/flex-probe-refs/Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv
aws s3 ls s3://scpca-nf-references/homo_sapiens/ensembl-98/annotation/Homo_sapiens.GRCh38.98.mitogenes.txt
aws s3 ls s3://scpca-nf-references/mus_musculus/ensembl-110/annotation/Mus_musculus.GRCm39.110.mitogenes.txt
aws s3 ls s3://scpca-nf-references/mus_musculus/ensembl-98/annotation/Mus_musculus.GRCm38.98.mitogenes.txt

2026-03-27 10:49:57        592 Homo_sapiens.GRCh38.110.mitogenes.txt
2026-03-26 16:48:42       6406 Homo_sapiens.GRCh38.dna.primary_assembly.fa.fai
2026-03-26 16:23:35    6492395 Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv
2026-03-27 10:45:07        592 Homo_sapiens.GRCh38.98.mitogenes.txt
2026-03-27 10:30:27        703 Mus_musculus.GRCm39.110.mitogenes.txt
2026-03-27 10:33:47        703 Mus_musculus.GRCm38.98.mitogenes.txt
```

main.nf
```diff
@@ -258,6 +258,8 @@ workflow {
       flex: it.technology in flex_techs
     }
 
+  runs_ch.spatial.view()
```
allyhawkins:
```suggestion
```

I assume this is leftover from testing?

sjspielman: nailed it 😬

references/flex-probes.json
```diff
@@ -1,6 +1,6 @@
 {
   "Homo_sapiens.GRCh38.110": {
-    "10xflex_v1.1_single": "Chromimum_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv",
-    "10xflex_v1.1_multi": "Chromimum_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv"
+    "10xflex_v1.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv",
```
allyhawkins, line 3: yikes, glad we have that check!

allyhawkins (approved):
LGTM

PR #1246, opened 2026-03-26
Revert examples change for now

sjspielman:
This PR reverts the changes in #1243. I decided to use `example_fastqs` prefix for visium as well for the time being.
This is a companion PR to https://github.com/AlexsLemonade/ScPCA-admin/pull/1305

Note that the example `diagnosis_celltypes_file` indeed now lives on S3 like the rest. I'll re-upload the run metadata once this goes in.


sjspielman:
I think it's a bit TBD if we're actually reverting this, but I'll request review in case.

sjspielman:
Noting doing this since it wasn't the issue with running example data: https://github.com/AlexsLemonade/ScPCA-admin/pull/1305#issuecomment-4143126786

Issue #1237, opened 2026-03-24
Add remaining flex probe sets & references


sjspielman:
The one other thing I might suggest is to add in all of the flex probe sets (mouse and human). I don't see why we would not want to support those, and adding the files to our references should be straightforward now (You can do this in a separate PR.)

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/scpca-nf/pull/1228#pullrequestreview-4000365755_
            
Edit with tasks:
- [x] add new probes to references jsons
- [x] add new probe files to S3

sjspielman:
Noting probes: https://www.10xgenomics.com/support/flex-gene-expression/documentation/steps/probe-sets/chromium-frp-probe-sets-overview

sjspielman:
Probes have been uploaded to S3:
```
upload: ./Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv to s3://scpca-nf-references/homo_sapiens/ensembl-98/flex-probe-refs/Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv
upload: ./Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv to s3://scpca-nf-references/homo_sapiens/ensembl-98/flex-probe-refs/Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv
upload: ./Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv to s3://scpca-nf-references/mus_musculus/ensembl-110/flex-probe-refs/Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv
upload: ./Chromium_Mouse_Transcriptome_Probe_Set_v1.0_mm10-2020-A.csv to s3://scpca-nf-references/mus_musculus/ensembl-98/flex-probe-refs/Chromium_Mouse_Transcriptome_Probe_Set_v1.0_mm10-2020-A.csv
upload: ./Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv to s3://scpca-nf-references/mus_musculus/ensembl-98/flex-probe-refs/Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv
```

PR #1248, opened 2026-03-27
Add additional flex probes

sjspielman:
Towards #1237 (will be closed once probes have been uploaded to S3 references bucket).

This PR updates the flex probes to include all of the probes that can be used with cellranger9. All the files I added to the references JSON are locally on my computer for upload to S3 after this goes in, but no new references need to be added (aka we don't need to run `build-index.nf`)

To determine which probe sets to get, I used the table at the bottom of this page: https://www.10xgenomics.com/support/flex-gene-expression/documentation/steps/probe-sets/chromium-frp-probe-sets-overview

Some details about this:
- To get older probe sets, I had to download cellranger7 & 8 and instead of being able to grab directly. Harumph. 
- It looks like there's a discrepancy with v1.1.0 probes between species, but there isn't - 
  - The mouse v1.1.0 had a bug so this was immediately patched to v1.1.1. The mouse v1.1.0 is no longer disseminated anywhere it seems. See the [2nd info box here](https://www.10xgenomics.com/support/flex-gene-expression/documentation/steps/probe-sets/chromium-frp-mouse-transcriptome-probe-set-1-1)
  - But, human had no such problem, so there is a human v1.1.0 but there is no human v1.1.1
- Note how I named tech versions (where my shorthand of new = newer genome and old = older genome) - 
  - NEW: `v1` (old human only)  <----- trailing 0 stripped off as we do for flex `v1.1` and visium `v1`, `v2`, `v2.1`
  - NEW: `v1.0.1` (old human & mouse)
  - `v1.1` (new human only)  <---- This is the one we have had already, and we don't expect a patch release for this
  - NEW `v1.1.1` (new mouse only) 
 - Finally, forward-thinking FYIs
   - Worth noting that we'll need to bump to cellranger10 if we get flex v2 data
   - But, if/when we move to cellranger10, `v1.0` and `v1.0.1` probes won't be supported but `v1.1` (used for skc) will be supported
   - Seems fine to put all this in now and circle back later if version changes are needed, since probe files are easy to manage!

### DIFF
--- a/references/flex-probes.json
+++ b/references/flex-probes.json
@@ -1,6 +1,20 @@
 {
+  "Homo_sapiens.GRCh38.98": {
+    "10xflex_v1.0_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
+    "10xflex_v1.0_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
+    "10xflex_v1.0.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv",
+    "10xflex_v1.0.1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv"
+  },
   "Homo_sapiens.GRCh38.110": {
     "10xflex_v1.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv",
     "10xflex_v1.1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv"
+  },
+  "Mus_musculus.GRCm38.98": {
+    "10xflex_v1.0.1_single": "Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv",
+    "10xflex_v1.0.1_multi": "Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv"
+  },
+  "Mus_musculus.GRCm39.110": {
+    "10xflex_v1.1.1_single": "Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv",
+    "10xflex_v1.1.1_multi": "Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv"
   }
 }
--- a/references/flex-probes.json
+++ b/references/flex-probes.json
@@ -1,7 +1,7 @@
 {
   "Homo_sapiens.GRCh38.98": {
-    "10xflex_v1.0_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
-    "10xflex_v1.0_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
+    "10xflex_v1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
+    "10xflex_v1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
     "10xflex_v1.0.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv",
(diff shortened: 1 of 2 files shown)


references/flex-probes.json
```diff
@@ -1,6 +1,20 @@
...
+    "10xflex_v1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
+    "10xflex_v1.0.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv",
+    "10xflex_v1.0.1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv"
+  },
   "Homo_sapiens.GRCh38.110": {
     "10xflex_v1.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv",
     "10xflex_v1.1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv"
+  },
+  "Mus_musculus.GRCm38.98": {
+    "10xflex_v1.0.1_single": "Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv",
```
allyhawkins, line 15: Why is there no `v1` here? It looks like there should be?

allyhawkins (commented):
This looks good, I just had one question about the mouse probes for mm10-2020-A. It looks like there should be `v1.0` and `v1.0.1`?

references/flex-probes.json
sjspielman: I know, but I couldn't find it! It's not disseminated with cellranger 7.0, and no v1s are disseminated with cellranger 7.1. Not on the cellranger downloads page, either.

```
➤ pwd                                                                              
/Users/sjspielman/Downloads/cellranger-7.0.0
➤ ls probe_sets/*                                                                  
probe_sets/Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv


➤ pwd                                                                              
/Users/sjspielman/Downloads/cellranger-7.1.0
➤ ls probe_sets/*                                                                  
probe_sets/Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv
probe_sets/Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv
```

But now, I was able to track it down now with a URL directly: `https://cf.10xgenomics.com/supp/cell-exp/probeset/Chromium_Mouse_Transcriptome_Probe_Set_v1.0_mm10-2020-A.csv`. This file has these comments at the top so this should be it:
```
#probe_set_file_format=1.0
#panel_name=Chromium Mouse Transcriptome Probe Set
#panel_type=predesigned
#reference_genome=mm10
#reference_version=2020-A
```

sjspielman:
v1 mouse has now been ~founded~ found* and added to the jsons

allyhawkins (approved):
LGTM! Thanks for digging up the mouse v1 probes.

references/flex-probes.json
jashapiro: I don't think there is any reason to include the `1.0` version here; I think we could designate the file as `10xflex_v1_*` but use the 1.0.1 file, as the change is purely the addition of a column. Similarly, we could use `10xflex_v1.1_*` where we have the 1.1.1 file, as there is no reason to keep the 1.1.0 version, which had a error: https://www.10xgenomics.com/support/flex-gene-expression/documentation/steps/probe-sets/chromium-frp-probe-set-files.

Essentially, we really only need the tag to reflect the probe set used, not bug fix changes to the annotation, and 10x seems to be doing a pretty good job with semantic versioning. As long as we record the actual file used, which I think we do, we should be fine to stick a single version for 1.0 and 1.1 within this file, which will simplify the technology naming a bit, with just one version of the technology string that goes along with each annotation version.

sjspielman: @jashapiro I'm not precisely following what you are asking for here since it seems like several items and/or some thoughts in progress. From what I can glean, I will say:
- I disagree about using the `10xflex_v1.1_` designation _precisely because of the bug_ and being explicit about the patch version makes the probe set entirely unambiguous. It's possible there is archival data where someone used 1.1.0 and we would not want to have any confusion here.
- I don't mind removing the mouse v1.0 and consolidating with just v1.0.1, if that is what you are getting at? But the fact that I'm not precisely sure what you're getting at with `1.1` is kind of my previous point - we need to be sure. 

Either way, it would be helpful if you could open a fresh issue since this comment is in a closed/merged PR and it sounds like you are requesting changes to block a release that I was planning on making today.

Issue #1236, opened 2026-03-24
Prepare for scpca-nf release `v0.10.0`


sjspielman:
### Preparing for the release

- [x] Are all of the issues planned for this release resolved? If there are any issues that are unresolved, mark this issue as blocked by those in the Science Team GitHub Project.
- [x] Update code and documentation with the latest version number in the `development` branch:
  - [x] [`nextflow.config`](https://github.com/AlexsLemonade/scpca-nf/blob/main/nextflow.config)
  - [x] [`internal-instructions.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/internal-instructions.md)
  - [x] [`external-instructions.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md)
  - [x] [`sb_nextflow_schema.yaml`](https://github.com/AlexsLemonade/scpca-nf/blob/main/cavatica/sb_nextflow_schema.yaml) and [`sb_doc.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/cavatica/sb_doc.md)
- [x] Update the tag for `scpca-nf` containers in `config/containers.config` and `nextflow-schema.json`
  _Note:_ Any processes that use those containers will not work until the new tag has been created.
- [x] Check that the `nextflow-schema.json` is up to date with `pixi run nf-core pipelines schema build`
- [x] Check that [`sb_nextflow_schema.yaml`](https://github.com/AlexsLemonade/scpca-nf/blob/main/cavatica/sb_nextflow_schema.yaml) entries are up to date (sadly, no automated tool for this)
- [x] Test that the workflow is in good working order by running the `development` branch through the workflow in `testing` mode
- [x] File a PR from the `development` branch to the `main` branch. This should include all of the changes that will be associated with the next release.
- [x] (Optional) Generate new example `scpca-nf` output files.
If updating the example output is not necessary for this release, check these boxes off for free.
  - [x] [Re-process the example data](https://github.com/AlexsLemonade/scpca-nf/blob/main/internal-instructions.md#processing-example-data) through the `scpca-nf` workflow and ensure it looks correct.
  - [x] Compress the example output in `scpca_out` to create `scpca_out.zip`, as described in the instructions, and ensure the file is set to public and read-only.

### Creating a release
- [ ] On the [releases page](https://github.com/AlexsLemonade/scpca-nf/releases), choose `Draft a new release`.
[...]

PR #1253, opened 2026-03-31
Bump version for release

sjspielman:
Towards #1236 

This PR updates v0.9.3 -> v0.10.0 in a variety of spots to prep for* release. I also updated the release template since apparently version info is no longer in internal-instructions.

Also, it looks to me like we need an `scpcaTools` release, so I will do that before updating the schema files - just want to confirm you agree! We did some pullthrough stuff. https://github.com/AlexsLemonade/scpcaTools/compare/v0.4.4...main

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-checklist.md
+++ b/.github/ISSUE_TEMPLATE/release-checklist.md
@@ -14,9 +14,9 @@ assignees: ''
 - [ ] Are all of the issues planned for this release resolved? If there are any issues that are unresolved, mark this issue as blocked by those in the Science Team GitHub Project.
 - [ ] Update code and documentation with the latest version number in the `development` branch:
   - [ ] [`nextflow.config`](https://github.com/AlexsLemonade/scpca-nf/blob/main/nextflow.config)
-  - [ ] [`internal-instructions.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/internal-instructions.md)
   - [ ] [`external-instructions.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md)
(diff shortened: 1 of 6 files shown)


allyhawkins:
> Also, it looks to me like we need an `scpcaTools` release, so I will do that before updating the schema files - just want to confirm you agree! We did some pullthrough stuff. [AlexsLemonade/scpcaTools@v0.4.4...main](https://github.com/AlexsLemonade/scpcaTools/compare/v0.4.4...main)

All of those changes look to be just related to the GHA so don't actually affect the docker files themselves. The changes that do matter are the `renv` changes, but I don't know that we need to include those here? If you do want to do that release, you'll have to test the workflow with `edge` first to make sure the new `renv` changes don't break anything. But again, I don't think it's totally necessary.

sjspielman:
> If you do want to do that release, you'll have to test the workflow with edge first to make sure the new renv changes don't break anything. 

Fair point about GHAs not being relevant! But since we actually been running with `edge` in `development` (which we've heavily run during infra testing and spaceranger updates!) is also why I thought we should bump `scpcaTools` https://github.com/AlexsLemonade/scpca-nf/blob/development/config/containers.config#L8

allyhawkins:
> But since we actually been running with edge in development (which we've heavily run during infra testing and spaceranger updates!) is also why I thought we should bump scpcaTools

Makes sense to me then!

sjspielman:
@allyhawkins this is ready for a real review now. I also bumped `scpcaTools` docker containers to 0.4.5. I'm still investigating the snapshot weirdness between renv and non-renv R CMD CHECK runs, but I'm not so worried about it that I think it should hold something up and probably we don't need to "fix" 0.4.5 either.

allyhawkins (approved):
LGTM

PR #1255, opened 2026-04-03
Update the schema file and other pre-release miscellany

sjspielman:
Towards #1236

This PR updates the nextflow schema file for the new release. I also fixed a few more release template spots (mostly change a typo `-` to `_`).

I looked over the cavatica schema yaml as well but I didn't see any spots to change. As part of review it would be helpful if you could have a glance and confirm that you agree no changes were needed there: https://github.com/AlexsLemonade/scpca-nf/blob/development/cavatica/sb_nextflow_schema.yaml

After this, I'll proceed to test the workflow and file a PR from `development` -> `main` for releasing early next week. Based on the [full diff](https://github.com/AlexsLemonade/scpca-nf/compare/main...development?expand=1), I don't think the example output needs to be re-run. But, when looking at this diff, I did catch one spot we need to fix - I didn't account in #1254 for this new tech string in `main.nf`, so that is also done here along with a comment because it really seems like a typo otherwise 🥴


allyhawkins (approved):
LGTM and confirmed the Cavatica schema does not need to be updated.

Issue #1257, opened 2026-04-06
Simplify flags for flex technologies (only specify minor versions, not patches)


jashapiro:
@jashapiro I'm not precisely following what you are asking for here since it seems like several items and/or some thoughts in progress. From what I can glean, I will say:
- I disagree about using the `10xflex_v1.1_` designation _precisely because of the bug_ and being explicit about the patch version makes the probe set entirely unambiguous. It's possible there is archival data where someone used 1.1.0 and we would not want to have any confusion here.
- I don't mind removing the mouse v1.0 and consolidating with just v1.0.1, if that is what you are getting at? But the fact that I'm not precisely sure what you're getting at with `1.1` is kind of my previous point - we need to be sure. 

Either way, it would be helpful if you could open a fresh issue since this comment is in a closed/merged PR and it sounds like you are requesting changes to block a release that I was planning on making today.

_Originally posted by @sjspielman in https://github.com/AlexsLemonade/scpca-nf/pull/1248#discussion_r3039480061_

I think my main point is that the tags we use for technologies and the precise files we use for processing need not be the same thing. The reason the 1.0.0 version was hard to find is because it should never be used for new processing. The files used for processing are recorded in the metadata, and that should be the source of truth for archival, not the tags we use. 

The discussion about 1.1 is similar: if someone is specifying `10xflex_v1.1` for technology, then the mouse version used for processing should be `v1.1.1`. There is no condition where we should be using the `v1.1.0` file, and therefore no reason to have both a `v1.1` tag and a `v1.1.1` tag.

sjspielman:
> I think my main point is that the tags we use for technologies and the precise files we use for processing need not be the same thing. The reason the 1.0.0 version was hard to find is because it should never be used for new processing. The files used for processing are recorded in the metadata, and that should be the source of truth for archival, not the tags we use.

This I'm on board with.

> The discussion about 1.1 is similar: if someone is specifying 10xflex_v1.1 for technology, then the mouse version used for processing should be v1.1.1. There is no condition where we should be using the v1.1.0 file, and therefore no reason to have both a v1.1 tag and a v1.1.1 tag.

[...]

PR #1256, opened 2026-04-03
Merge development into main for release

sjspielman:
Towards #1236 

This PR merges `development` into `main` for the `v0.10.0` release. I'm filing this on a Friday, but don't want to merge this in until next week since I don't want to do the release today and don't want to be tempted ;) 

Still if folks who are in today have time to review, feel free!

PR #1258, opened 2026-04-06
Simplify probe tags

jashapiro:
Since the change is minor, here is a draft PR to address the changes I am suggesting in #1257. 

I assume there are some docs changes that would also be required, but I wanted the specific changes I was suggesting to be clear.

### DIFF
--- a/references/flex-probes.json
+++ b/references/flex-probes.json
@@ -1,22 +1,18 @@
 {
   "Homo_sapiens.GRCh38.98": {
-    "10xflex_v1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
-    "10xflex_v1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
-    "10xflex_v1.0.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv",
-    "10xflex_v1.0.1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv"
+    "10xflex_v1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv",
+    "10xflex_v1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv"
   },
   "Homo_sapiens.GRCh38.110": {
     "10xflex_v1.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv",
     "10xflex_v1.1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.1.0_GRCh38-2024-A.csv"
   },
   "Mus_musculus.GRCm38.98": {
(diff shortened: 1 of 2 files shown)


references/flex-probes.json
```diff
@@ -1,22 +1,18 @@
 {
   "Homo_sapiens.GRCh38.98": {
-    "10xflex_v1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
-    "10xflex_v1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0_GRCh38-2020-A.csv",
-    "10xflex_v1.0.1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv",
-    "10xflex_v1.0.1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv"
+    "10xflex_v1_single": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv",
+    "10xflex_v1_multi": "Chromium_Human_Transcriptome_Probe_Set_v1.0.1_GRCh38-2020-A.csv"
```
sjspielman, line 4: I'm on board with this and the mouse v1 change. Once we finalize here I would like to add some corresponding docs here to both explain decisions made and support ufu: https://github.com/AlexsLemonade/ScPCA-admin/blob/main/data-processing-and-handoff/README.md#genome-annotations-and-probe-files

Mostly, I would like to note that we use version 2 of probe file formats instead of version 1 where available.

references/flex-probes.json
```diff
@@ -1,22 +1,18 @@
...
-    "10xflex_v1.0.1_single": "Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv",
-    "10xflex_v1.0.1_multi": "Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv"
+    "10xflex_v1_single": "Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv",
+    "10xflex_v1_multi": "Chromium_Mouse_Transcriptome_Probe_Set_v1.0.1_mm10-2020-A.csv"
   },
   "Mus_musculus.GRCm39.110": {
-    "10xflex_v1.1.1_single": "Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv",
-    "10xflex_v1.1.1_multi": "Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv"
+    "10xflex_v1.1_single": "Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv",
+    "10xflex_v1.1_multi": "Chromium_Mouse_Transcriptome_Probe_Set_v1.1.1_GRCm39-2024-A.csv"
```
sjspielman, line 16: 👍 
For posterity, we have discussed that this isn't an issue if someone used v1.1.0 mouse in their experiment because the bug fixed in this patch was just that gene mapping to probes was fixed. The probes themselves are the same so using this for mapping won't be an issue if 1.1.0 was used.

sjspielman (approved):

sjspielman:
Once this goes in, I will update the branch in #1256 to merge it all into `main`, and then release.

PR #1259, opened 2026-04-06
remove AWS batch scripts and config

jashapiro:
Related to  https://github.com/AlexsLemonade/ScPCA-admin/issues/1329

Here I am removing the scripts  and config associated with AWS CodeDeploy from this repository as we now manage these elsewhere. I left the nextflow-runner login script as I can imagine internal folks might still want to use it while working in this repo, but could remove that as well, if desired.

The submission GHA was already removed, so that should be fine.


sjspielman (approved):
> I left the nextflow-runner login script as I can imagine internal folks might still want to use it while working in this repo, but could remove that as well, if desired.

I lean towards removing this, since one copy is better to maintain than two copies. Our processing/handoff instructions in `scpca-admin` only point to the one in that repo, and it's enough to `cd` there as needed. But if you personally enjoy having this for your own muscle memory I don't care much.

sjspielman:
@jashapiro is this ready to go in? If so, I'll make the release tomorrow morning. Thanks!

jashapiro:
Yeah, sorry, I thought I had pushed the removing the second part as well. Done now.

Issue #1265, opened 2026-04-13
Consider updating workflow when there are too few cells for celltyping


sjspielman:
Related to https://github.com/AlexsLemonade/scpca-nf/issues/1264

This is a placeholder issue for potentially circling back to later.
There are some cases (too few cells) when CellAssign & SCimilarity cell type processes should probably be skipped entirely. This is unlikely to affect most libraries, but it may be helpful not enter those processes if we don't need to run those methods. This issue tracks revisiting this.

sjspielman:
Noting this comment where the need 30 cells for cellassign came from: https://github.com/AlexsLemonade/scpca-nf/pull/1266#discussion_r3080312299

Issue #1261, opened 2026-04-07
Track down error in filter process


sjspielman:
While processing new samples with `v0.10.0` I encountered a bug. Still looking into it but wanted to post the context.
First, here's the seqera run: https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/4pQsWG6xCkiNP8/logs

The failed sample was `SCPCL001234` (`SCPCS000860`) which ran through previous versions of scpca-nf fine but is not currently with an I/O error reading in the filtered file. 

### Checkpoints

I confirmed these were properly copied over to staging:
```
 ➤ aws s3 ls s3://ccdl-scpca-nf-external-input/portal/wernig-zorc_2025-09-29/scpca_out/checkpoints/rad/SCPCL001234/SCPCR001234-rna/
                           PRE alevin/
                           PRE aux_info/
                           PRE logs/
2026-02-02 13:56:58        321 cmd_info.json
2026-02-02 13:56:58   11790695 map.rad
2026-02-02 13:56:59       2154 scpca-meta.json
2026-02-02 13:56:59   34259292 unmapped_bc_count.bin
➤ aws s3 ls s3://ccdl-scpca-nf-results-staging/checkpoints/rad/SCPCL001234/SCPCR001234-rna/
                           PRE alevin/
                           PRE aux_info/
                           PRE logs/
2026-04-07 11:22:09        321 cmd_info.json
2026-04-07 11:22:09   11790695 map.rad
2026-04-07 11:22:09       2154 scpca-meta.json
2026-04-07 11:22:10   34259292 unmapped_bc_count.bin
```                                   

### Workdir

This process was in the workdir `s3://ccdl-scpca-nf-workdir/work/d6/e3bf14dc18ff591616f63219586bf7/`. The `SCPCL001234_unfiltered.rds` file present there has 0 cells in the first place which we don't expect given its previous successful runs in earlier versions (see here: `s3://ccdl-scpca-nf-external-input/portal/wernig-zorc_2025-09-29/scpca_out/results/SCPCP000027/SCPCS000860/`). 

The immediate error is probably coming from the droplets script when trying to read in a size 0 filtered file; we should fix this either way, but there's definitely an upstream bug that led to this having 0 cells. This issue tracks finding it. It's worth noting that other libraries from this project ran through the filter process without this error.

sjspielman:
[...]

PR #1266, opened 2026-04-14
Account for too few cells to run scimilarity

### LINKED ISSUE #1264: Address bug when there are too few cells for cell typing
We have encountered issues running the library SCPCL001295 through the pipeline. Here's a run down of what happened:

1. The unfiltered file is fine - plenty of _droplets_, but it gets filtered down to 82 cells in the filtered process
2. Then, it goes down to 1 cell in post processing with this filtering (no miQC on board for this one):
https://github.com/AlexsLemonade/scpca-nf/blob/ccb3970d64a77d49cb26c385a8beb132e29c1afa/bin/post_process_sce.R#L86-L90

As a consequence of having only 1 cell, cell typing doesn't work for CellAssign or SCimilarity (but SingleR does work). To address this:
- We'll need to impose some kind of minimum to head into certain cell typing processes, or have more robust checks in add_celltypes_to_sce
  - I'd say at least 50 cells for some amount of robustness with CellAssign, but we can do lower numbers probably to technically run _something_. 
- We also need to ensure there's no attempt to assign consensus cell types if all 3 methods weren't successfully run. We could devise a different approach when e.g. only 2/3 methods succeeded but I have a pretty decent anti-preference for that since it seems like too much investment for something that might mislead, and also docs could get confusing. 

Noting that we may wish to handle this library differently internally, but we would like the workflow not to fail if/when these cases with too few cells for celltyping are encountered.

sjspielman:
Closes #1264 

This PR makes several changes to `scpca-nf`, including:

- Following our threshold of 30 cells for CellAssign, I updated the SCimilarity script to only run if there are at least 30 cells. This remains kind of an arbitrary number but seems fine at least for now; can always revisit as part of #1265 
- This led to a variety of other small fixes!
  - I updated `add_celltypes_to_sce.R` to add "Not run" for scimilarity in same way we do for cellassign
  - I noticed that `add_celltypes_to_sce.R` also had some leftover `has_cellassign` variables that are no longer used to determine if we want consensus cell types, as we have switched over to using `length(automated_methods) > 1` to determine if we should do this. I removed those unused variables.
  - I updated to the QC report since we have to check for "Not run" for scimilarity as well - this involved some slight function modification to make our existing function that checks for cellassign be more general to be used for scimilarity
  - I added an if to not print a sentence about "go see the cell type report" if we're not actually rendering the cell type report - in the `sce_qc_report.R` script we only render that report if there's >1 cell, so I mirrored that in the report 
  - I caught two more bugs:
    - We need to have `dotplot_width` & `dotplot_height` defined as defaults in case of no consensus
    - There was a typo in the infercnv reporting section. We had duplicated one of the status strings that should have been "no_consensus" in the if checks, fixed!
- Finally, I updated one small item I had noted in a comment in #1261: if `filter_sce.R` spits out a fully empty `filtered.rds` to send into doublets, that script will fail to read in the sce. I therefore added a guard against that to spit back out an empty rds from the doublets script if needed

Here's a report for proof of concept that this runs! The report is pretty minimal, as expected given 1 processed cell.
[SCPCL001295_qc.html](https://github.com/user-attachments/files/26715067/SCPCL001295_qc.html)

Once this goes in we can go back to removing this problematic library: https://github.com/AlexsLemonade/ScPCA-admin/pull/1346

### DIFF
--- a/bin/run_scimilarity.py
+++ b/bin/run_scimilarity.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Script to annotate processed ScPCA objects with SCimilarity
+# Script to annotate processed ScPCA objects with > 30 cells with SCimilarity
 # Follows this tutorial: https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html
 
 # Adapted from: https://github.com/AlexsLemonade/OpenScPCA-nf/blob/v0.1.6/modules/cell-type-scimilarity/resources/usr/bin/run-scimilarity.py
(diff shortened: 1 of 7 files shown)


bin/run_scimilarity.py
```diff
@@ -145,44 +145,56 @@ def main() -> None:
...
     processed_anndata = anndata.read_h5ad(arg.processed_h5ad_file)
-    processed_anndata = format_scimilarity(processed_anndata)
 
-    # Preprocess the data
-    # Align the query dataset to the reference model
-    processed_anndata = align_dataset(processed_anndata, scimilarity_model.gene_order)
-    # Log-normalize the counts
-    processed_anndata = lognorm_counts(processed_anndata)
+    # Before proceeding, ensure sufficient cells
+    # We use 30 here to match what we do for CellAssign
```
allyhawkins, line 150: Just noting that the reason we use 30 for CellAssign is because it explicitly fails with fewer cells than 30 without changing arguments. I think it's fine to use 30 for this too and be consistent, but I just wanted to write that down somewhere.

allyhawkins: One other note is that we might change some results for libraries with fewer than 30 cells by doing this. I don't think that's a big deal since they are poor quality to begin with, but I did want to note that there could be a change there.

allyhawkins (commented):
LGTM, thanks for doing this. Note that we also should update the docs to include a comment about the minimum number of cells required for SCimilarity.

bin/run_scimilarity.py
sjspielman: > One other note is that we might change some results for libraries with fewer than 30 cells by doing this. 

That's later down the line though, right? Unless we're doing a patch release with this now.

sjspielman: (also, did you mean to approve?)

allyhawkins: > That's later down the line though, right? Unless we're doing a patch release with this now.

Yes. Whenever we re-process and update the Portal there may be small changes, but not a concern for right now.

allyhawkins (approved):

PR #1268, opened 2026-04-16
Add script to check format of individual SCE objects

### LINKED ISSUE #1165: Script for checking format of SCE object
As mentioned in https://github.com/AlexsLemonade/scpca-nf/issues/1148#issuecomment-3781463940, we'll want to write a script that takes as input an SCE object and the reference formatting JSON (#1164) and checks for any formatting errors. The output should be an empty file if no formatting errors or a file with a message listing all the items that are missing. 

The script should follow this general outline: 

- Read in the object
- Subset the reference file to pull out the expected contents for that object type
- Check counts are rounded
- check assay names
- check colData column names and type
- check rowData column names and type
- check metadata names
- if altExp, check for altExp assay
- if processed, check for reduced dimensions, celltyping columns, and infercnv columsn

Note that for processed objects, we will need to grab the expected celltyping columns and inferCNV status from the metadata of each object. 

I think it probably makes sense to first write the script for the basic filtered/unfiltered SCE objects and then in separate PRs expand to the processed object and specialized scenarios (cite seq, cellhash).

allyhawkins:
Closes #1165 

This PR adds in a script to check the format of an SCE object. This takes as input a single SCE file, the object type (unfiltered, filtered, processed), and the formatting reference and outputs either an empty text file or a file with a description of any mismatches/errors found. Ultimately this will be placed inside a process and I would like to concatenate the error messages across libraries. 

In the script I check the assays, colData, rowData, and metadata for both the main experiment and any altExps that may be present. If the object type is processed then we also look for reducedDimNames. 

The trickiest part here was handling the conditional items that may be present. To do that I make a vector of all possible conditions for the main experiment and each altExp type. In a lot of cases we are using the contents of metadata to check for things, but the cellhashing demultiplexing types have to be checked by explicitly looking for certain column names. The `demux_methods` slot only contains `vireo`. **This might be an error we need to look into separately?**

I also had to handle a few "special" situations described here: 

- In some cases, class reports `data.frame tbl tbl_df` for a `data.frame`. So I have a check there for if one of the types that is returned is `data.frame` then to accept it. 
- We store `numeric` for `colData(sce)$detected`, but when I was working with the object here `class(sce$detected)` returns `integer`. We could change that in the reference, or do what I did which was say that `integer` is equivalent to `numeric`. (Note that the objects I used for testing were downloaded directly from the current prod bucket). 
- For multiplexed samples, `sample_type` is a list and not a character, so I accepted both. 
- `metadata(sce)$prob_compromised_cutoff` is a weird situation where we have it listed in both the main metadata checks and the conditional metadata checks. Basically, if `has_miQC` is false then this is a logical because its set as `NA`. Otherwise it gets set to a `numeric`. This means there are two rows that get returned for the checks and one is considered an error. To get around this, I removed the logical check if it's present twice. 
  - Related to this, we can only check if `miQC` is present if `metadata(sce)$prob_compromised_cutoff` is `numeric`. This seems a little circular, so maybe we want to explicitly add a `has_miQC` to the metadata? 

While working on this I also caught a few errors in the reference file that I fixed: 

- We forgot to include the mapping tool options in the main experiment metadata. 
- We don't have `sample_metadata` or `sample_type` in the altExp metadata for ADTs. 
- `ambient_profile` was incorrectly classified as a "character" and not "numeric". 
- Lastly, you'll notice in the examples I attached that `singler_gene_set_version` and `singler_date` are getting considered characters even though they are NA. Since they are NA right now we had temporarily put in `logical`, but that doesn't seem to be the case? I went ahead and updated the reference to instead just say character, but I can revert that if we prefer? 

I ran a few different types of libraries (single cell, with submitter, multiplexed, and cite) all at the processed stage. Here's some example error sheets, which have now been corrected with these changes just so you can see an idea of what to expect. 
[SCPCL000001_formatting_errors.txt](https://github.com/user-attachments/files/26794813/SCPCL000001_formatting_errors.txt)
[SCPCL000557_formatting_errors.txt](https://github.com/user-attachments/files/26794818/SCPCL000557_formatting_errors.txt)
[SCPCL000697_formatting_errors.txt](https://github.com/user-attachments/files/26794809/SCPCL000697_formatting_errors.txt)


bin/sce-formatting-checks.R
```diff
@@ -0,0 +1,369 @@
...
+          is_type_match <- TRUE
+        } else if (obs_type == "integer" && type == "numeric") {
+          # integers are valid numerics
+          is_type_match <- TRUE
+        } else if (name == "sample_type" && obs_type == "list") {
+          # if multiplex then sample type is a list not character so this is valid
+          is_type_match <- TRUE
+        } else {
+          is_type_match <- obs_type == type
+        }
```
jashapiro, line 42: I think you can do this more simply with the `is()` function. Is there a reason not to do so? It should handle everything but the multiplexed data. Definitely works for tibbles as data frames and integers as numeric. 
```suggestion
        if (name == "sample_type" && is(data[[name]], "list") =) {
           # sample type may be a list for multiplexed
           is_type_match <- is(data[[name]][[0]], type) 
        } else {
          is_type_match <- is(data[[name]], type)
        }
```

jashapiro:
> `metadata(sce)$prob_compromised_cutoff` is a weird situation where we have it listed in both the main metadata checks and the conditional metadata checks. Basically, if `has_miQC` is false then this is a logical because its set as `NA`. Otherwise it gets set to a `numeric`. This means there are two rows that get returned for the checks and one is considered an error. To get around this, I removed the logical check if it's present twice.
> 
> - Related to this, we can only check if `miQC` is present if `metadata(sce)$prob_compromised_cutoff` is `numeric`. This seems a little circular, so maybe we want to explicitly add a `has_miQC` to the metadata?

Hmm, so the _best_ way to handle this might have been to make sure we set it as `NA_real` so that it has a consistent type. Or we could add a bit of extra logic so we check optional items as `is(item, type) || is.na()`. I think the second one makes sense for this use case. 

I do think adding `has_miQC` to the metadata makes sense though. So probably some combination of these?

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,362 @@
...
+#' @param conditionals_vec A named logical vector indicating which conditions
+#'   are TRUE for the object being checked. Should be defined using the
+#'   top-level SCE (not an altExp) since many conditions depend on top-level
+#'   metadata (e.g., altExpNames, celltype_methods)
+#'
+#' @return A data frame combining required and conditional check results,
+#'   with a `slot` column indicating the source slot and a `conditional` column
+#'   (NA for required checks)
+check_sce_object <- function(sce_exp, ref_list, conditionals_vec) {
+  # required colData and rowData checks ----------------------------------------
```
sjspielman: Super minor but I don't love including this kind of comment _inside_ a function since it's meant to act as a script anchor for navigating and it will appear in the "toc" but really we'd only want the function there

But we can make it stand out with eg

```suggestion
  ############ required colData and rowData checks ############
```

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,362 @@
...
+    "colData" = list(data = colData(sce_exp), ref = ref_list[["colData"]]),
+    "rowData" = list(data = rowData(sce_exp), ref = ref_list[["rowData"]]),
+    "metadata" = list(data = metadata(sce_exp), ref = ref_list[["metadata"]])
+  ) |>
+    # skip slots that have no reference defined (e.g. row/colData absent for altExps)
+    purrr::keep(\(x) length(x$ref) > 0) |>
+    purrr::map(\(x) check_names_and_types(x$data, x$ref)) |>
+    dplyr::bind_rows(.id = "slot")
+
+  # conditional colData and metadata checks ------------------------------------
```
sjspielman:
```suggestion
  ####### conditional colData and metadata checks #######
```

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,362 @@
...
+
+  # type mismatch errors
+  mismatch_df <- dplyr::filter(match_df, is_present, !is_type_match)
+  if (nrow(mismatch_df) > 0) {
+    errors <- c(
+      errors,
+      glue::glue("
+      Type mismatch in '{mismatch_df$name}' from {label} {mismatch_df$slot}.
+      Expected {mismatch_df$expected_type}, but found {mismatch_df$observed_type}.
+
```
sjspielman:
```suggestion
```

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,362 @@
...
+# initialize empty error string
+errors <- ""
+
+# Check main SCE assays --------------------------------------------------------
+
+errors <- c(errors, check_assays(sce, all_ref_list$assayNames, label = "main SCE"))
+
+# check that counts assay contains rounded integers
+all_values <- as.vector(counts(sce))
+if (any(all_values != floor(all_values))) {
```
sjspielman: I think we should try to do this in sparse Matrix land for efficiency. Something like this?
```
all(counts(sce)@x == floor(counts(sce)@x))
```

I have learned that `x`** brings out the non-zero values from a sparse matrix. Super quick and less in memory

sjspielman (commented):
Had a first look but need a bit to time to absorb this all for a 2nd round probably... leaving initial comments now since I see this is being worked on ("refresh" keeps popping up, nbd!)

allyhawkins:
@jashapiro I implemented the solution to check using the `is` function, which was definitely helpful thank you! 

> Hmm, so the best way to handle this might have been to make sure we set it as NA_real so that it has a consistent type. Or we could add a bit of extra logic so we check optional items as is(item, type) || is.na(). I think the second one makes sense for this use case.

So using the `|| is.na()` doesn't really work. The problem is that if miQC is present, then it's NOT NA and is instead a numeric. When we do the non-conditional checks it will report a mismatch. Then we do the conditional checks and it reports a match because the numeric type is appearing as expected. 

So I think the solution here is to set this specific value to `NA_real_` since it's a unique scenario. I went ahead and did that and then also added `has_miQC` to the objects. Note that I did leave the code in for now depending on what we decide here. If we do accept these changes we'll also need to update the formatting reference file.

bin/sce_formatting_checks.R
allyhawkins: This is much better thank you!

sjspielman:
> So I think the solution here is to set this specific value to NA_real_ since it's a unique scenario. I went ahead and did that and then also added has_miQC to the objects. 

Noting I'm definitely on board with this option.

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,366 @@
...
+#' @return A data frame with the same columns as check_names_and_types, plus
+#'   a `conditional` column indicating which condition triggered the check
+check_conditional_names_and_type <- function(conditions_present, data, ref) {
+  conditions_present |>
+    purrr::map(\(condition) {
+      df <- check_names_and_types(data, ref[[condition]])
+      df$conditional <- condition
+      return(df)
+    }) |>
+    dplyr::bind_rows()
```
sjspielman: up to you if you like this more, should be equivalent

```suggestion
    purrr::set_names() |>
    purrr::map(\(condition) {
      check_names_and_types(data, ref[[condition]])
    }) |>
    dplyr::bind_rows(.id = "conditional")
```

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,366 @@
...
+
+# initialize empty error string
+errors <- ""
+
+# Check main SCE assays --------------------------------------------------------
+
+errors <- c(errors, check_assays(sce, all_ref_list$assayNames, label = "main SCE"))
+
+# check that counts assay contains rounded integers
+if (all(counts(sce)@x == floor(counts(sce)@x))) {
```
sjspielman: All equal seems good though..
I realize my suggestion wasn't ideal since it didn't match your logic. But with our powers combined...

```suggestion
if (any(counts(sce)@x != floor(counts(sce)@x))) {
```

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,366 @@
...
+  has_clusters = "cluster" %in% names(colData(sce)),
+  has_consensus = "consensus_celltype_annotation" %in% names(colData(sce)),
+  has_singler = "singler" %in% metadata(sce)$celltype_methods,
+  has_cellassign = "cellassign" %in% metadata(sce)$celltype_methods,
+  has_scimilarity = "scimilarity" %in% metadata(sce)$celltype_methods,
+  has_openscpca = "openscpca" %in% metadata(sce)$celltype_methods,
+  # submitter annotations are only valid if not all NA
+  has_submitter = "submitter" %in% metadata(sce)$celltype_methods &&
+    !all(is.na(sce$submitter_celltype_annotation)),
+  has_infercnv = !is.null(metadata(sce)$infercnv_status),
```
sjspielman: Only this condition leads to infercnv results in metadata & colData slots. The other values here are various error messages about why there's no infercnv results.

```suggestion
  has_infercnv = metadata(sce)$infercnv_status == "success"
```

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,366 @@
...
+  # check assay errors first
+  errors <- c(errors, check_assays(cellhash_sce, cellhash_ref$assayNames, label = "cellhash altExp"))
+  # now check col/row/metadata
+  errors <- collect_match_errors(cellhash_match_df, errors, label = "cellhash altExp")
+}
+
+# Check reducedDims (processed objects only) -----------------------------------
+
+if (opt$object_type == "processed") {
+  if (all(reducedDimNames(sce) != all_ref_list$reducedDimNames)) {
```
sjspielman: optional suggestion
```suggestion
  if (!setequal(reducedDimNames(sce), all_ref_list$reducedDimNames)) {
```

jashapiro:
> So using the || is.na() doesn't really work. The problem is that if miQC is present, then it's NOT NA and is instead a numeric. 

Unless I am misunderstanding, I would still expect this to work?

```r
> a <- .5
> b <- NA
> is(a, "numeric") || is.na(a)
[1] TRUE
> is(b, "numeric") || is.na(b)
[1] TRUE
```

allyhawkins:
> > So using the || is.na() doesn't really work. The problem is that if miQC is present, then it's NOT NA and is instead a numeric.
> 
> Unless I am misunderstanding, I would still expect this to work?
> 
> ```r
> > a <- .5
> > b <- NA
> > is(a, "numeric") || is.na(a)
> [1] TRUE
> > is(b, "numeric") || is.na(b)
> [1] TRUE
> ```

That does work but the problem is we expect it be logical for the first check and then numeric for the second check. The problem is in the reference file itself which will be solved with setting it to `NA_real_` so its always numeric.

bin/sce_formatting_checks.R
jashapiro: I think @sjspielman's suggestion is actually what you want (or replace the `all` with `any`)?

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,365 @@
...
+  has_clusters = "cluster" %in% names(colData(sce)),
+  has_consensus = "consensus_celltype_annotation" %in% names(colData(sce)),
+  has_singler = "singler" %in% metadata(sce)$celltype_methods,
+  has_cellassign = "cellassign" %in% metadata(sce)$celltype_methods,
+  has_scimilarity = "scimilarity" %in% metadata(sce)$celltype_methods,
+  has_openscpca = "openscpca" %in% metadata(sce)$celltype_methods,
+  # submitter annotations are only valid if not all NA
+  has_submitter = "submitter" %in% metadata(sce)$celltype_methods &&
+    !all(is.na(sce$submitter_celltype_annotation)),
+  has_infercnv = metadata(sce)$infercnv_status == "success"
```
jashapiro: Missing comma preventing passing (and run, presumably)
```suggestion
  has_infercnv = metadata(sce)$infercnv_status == "success",
```

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,365 @@
...
+    # skip slots that have no reference defined (e.g. row/colData absent for altExps)
+    purrr::keep(\(x) length(x$ref) > 0) |>
+    purrr::map(\(x) check_names_and_types(x$data, x$ref)) |>
+    dplyr::bind_rows(.id = "slot")
+
+  ####### conditional colData and metadata checks #######
+  # only check conditions that are true for this object
+  true_conditions <- names(conditionals_vec[which(conditionals_vec)])
+
+  if (length(true_conditions) > 1) {
```
jashapiro: What is the case where true conditions has length 1? It seems like the result there is that you skip the conditional matches, which seems odd; I would expect you would skip for 0 but not for 1.

Either way, I would reverse the if/else order here to keep the short block first.

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,365 @@
...
+#' @param label A string used in error messages to identify the object
+#'   (e.g., "main SCE", "adt altExp"). Default is "SCE".
+#'
+#' @return A character vector of error messages for any missing assays,
+#'   or an empty character vector if all assays are present
+check_assays <- function(sce_exp, ref_assay_names, label = "SCE") {
+  missing_assays <- ref_assay_names[!ref_assay_names %in% assayNames(sce_exp)]
+
+  if (length(missing_assays) > 0) {
+    glue::glue("Missing assay '{missing_assays}' from {label}.")
```
jashapiro, line 152:
```suggestion
    glue::glue("Missing assay(s) '{missing_assays}' from {label}.")
```

bin/sce_formatting_checks.R
allyhawkins: Wow thank you

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,365 @@
...
+#' @param match_df A data frame returned by check_sce_object
+#' @param errors A string of accumulated error messages to append to
+#' @param label A string used in error messages to identify the object
+#'   (e.g., "main SCE", "adt altExp"). Default is "SCE".
+#'
+#' @return An updated error string with any new errors appended
+collect_match_errors <- function(match_df, errors, label = "SCE") {
+  # TODO: Remove this if we accept using NA_real_
+  # remove double prob_compromised_cutoff if its present for conditional and regular metadata
+  # this value is present with or without miQC with a different type so we need to keep both checks in the reference
```
jashapiro: Looking at this a bit more: do we really need to record prob_compromised_cutoff in both the full and conditional part? If we always have it, why is it in the conditional section? Is that because the type was changing? 

I still think the better way to handle this would be to make the type checks generally allow `NA` values (or specify fields allowed to be `NA`) rather than getting caught by the type of the NA. I have a feeling this is going to come up for Python checks as well (where `None` is it's own type).

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  # missing field errors
+  missing_df <- dplyr::filter(match_df, !is_present)
+  if (nrow(missing_df) > 0) {
+    errors <- c(
+      errors,
+      glue::glue("
+      Missing '{missing_df$name}' from {label} {missing_df$slot}.
+        ")
```
jashapiro: formatting here seems odd

bin/sce_formatting_checks.R
```diff
@@ -0,0 +1,365 @@
...
+conditionals_vec <- c(
+  # mapping tools
+  "alevin-fry" = metadata(sce)$mapping_tool == "alevin-fry",
+  "cellranger-multi" = metadata(sce)$mapping_tool == "cellranger multi",
+
+  # preprocessing
+  umi_filtering = metadata(sce)$filtering_method == "UMI cutoff",
+  # the only way to confirm miQC was run successfully is if this cutoff is present and numeric
+  # TODO: use the has_miQC entry instead
+  has_miQC = class(metadata(sce)$prob_compromised_cutoff) == "numeric",
```
jashapiro: I feel like this is maybe more complete?  (I agree with using `has_miQC` more generally, but for now)
```suggestion
  has_miQC = "prob_compromised_cutoff" %in% names(metadata(sce)) &&
    !is.na(metadata(sce)$prob_compromised_cutoff) &&
    is(metadata(sce)$prob_compromised_cutoff), "numeric"),
```

jashapiro (commented):
Overall this looks good, but I am still confused about the NA/numeric situation. I'm sure I'm missing something, but it seems to me we should always have the `prob_compromised_cutoff` metadata value, and it should not be a conditional? In that case, the type should always be numeric, but we can allow it to be NA (of any type).

Are there any other places where we might have `NA` values or want to allow them? Or do we want to generally not allow `NA` in the type checking? As I said in one of my comments, I think we are going to run into this again with Python, where a typed NA value isn't going to solve the problem.

bin/sce_formatting_checks.R
allyhawkins: > Looking at this a bit more: do we really need to record prob_compromised_cutoff in both the full and conditional part? If we always have it, why is it in the conditional section? Is that because the type was changing?

Yes the type was changing, but now it should always be numeric. I think that's where your confusion was coming in because I updated the type but didn't update this to remove this exception handling. We don't need to check for NA in this case (at least not for SCEs). So the reference can be updated to just always be numeric. 

I do think it could be helpful to have an explicit list of values that are allowed to be NA, but we should probably encode those fields in the reference json.

bin/sce_formatting_checks.R
allyhawkins: Since I'm adding `has_miQC` here, I think we can actually just change this to use that directly since anything we process and do checks for will now have that field present.

bin/sce_formatting_checks.R
allyhawkins: Just noting that since this is using glue, only one assay is printed per line.

allyhawkins:
@jashapiro I cleaned up the checking and reference to hopefully minimize the confusion with the prob compromised cutoff. The SCE reference now always has that has `numeric` for the type and it only gets checked once. That's really the only entry where we have NA as an acceptable possibility, so I don't think it makes sense to separately allow NA. Maybe as we test more libraries we might see other spots that I'm missing and then we can update it, but for now I think this is okay. 

For the python checks, I ended up updating them to have either NoneType or float specifically for the prob compromised cutoff. So when we write up that script we'll just have to account for multiple types as possibilities, but I think that's reasonable. Let me know if you disagree or have another idea to handle this.

references/scripts/create-formatting-reference.py
```diff
@@ -565,19 +569,25 @@ def convert_experiment_metadata_types(metadata):
 filtered_uns_metadata = {
     **convert_experiment_metadata_types(copy.deepcopy(filtered_experiment_metadata)),
     **anndata_uns_metadata,
+    # this is NA or numeric so account for both possibilities in the reference
+    "prob_compromised_cutoff": "NoneType,float",
```
jashapiro: I'm not sure if this will cause problems later, but it seems to me that as long as we outputting in json, we should just make this a list.
```suggestion
    "prob_compromised_cutoff": ["NoneType", "float"],
```

bin/filter_sce.R
sjspielman: We can get rid of `miQC_worked` now
```suggestion
```

edit - I'm not sure this suggestion worked properly? I tried to delete `miQC_worked <- FALSE`

bin/filter_sce.R
```diff
@@ -124,14 +124,16 @@ try({
     enforce_left_cutoff = opt$enforce_left_cutoff
   )
   metadata(filtered_sce)$prob_compromised_cutoff <- opt$prob_compromised_cutoff
+  metadata(filtered_sce)$has_miQC <- TRUE
   miQC_worked <- TRUE
 })
 # set prob_compromised to NA if miQC failed
 if (!miQC_worked) {
```
sjspielman:
```suggestion
  metadata(filtered_sce)$has_miQC <- TRUE
})
# set prob_compromised to NA if miQC failed
if (!metadata(filtered_sce)$has_miQC) {
```

references/anndata-formatting-reference.json
```diff
@@ -176,18 +176,31 @@
                 "sample_metadata": "pandas.DataFrame",
                 "sample_type": "str",
                 "filtering_method": "str",
-                "prob_compromised_cutoff": "bool",
+                "prob_compromised_cutoff": "NoneType,float",
```
sjspielman: Why a single string here vs a list of possible types? Not a big comment, just a thought since it seems easier to deal with a list.

sjspielman (approved):
This lgtm! I did catch a spot where we can clean up a bit in the `filter_sce.R` script, and I'm good with however you prefer to deal with `prob_compromised_cutoff`.

jashapiro (approved):
I wrote my one comment a while ago, but apparently hit the wrong button. Seeing @sjspielman make the same comment made me realize my approval hadn't gone through.

There is an argument that all of the type values should be turned into lists (most only length 1) for consistency, but I don't think we need to be pure about it; a conditional in the code to allow for a list of types seems fine when you get there.

Issue #1270, opened 2026-04-17
Add GPU support for CellAssign


jashapiro:
CellAssign is the slowest step in our workflow, with the largest compute requirements, so we should consider making it run on GPU instances, which should in theory make it more efficient.

I don't expect this to happen soon, but there are some changes we could make to the processes to make this easier, like separating H5AD conversion into its own process, which would allow the CellAssign process to only require Python dependencies in the container. That should make building a GPU-enabled docker image easier. 

Running this will also require GPU instances in our batch setup, as well as continued support for CPU-only usage. 

I don't have a timeline for this, but if it works out we could have a template for future modules that also use GPU compute.  I would expect this to turn into an epic if we decide to pursue it.

PR #1271, opened 2026-04-17
Separate out H5AD conversion for cellassign and scimilarity

jashapiro:
Draft PR relevant to #1270

Here I separated out the H5AD conversion for scimilarity and cellassign into a separate process. This will make it easier (in the future) to deprecate the docker containers with both ScPCATools and python packages in favor of smaller images wiht just the python requirements, which can also be built more easily (hopefully) with GPU support. 

Because you can't use the same process name twice, and I didn't want to mess with the structure too much, I had to create a separate file for the process, and then include it with aliases. (I realize as I write this that it might be worth messing with the structure to convert only once, but we'll see)

Because we had some code for what to do if the conversion fails, I tried to maintain that, but move the checks out to the workflow itself.

One thing I was thinking might be nice is to have the conversion script strip out any extra data that isn't needed for CellAssign or Scimilarity, but for now, I just wanted a proof of concept. 

At the moment, I haven't tested this, but it passes (local) stub checks. Testing will come.

### DIFF
--- a/modules/classify-celltypes.nf
+++ b/modules/classify-celltypes.nf
@@ -1,5 +1,8 @@
 
 include { getVersions; makeJson; readMeta; getMetaVal; parseNA; pullthroughContainer } from '../lib/utils.nf'
+// import utility process with aliases to allow multiple uses in this workflow
+include { convert_anndata as convert_cellassign } from './convert-anndata.nf'
+include { convert_anndata as convert_scimilarity } from './convert-anndata.nf'
 
 process classify_singler {
   container "${pullthroughContainer(params.scpcatools_container, params.pullthrough_registry)}"
@@ -55,7 +58,7 @@ process classify_cellassign {
   label 'long_running'
   tag "${meta.unique_id}"
   input:
-    tuple val(meta), path(processed_rds), path(cellassign_reference_file)
+    tuple val(meta), path(processed_h5ad), path(cellassign_reference_file)
   output:
     tuple val(meta.unique_id), path(cellassign_dir)
   script:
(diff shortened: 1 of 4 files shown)


jashapiro:
I lied, the stub isn't working yet. (now it is)

jashapiro:
Seems to be working over at https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/4qVtBhBQoUpnfe/tasks

modules/classify-celltypes.nf
```diff
@@ -50,12 +53,12 @@ process classify_cellassign {
       path: "${meta.celltype_checkpoints_dir}",
       mode: 'copy'
     )
-  label 'mem_max'
+  label 'mem_128'
```
allyhawkins, line 56: Why did you decide to change this?

allyhawkins (commented):
This all looks good to me, I just ad one question about the memory change you made before I approve.

modules/classify-celltypes.nf
jashapiro: 192 seemed really high, and really limits how many jobs can run. So I thought it might be worth it to try a bit lower to start. I went back through the changes and couldn't really find a justification for when we went up from 96 to 192. I also think we might be able to decrease memory usage with a change to the script, but I'm waiting to do that for the GPU testing.

jashapiro: I'm waiting on some more test runs as well, but the peak memory I have seen looking at cellassign runs seems to be much smaller, so it seems wasteful to start as high as we do. 

I expect to do more testing on this as we add GPUs though.

allyhawkins: That makes sense to me. I will say we saw a lot of OOM errors when we first tried to run `CellAssign` on everything which was part of the change. But I'm fine to try this.

allyhawkins (approved):

PR #1274, opened 2026-04-22
Skip infercnv if no gene order file 

### LINKED ISSUE #1273: [BUG] file() called with null arg
**Describe the bug**
In some recent workflow runs (e.g., https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/1kmbkrKCmeCwM6/logs), we have gotten an `Argument of `file()` function cannot be null` error.

**To reproduce**
See https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/1kmbkrKCmeCwM6/logs
- Launched via scpca-admin workflow.

**Expected behavior**
Workflow should complete wihtout errors! And certainly no `file()` should be null

**Version information**
scpca-nf v0.10.0

**Additional context**

My best guess looking at the error and commonalities between the failure for https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/1kmbkrKCmeCwM6/logs and https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/4wH8IuypD34m2b/logs is that the failure is coming at the following call for the gene order file: 
https://github.com/AlexsLemonade/scpca-nf/blob/ccb3970d64a77d49cb26c385a8beb132e29c1afa/modules/call-cnvs.nf#L116

One part of the solution would be a more robust `file()` call here to include `[]` if the gene order file is not specified. But we should also probably be checking earlier if the gene order file is found and not `null`. 

This specific failure may also be due to bad data encoded in a `scpca-meta.json` file from an earlier workflow run and cached in `checkpoints`. I have not investigated those files as yet.

jashapiro:
closes #1273 (but importantly https://github.com/AlexsLemonade/ScPCA-admin/issues/1358)

Here I am adding a couple of little checks to make sure the workflow doesn't fail if the gene order file doesn't exit.

The first is our usual pattern of replacing the file object with `[]` if the file path does not exist or is blank. The second part skips infercnv if the file object doesn't exist. 

This adds to the robustness of the workflow, though it can result in infercnv being "silently" skipped if we don't pay attention. I still think that is better than the workflow failing with an error that doesn't say where it happened!

### DIFF
--- a/modules/call-cnvs.nf
+++ b/modules/call-cnvs.nf
@@ -111,9 +111,10 @@ workflow call_cnvs {
         meta.infercnv_heatmap_file = "${meta.infercnv_checkpoints_dir}/${meta.unique_id}_infercnv-heatmap.png"
         meta.infercnv_results_file = "${meta.infercnv_checkpoints_dir}/${meta.unique_id}_infercnv-results.rds"
         meta.infercnv_table_file = "${meta.infercnv_checkpoints_dir}/${meta.unique_id}_infercnv-table.txt"
+        gene_order_file = meta.infercnv_gene_order ? file(meta.infercnv_gene_order, checkIfExists: true) : []
 
         // return simplified input with gene order file
-        [meta, processed_sce, file(meta.infercnv_gene_order, checkIfExists: true)]
+        [meta, processed_sce, gene_order_file]
       }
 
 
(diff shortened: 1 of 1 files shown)


modules/call-cnvs.nf
```diff
@@ -125,14 +126,16 @@ workflow call_cnvs {
...
-        )
+          (
+            !params.repeat_cnv_inference
+            && file(it[0].infercnv_heatmap_file).exists()
+            && file(it[0].infercnv_results_file).exists()
+            && file(it[0].infercnv_table_file).exists()
+            && it[0].infercnv_reference_cell_hash == stored_cell_hash
+          )
+          || it[0].infercnv_reference_cell_count < params.infercnv_min_reference_cells
+          || !(it[2] in Path && it[2].exists()) // if the gene order file doesn't exist, we can't run infercnv
```
sjspielman, line 141: Registering a complaint that this is `in` and not `is`, but that is rather above anyone's paygrade to change.

sjspielman (commented):
This looks good to me, but I would suggest also throwing in a `log.warn()` ariound here. I don't think this is situation merits a warning in the qc report the way we do for data-related edge cases for inferCNV, but nextflow logging a warning that "the gene order file wasn't found so inferCNV isn't being run" would be useful.

modules/call-cnvs.nf
```diff
@@ -111,9 +111,14 @@ workflow call_cnvs {
         meta.infercnv_heatmap_file = "${meta.infercnv_checkpoints_dir}/${meta.unique_id}_infercnv-heatmap.png"
         meta.infercnv_results_file = "${meta.infercnv_checkpoints_dir}/${meta.unique_id}_infercnv-results.rds"
         meta.infercnv_table_file = "${meta.infercnv_checkpoints_dir}/${meta.unique_id}_infercnv-table.txt"
+        gene_order_file = meta.infercnv_gene_order ? file(meta.infercnv_gene_order, checkIfExists: true) : []
+
+        if  (!gene_order_file){
```
sjspielman:
```suggestion
        if (!gene_order_file) {
```

sjspielman (approved):
Just a little spacing, thanks for adding this!

PR #1275, opened 2026-04-23
Add cellassign docker containers

jashapiro:
Draft: 

Here I am adding a docker image, created via a pixi environment, that should include support for GPU-accelerated cellassign. I am submitting the draft for now against `development`, even though it should really be stacked on #1271, because I want to test that the docker environment builds correctly.

Once that is tested, I will move my work updating the workflow to a separate branch that we can run for testing and comparison between the existing implementation and this new container. But for now, let's make sure the image builds!


.github/workflows/docker_scvi.yaml
```diff
@@ -0,0 +1,131 @@
...
+        with:
+          images: |
+            ${{ env.GH_REGISTRY }}/${{ env.GH_REGISTRY_USER }}/${{ env.IMAGE_NAME }}
+            ${{ env.AWS_REGISTRY }}/${{ env.AWS_PREFIX }}/${{ env.GH_REGISTRY_USER }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=semver,pattern={{raw}},suffix=-${{ matrix.pixi_env }}
+            type=edge,branch=development,suffix=-${{ matrix.pixi_env }}
+            type=ref,event=branch,suffix=-${{ matrix.pixi_env }}
+
+      - name: Docker metadata (no suffix))
```
allyhawkins:
```suggestion
      - name: Docker metadata (no suffix)
```

allyhawkins (approved):
LGTM, let's see how this goes!

PR #1278, opened 2026-05-04
Update Cell Browser naming

jashapiro:
I made a few small changes to the CellBrowser build script to address the following two issues:
- https://github.com/AlexsLemonade/ScPCA-admin/issues/1375
- https://github.com/AlexsLemonade/ScPCA-admin/issues/1377

The changes are 
- add the project title to the short name so the main page is more descriptive when selecting projects
- run `cbMarkerAnnotate` on the marker list (generated from the primary cell type/cluster annotation)

The second of these was the easiest way to convert and get the marker gene list to show gene symbols instead of Ensembl IDs. It comes with a number of extras that we don't necessarily care about, like linking genes to the expression in the Allen Brain Atlas, but I couldn't find a way to easily turn those things off, so I didn't. I don't think they hurt anything at all. 

Originally, I was planning to do the annotation step at the library level for parallelization, but it includes downloading some reference databases, and I preferred to do that only once, rather than for every library separately. It would be possible to pre-fetch these references, and maintain a copy ourselves, but that really felt like more than was necessary at this point. If we want to revisit that decision later, we could, but I wanted to have a working version up as soon as possible. It isn't too slow anyway. 

In testing this I did find some other little changes to the process to make sure the input data isn't modified, mostly by using the `stageInMode 'copy'` flag for the final build process. 

Finally, I made a few little changes so checks will run when we merge to the `manuscript` branch, which I created as the location for any files being modified before publication. I did _not_ update the docker images we are using, despite the expectation that there will be a new release/tag this is partly because of testing, but also because we are not making any changes that should affect the images. 

Note that a test run is running at https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/1Om7MHa42y5dAk/tasks, but it is losing the scheduling battle for now with the full run in progress.

### DIFF
--- a/modules/cellbrowser.nf
+++ b/modules/cellbrowser.nf
@@ -68,6 +68,12 @@ process cellbrowser_site {
     library_dirs=(${library_dirs.join(" ")})
     project_ids=(${project_ids.join(" ")})
     for i in \${!library_dirs[@]}; do
+      # annotate the marker file and if it succeeds, replace the unannotated version with the annotated version.
+      cbAnnotateMarkers "\${library_dirs[\$i]}/markers.tsv" "\${library_dirs[\$i]}/markers_annotated.tsv" \
+        && mv "\${library_dirs[\$i]}/markers_annotated.tsv" "\${library_dirs[\$i]}/markers.tsv" \
+        || true # if annotation fails, keep the original file and continue with the build
+
+      # move files into place for cellbrowser build
       library_id=\$(basename \${library_dirs[\$i]})
       mv \${library_dirs[\$i]} "cb_data/\${project_ids[\$i]}/\${library_id}"
     done
--- a/modules/cellbrowser.nf
+++ b/modules/cellbrowser.nf
@@ -84,7 +84,7 @@ process cellbrowser_site {
     library_dirs=(${library_dirs.join(" ")})
     project_ids=(${project_ids.join(" ")})
     for i in \${!library_dirs[@]}; do
-      # copy files into place for cellbrowser build
+      # move files into place for cellbrowser build
       library_id=\$(basename \${library_dirs[\$i]})
       project_dir="cb_data/\${project_ids[\$i]}/"
       mkdir -p "\${project_dir}"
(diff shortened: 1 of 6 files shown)


jashapiro:
Oh, note that this PR is to the `manuscript` branch, which is branched off of `main`. When we finish all changes, we will merge `manuscript` into both `main` and `development`

jashapiro:
Just to show what the popup looks like now, confirming that it shows gene symbols not Ensembl ids: 
Screenshot 2026-05-04 at 3 25 47 PM

jashapiro:
Hmmm.... there might be something I am missing, as the marker list tables that are generated are not showing the additional annotations. (They show in the headers, but not the data rows, for some reason)

allyhawkins (commented):
The code here and decisions you made all seems fine to me, but your last comment makes me think something isn't working right, so I'm going to hold off on approving.

jashapiro:
@allyhawkins Okay, this was more of an adventure than I thought, because I didn't want to leave it half-done. There seems to be an issue with the marker annotation script where after running it cbBuild doesn't correctly create the marker tables for each cluster. I tried very hard to find a solution within cellBrowswer tools, which started with some rearrangement of the input Anndata files. That wasn't ultimately successful, but I left some of it as I think it better matches the expectation that CellBrowser has. 

After a lot of fighting, I ultimately decided to just rebuild the marker tables myself, so that is what the new python script does. It goes through all the libraries and rewrites the files with the expected columns and contents. I decided while doing this to drop the links to the Allen Brain atlas expression data, as that seemed overall not that useful, but I left the protein classification and gene list links. So the final marker tables look like this: 

Screenshot 2026-05-05 at 12 55 32 PM

Ready for another look!

modules/cellbrowser.nf
```diff
@@ -21,12 +21,27 @@ process cellbrowser_library {
...
+adata = anndata.read_h5ad("${h5ad_file}")
+adata.var = adata.var.rename(columns={"gene_ids": "ensembl_id"})
+if "gene_symbol" in adata.var.columns:
+    adata.var["gene_symbols"] = [
+        symbol if isinstance(symbol, str) else ensembl_id
+        for symbol, ensembl_id in zip(adata.var["gene_symbol"], adata.var["ensembl_id"])
+    ]
+adata.uns = {}  # remove uns to prevent export error
+adata.write_h5ad("cb_input.h5ad")
+EOF
```
allyhawkins, line 37: Is this the correct spacing? It seems off to me but it's hard to tell.

allyhawkins (approved):
This all looks fine to me. I had one question about spacing, but I think it might be okay and just looks weird.

modules/cellbrowser.nf
jashapiro: it's a heredoc, so a bit odd, but yes.

jashapiro:
run succeeded, merging.

PR #1280, opened 2026-05-05
Update inferCNV sentence 

### LINKED ISSUE #1272: Make inferCNV report text more generalizable
This always says "were," even if only 0 or 1 cells. It would be a nice treat to make the grammar generally work.

https://github.com/AlexsLemonade/scpca-nf/blob/ccb3970d64a77d49cb26c385a8beb132e29c1afa/templates/qc_report/infercnv_qc.rmd#L68-L71

sjspielman:
Closes #1272

Here is my PR-as-a-treat where I make the verb work for 0, 1, or more cells. Tested locally, works as expected.

### DIFF
--- a/templates/qc_report/infercnv_qc.rmd
+++ b/templates/qc_report/infercnv_qc.rmd
@@ -64,7 +64,6 @@ if (infercnv_status == "unknown_diagnosis") {
     </div>
   ")
 } else if (infercnv_status == "insufficient_reference_cells") {
-  
   n_ref_cells <- sum(processed_sce$is_infercnv_reference)
   if (n_ref_cells == 0) {
     cell_text <- "no normal cells were"
--- a/templates/qc_report/infercnv_qc.rmd
+++ b/templates/qc_report/infercnv_qc.rmd
@@ -64,10 +64,19 @@ if (infercnv_status == "unknown_diagnosis") {
     </div>
   ")
 } else if (infercnv_status == "insufficient_reference_cells") {
+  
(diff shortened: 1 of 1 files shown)


jashapiro:
I am changing this to stack on #1278, which already has the temporary fix for Nextflow versions. (and will target `manuscript` not `main`)

templates/qc_report/infercnv_qc.rmd
```diff
@@ -64,10 +64,18 @@ if (infercnv_status == "unknown_diagnosis") {
...
   ")
 } else if (infercnv_status == "insufficient_reference_cells") {
+  n_ref_cells <- sum(processed_sce$is_infercnv_reference)
+  if (n_ref_cells == 0) {
+    cell_text <- "no normal cells were"
+  } else if (n_ref_cells == 1) {
+    cell_text <- "only one normal cell was"
+  } else {
+    cell_text <- glue::glue("only {n_ref_cells} normal cells were")
+  }
```
jashapiro: Might as well not have those chained if/elses if we can avoid it.
```suggestion
  cell_text <- dplyr::case_when(
    n_ref_cells == 0 ~ "no normal cells were", 
    n_ref_cells == 1 ~ "only one normal cell was",
    .default = glue::glue("only {n_ref_cells} normal cells were")
  )
```

jashapiro (commented):
Probably fine, but as long as we are being recreational, let's keep it pretty.

PR #1281, opened 2026-05-05
Fix export of reference cell count and hash to use write lines

### LINKED ISSUE #1277: Look into inferCNV branching behavior
We are currently doing a full run of ScPCA portal libraries here:  https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/KLY7vt3T3mGeO/tasks

Looking at some of this, I am surprised to see so many libraries going through the `run_infercnv` process that I would have expected to be checkpointed and hence skip this process. I looked at a few libraries, and for some the normal reference has indeed changed (e.g. SCPCL000002 now has 671 instead of 670 reference cells), but for other libraries (e.g. SCPCL00038) the normal reference cells didn't change at all (same hash).  

I'd like to circle back and confirm some of the inferCNV branching behavior here.

allyhawkins:
Possibly closes #1277 

Here I switched to using `write_lines` for exporting the reference cell count and hash to hopefully fix the issue with skipping inferCNV. I tested this locally by exporting the file and running `cat` and it now just prints out the correct string without the `%` at the end.

### DIFF
--- a/bin/add_celltypes_to_sce.R
+++ b/bin/add_celltypes_to_sce.R
@@ -733,5 +733,5 @@ if (length(automated_methods) > 1) {
 readr::write_rds(sce, opt$output_sce_file, compress = "bz2")
 
 # export normal cell count and hash
-readr::write_file(reference_cell_count, opt$reference_cell_count_file)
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM

PR #1286, opened 2026-05-06
Bugfix + changelog  + 0.10.2 updates

jashapiro:
Here I started by fixing a bug in the parsing of cluster names for the case where we don't have cell types so the clusters are the automatic ones (cell lines, I assume), which are being stored as numbers. 

Then I decided to go ahead and add the changelog I was working on, which I did by having Claude parse the release notes. I started a bit more organized changes for future updates as well, but I am not wedded to that format. It just seems kind of nice (it's what nf-core uses).

We'll want to add some updates to the release checklist to go with that, but for now I am just trying to get this done!


allyhawkins (approved):
LGTM

PR #1288, opened 2026-05-14
Only celltype on human samples

sjspielman:
"Updated" version of https://github.com/AlexsLemonade/scpca-nf/pull/1287

This PR updates the skipping logic for cell typing to also skip samples where the organism is not human. If we extend cell typing in the future, I figure we can make a list of organisms to cell type and update the code then; this is me saying "I think it's fine to hard-code Homo sapiens in here at this time."

As in #1287, I also removed the cell type true default for the ccdl testing profile.

I also updated the stub files to include a non-human sample to confirm the behavior, which works as expected for me locally running stub. We had `NA` in the organism sample metadata for the stub samples since this variable wasn't previously used but it is now, so I went ahead and set every sample except the one I added to human. I also cleaned up a few spots - some ids were missing from some of the stub files.

I'll note that another way to have done this is - we have an organism column in our library metadata, which could instead be parsed out and stored in `meta` in `main.nf` to use for checking here. I elected here to use the sample metadata file since we were reading it in anyways for the cell line info, but it's fine with me if anyone has a preference for storing the library metadata version in `meta` and checking with that instead. The tiny caveat to be aware of are that the library organism column has an underscore in it (but does it need to? why does it? not a huge deal.), and we're not using this info anywhere else in the workflow at this time - but that could change in the future is an argument for grabbing it out of library!


sjspielman:
One other note - to use this to process data in `staging`, we'll need another point release with this code as well.

jashapiro:
> One other note - to use this to process data in `staging`, we'll need another point release with this code as well.

We do not need a tagged release to do an initial processing, but we would need one before processing for the portal.

jashapiro (approved):
LGTM

sjspielman:
> We do not need a tagged release to do an initial processing, but we would need one before processing for the portal.

Ah, right. I got mixed up - `development` indeed fine for `staging`! Pasting here as reminder to self.
https://github.com/AlexsLemonade/ScPCA-admin/blob/a412f81d34819e3c20f83a97595b1268dd42ff05/.github/workflows/scripts/create_scpca-nf_tmux.sh#L54-L58

PR #1289, opened 2026-05-14
Make plotting genes more flexible in report

sjspielman:
Another bug discovered when processing mouse data - we hardcoded `ENSG` to refer to hvgs in the report for UMAP plotting, but we no longer have this prefix with mouse ids. So I just updated this to be generally flexible and use all the genes, regardless of the prefix.

I made an analogous change in the celltype report for future-proofing just in case.

I tested changes locally with both the mouse library that failed and SCPCL000001 (aka a library with cell types), and it renders as expected.


allyhawkins (approved):
👍

PR #1291, opened 2026-05-15
Add module and process for formatting checks

### LINKED ISSUE #1276: Create a process to run formatting checks on all published SCE files
At the very end of the workflow, we will want a process that reads in the files that get published and runs formatting checks using the script being added in #1268. This process will only read in SCE objects and we will make a separate process for AnnData checks later.

allyhawkins:
Closes #1276 

Here I'm adding in a process to run the formatting checks specifically for SCE objects and export any errors that are found. The process itself is fairly straightforward as we read in the SCE file, the object type, and the formatting reference and then just run the script for checking. The output is a text file with errors or an empty text file. 

For reporting the errors, I chose to just print them out to the log file as a warning. What do we think of that or should I use a different approach? 

In terms of design, I'm thinking that we will read in the sce channel and the anndata channel to the `format_checks` workflow and then run each process for SCE and AnnData separately. 

I'm doing a test right now, so will keep this in a draft state until I check that it's fully functional.


allyhawkins:
This is now ready for review. I did a test run with SCPCL000001 and then checked the log file and saw this message. 

```
May-18 16:20:19.526 [Actor Thread 46] ERROR nextflow.Nextflow - Formatting errors for library SCPCL000001 filtered:

Type mismatch in 'miQC_model' from main SCE metadata.
Expected S4, but found flexmix.
```

I will say I don't think this is the most convenient way of exposing this information. I think it might make more sense to concatenate all the formatting error files together and then actually publish the file to the output directory? If others agree, then I'll make that change instead of outputting to the log file.

jashapiro:
> I will say I don't think this is the most convenient way of exposing this information. I think it might make more sense to concatenate all the formatting error files together and then actually publish the file to the output directory? If others agree, then I'll make that change instead of outputting to the log file.

I think both places make sense.

We should presumably fix the flexmix error though too?

modules/format-checks.nf
```diff
@@ -0,0 +1,58 @@
...
+    """
+  stub:
+    """
+    touch ${meta.unique_id}_formatting_errors.txt
+    """
+}
+
+workflow format_checks {
+  take: 
+    sce_ch // [ meta, unfiltered sce, filtered sce, processed sce, metadata json]
```
jashapiro, line 59: Are you planning to have this also take anndata_ch?

modules/format-checks.nf
```diff
@@ -0,0 +1,58 @@
...
+  main: 
+
+    sce_format_ch = sce_ch
+      // drop the metadata json
+      .map{ meta, unfiltered_sce, filtered_sce, processed_sce, metadata_json -> 
+      [meta, [unfiltered_sce, filtered_sce, processed_sce]]
```
jashapiro:
```suggestion
        [meta, [unfiltered_sce, filtered_sce, processed_sce]]
```

modules/format-checks.nf
```diff
@@ -0,0 +1,58 @@
...
+      } 
+      .transpose() // [ [meta, unfiltered sce], [meta, filtered sce], [meta, processed sce] ]
+      .map{ meta, sce -> 
+        def object_type = sce.baseName.replace("${meta.library_id}_", "")
```
jashapiro: Why not do this inside the process and just pass in the sce?

modules/format-checks.nf
```diff
@@ -0,0 +1,58 @@
...
+        [meta, sce, object_type]
+      }
+
+    check_sce(sce_format_ch, file(params.sce_format_reference_file))
```
jashapiro: A minor thing that we could change for future compliance is to move this to be one of the things that is passed to the workflow itself (one of the `takes`) and then we would pass in the file from the main workflow. 

The reason for this is to start to consolidate `params` so they don't appear inside modules. It's not a big deal either way; we would have a lot to change across the workflow as a whole.

main.nf
```diff
@@ -460,10 +461,14 @@ workflow {
     file(params.validation_palette_file)
   )
 
+  // check formatting for published sce files and generate error reports
+  format_checks(qc_publish_sce.out.data)
```
jashapiro: Following the previous suggestion, if we want to do this now (start to consolidate params)
```suggestion
  format_checks(qc_publish_sce.out.data, file(params.sce_format_reference_file))
```

jashapiro (commented):
Overall this looks good. I had a minor future-looking suggestion, but it is not important to this PR at all. 

I would support writing all of the errors to an output file as well as to the log (they belong in both, I think)

modules/format-checks.nf
allyhawkins: Yes that's my plan. And then I want a separate process for running the anndata checks.

allyhawkins:
Thanks for the review @jashapiro. I updated this to both output the errors to the log file and to a `format_check_errors.txt` file, which will be written to the root of the output directory. I also took your suggestion of inputting the reference file to the workflow using `take`. This should be ready for another look.

modules/format-checks.nf
```diff
@@ -0,0 +1,85 @@
...
+    touch ${meta.unique_id}_formatting_errors.txt
+    """
+}
+
+process compile_errors {
+  publishDir "${params.outdir}", mode: 'copy'
+  label 'mem_8'
+  input:
+   // more than one error file per library (unfiltered, filtered, processed), so use stageAs
+   path error_files, stageAs: "error_file_??.txt"
```
jashapiro: You probably want to use a * here for an arbitrary number of files.
```suggestion
   path error_files, stageAs: "error_file_*.txt"
```

modules/format-checks.nf
```diff
@@ -0,0 +1,85 @@
...
+    path "format_check_errors.txt"
+  script:
+    """
+    # check if all files are empty, if so print a success message
+    # otherwise concatenate all error files into one output file
+    if [ -z "\$(cat ${error_files})" ]; then
+      echo "No formatting errors found." > format_check_errors.txt
+    else
+      cat ${error_files} > format_check_errors.txt
+    fi
```
jashapiro: Minor, but  since there won't always be errors, rename as results?
Also only `cat`ting once. (I, know, you'd rather triple 🐈‍⬛)
```suggestion
    errors="\$(cat ${error_files})"
    if [ -z "\${errors}" ]; then
      echo "No formatting errors found." > format_check_results.txt
    else
      cat \${errors} > format_check_results.txt
    fi
```

modules/format-checks.nf
```diff
@@ -0,0 +1,85 @@
...
+}
+
+process compile_errors {
+  publishDir "${params.outdir}", mode: 'copy'
+  label 'mem_8'
+  input:
+   // more than one error file per library (unfiltered, filtered, processed), so use stageAs
+   path error_files, stageAs: "error_file_??.txt"
+  output:
+    path "format_check_errors.txt"
```
jashapiro:
```suggestion
    path "format_check_results.txt"
```

modules/format-checks.nf
```diff
@@ -0,0 +1,85 @@
...
+    # otherwise concatenate all error files into one output file
+    if [ -z "\$(cat ${error_files})" ]; then
+      echo "No formatting errors found." > format_check_errors.txt
+    else
+      cat ${error_files} > format_check_errors.txt
+    fi
+    """
+  stub: 
+    """
+    touch format_check_errors.txt
```
jashapiro:
```suggestion
    touch format_check_results.txt
```

modules/format-checks.nf
```diff
@@ -0,0 +1,85 @@
...
+        [meta, [unfiltered_sce, filtered_sce, processed_sce]]
+      } 
+      .transpose() // [ [meta, unfiltered sce], [meta, filtered sce], [meta, processed sce] ]
+
+    check_sce(sce_format_ch, sce_format_reference_file)
+
+    // collect all error files and concatenate to print to a formatting errors output file
+    error_input_ch = check_sce.out
+      .map{ meta, error_file -> error_file } // get just the error file paths
+      .collect() // collect into a list of error files
```
jashapiro: I'm pretty sure you can give `collect` a closure and do these in one step
```suggestion
      .collect{ meta, error_file -> error_file } // collect into a list of just the error files
```

jashapiro (approved):
Looks good. I had a few comments, all pretty minor, so I don't need to see it again.

modules/format-checks.nf
allyhawkins:
```suggestion
    errors="\$(cat ${error_files})"
    if [ -z \${errors} ]; then
      echo "No formatting errors found." > format_check_results.txt
    else
      \${errors} > format_check_results.txt
    fi
```
I think this is what you mean so we can actually only `cat` once (sadly)

jashapiro: Maybe `echo \${errors} > format_check_results.txt`?

PR #1293, opened 2026-05-20
Bug fixes in the formatting checks script

### LINKED ISSUE #1292: Fix error for checking SCE contents
It looks like R is complaining when we run the `if (errors == "")` statement if any errors are actually found. This works fine for when there are no errors but not if the `errors` variable has content. See https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/2NISuG6yAHRuox/logs.

sjspielman:
Context: https://github.com/AlexsLemonade/ScPCA-admin/issues/1360#issuecomment-4501820329

This PR fixes some small bugs in the formatting check script; the `if (errors == 0)` fails when length >1, so I did some re-jiggering. I tested this out with a local SCE I broke, and got some _one_ of the expected errors I coded up...

```
 ➤ cat out.txt
Formatting errors found for SCPCL001314 processed:

Type mismatch in 'detected' from main SCE rowData.
Expected numeric, but found logical.
```

But (by chance!) I also had changed sizeFactor -> character and it  wasn't flagged! I tracked this down as well - we were not defining the `has_normalization` variable properly. After defining this with actual values we use in the workflow, I now get:

```
 ➤ cat out.txt                        
Formatting errors found for SCPCL001314 processed:

Type mismatch in 'detected' from main SCE rowData.
Expected numeric, but found logical.
Type mismatch in 'sizeFactor' from main SCE colData.
Expected numeric, but found character.
```

Finally, I caught a spot that had `params$` but should be `opt$`. I also changed `file.create()` to `fs::file_create()` because it was returning and printing out  `TRUE` which marginally annoyed me ;).

Either way, the error output is both formatted properly and works with(out) errors.

### DIFF
--- a/bin/sce_formatting_checks.R
+++ b/bin/sce_formatting_checks.R
@@ -266,7 +266,7 @@ conditionals_vec <- c(
   # preprocessing
   umi_filtering = metadata(sce)$filtering_method == "UMI cutoff",
   has_miQC = metadata(sce)$has_miQC,
-  has_normalization = metadata(sce)$normalization == "normalization",
+  has_normalization = metadata(sce)$normalization %in% c("deconvolution", "log-normalization"),
 
   # clustering and cell typing
   has_clusters = "cluster" %in% names(colData(sce)),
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
Thanks for doing this

PR #1296, opened 2026-05-22
Allow for generation of metrics report with SSO creds

allyhawkins:
Related to https://github.com/AlexsLemonade/ScPCA-admin/issues/1392, I think I have a solution to our problem using the presigned urls. I updated the reports to use `paws` instead of `s3fs` for grabbing the files since that seems to allow for use with SSO creds. See the documentation here: https://github.com/paws-r/paws/blob/main/docs/credentials.md#use-aws-single-sign-on-sso

I first tried to use `aws.s3` but then ran into the same issue of credentials not being used properly, which is how I came across `paws` 🐶 . As part of this I also added an argument to set the AWS profile, which by default will use the profile in the environment. 

This seems to work both interactively and at the command line when running the full script. I am hitting an error at the very end in the inferCNV section, which is a problem for another day.

### DIFF
--- a/scripts/compare-metrics/compare-metrics-template.rmd
+++ b/scripts/compare-metrics/compare-metrics-template.rmd
@@ -39,16 +39,35 @@ set.seed(1234)
 # Functions
 #| include: false
 
-read_metrics_files <- function(urls) {
-  # read a set of metrics files
-  # url: URL of the metrics files, each in json format
-  # returns: data frame (tibble) of combined metrics
-  purrr::map(urls, \(url) {
-    jsonlite::read_json(url, simplifyVector = TRUE) |>
-      # handle bare NAs, which may be quoted
-      purrr::modify_if(\(x){
-        length(x) == 1 && !is.na(x) && x == "NA"
-      }, ~NA)
+# parse S3 uri into bucket and key
+parse_s3_path <- function(path) {
+  parsed <- httr::parse_url(path)
+  list(
+    bucket = parsed$hostname,
+    key    = sub("^/", "", parsed$path) # strip leading slash
+  )
+}
+
+# list all objects that match the regex
+list_s3_metrics <- function(s3_uri, metrics_regex) {
+  s3 <- paws::s3()
+  p <- parse_s3_path(s3_uri)
+
+  s3$list_objects_v2(Bucket = p$bucket, Prefix = p$key) |>
+    paws.common::paginate() |> # return more than 1000 entries
+    purrr::map(\(page) purrr::map_chr(page$Contents, \(obj) obj$Key)) |>
+    purrr::list_c() |>
+    grep(metrics_regex, x = _, value = TRUE) |>
+    paste0("s3://", p$bucket, "/", x = _)
+}
+
+read_metrics_files <- function(paths) {
+  s3 <- paws::s3()
+  purrr::map(paths, \(path) {
+    p <- parse_s3_path(path)
+    raw <- s3$get_object(Bucket = p$bucket, Key = p$key)
+    jsonlite::fromJSON(rawToChar(raw$Body), simplifyVector = TRUE) |>
+      purrr::modify_if(\(x) length(x) == 1 && !is.na(x) && x == "NA", ~NA)
   }) |>
     purrr::list_transpose(default = NA) |>
     tibble::as_tibble()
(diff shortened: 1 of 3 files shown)


scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -39,16 +39,35 @@ set.seed(1234)
...
-  # returns: data frame (tibble) of combined metrics
-  purrr::map(urls, \(url) {
-    jsonlite::read_json(url, simplifyVector = TRUE) |>
-      # handle bare NAs, which may be quoted
-      purrr::modify_if(\(x){
-        length(x) == 1 && !is.na(x) && x == "NA"
-      }, ~NA)
+# parse S3 uri into bucket and key
+parse_s3_path <- function(path) {
+  parsed <- httr::parse_url(path)
```
jashapiro: These seem to work exactly the same way, but I've been working with the 2 version, which is 1 more.
```suggestion
  parsed <- httr2::url_parse(path)
```

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -39,16 +39,35 @@ set.seed(1234)
...
+  list(
+    bucket = parsed$hostname,
+    key    = sub("^/", "", parsed$path) # strip leading slash
+  )
+}
+
+# list all objects that match the regex
+list_s3_metrics <- function(s3_uri, metrics_regex) {
+  s3 <- paws::s3()
```
jashapiro: Pretty minor. but probably a little better to do this outside the function and pass in the s3 connection.

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -39,16 +39,35 @@ set.seed(1234)
...
+  p <- parse_s3_path(s3_uri)
+
+  s3$list_objects_v2(Bucket = p$bucket, Prefix = p$key) |>
+    paws.common::paginate() |> # return more than 1000 entries
+    purrr::map(\(page) purrr::map_chr(page$Contents, \(obj) obj$Key)) |>
+    purrr::list_c() |>
+    grep(metrics_regex, x = _, value = TRUE) |>
+    paste0("s3://", p$bucket, "/", x = _)
```
jashapiro: Should be the same thing, but prettier?
```suggestion
    stringr::str_subset(metrics_regex) |>
    glue::glue("s3://{p$bucket}/{.}")
```

scripts/compare-metrics/compare-metrics.R
```diff
@@ -35,6 +35,12 @@ option_list <- list(
     default = "",
     help = "Output file name for the rendered report. If not provided, a name will be generated with the current date and project ID(s)."
   ),
+  make_option(
+    c("--profile"),
+    type = "character",
+    default = Sys.getenv("AWS_PROFILE", unset = "default"),
```
jashapiro: Probably better to leave this blank by default and then just not set it if nothing was passed in. 
```suggestion
    default = "",
```

scripts/compare-metrics/compare-metrics.R
```diff
@@ -46,6 +52,10 @@ option_list <- list(
 opt_parser <- OptionParser(option_list = option_list)
 opt <- parse_args(opt_parser)
 
+# set AWS profile for use with paws
+# see https://github.com/paws-r/paws/blob/main/docs/credentials.md#use-aws-single-sign-on-sso
+Sys.setenv(AWS_PROFILE = opt$profile)
```
jashapiro:
```suggestion
if (opt$profile != "") {
  Sys.setenv(AWS_PROFILE = opt$profile)
}
```

scripts/compare-metrics/compare-metrics.R
jashapiro: do we want to remove s3fs here too?

scripts/compare-metrics/renv.lock
jashapiro, line 1: We may want to update this renv to R 4.5?

jashapiro (commented):
Overall looks good, but I had a number of little comments throughout... I'm not sure how much time we really want to spend on this though, so some of these (e.g., refactoring out the S3 connection to be passed in rather than creating multiple) may not really be necessary. I probably would go through and remove all use of `s3fs` though.

allyhawkins:
I went ahead and updated this to remove`s3fs` throughout. It's a little more complicated than using the `s3fs::s3_dir_exists()` function, but at least it makes things more consistent. I went ahead and updated to R 4.5, but let me know if you want to keep or revert that. 

The last thing I updated was handling the error with checking the inferCNV number of reference cells. It looks like the metrics files in prod don't have that number, so it was causing errors with trying to create the data frame that calculates the change in number of cells. I went ahead and added a check to make sure both columns (ref and comp) exist, otherwise there are no changes so everything else is skipped. 

We now have a report for SCPCP000013 🎉

scripts/compare-metrics/compare-metrics-template.rmd
```diff
@@ -2,7 +2,7 @@
 params:
   reference_s3: s3://ccdl-scpca-nf-results-prod/results
   comparison_s3: s3://ccdl-scpca-nf-results-staging/results
-  project_id: "all"
+  project_id: "SCPCP000013"
```
jashapiro: Probably want to change this back?
```suggestion
  project_id: "all"
```

jashapiro (approved):
LGTM, with just a possible default change. Thanks for handling this!

PR #1298, opened 2026-05-27
Script for checking formatting of AnnData objects

### LINKED ISSUE #1166: Script for checking format of AnnData object
We will need a parallel script to #1165 that is used for AnnData objects. This script should be in Python and read in the AnnData object and reference file to check for formatting. A lot of the structure should mirror what's being added in #1165 but for the AnnData objects.

allyhawkins:
Closes #1166

This PR adds in a script that will be used to check the formatting of an individual AnnData object. Just like with the SCE object checking, this script takes in a single object and the object type. The main difference here is that we will need to check both the RNA and ADT files. Right now, I'm grabbing that information from the filename, but let me know if you think it would be better to parse that in Nextflow and add that as an argument? 

I tried my best to use `isinstance()` instead of `type()` for the matching, but did hit a few issues along the way. 

- For the `obs` and `var` slots, which are Pandas data frames, `isinstance(data["col"], int)` and `isinstance(data["col"], int32)` both give failures. Because of that we need to use the `series.dtype.name` and then check that against the expected type. 
- For the `uns` slots, the names we have in our reference didn't exactly match what's expected since they are all numpy types. I updated the reference to reflect this so that we could use `isinstance` here. 

There was one other error where we still had factor in the reference as type, so I fixed that to be set to `category`.

### DIFF
--- a/bin/anndata_formatting_checks.py
+++ b/bin/anndata_formatting_checks.py
@@ -18,38 +18,23 @@ import numpy
 import pandas
 
 
-# Maps reference type strings to Python types for uns value checking.
-# The reference uses these strings but actual uns values are numpy scalars/arrays,
-# so isinstance is needed rather than comparing type names directly.
-_UNS_TYPE_MAP = {
+_BUILTIN_TYPES = {
     "str": str,
-    "int": (int, numpy.integer),
-    "float": (float, numpy.floating),
-    "bool": (bool, numpy.bool_),
-    "numpy.ndarray": numpy.ndarray,
-    "pandas.DataFrame": pandas.DataFrame,
+    "int": int,
+    "float": float,
+    "bool": bool,
     "NoneType": type(None),
     "dict": dict,
 }
 
 
-def _uns_type_matches(val, expected):
-    """Check a uns value against one or more expected type strings."""
-    allowed = expected if isinstance(expected, list) else [expected]
-    for t in allowed:
-        if t is None:
-            continue
-        # float64 in nested uns (e.g. pca variance/variance_ratio) is a numpy float64 array
-        if t == "float64" and isinstance(val, numpy.ndarray):
-            if val.dtype == numpy.float64:
-                return True
-            continue
-        types = _UNS_TYPE_MAP.get(t)
-        if types is None:
-            return True  # unknown type string, skip
-        if isinstance(val, types):
-            return True
-    return False
+def _resolve_type(type_str):
+    """Resolve a reference type string to a Python type for isinstance checking."""
+    if type_str.startswith("numpy."):
+        return getattr(numpy, type_str[6:], None)
+    if type_str.startswith("pandas."):
+        return getattr(pandas, type_str[7:], None)
+    return _BUILTIN_TYPES.get(type_str)
 
 
 def check_names_and_types(data, ref, label, slot):
@@ -71,7 +56,13 @@ def check_names_and_types(data, ref, label, slot):
 
         elif expected_type is not None:
             if slot_is_uns:
-                if not _uns_type_matches(data[col], expected_type):
+                allowed = (
+                    expected_type
+                    if isinstance(expected_type, list)
+                    else [t.strip() for t in str(expected_type).split(",")]
+                )
+                py_types = [_resolve_type(t) for t in allowed if t]
+                if not any(isinstance(data[col], t) for t in py_types if t is not None):
                     obs_type = type(data[col]).__name__
                     errors.append(
                         f"Type mismatch in '{col}' from {label} {slot}. "
--- a/bin/anndata_formatting_checks.py
+++ b/bin/anndata_formatting_checks.py
@@ -14,9 +14,20 @@ import sys
 from pathlib import Path
 
 import anndata
-import numpy
-import pandas
+import numpy as np
+import pandas as pd
 
+
+# Maps simplified type names to pandas dtype-checking functions.
+# These work for both numpy arrays and pandas Series regardless of backend.
+_DTYPE_CHECKERS = {
+    "int": pd.api.types.is_integer_dtype,
+    "float": pd.api.types.is_float_dtype,
+    "string": pd.api.types.is_string_dtype,
+    "bool": pd.api.types.is_bool_dtype,
+    # account for deprecated pd.CategoricalDtype in older pandas versions
+    "category": lambda obj: isinstance(getattr(obj, "dtype", obj), pd.CategoricalDtype),
+}
 
 _BUILTIN_TYPES = {
     "str": str,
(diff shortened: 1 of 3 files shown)


references/anndata-formatting-reference.json
```diff
@@ -358,22 +358,25 @@
                 "sample_metadata": "pandas.DataFrame",
                 "sample_type": "str",
                 "filtering_method": "str",
-                "prob_compromised_cutoff": "NoneType,float",
+                "prob_compromised_cutoff": [
+                    "NoneType",
+                    "float"
```
jashapiro: is this okay staying as a float or should it also be a numpy.floating?

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,281 @@
...
+
+    for layer in ref.get("layers", []):
+        if layer not in adata.layers:
+            errors.append(f"Missing layer '{layer}' from {label}.")
+
+    # get conditionals dict based on adata contents and uns values, to determine which conditional checks to run
+    conditions = get_conditionals(adata)
+    # only keep true conditions and return as a list of conditionals
+    true_conditions = [cond for cond, is_true in conditions.items() if is_true]
+
```
jashapiro, line 213: I would collapse these two parts so that the get_conditionals function returns just a set of true conditions, rather than having to sort through it after the return. Since you only use this function once, you may as well have it give exactly the value you need. Alternatively, keep the dictionary, and save the lookups for later. (Rather than doing `condition in true_conditions` you should be able to do just `conditions.get(condition)` (a `None` value will act as False).

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,281 @@
...
+        # additional modalities
+        "has_adt": has_adt,
+        "has_cellhash": False,
+        "has_hashedDrops": any("hashedDrops" in c for c in obs_cols),
+        "has_HTODemux": any("HTODemux" in c for c in obs_cols),
+        "has_vireo": any("vireo" in c for c in obs_cols),
+        "is_multiplexed": is_multiplexed,
+        "has_negative_control": bool(has_neg_ctrl),
+        "no_negative_control": bool(no_neg_ctrl),
+    }
```
jashapiro:
```suggestion
    condition_tests =  {
        # mapping tools
        "alevin-fry": uns.get("mapping_tool") == "alevin-fry",
        "cellranger-multi": uns.get("mapping_tool") == "cellranger multi",
        # preprocessing steps
        "umi_filtering": uns.get("filtering_method") == "UMI cutoff",
        "has_normalization": "sizeFactor" in obs_cols,
        # clustering and cell type annotation
        "has_clusters": "cluster" in obs_cols,
        "has_consensus": "consensus_celltype_annotation" in obs_cols,
        "has_singler": "singler" in celltype_methods,
        "has_cellassign": "cellassign" in celltype_methods,
        "has_scimilarity": "scimilarity" in celltype_methods,
        "has_openscpca": "openscpca" in celltype_methods,
        "has_submitter": has_submitter,
        "has_infercnv": safe_uns_str(uns, "infercnv_status") == "success",
        # additional modalities
        "has_adt": has_adt,
        "has_cellhash": False,
        "has_hashedDrops": any("hashedDrops" in c for c in obs_cols),
        "has_HTODemux": any("HTODemux" in c for c in obs_cols),
        "has_vireo": any("vireo" in c for c in obs_cols),
        "is_multiplexed": is_multiplexed,
        "has_negative_control": bool(has_neg_ctrl),
        "no_negative_control": bool(no_neg_ctrl),
    }
    
    return set(cond for cond, is_true in condition_tests.items() if is_true)
```

This is the simplest version of my suggestion to have the get_conditionals function return the actual set you need.

references/anndata-formatting-reference.json
```diff
@@ -358,22 +358,25 @@
...
                 "normalization": "str",
                 "highly_variable_genes": "numpy.ndarray",
                 "X_name": "str",
                 "schema_version": "str",
                 "pca": {
                     "param": "dict",
-                    "variance": "float64",
-                    "variance_ratio": "float64"
+                    "variance": "numpy.ndarray",
+                    "variance_ratio": "numpy.ndarray"
```
jashapiro: I think we want the data type here, not just that it is an array?

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,281 @@
...
+                if not any(isinstance(data[col], t) for t in py_types if t is not None):
+                    obs_type = type(data[col]).__name__
+                    errors.append(
+                        f"Type mismatch in '{col}' from {label} {slot}. "
+                        f"Expected {expected_type}, but found {obs_type}."
+                    )
+            else:
+                # need to use dtype.name for obs/var columns since they are numpy dtypes
+                obs_type = data[col].dtype.name
+                if obs_type != expected_type:
```
jashapiro: All of these are stored as `int32` etc, but I think we want to be more flexible here, in case the number of bits changes. (I know this means changing the json and the create reference script)

I think we can do that with something like:
```
numpy.issubdtype(a.dtype, getattr(numpy, "integer"))
```

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,281 @@
...
+def check_names_and_types(data, ref, label, slot):
+    """
+    Check obs or var columns or uns dict for presence and correct type.
+    If slot == "uns", data is a dict and types are checked with isinstance.
+    If slot == "obs" or "var", data is a DataFrame and types are checked via dtype.name.
+    """
+    errors = []
+    slot_is_uns = slot == "uns"
+
+    for col, expected_type in ref.items():
```
jashapiro: I was very confused by `col` here for a while since these are not necessarily columns.
```suggestion
    for key, expected_type in ref.items():
```

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+def check_names_and_types(data, ref, label, slot):
+    """
+    Check obs or var columns or uns dict for presence and correct type.
+    If slot == "uns", data is a dict and types are checked with isinstance.
+    If slot == "obs" or "var", data is a DataFrame and types are checked via dtype.name.
+    """
+    errors = []
+    slot_is_uns = slot == "uns"
```
jashapiro: I think I would change this to pass in more directly the thing you are interested in: rather than passing `slot`, pass a boolean for `is_numpy` or similar?

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,281 @@
...
+    return val if isinstance(val, str) else None
+
+
+# for any uns key (cell type methods) that may be a list or array, we want to safely check and convert to list for conditional checks
+def safe_uns_array(uns, key):
+    val = uns.get(key)
+    if val is None:
+        return []
+    if isinstance(val, numpy.ndarray):
+        return val.tolist()
```
jashapiro: If it is an ndarray, should we use dtypes to check it?

bin/anndata_formatting_checks.py
jashapiro: Update: I think we may want to be even more flexible, and use the `pandas.api.types` functions.

jashapiro (commented):
This looks like a good start, but I am worried it is still kind of brittle. One worry I have is that `pandas` has a number of backends (ask me how I know), and there could be changes in the exact types used, which we would like to avoid errorring on.

I think the way I would go about it is to be a bit more dynamic in how the type checking is done, which I think can ultimately simplify the whole thing. Rather than having pre-determined paths for uns, obs, etc., we can check each object to determine if it is a `pandas` or `numpy` object with something like (Note I am assuming `import pandas as pd`, etc. for the code here):
```
isinstance(obj, (pd.Series, np.ndarray))`
```

Then if we are in one of those cases, we can use pandas functions to test the particular data types: `pd.api.types.is_integer_dtype()`, `pd.api.types.is_float_dtype()`, `pd.api.types.is_string_dtype()`, `pd.api.types.is_bool_dtype()`. Those will work for both numpy arrays and columns from pandas, regardless of what backend is being used for the DataFrame. 

If we are not in one of those objects, we just use the standard `isinstance()` with python data types. 

Part of this involves updating the JSON/formatting script to simplify the types recorded there to just "integer", "float", etc., which we can translate to the appropriate pandas functions using a dict like you have for have for `_BUILTIN_TYPES` (we could even change the builtins to use the same names).

allyhawkins:
Thanks so much for the helpful suggestions here @jashapiro. I updated this to now have two dictionaries that are used to map types - one for the built in types and one for the dtypes. I also added a new function that explicitly does the check. This first checks for `numpy.ndarray` since they can be in uns and the regular built in checks don't catch them. Then I check if the object is a series or array and if so use the `pd.api.types` set of functions to check. If not, then I just use `isinstance` with my built in types map. 

I also went ahead and changed the reference to use regular int/float and instead of category for "character" in R, I'm using "string", which fits the expectations. I think this should be ready for another look.

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,313 @@
...
+    type checking is robust across pandas backends and numpy dtype variations.
+    For plain Python objects, falls back to isinstance with built-in types.
+
+    Returns True if the type matches, False otherwise.
+    """
+
+    # first check for numpy arrays, since these can be present in uns
+    # and require a different check than plain Python objects or pandas Series
+    if expected_type == "numpy.ndarray":
+        return isinstance(obj, np.ndarray)
```
jashapiro, line 57: It looks like the only one of these we have is the transcript_types. Can we just set that as string (or whatever the underlying values are)?  No problem if not, but I thought worth a check.

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,313 @@
...
+def check_names_and_types(data, ref, label, slot):
+    """
+    Check obs/var columns or uns dict for presence and correct type.
+
+    For obs/var slots, data is a DataFrame; column values are pd.Series.
+    For uns slot, data is a dict; values may be plain Python objects,
+    numpy arrays, or pandas Series — _check_type handles all cases.
+    """
+    errors = []
+    keys = data.keys() if slot == "uns" else data.columns
```
jashapiro, line 87: Part of me doesn't love this and would rather it be a dynamic check based on what type `data` is, but it makes sense to have the slot in the error messages, so this seems fine.

bin/anndata_formatting_checks.py
```diff
@@ -0,0 +1,313 @@
...
+
+    for key, expected_type in ref.items():
+        if key not in keys:
+            errors.append(f"Missing '{key}' from {label} {slot}.")
+
+        elif isinstance(expected_type, dict):
+            # nested dict: recurse (used for uns keys like "pca")
+            errors.extend(check_names_and_types(data[key], expected_type, label, slot))
```
jashapiro: I wonder if you want to relabel the slot when recursing?
```suggestion
            errors.extend(check_names_and_types(data[key], expected_type, label, f"{slot}:{key}"))
```

jashapiro (approved):
This looks good to me. I had just a couple minor questions/further thoughts, but I don't think I should need to see it again, unless you want me to look over any particular changes.

bin/anndata_formatting_checks.py
allyhawkins: It also affects `celltype_methods`, `highly_variable_genes`, and `consensus_celltype_methods`. I tried to set it as a string, and there's still a type mismatch so I'm going to leave it as is.

PR #1299, opened 2026-05-29
Add process for AnnData formatting checks

### LINKED ISSUE #1295: Process for running formatting checks for AnnData objects
Once we have the script for checking AnnData objects (see #1166), we will need to run that in a process in the existing `check_formats` workflow.

allyhawkins:
Closes #1295 

This PR adds in the process for running the script to check formatting of AnnData for every object. The process itself was very straightforward and followed the same setup as the SCE process. 

For outputting the final combined file, I ended up just putting all the error files in one channel and then exporting as a single file. The problem with that is we could have errors for a library in both the AnnData and SCE version of an object. To get around that, I added the type of object to the header that gets printed for each object with errors. Are we okay with that, or should we make two separate formatting results files, one for AnnData and one for SCE? 

The first time this ran through successfully, I got a file with a decent amount of errors shown here: 
[format_check_results_test.txt](https://github.com/user-attachments/files/28409998/format_check_results_test.txt)

I then updated the reference to account for those errors and now the output is empty with no reported errors. 

- We don't include `sample_metadata` in the `uns` of AnnData objects, so that is removed entirely. 
- Some of the `uns` contents seem to be stored as numpy type objects (so int64, float64, etc). To account for that, I added the numpy types to the `_BUILTIN_TYPES` dictionary. 
- `highly_variable` in the `var` should only be in the processed objects, so I removed it from unfiltered and filtered. 
- And finally, the `detected` column in the `var` slot should be a float, but because we have `detected` in the `obs` column that's an `int`, we are overwriting it during our conversion. Since the `obs` one is the one that has the weird type, I ended up overwriting that for each of the `obs` dictionary and removing the exception.

### DIFF
--- a/bin/anndata_formatting_checks.py
+++ b/bin/anndata_formatting_checks.py
@@ -31,9 +31,9 @@ _DTYPE_CHECKERS = {
 
 _BUILTIN_TYPES = {
     "str": str,
-    "int": int,
-    "float": float,
-    "bool": bool,
+    "int": (int, np.integer),  # covers int64, int32, etc.
+    "float": (float, np.floating),  # covers float64, float32, etc.
+    "bool": (bool, np.bool_),  # covers numpy.bool_
     "NoneType": type(None),
     "dict": dict,
     "numpy.ndarray": np.ndarray,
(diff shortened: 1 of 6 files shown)


jashapiro (approved):
LGTM! I think the choices you made for updates seem quite reasonable. Thanks for handling all of this.

PR #1300, opened 2026-06-02
Rename FASTQ files for space ranger input

### LINKED ISSUE #1297: Ensure spatial FASTQ file names are compatible with space ranger
Space ranger requires input names to be in the format `<sample>_S001_L001_R1_001.fastq.gz`, or more minimally without the lane number: `<sample>_S1_R1_001.fastq.gz`.

There are circumstances when the input files would not follow this naming convention. This issue tracks modifying the spaceranger module to rename files when necessary for input to space ranger.

sjspielman:
Closes #1297 

This (Claude assisted, heavily babysat, hence also a `CLAUDE.md`) PR adds functionality to rename the fastq files for space ranger as needed. Here is how I ended up implementing this:

- Rather than removing the existing `getCRsamples()` function, I replaced with a different function `getSpaceRangerPrefix()`. This allows us to do some error handling in the nextflow code itself rather than shunting everything to Python. This function grabs the prefix if files are already properly formatted as spaceranger suggests, or returns `""` the files are an acceptable format which need to be renamed; right now I allow `_R1/R2.fastq.gz` & `1/2.fastq.gz`. Otherwise if the files have a different name that can't be parsed clearly, it returns `null` and the associated run is skipped entirely. The prefix gets added to `meta` along the way
  - _Review question: Do we want to be more flexible than this for allowed input file names?_
- In the spaceranger processes, we can then use the `meta.spaceranger_prefix` to control whether files need to be renamed and associated args to spaceranger
- The Python script renames by making a new directory `fastq_staged` and making symlinks with names spaceranger will approve of, using the library id as the prefix.

I've tested this with the following file names for a dataset with 4 pairs of reads -
  - Space ranger formatted --> Runs as expected
  - Allowed pattern: files named `{SRR}_R1.fastq.gz / {SRR}_R2.fastq.gz` --> Runs as expected
  - Allowed pattern: files named `{SRR}_1.fastq.gz / {SRR}_2.fastq.gz` --> Runs as expected
  - Bad file names: files named `{SRR}_read1.fastq.gz / {SRR}_read2.fastq.gz`

You can see a workdir for example:

```
# inputted files
scpca-nf ➤ aws s3 ls s3://ccdl-scpca-nf-results-testing/example-data/visium-test-runs/runs-nonconformant-names/SCPCR210000/fastq/
2026-06-02 11:37:11        220 SRR21000001_R1.fastq.gz
2026-06-02 11:37:11        362 SRR21000001_R2.fastq.gz
2026-06-02 11:37:12        217 SRR21000002_R1.fastq.gz
2026-06-02 11:37:13        362 SRR21000002_R2.fastq.gz
2026-06-02 11:37:13        230 SRR21000003_R1.fastq.gz
2026-06-02 11:37:14        365 SRR21000003_R2.fastq.gz
2026-06-02 11:37:15        233 SRR21000004_R1.fastq.gz
2026-06-02 11:37:15        378 SRR21000004_R2.fastq.gz

# renamed by the python script 
 ➤ aws s3 ls s3://ccdl-scpca-nf-workdir/work/d7/3962255e4b5b422dd8da924d8b8b6f/fastq_staged/                git:sjspielman/1297-fastq-names-spaceranger
2026-06-02 12:12:41        287 .fusion.symlinks
2026-06-02 11:58:32        101 SCPCL210000_S1_L001_R1_001.fastq.gz
2026-06-02 11:58:32        101 SCPCL210000_S1_L001_R2_001.fastq.gz
2026-06-02 11:58:32        101 SCPCL210000_S1_L002_R1_001.fastq.gz
2026-06-02 11:58:32        101 SCPCL210000_S1_L002_R2_001.fastq.gz
2026-06-02 11:58:32        101 SCPCL210000_S1_L003_R1_001.fastq.gz
2026-06-02 11:58:32        101 SCPCL210000_S1_L003_R2_001.fastq.gz
2026-06-02 11:58:31        101 SCPCL210000_S1_L004_R1_001.fastq.gz
2026-06-02 11:58:31        101 SCPCL210000_S1_L004_R2_001.fastq.gz
```


jashapiro:
> * Rather than removing the existing `getCRsamples()` function, I replaced with a different function `getSpaceRangerPrefix()`. This allows us to do some error handling in the nextflow code itself rather than shunting everything to Python. This function grabs the prefix if files are already properly formatted as spaceranger suggests, or returns `""` the files are an acceptable format which need to be renamed; right now I allow `_R1/R2.fastq.gz` & `1/2.fastq.gz`. Otherwise if the files have a different name that can't be parsed clearly, it returns `null` and the associated run is skipped entirely. The prefix gets added to `meta` along the way
>   
>   * _Review question: Do we want to be more flexible than this for allowed input file names?_
> 
> * In the spaceranger processes, we can then use the `meta.spaceranger_prefix` to control whether files need to be renamed and associated args to spaceranger

I haven't looked at the code yet, but I think this is just a bit of extra complexity that is probably more confusing than helpful. I would always populate the prefix if parsing succeeds, and then let the Python script deal with whether the files need to be renamed or not. In other words: always run the python script and have it no-op if the files are already in spaceranger format.

bin/prepare_spaceranger_fastqs.py
```diff
@@ -0,0 +1,92 @@
...
+# Input files must end in _<R1/R2/1/2>.fastq.gz.
+# Files are grouped by their original sample prefix; each group gets its own lane.
+# Prints the sample_name prefix to stdout.
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+ALLOWED_PATTERN = re.compile(r"^(.+)_(R?[12])\.fastq\.gz$")
```
jashapiro: I would expand this to allow `_001` etc. suffixes as well. I'm also a big fan of named groups in python regexes. So I would do something like: 
```suggestion
ALLOWED_PATTERN = re.compile(
    r"^(?P<prefix>.+)_(?P<read>R?[12])(?:_(?P<lane>\d{3}))?\.fastq\.gz$"
)
```

Then your code below can `m.group("prefix")` etc. which makes that part of the code more clear and robust.

Expanding this to capture all the possible parts (Sample index and lane number) is something we could do as well (to allow for the expected spaceranger format as a no-op)

bin/prepare_spaceranger_fastqs.py
jashapiro, line 1: minor: can we call this `prepare_ranger_fastqs.py` so that it can be reused for Cell Ranger as needed?

bin/prepare_spaceranger_fastqs.py
```diff
@@ -0,0 +1,92 @@
...
+        sys.exit(1)
+
+    # List all FASTQ files in order
+    # Note their existence was confirmed in nextflow code already
+    fastqs = sorted(fastq_dir.glob("*.fastq.gz"))
+
+    # Create directory where we will stage the renamed symlinks
+    staged_dir.mkdir(exist_ok=True)
+
+    # Group files by original sample prefix, preserving sort order for lane assignment
```
jashapiro: Minor, but I don't think you need to worry about the sort order?

bin/prepare_spaceranger_fastqs.py
```diff
@@ -0,0 +1,92 @@
...
+    groups = {}
+    for f in fastqs:
+        orig_prefix, read_pair = parse_fastq(f.name)
+        groups.setdefault(orig_prefix, {})[read_pair] = f
+
+    # Create the symlinks with new names according to the Space Ranger convention
+    for lane, orig_prefix in enumerate(sorted(groups), start=1):
+        for read_pair, f in sorted(groups[orig_prefix].items()):
```
jashapiro: Whatever you decide about sorting, you definitely don't need sorting here.
```suggestion
        for read_pair, f in groups[orig_prefix].items():
```

modules/spaceranger.nf
```diff
@@ -21,12 +21,21 @@ process spaceranger {
...
+
+    // if spaceranger_prefix is set, files are already conformant: use fastq_dir directly.
+    // otherwise, run the Python script to stage renamed symlinks and capture the prefix.
+    prefix_cmd = meta.spaceranger_prefix ?
+      "spaceranger_prefix=${meta.spaceranger_prefix}" :
+      "spaceranger_prefix=\$(prepare_spaceranger_fastqs.py ${fastq_dir} ${meta.library_id} fastq_staged) || exit 1"
+    fastqs_dir = meta.spaceranger_prefix ? "${fastq_dir}" : "fastq_staged"
     """
+    ${prefix_cmd}
+
```
jashapiro, line 30: This is the extra complexity I was talking about. Just run the script always and have it handle "correctly" named files; it is way too fast to complicate with spreading out the logic into multiple places.
```suggestion
"""
    prepare_spaceranger_fastqs.py ${fastq_dir} ${meta.library_id} fastq_staged
```

modules/spaceranger.nf
```diff
@@ -21,12 +21,21 @@ process spaceranger {
...
     spaceranger count \
       --id=${out_id} \
       --transcriptome=${index} \
-      --fastqs=${fastq_dir} \
-      --sample=${meta.cr_samples} \
+      --fastqs=${fastqs_dir} \
```
jashapiro: If you take my suggestion, this is a constant location.
```suggestion
      --fastqs=fastq_staged \
```

modules/spaceranger.nf
jashapiro: The suggestion here didn' t quite take because of the one unchanged line. but I am deleting all of the prefix_cmd definition and logic, as well as the `fastqs_dir` variable.

modules/spaceranger.nf
```diff
@@ -217,24 +235,37 @@ process spaceranger_publish {
...
-  def samples = []
-  fastq_files.each{ it ->
-    // append sample names to list, using regex to extract element before S001, etc.
-    // [0] for the first match set, [1] for the first extracted element
-    // use .name to ensure we aren't getting a full path
-    samples << (it.name =~ /^(.+)_S.+_L.+_[R|I].+.fastq.gz$/)[0][1]
+  assert fastq_files : "No FASTQ files found in ${files_dir}."
+
+  def spaceranger_pattern = /^(.+)_S\d+_(L\d+_)?(R[12]|I[12])_001\.fastq\.gz$/
+  def allowed_pattern = /^.+_(R?[12])\.fastq\.gz$/
```
jashapiro: I suggest just the allowed pattern (but make it allow the `_00X` suffix)

modules/spaceranger.nf
```diff
@@ -217,24 +235,37 @@ process spaceranger_publish {
...
   def fastq_files = files("${files_dir}/*.fastq.gz")
-  def samples = []
-  fastq_files.each{ it ->
-    // append sample names to list, using regex to extract element before S001, etc.
-    // [0] for the first match set, [1] for the first extracted element
-    // use .name to ensure we aren't getting a full path
-    samples << (it.name =~ /^(.+)_S.+_L.+_[R|I].+.fastq.gz$/)[0][1]
+  assert fastq_files : "No FASTQ files found in ${files_dir}."
+
+  def spaceranger_pattern = /^(.+)_S\d+_(L\d+_)?(R[12]|I[12])_001\.fastq\.gz$/
```
jashapiro: This would need to be made more flexible, as 001 is not the only possible suffix.

bin/prepare_spaceranger_fastqs.py
```diff
@@ -0,0 +1,92 @@
...
+    if not staged_dir.name:
+        print("Error: staged_dir must not be empty.", file=sys.stderr)
+        sys.exit(1)
+
+    # List all FASTQ files in order
+    # Note their existence was confirmed in nextflow code already
+    fastqs = sorted(fastq_dir.glob("*.fastq.gz"))
+
+    # Create directory where we will stage the renamed symlinks
+    staged_dir.mkdir(exist_ok=True)
```
jashapiro: It is probably not okay if the directory exists, as we don't want it to have other files in it.
```suggestion
    staged_dir.mkdir()
```

bin/prepare_spaceranger_fastqs.py
```diff
@@ -0,0 +1,92 @@
...
+        print(
+            f"Error: fastq_dir {fastq_dir} does not exist or is not a directory.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    if not sample_name:
+        print("Error: sample_name must not be empty.", file=sys.stderr)
+        sys.exit(1)
+    if not staged_dir.name:
+        print("Error: staged_dir must not be empty.", file=sys.stderr)
```
jashapiro: This is a bit confusing, as it makes me think the directory should exist and have something in it. In truth, I would probably just use a default name here.
```suggestion
        print("Error: staged_dir must not be provided.", file=sys.stderr)
```

bin/prepare_spaceranger_fastqs.py
```diff
@@ -0,0 +1,92 @@
...
+    parser = argparse.ArgumentParser(
+        description="Stage and rename allowed FASTQ files for Space Ranger."
+    )
+    parser.add_argument("fastq_dir", type=Path, help="Directory containing FASTQ files")
+    parser.add_argument(
+        "sample_name", help="Library ID to use as the shared sample prefix"
+    )
+    parser.add_argument(
+        "staged_dir", type=Path, help="Directory to create with renamed symlinks"
+    )
```
jashapiro: I prefer named arguments (note that argparse will convert the kebab-case to snake_case for the actual variable names, but I like kebab-case on the command line) 
```suggestion
    parser.add_argument("--fastq-dir", type=Path, help="Directory containing FASTQ files")
    parser.add_argument(
        "--sample-name", help="Library ID to use as the shared sample prefix"
    )
    parser.add_argument(
        "--staged-dir", type=Path, help="Directory to create with renamed symlinks"
    )
```

bin/prepare_spaceranger_fastqs.py
```diff
@@ -0,0 +1,92 @@
...
+    args = parser.parse_args()
+
+    # Parse args and checks
+    fastq_dir = args.fastq_dir
+    sample_name = args.sample_name
+    staged_dir = args.staged_dir
```
jashapiro: This step is not needed: just use the `args.` directly.

jashapiro (commented):
My main comment remains the one about not getting too fancy: Just have the Nextflow code validate that there are files you will be able to convert, but then let Python handle everything about the conversion. Part of that simplification will be to always make symlinks and use a constant directory name for them. 


Looks like you may have already partly responded to that. I will look again, but I had a few other style-type comments as well.

sjspielman:
> Looks like you may have already partly responded to that. I will look again, but I had a few other style-type comments as well.

Yes, sorry, mid work now on your opening comment. Other changes planned to land as well! I'll tag you back in.

bin/prepare_spaceranger_fastqs.py
sjspielman: Can I counter with `prepare_10x_fastqs.py`?

jashapiro: Sure, but isn't mine more fun?

sjspielman: 🤠

bin/prepare_spaceranger_fastqs.py
sjspielman: > Expanding this to capture all the possible parts (Sample index and lane number) is something we could do as well (to allow for the expected spaceranger format as a no-op)

Unless you feel really strongly, I'm hesitant to do this because of how long and increasingly inscrutable the regex gets. I'm updating this now (pending testing) to always symlink into `staged_fastq`, just only rename if it matches ALLOWED_PATTERN vs a separately defined CONFORMANT_PATTERN regex.

jashapiro: I actually feel pretty strongly about named groups, especially for predefined regexes. For readability, a multiline regex can be commented, e.g.:
```
re.compile(
    r"""
    ^(?P<sample>.+) # capture the initial prefix
    _(?P<read>R?[12]) # read of form R1
    (?:_(?P<lane>\d{3}))?  # optional lane, captures digits only
    \.fastq\.gz$ # suffix
    """,
    re.VERBOSE,
)
```

sjspielman: I am keeping the named groups for sure, just not combining regexes for both the "need to change prefixes" and "already fine for spaceranger prefixes" which is what I thought you were asking here.

jashapiro: Oh yeah, that should be fine.

sjspielman:
This has been updated/simplified and has been run through successfully with variously named input files - ready for review! @jashapiro

bin/prepare_10x_fastqs.py
```diff
@@ -0,0 +1,106 @@
...
+# sample_name is printed to stdout.
+#
+# Exits non-zero if files don't match either recognized pattern.
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+CONFORMANT_PATTERN = re.compile(r"^(.+)_S\d+_(?:L\d+_)?(?:R[12]|I[12])_001\.fastq\.gz$")
```
jashapiro: No need for non-capture groups since you just get the first for this one. Also we want to accommodate multiple files from the same library in case they were split. 
```suggestion
CONFORMANT_PATTERN = re.compile(r"^(.+)_S\d+_(L\d+_)?[RI][12]_\d{3}\.fastq\.gz$")
```

modules/spaceranger.nf
```diff
@@ -253,7 +246,11 @@ workflow spaceranger_quant{
       // add sample names and spatial output directory to metadata
       .map{ meta_in ->
         def meta = meta_in.clone()
-        meta.cr_samples = getCRsamples("${meta.files_directory}/fastq")
+        def sr_pattern = /^.+_S\d+_(L\d+_)?(R[12]|I[12])_001\.fastq\.gz$/
+        def allowed_pattern = /^.+_(R?[12])(_\d{3})?\.fastq\.gz$/
+        meta.fastq_format_recognized = files("${meta.files_directory}/fastq/*.fastq.gz").every { f ->
+          (f.name =~ sr_pattern) || (f.name =~ allowed_pattern)
+        }
```
jashapiro: I don't think you need both patterns? 
```suggestion
        def allowed_pattern = /^.+_(R?[12])(_\d{3})?\.fastq\.gz$/
        meta.fastq_format_recognized = files("${meta.files_directory}/fastq/*.fastq.gz")
          .every { f -> f.name =~ allowed_pattern }
```

modules/spaceranger.nf
```diff
@@ -263,6 +260,7 @@ workflow spaceranger_quant{
         def stored_ref_assembly = getMetaVal(file("${it.spaceranger_results_dir}/scpca-meta.json"), "ref_assembly")
         def stored_tech = getMetaVal(file("${it.spaceranger_results_dir}/scpca-meta.json"), "technology") ?: ""
         // branch for invalid cases
+        unrecognized_fastq_format: !it.fastq_format_recognized
```
jashapiro: rather than adding another value to `meta`, why don't we move the pattern test here to the `.branch` where it is used?

bin/prepare_10x_fastqs.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Create directory where we will stage all symlinks
+    args.staged_dir.mkdir()
+
+    if all(CONFORMANT_PATTERN.match(f.name) for f in fastqs):
+        # Files are already conformant; symlink with original names and return the unique sample prefixes
+        prefixes = set(CONFORMANT_PATTERN.match(f.name).group(1) for f in fastqs)
+        for f in fastqs:
+            (args.staged_dir / f.name).symlink_to(f.absolute())
+        print(",".join(sorted(prefixes)))
```
jashapiro: Returning values via STDOUT is potentially a bit brittle but I do think it seems reasonable enough here. But I would move the actual print statement out of the if/else and have just one. I was a little confused by what was going on here and whether there might be two prints to stdout.

bin/prepare_10x_fastqs.py
```diff
@@ -0,0 +1,106 @@
...
+    else:
+        # Group files by original sample prefix; each group gets its own lane
+        groups = {}
+        for f in fastqs:
+            orig_prefix, read_pair = parse_allowed_fastq(f.name)
+            groups.setdefault(orig_prefix, {})[read_pair] = f
+
+        # Create the symlinks with new names according to the 10x Genomics convention
+        for lane, orig_prefix in enumerate(groups, start=1):
```
jashapiro: Do they need to be separate lanes? If the prefixes are unique (which you are assuming by making a dict) there is no risk of overlap: Can you not just use the `orig_prefix` and return them as a list the same way as with the conformant names? It would seem that this would make tracking a bit easier than renaming them all to `sample_name` (you also save an input arg).  Just add  `_S1_`  and `_001` since the lane is optional, and be done?

jashapiro (commented):
The updates look good so far. The main theme of this review is that we can still go simpler! (I do find that Claude will often overcomplicate things in its desire to make things work. There is a `/simplify` skill that sometimes helps it revise its work, but not always (and it tends to remove comments, which I don't like))

The two main simplifications I am suggesting:
- Do we need to rename to sample name? Why not just keep the original prefix and simply insert `_S1_` and `_001` to make the files conformant? Then we don't have to worry about different lanes for uniqueness. It also makes the two code paths more similar and we can consolidate the print to stdout to a single statement.
- Move the file test logic to the `.branch` statement so we don't have to add anything to `meta` (there is also a small simplification of the name test that I think is okay?)

sjspielman:
> (and it tends to remove comments, which I don't like))

I have actually added a statement (aka in CLAUDE.md) to a personal project I use claude for about this. It's a menace about deleting my comments, and we know how I love comments.

bin/prepare_10x_fastqs.py
sjspielman: My read of the space ranger docs was that the `001` suffix specifically (and not other numbers) is expected, but I haven't tested this - I'll test as part of responding to reviews.

sjspielman: It's indeed more flexible than than 001!

sjspielman:
Much simplified and tested ✅ 
I also changed the name 😄

bin/prepare_ranger_fastqs.py
```diff
@@ -0,0 +1,96 @@
...
+# The comma-separated list of sample prefixes is printed to stdout.
+# Exits non-zero if any file does not match either recognized pattern.
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+CONFORMANT_PATTERN = re.compile(r"^(.+)_S\d+_(L\d+_)?[RI][12]_\d{3}\.fastq\.gz$")
+ALLOWED_PATTERN = re.compile(r"^(?P<prefix>.+)_(?P<read>R?[12])(?:_\d{3})?\.fastq\.gz$")
```
jashapiro: Maybe allow `I` reads?
```suggestion
ALLOWED_PATTERN = re.compile(r"^(?P<prefix>.+)_(?P<read>[RI]?[12])(?:_\d{3})?\.fastq\.gz$")
```

modules/spaceranger.nf
```diff
@@ -262,7 +252,13 @@ workflow spaceranger_quant{
       .branch{ it ->
         def stored_ref_assembly = getMetaVal(file("${it.spaceranger_results_dir}/scpca-meta.json"), "ref_assembly")
         def stored_tech = getMetaVal(file("${it.spaceranger_results_dir}/scpca-meta.json"), "technology") ?: ""
+
+        def allowed_pattern = /^.+_(R?[12])(_\d{3})?\.fastq\.gz$/
```
jashapiro: You definitely allow this for space ranger, so we should not die if we find `I` reads.
```suggestion
        def allowed_pattern = /^.+_([RI]?[12])(_\d{3})?\.fastq\.gz$/
```

jashapiro (approved):
Looks good. Just one consistency thing to allow `I` reads throughout.

PR #1304, opened 2026-06-15
Add merged object information to reference SCE and AnnData files

### LINKED ISSUE #1195: Formatting reference file: Merged objects

allyhawkins:
Closes #1195 

This PR updates the existing reference files for SCE and AnnData objects to now include a slot for merged objects. I would like to use the same scripts that we have set up for checking these objects, which would mean just providing these reference files and the appropriate object type, which in this case is `merged`. 

- The cell metadata should have everything present in the processed objects, plus some additional sample metadata information that we ported over from the `sample_metadata` that was originally saved in `metadata(sce)`. 
- The feature metadata is the same, except we now have `library_id-mean` (`library_id.mean` in python) and `library_id-detected`. I think what we need to do is update the script to check all these columns if the object type is merged rather than try and address that in the reference file here. 
- The experiment metadata now only has `sample_id`, `library_id`, `library_metadata` (which is a list of metadata for each library), and then `merged_highly_variable_genes` for SCE objects only. For the `library_metadata`, I chose not to check each of the individual slots since that should be the same as what's in the processed metadata for each object. Again, this is something I think we can adjust in the script itself if we want to check this information.

I think everything else was fairly straightforward, except I noticed that I was manually using `category` as the type for the AnnData specific cell metadata columns. Using `string` consistently for all cell metadata column types that are characters in R seems to be okay. I did a few manual checks using our dtype checker and it worked. The only time we should see `category` is for factors.

### DIFF
--- a/references/scripts/create-formatting-reference.py
+++ b/references/scripts/create-formatting-reference.py
@@ -13,7 +13,7 @@ anndata_ref_file = "../anndata-formatting-reference.json"
 # --------------------- SCE ---------------------------
 # assays
 assays = ["counts", "spliced"]
-processed_assays = ["counts", "spliced", "logcounts"]
+processed_assays = ["logcounts"]
 adt_assays = ["counts"]
 cellhash_assays = ["counts"]
 
@@ -430,7 +430,7 @@ filtered_sce = {
 # build processed SCE  ------
 
 processed_sce = {
-    "assayNames": processed_assays,
+    "assayNames": assays + processed_assays,
     "colData": filtered_cell_metadata,
     "rowData": feature_metadata,
     "colData_conditional": processed_cell_metadata_conditional,
@@ -457,7 +457,7 @@ processed_sce = {
 # build merged sce -----
 
 merged_sce = {
-    "assayNames": processed_assays,
+    "assayNames": assays + processed_assays,
     "colData": merged_cell_metadata,
     "rowData": feature_metadata,
     "colData_conditional": processed_cell_metadata_conditional,
--- a/references/scripts/create-formatting-reference.py
+++ b/references/scripts/create-formatting-reference.py
@@ -139,7 +139,8 @@ merged_cell_metadata = {
 
 # row metadata ----------
 # this is the same for all object types
-# for merged objects, check that library_id-mean and library_id-detected are correct in the script
+# for merged objects, check that {library_id}-mean and {library_id}-detected are correct in the script
+# these will use the same types as `mean` and `detected`
 feature_metadata = {
     "gene_ids": "character",
     "gene_symbol": "character",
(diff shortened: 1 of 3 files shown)


references/scripts/create-formatting-reference.py
```diff
@@ -13,7 +13,7 @@
 # --------------------- SCE ---------------------------
 # assays
 assays = ["counts", "spliced"]
-processed_assays = ["logcounts"]
+processed_assays = ["counts", "spliced", "logcounts"]
```
jashapiro: I had assumed we were merging `assays` and `processed_assays`, but either way is really fine.

references/scripts/create-formatting-reference.py
```diff
@@ -105,8 +105,41 @@
...
+    "disease_ontology_term_id": "character",
+    "tissue_ontology_term_id": "character",
+    "tech_version": "character",
+    "assay_ontology_term_id": "character",
+    "suspension_type": "character",
+}
+
 # row metadata ----------
 # this is the same for all object types
+# for merged objects, check that library_id-mean and library_id-detected are correct in the script
```
jashapiro: Just to be more clear that these are variables, and not the actual data, and a note about the types that these should be. 
```suggestion
# for merged objects, check that {library_id}-mean and {library_id}-detected are correct in the script
# these will use the same types as `mean` and `detected`
```

references/scripts/create-formatting-reference.py
```diff
@@ -375,7 +429,7 @@
 # build processed SCE  ------
 
 processed_sce = {
-    "assayNames": assays,
+    "assayNames": processed_assays,
```
jashapiro: I guess this is what I would have expected from my earlier comment, but again, either way is fine, which is why I am not making this a suggestion. 
```
    "assayNames": assays + processed_assays,
```

references/scripts/create-formatting-reference.py
```diff
@@ -601,6 +684,14 @@ def convert_experiment_metadata_types(metadata):
     copy.deepcopy(processed_experiment_metadata_conditional)
 )
 
+# merged metadata is the same as sce but no library_metadata list
+# these are character vectors in R which get converted to numpy.ndarrays in Python
+merged_uns_metadata = {
+    "library_id": "numpy.ndarray",
+    "sample_id": "numpy.ndarray",
```
jashapiro: I thought we had the ability to check the types within `ndarrays` too?

https://github.com/AlexsLemonade/scpca-nf/blob/0fe77c915cef898cd8b3b2baf32c044863cdabd9/bin/anndata_formatting_checks.py#L59-L65

jashapiro (commented):
Overall this looks good. My one real question is about the use of `ndarray` vs. the actual type for the uns fields. I would have thought the code we have is able to test the types within those arrays?

references/scripts/create-formatting-reference.py
allyhawkins: We sure do, thanks for catching this. I updated these entries to now use `string` instead. This should be ready for another look.

jashapiro (approved):
👍🏼

PR #1305, opened 2026-06-17
Add process for checking format of merged objects

### LINKED ISSUE #1168: Add process for checking merged objects
We will also want to check formatting for merged objects. I think we should be able to use the same scripts that we write for individual objects to do this, but we will still need to add a process to the merge workflow for checking. When we get to this we might expand on what else is needed, but I'm filing this as a placeholder for now.

allyhawkins:
Closes #1168 

This PR adds the processes to the merge workflow to check and report formatting errors. The actual process is the same as what happens in the main workflow, but instead of extracting the object type from the input file, we just use merged as the object type. 

The trickiest part here was figuring out how to deal with checking the `libraryid-mean` and `libraryid-detected` columns. I ended up adding a helper function that finds all `mean` and `detected` columns and expands the reference file to include the exact column names that are present and then do the checking. This is done for both the SCE and AnnData checks. 

When I first ran this through, I was getting this output: 
```
Formatting errors found for SCPCL000001 merged SingleCellExperiment:
Formatting errors found for SCPCL000002 merged SingleCellExperiment:
Formatting errors found for SCPCL000003 merged SingleCellExperiment:

Missing 'organism_ontology_term_id' from main SCE colData.

Formatting errors found for ['SCPCL000001' 'SCPCL000002' 'SCPCL000003'] merged rna AnnData object:

Missing 'organism_ontology_term_id' from RNA AnnData obs.
Missing 'is_primary_data' from RNA AnnData obs.
Missing 'feature_is_filtered' from RNA AnnData var.
```

So I updated the scripts to pass in the `object_id` which is either the library ID or project ID and edited the header for the error messages so we don't print out a bunch of ids. 

I also fixed the reference file to remove `is_primary_data` and `feature_is_filtered` from the AnnData merged objects only and fixed the typo for `organism_ontology_term_id`, which is just `organism_ontology_id`. For the project that I tested (SCPCP000001), there are no formatting errors now.

### DIFF
--- a/bin/anndata_formatting_checks.py
+++ b/bin/anndata_formatting_checks.py
@@ -208,7 +208,9 @@ def expand_merged_var_ref(var, ref_var):
     mean_type = ref_var.get("mean")
     detected_type = ref_var.get("detected")
 
-    expanded = dict(ref_var)
+    # remove mean and detected so they don't get flagged as missing
+    expanded = {k: v for k, v in ref_var.items() if k not in ("mean", "detected")}
+
     for col in var.columns:
         if col.endswith(".mean"):
             expanded[col] = mean_type
--- a/bin/anndata_formatting_checks.py
+++ b/bin/anndata_formatting_checks.py
@@ -206,7 +206,7 @@ def expand_merged_var_ref(var, ref_var):
     with the same expected types as `mean` and `detected`.
     """
     expanded = ref_var.copy()
-    
+
     # get mean and detected values, while removing them so they don't get flagged as missing
     mean_type = expanded.pop("mean")
     detected_type = expanded.pop("detected")
--- a/bin/anndata_formatting_checks.py
+++ b/bin/anndata_formatting_checks.py
@@ -196,7 +196,29 @@ def get_conditionals(adata):
     return set(cond for cond, is_true in condition_tests.items() if is_true)
 
 
-def check_anndata(adata, ref, label):
+def expand_merged_var_ref(var, ref_var):
+    """
+    Expand the var reference for merged AnnData objects to include per-library columns.
+
+    For merged objects, var contains `libraryid.mean` and `libraryid.detected`
+    columns for each library.
+    This function finds those columns and adds them to the reference
+    with the same expected types as `mean` and `detected`.
+    """
+    mean_type = ref_var.get("mean")
+    detected_type = ref_var.get("detected")
+
+    expanded = dict(ref_var)
+    for col in var.columns:
+        if col.endswith(".mean"):
+            expanded[col] = mean_type
+        elif col.endswith(".detected"):
+            expanded[col] = detected_type
+
+    return expanded
+
+
+def check_anndata(adata, ref, label, object_type):
     """Run all formatting checks for an AnnData object."""
     errors = []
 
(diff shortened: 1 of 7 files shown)


bin/anndata_formatting_checks.py
```diff
@@ -196,7 +196,31 @@ def get_conditionals(adata):
...
+    For merged objects, var contains `libraryid.mean` and `libraryid.detected`
+    columns for each library.
+    This function finds those columns and adds them to the reference
+    with the same expected types as `mean` and `detected`.
+    """
+    mean_type = ref_var.get("mean")
+    detected_type = ref_var.get("detected")
+
+    # remove mean and detected so they don't get flagged as missing
+    expanded = {k: v for k, v in ref_var.items() if k not in ("mean", "detected")}
```
jashapiro: You can do both of these steps at the same time, while saving the parsing through the whole list.
```suggestion
    expanded = ref_var.copy()
    
    # get mean and detected values, while removing them so they don't get flagged as missing
    mean_type = expanded.pop("mean")
    detected_type = expanded.pop("detected")
```

bin/sce_formatting_checks.R
```diff
@@ -204,10 +237,16 @@ option_list <- list(
     default = "",
     help = "Path to RDS file containing a SingleCellExperiment object from scpca-nf"
   ),
+  make_option(
+    opt_str = c("--object_id"),
+    type = "character",
+    default = "",
+    help = "Identifier for the object (e.g. library ID) to include in error messages"
+  ),
```
jashapiro: I'd kind of rather the script just picked this, rather than making the user do it. Can we just use project_id for merged data?

jashapiro (commented):
Overall this looks good, but I had a few little comments, with the biggest mostly about trying to avoid adding an extra argument (object_id) if we can. I only made the comment in one place, but the same logic would apply generally.

allyhawkins:
> Overall this looks good, but I had a few little comments, with the biggest mostly about trying to avoid adding an extra argument (object_id) if we can. I only made the comment in one place, but the same logic would apply generally.

I updated this to remove the `object_id` argument and instead grab the project id from the merged file name. We don't store the `project_id` in the metadata of the object, so this seemed like the best way to do it. This should be ready for another look.

jashapiro:
> I updated this to remove the `object_id` argument and instead grab the project id from the merged file name. We don't store the `project_id` in the metadata of the object, so this seemed like the best way to do it. This should be ready for another look.

Seems like an oversight, but not for now.

references/scripts/create-formatting-reference.py
```diff
@@ -627,6 +630,11 @@ def convert_experiment_metadata_types(metadata):
 # highly variable is only in the processed object
 processed_var_metadata = {**unfiltered_var_metadata, "highly_variable": "bool"}
 
+# merged var contains everything but feature_is_filtered
+merged_var_metadata = {
+    k: v for k, v in processed_var_metadata.items() if k != "feature_is_filtered"
+}
```
jashapiro: Moving to pop here too, potentially (with the second None arg in case `feature_is_filtered` is not defined.
```suggestion
merged_var_metadata = processed_var_metadata.copy()
merged_var_metadata.pop("feature_is_filtered", None)
```

jashapiro (approved):
LGTM

PR #1307, opened 2026-06-23
Allow for extra columns to be added besides submitter cell type annotations

### LINKED ISSUE #1306: Accommodate more columns in the submitter cell types file
We are planning to add in additional information for cells based on items provided by submitters, such as columns containing scores for metaprograms. To do this, we need to update the `add_submitter_annotations.R` script to keep all columns that are in the submitter cell types file in addition to the provided cell type annotations. 

For these columns, we will want to have a function to clean up the names to replace `-`, `.`, or spaces with `_` and ensure everything is lowercase. We then will want to prefix the column names to have `submitter_` attached to them.

allyhawkins:
Closes #1306 

This PR modifies the `add_submitter_annotations.R` script to include any extra columns beyond `submitter_celltype_annotation` in all three object types. 

First, I checked for any extra columns that we may want to include in the output beyond the required columns (`scpca_library_id`, `cell_barcode` and `cell_type_assignment`), the optional `CL_ontology_id` column, and any extra identifying columns that might have been included in the submitter cell types file. This last filtering step was included because in looking at some of the existing submitter cell types file, at least one of them has a few extra columns that we would not want to include as those will cause joining issues later on. 

Any of the extra columns get renamed and reformatted: 

- all lowercase
- dashes, periods, and spaces are replaces with underscores
- prefix with `submitter_`

For all of the columns being added, including the `submitter_celltype_annotation` column, I changed the value to `Submitter-excluded` if the barcode isn't present in the submitters file. Previously this used to check for `NA` values, but we don't know what the contents of the other columns are and we really could have an `NA` which is different from the submitter didn't even provide information for that cell. 

Finally, while I was here, I accounted for the optional `CL_ontology_id` column, which we actually have for one of our submitted datasets but have previously not included in the objects. I treated this one a little differently since we explicitly want to name it `submitter_celltype_ontology`. 

I tested this locally with a modified submitter cell types file with extra columns and things worked as expected. It might be worth doing a full workflow test with an existing object and submitter annotations to ensure we haven't broken anything, but I'll need to add a library with submitter annotations to our testing sandbox in order to do that. I am happy to do this if reviewers think it is worth it.

### DIFF
--- a/bin/add_submitter_annotations.R
+++ b/bin/add_submitter_annotations.R
@@ -79,7 +79,7 @@ extra_cols <- setdiff(
 renamed_extra_cols <- extra_cols |>
   tolower() |>
   stringr::str_replace_all("[\\-\\.\\s]+", "_") |>
-  (\(x) paste0("submitter_", x))()
+  (\(x) paste0("submitter_data_", x))()
 
 # Build a named vector for renaming: old name -> new name
 extra_col_rename <- setNames(extra_cols, renamed_extra_cols)
--- a/bin/add_submitter_annotations.R
+++ b/bin/add_submitter_annotations.R
@@ -28,6 +28,17 @@ option_list <- list(
 
 opt <- parse_args(OptionParser(option_list = option_list))
 
+# define column lists for required and optional
+required_cols <- c(
+  "scpca_library_id",
+  "cell_barcode",
+  "cell_type_assignment"
+)
+
+optional_cols <- c(
+  "CL_ontology_id"
+)
+
 # check that output file name ends in .rds
 if (!(stringr::str_ends(opt$sce_file, ".rds"))) {
   stop("SingleCellExperiment file name must end in .rds")
@@ -52,26 +63,28 @@ submitter_df <- readr::read_tsv(
 )
 
 # Check required columns before proceeding for faster failing:
-if (!all(c("cell_barcode", "cell_type_assignment") %in% names(submitter_df))) {
-  stop("The submitter TSV file must contain columns `cell_barcode` and `cell_type_assignment`.")
+if (!all(required_cols %in% names(submitter_df))) {
+  stop(
+    glue::glue(
+      "The submitter TSV file must contain the following columns: {paste0(required_cols, collapse = ',')}."
+    )
+  )
 }
 
 # Now that we are confident to proceed, read in the sce
 sce <- readr::read_rds(opt$sce_file)
+
+# define list of columns to exclude based on columns in the sample metadata
+# anything that's in the sample metadata that's not required must be excluded
+sample_df_cols <- names(metadata(sce)$sample_metadata)
+exclude_cols <- sample_df_cols[!required_cols]
 
 # Identify any extra columns beyond the required ones in the submitters file
 # and CL ontology Id which will get its own name if present
 # specify any columns that might be present that we don't need to include in the output object (e.g., scpca_sample_id or submitter_id)
 extra_cols <- setdiff(
   names(submitter_df),
-  c(
-    "scpca_library_id",
-    "submitter_id",
-    "scpca_sample_id",
-    "cell_barcode",
-    "cell_type_assignment",
-    "CL_ontology_id"
-  )
+  c(required_cols, exclude_cols)
 )
 
 # Rename extra columns with `submitter_` prefix, replacing dashes/periods/spaces
--- a/bin/add_submitter_annotations.R
+++ b/bin/add_submitter_annotations.R
@@ -66,7 +66,7 @@ submitter_df <- readr::read_tsv(
 if (!all(required_cols %in% names(submitter_df))) {
   stop(
     glue::glue(
-      "The submitter TSV file must contain the following columns: {paste0(required_cols, collapse = ',')}."
+      "The submitter TSV file must contain the following columns: {paste0(required_cols, collapse = ', ')}."
     )
   )
 }
@@ -84,7 +84,7 @@ exclude_cols <- setdiff(sample_df_cols, c(required_cols, optional_cols))
 # specify any columns that might be present that we don't need to include in the output object (e.g., scpca_sample_id or submitter_id)
 extra_cols <- setdiff(
   names(submitter_df),
-  c(required_cols, optional_cols, exclude_cols)
+  c(required_cols, optional_cols, sample_df_cols)
 )
 
 # Rename extra columns with `submitter_` prefix, replacing dashes/periods/spaces
--- a/bin/add_submitter_annotations.R
+++ b/bin/add_submitter_annotations.R
@@ -28,6 +28,10 @@ option_list <- list(
 
 opt <- parse_args(OptionParser(option_list = option_list))
 
+opt$sce_file <- "~/Documents/ALSF/git_repos/ews-nf/scpca-nf-out/SCPCP000015/SCPCS000490/SCPCL000822_processed.rds"
+opt$submitter_cell_types_file <- "../scpca-nf/SCPCL000822-SCPCS000490_annotations.tsv"
+opt$library_id <- "SCPCL000822"
+
 # check that output file name ends in .rds
 if (!(stringr::str_ends(opt$sce_file, ".rds"))) {
   stop("SingleCellExperiment file name must end in .rds")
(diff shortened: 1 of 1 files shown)


bin/add_submitter_annotations.R
```diff
@@ -28,6 +28,10 @@ option_list <- list(
 
 opt <- parse_args(OptionParser(option_list = option_list))
 
+opt$sce_file <- "~/Documents/ALSF/git_repos/ews-nf/scpca-nf-out/SCPCP000015/SCPCS000490/SCPCL000822_processed.rds"
+opt$submitter_cell_types_file <- "../scpca-nf/SCPCL000822-SCPCS000490_annotations.tsv"
+opt$library_id <- "SCPCL000822"
+
```
jashapiro: Guessing you meant to remove these before PR
```suggestion
```

bin/add_submitter_annotations.R
```diff
@@ -51,26 +55,84 @@ submitter_df <- readr::read_tsv(
   na = character()
 )
 
-# Check columns before proceeding for faster failing:
-# TODO: WHAT IF THEY PROVIDE AN ONTOLOGY COLUMN? WE NEED TO PARSE THAT TOO. WILL NEED MORE IF BELOW.
+# Check required columns before proceeding for faster failing:
```
jashapiro, line 65: I wonder if we want to require that they include the library id too? I feel like it would potentially be more convenient to have that as a double check?

bin/add_submitter_annotations.R
```diff
@@ -51,26 +55,84 @@ submitter_df <- readr::read_tsv(
...
+# specify any columns that might be present that we don't need to include in the output object (e.g., scpca_sample_id or submitter_id)
+extra_cols <- setdiff(
+  names(submitter_df),
+  c(
+    "scpca_library_id",
+    "submitter_id",
+    "scpca_sample_id",
+    "cell_barcode",
+    "cell_type_assignment",
+    "CL_ontology_id"
```
jashapiro: I am worried about the maintainability of this list over time, since we don't know what a user might include. I have a few thoughts: 

- First, I would separate expected/required columns from optional ones
- Define those two lists of columns at the top of the script as constants, so you can use the expected columns both here and in the expected column check
- Separately, and more ambitiously/more fraught: maybe we exclude all columns already present in the SCE? I can see this going wrong too, so I am not really sold on it fully; I just feel like we might need something more that just the ones we have seen so far.

bin/add_submitter_annotations.R
```diff
@@ -51,26 +55,84 @@ submitter_df <- readr::read_tsv(
...
+  )
+)
+
+# Rename extra columns with `submitter_` prefix, replacing dashes/periods/spaces
+# with underscores and converting to lower case
+renamed_extra_cols <- extra_cols |>
+  tolower() |>
+  stringr::str_replace_all("[\\-\\.\\s]", "_") |>
```
jashapiro: In case there are multiple separators, we can collapse them too?
```suggestion
  stringr::str_replace_all("[\\-\\.\\s]+", "_") |>
```

bin/add_submitter_annotations.R
```diff
@@ -51,26 +55,84 @@ submitter_df <- readr::read_tsv(
...
+  (\(x) paste0("submitter_", x))()
+
+# Build a named vector for renaming: old name -> new name
+extra_col_rename <- setNames(extra_cols, renamed_extra_cols)
```
jashapiro, line 96: I was going to suggest we would want to go the other way here, but then I saw how you were using it with rename, so it seems this is correct. I usually don't like `!!!`, but this does seem a good use.

bin/add_submitter_annotations.R
```diff
@@ -51,26 +55,84 @@ submitter_df <- readr::read_tsv(
...
   ) |>
-  dplyr::distinct()
+  # rename extra columns with submitter_ prefix
+  dplyr::rename(!!!extra_col_rename) |>
+  dplyr::distinct() |>
+  dplyr::relocate(
+    # make sure barcodes and submitter celltypes are the first columns
+    c("barcodes", "submitter_celltype_annotation"),
+    .before = 0
+  )
```
jashapiro: Since you use `select()` right after this, I think you can skip the `relocate()`.
```suggestion
  dplyr::distinct() 
```

bin/add_submitter_annotations.R
```diff
@@ -51,26 +55,84 @@ submitter_df <- readr::read_tsv(
...
+    "CL_ontology_id"
+  )
+)
+
+# Rename extra columns with `submitter_` prefix, replacing dashes/periods/spaces
+# with underscores and converting to lower case
+renamed_extra_cols <- extra_cols |>
+  tolower() |>
+  stringr::str_replace_all("[\\-\\.\\s]", "_") |>
+  (\(x) paste0("submitter_", x))()
```
jashapiro: I wonder if we might want to use a double underscore or something else to make these a bit more distinct from the cell type columns? Maybe `submitter_data_` or `submitter_info_` or `submitter_extra_`? This suggestion might be a bit late, if engineering already has expectations, but I though I would throw it out there anyway.
```suggestion
  (\(x) paste0("submitter_data_", x))()
```

jashapiro (commented):
This looks good overall, but we might want to think a bit more about how we decide if a column should be included or not. I am worried that maintaining a list of columns will be pretty brittle, though my other suggestion (excluding any existing SCE coldata column) might also have problems of its own. I'd be curious to hear your thoughts.

I also had a prefix thought that might be too late.

bin/add_submitter_annotations.R
allyhawkins: These columns will only be present in the objects and won't get surfaced to engineering so we can name them whatever we want. I like the idea of making these separate from the cell type columns and this reminds me to check on how we grab submitter cell types for the report because I think we just look for columns with "submitter" in them. 

I do like `submitter_data` though, so I'll add that suggestion in for now.

bin/add_submitter_annotations.R
allyhawkins: > Separately, and more ambitiously/more fraught: maybe we exclude all columns already present in the SCE? I can see this going wrong too, so I am not really sold on it fully; I just feel like we might need something more that just the ones we have seen so far.

So we can't do this option exactly because these are the unfiltered objects. The main problem with having these columns here is that before objects get merged we add the contents of `metadata(sce)$sample_metadata` as columns in the `colData`. So if we have duplicates of those columns here, then we would have a problem. One option is to exclude any columns already present in `sample_metadata`? 

Another option is to remove any extraneous columns from the file itself as a step when we add new submitter annotations or new contributed data. This also could be error-prone though, so I'm not sure which one I like better. 

We could temporarily just separate them into expected and required columns as you mentioned and then leave a TODO to revisit?

allyhawkins: Okay, I ended up doing a little bit of a combination of these. I first define the vectors of required and optional columns to include. I then define a vector of columns to exclude by taking the difference in the required/optional columns and the columns present in the sample metadata. I think it's safe to say we don't want to include those columns if they are defined in our main metadata. 

This should be ready for another look.

bin/add_submitter_annotations.R
```diff
@@ -51,27 +62,82 @@ submitter_df <- readr::read_tsv(
...
 
-# Check columns before proceeding for faster failing:
-# TODO: WHAT IF THEY PROVIDE AN ONTOLOGY COLUMN? WE NEED TO PARSE THAT TOO. WILL NEED MORE IF BELOW.
-if (!all(c("cell_barcode", "cell_type_assignment") %in% names(submitter_df))) {
-  stop("The submitter TSV file must contain columns `cell_barcode` and `cell_type_assignment`.")
+# Check required columns before proceeding for faster failing:
+if (!all(required_cols %in% names(submitter_df))) {
+  stop(
+    glue::glue(
+      "The submitter TSV file must contain the following columns: {paste0(required_cols, collapse = ',')}."
```
jashapiro:
```suggestion
      "The submitter TSV file must contain the following columns: {paste0(required_cols, collapse = ', ')}."
```

bin/add_submitter_annotations.R
```diff
@@ -51,27 +62,82 @@ submitter_df <- readr::read_tsv(
...
+# anything that's in the sample metadata that's not required must be excluded
+sample_df_cols <- names(metadata(sce)$sample_metadata)
+exclude_cols <- setdiff(sample_df_cols, c(required_cols, optional_cols))
+
+# Identify any extra columns beyond the required ones in the submitters file
+# and CL ontology Id which will get its own name if present
+# specify any columns that might be present that we don't need to include in the output object (e.g., scpca_sample_id or submitter_id)
+extra_cols <- setdiff(
+  names(submitter_df),
+  c(required_cols, optional_cols, exclude_cols)
```
jashapiro: Since this is a set operation, you don't really need the previous line where you removed the required and optional cols from the sample_df_cols: you can just use that straight here and remove `exclude_cols` completely. (You are removing and adding back the same thing!)
```suggestion
  c(required_cols, optional_cols, sample_df_cols)
```

jashapiro (approved):
Thank you for these updates. I think I like where you ended up. The only comment is that you have a bit of redundant removing then adding back in there. My suggestion only includes one part of the full change, as I figured there would probably be some comment rewriting to go along with the changes that I thought best left to you. But I don't think I should need to see this again.

PR #1308, opened 2026-06-24
Remove dash in cellranger-multi mapping tool in report

allyhawkins:
In working on the test data, I was getting an error where the `mapping_tool` was not `cellranger-multi` but it's `cellranger multi` with the space. We changed this when writing docs a little while ago and I changed it in the place where we add the information to the metadata, but not in the report: https://github.com/AlexsLemonade/scpca-nf/blob/89d3f11c65900f93d4ea9df4d4018fdbad34e600/bin/generate_unfiltered_sce_cellranger.R#L164

This makes the change in the report. I also searched the repo for other instances and this should be the only one.


jashapiro (approved):
LGTM

PR #1309, opened 2026-06-24
Use the unique id for the filename for formatting checks

allyhawkins:
I found one more small issue while generating test data. We need to use the `unique_id` not `library_id` for the filenames in the formatting checks since at this point they have been renamed.


jashapiro (approved):
👍

PR #1310, opened 2026-06-25
Fix errors in formatting checks

allyhawkins:
Some more errors showing up in test data that we didn't catch before. The main change here is that we need to flatten the list of anndata objects so that we have `[meta, unfiltered rna, unfiltered adt, ... ]` instead of nested lists for each object type. This makes it so we are only passing one file into the anndata formatting checks. I made this fix for both the individual processing and merged objects.


jashapiro (approved):
LGTM

PR #1312, opened 2026-06-26
Fix GRCm38 assembly name to mm10

### LINKED ISSUE #1303: Change assembly name in build-index: GRCm38 -> mm10
> > We could add some hardcoded logic in here to change the assembly name when it's GRCm38 -> mm10
> 
> This seems fine with me. I would just update the workflow to change the assembly within that process if this is the case. I wouldn't mess with the assembly that get's stored within meta, you just want to make sure the name of the output folder that is used when making the index matches.  

 _Originally posted by @allyhawkins in [#1330](https://github.com/AlexsLemonade/ScPCA-admin/issues/1330#issuecomment-4624715224)_

This issue tracks updating the assembly name for this reference to appease space ranger's checks. It won't affect other uses of this reference.

sjspielman:
Closes #1303 

This PR updates the build-index.nf workflow to hardcode in an edge case for spaceranger appeasement. The associated spaceranger probe file records the ref name as `mm10`, so we need to do the same to match. This won't affect other situations that choose to use this reference (although I don't imagine that will be a common choice in general!). 

As you can see from various reverted commits, I tested this out to confirm that the relevant Visium data set we discovered this situation for now initiates successfully. 

One this is approved, I will trigger the `build-index.nf` workflow to run for this reference name in `staging` mode; currently it has been run through `testing` only.

### DIFF
--- a/build-index.nf
+++ b/build-index.nf
@@ -82,9 +82,12 @@ process cellranger_index {
   script:
     cellranger_index = file(meta.cellranger_index).name
     // extract assembly from ref_name
-    // if ref_name is Mus_musculus.GRCm38.98, use mm10 and not GRCm38 to match spaceranger 
-    // probe files; the final output directory will still have GRCm38 in it though
-    assembly = ref_name == "Mus_musculus.GRCm38.98" ? "mm10" : ref_name.split("\\.")[1] 
+    assembly = ref_name.split("\\.")[1] 
+    
+    //  use mm10 instead of GRCm38 to match spaceranger probe files; 
+    // the final output directory will still have GRCm38 in it though
+    if (assembly == "GRCm38") assembly = "mm10"
+``
     """
     gunzip -c ${fasta} > genome.fasta
     gunzip -c ${gtf} > genome.gtf
--- a/build-index.nf
+++ b/build-index.nf
@@ -85,7 +85,7 @@ process cellranger_index {
     assembly = ref_name.split("\\.")[1] 
     
     //  use mm10 instead of GRCm38 to match spaceranger probe files; 
-    // the final output directory will still have GRCm38 in it though
+    // the final output directory will still be named with GRCm38
     if (assembly == "GRCm38") assembly = "mm10"
 ``
     """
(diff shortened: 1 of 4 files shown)


build-index.nf
```diff
@@ -81,7 +81,10 @@ process cellranger_index {
     path cellranger_index
   script:
     cellranger_index = file(meta.cellranger_index).name
-    assembly = ref_name.split("\\.")[1] // extract assembly from ref_name
+    // extract assembly from ref_name
+    // if ref_name is Mus_musculus.GRCm38.98, use mm10 and not GRCm38 to match spaceranger 
+    // probe files; the final output directory will still have GRCm38 in it though
+    assembly = ref_name == "Mus_musculus.GRCm38.98" ? "mm10" : ref_name.split("\\.")[1]
```
jashapiro: I might do this in two steps to be just a bit more clear? The only reason not to do this would be if there are other expected cases where we would not want to use `mm10` (pretty sure not, given the rest). 

```suggestion
    // extract assembly from ref_name
    assembly = ref_name.split("\\.")[1] 
    
    //  use mm10 instead of GRCm38 to match spaceranger probe files; 
    // the final output directory will still have GRCm38 in it though
    assembly = assembly == "GRCm38" ? "mm10" : assembly
```


or maybe a bit more clear:

```suggestion
    // extract assembly from ref_name
    assembly = ref_name.split("\\.")[1] 
    
    //  use mm10 instead of GRCm38 to match spaceranger probe files; 
    // the final output directory will still have GRCm38 in it though
    if (assembly == "GRCm38") assembly = "mm10"
``

File: build-index.nf [hunk 2]
```diff
@@ -81,7 +81,13 @@ process cellranger_index {
     path cellranger_index
   script:
     cellranger_index = file(meta.cellranger_index).name
-    assembly = ref_name.split("\\.")[1] // extract assembly from ref_name
+    // extract assembly from ref_name
+    assembly = ref_name.split("\\.")[1] 
+    
+    //  use mm10 instead of GRCm38 to match spaceranger probe files; 
+    // the final output directory will still have GRCm38 in it though
```
sjspielman:
```suggestion
    // the final output directory will still be named with GRCm38
```

jashapiro (approved):
👍🏼

PR #1315, opened 2026-07-06
Use air for precommit formatting

jashapiro:
Tiny little PR to follow #1260 

Here I am replacing the R precommit hooks with `air`. This should make it run and install updates faster, as the R code was pretty heavy. Note that `air format` will error for unparsable R, so we don't need that check from the package either.


jashapiro:
Hmmm... I wasn't expecting it to actually run on everything, but we can deal with that, I guess.

sjspielman (approved):
👍

PR #1316, opened 2026-07-06
Remove legacy nextflow profiles/config

### LINKED ISSUE #1314: Remove legacy profiles and config
With the removal of the legacy batch infrastructure, we should also remove the profiles and configs associated with it from the scpca-nf workflow.

jashapiro:
Closes #1314

Here I am removing all of the config files and profile settings associated with legacy configurations. I also removed the reference to these profiles in the main nextflow.config. I believe that is the only one left. 

I updated the readme in the config file as that was more than a bit out of date.


allyhawkins (approved):
LGTM

PR #1319, opened 2026-07-21
Fix nf-stub check action

jashapiro:
I made the nextflow version more flexible, keeping it only at `25.10` to allow for point updates.

Also updated other actions along the way


jashapiro:
error is just https://github.com/nf-core/website/issues/4317

PR #1321, opened 2026-07-29
Build CellBrowser in parallel

jashapiro:
With help from Claude, I was able to formulate and implement a plan to parallelize the building of CellBrowser sites.

The key change is that we build the full hierarchy first in the `cellbrowser_project` step and uses that to create a skeleton that is passed with each individual library to the `cellbrowser_dataset` step, which can now run in parallel.

The final  `cellbrowser_site` step, assembles all of those individual `cellbrowser_dataset` results back into the final site.

That last step is still going to be a bit slow as there will be a lot of file copying to do, but the wall clock time improvement should be quite significant since the per-library conversions are the most computationally intensive part of the process.

I have tested this in the testing environment and it seemed to work as expected.
You can see the test run at https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/2iaE11sT9GPPwS/tasks and the output is at s3://ccdl-scpca-nf-results-testing/sandbox/cellbrowser


allyhawkins (approved):
LGTM

PR #1322, opened 2026-07-31
Handle failures in scDblFinder

allyhawkins:
I received a report with the following error after trying to run `scpca-nf` indicating a failure in `scDblFinder` because of too few cells. 

> Creating ~1500 artificial doublets...
Error in fixupDN.if.valid(value, x@Dim) :
  length of Dimnames[[2]] (1) is not equal to Dim[2] (0)
Calls: <Anonymous> ... colnames<- -> dimnames<- -> dimnames<- -> fixupDN.if.valid
In addition: Warning message:
In scDblFinder::scDblFinder(sce, BPPARAM = bp_param, returnType = "table") :
  scDblFinder might not work well with very low numbers of cells.

We don't want the whole workflow to fail if this is the case, so we should handle this failure with a `tryCatch`. I do that here and then set the added `scDblFinder` columns to `NA` in the `colData`. Let me know if you prefer we use something other than NA to indicate an actual failure rather than just NA.

### DIFF
--- a/bin/detect_doublets.R
+++ b/bin/detect_doublets.R
@@ -64,24 +64,37 @@ if (file.size(opt$input_sce_file) == 0) {
 sce <- readr::read_rds(opt$input_sce_file)
 
 # run scDblFinder -----------
-doublet_result <- scDblFinder::scDblFinder(
-  sce,
-  BPPARAM = bp_param,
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
Seems reasonable to me.

Issue #1324, opened 2026-07-31
Prepare for scpca-nf release `v0.10.3`


allyhawkins:
## Steps for a new release of `scpca-nf`

This is just a patch release to include the fix for `scDblFinder` from #1322 

### Preparing for the release

- [x] Are all of the issues planned for this release resolved? If there are any issues that are unresolved, mark this issue as blocked by those in the Science Team GitHub Project.
- [x] Update code and documentation with the latest version number in the `development` branch:
  - [ ] [`nextflow.config`](https://github.com/AlexsLemonade/scpca-nf/blob/main/nextflow.config)
  - [ ] [`external-instructions.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md)
  - [ ] [`cavatica/sb_nextflow_schema.yaml`](https://github.com/AlexsLemonade/scpca-nf/blob/main/cavatica/sb_nextflow_schema.yaml)
  - [ ] [`cavatica/sb_doc.md`](https://github.com/AlexsLemonade/scpca-nf/blob/main/cavatica/sb_doc.md)
- [x] Update the tag for `scpca-nf` containers in `config/containers.config` and `nextflow_schema.json`
  _Note:_ Any processes that use those containers will not work until the new tag has been created.
- [x] Check that the `nextflow_schema.json` is up to date with `pixi run nf-core pipelines schema build`
- [x] Check that [`cavatica/sb_nextflow_schema.yaml`](https://github.com/AlexsLemonade/scpca-nf/blob/main/cavatica/sb_nextflow_schema.yaml) entries are up to date (sadly, no automated tool for this)
- [x] Test that the workflow is in good working order by running the workflow in `testing` mode from the `development` branch using [the `run-scpca-nf.yaml` workflow](https://github.com/AlexsLemonade/ScPCA-admin/blob/main/.github/workflows/run-scpca-nf.yaml)
- [x] File a PR from the `development` branch to the `main` branch. This should include all of the changes that will be associated with the next release.
- [x] (Optional) Generate new example `scpca-nf` output files.
If updating the example output is not necessary for this release, check these boxes off for free.
  - [ ] [Re-process the example data](https://github.com/AlexsLemonade/scpca-nf/blob/main/internal-instructions.md#processing-example-data) through the `scpca-nf` workflow and ensure it looks correct.
[...]

PR #1318, opened 2026-07-21
Update ScPCA citation

allyhawkins:
I updated the citation for ScPCA and added a new badge. I kept the biorxiv badge, but let me know if you think I should remove it.

PR #1301, opened 2026-06-04
Add visium2_v1

sjspielman:
We were missing this sample's (mouse-only) technology - https://github.com/AlexsLemonade/ScPCA-admin/issues/1330 😬

PR #1326, opened 2026-08-04
Merge changes for `v0.10.3` into `main`

allyhawkins:
Towards #1324 

This PR includes the updates to all the tags to use `v0.10.3` and cherry-picked commits from the following PRs which fix handling `scDblFinder` and include fixes for processing public data: 

#1322 
#1288 
#1289 
#1300
#1301 
#1312 
#1313 

I also included some other changes that I thought were worth releasing: 

#1318 which is the updated citation badge
#1308 fixes an error in how we record cellranger multi in the mapping tool 
#1296 allows us to run the metrics report using sso with AWS rather than `op run`


sjspielman (approved):
This looks good to me, but I want to just note that in the release notes we should probably include saying a claude code guideline file was added.

Issue #1279, opened 2026-05-04
Make compatible with strict syntax


jashapiro:
The workflow does not currently run with Nextflow 26.04, failing immediately with the following error: 
```
ERROR ~ Unknown process directive: `pullthroughContainer`

 -- Check script 'modules/af-rna.nf' at line: 7 or see 'stub-run.log' file for more details
```

(see https://github.com/AlexsLemonade/scpca-nf/actions/runs/25338223460/job/74288839350?pr=1278)

I suspect we can fix this particular error by making the `container` directive a closure, rather than a bare interpolated string, but I have yet to confirm this.

For now, the fix can be to set the testing version to `25.10.4`, but we should bite the bullet sooner rather than later to make this forward compatible.

sjspielman:
I just came to open an issue about this and was glad to notice this one is here already! Go team.

PR #1330, opened 2026-08-14
Add max nextflow version

jashapiro:
Because we do not support strict syntax yet, (see #1279), I wanted to put a hard upper bound on versions so we get a version error rather than a mysterious one when the syntax is off.


allyhawkins (approved):
LGTM

Issue #1329, opened 2026-08-14
Accommodate multiome kits barcode list


allyhawkins:
In order to be able to properly process RNA libraries generated using the multiome kit from 10x, we need to include the appropriate barcode list, `737k-arc-v1.txt.gz` (see https://kb.10xgenomics.com/s/article/115004506263-What-is-a-barcode-inclusion-list-formerly-barcode-whitelist). 

This fix involves: 

- [ ] Adding the barcode list to our references bucket
- [ ] Adding a new technology option to indicate when the multiome barcode list should be used 
- [ ] Updating the technology to barcodes file map in https://github.com/AlexsLemonade/scpca-nf/blob/51eb1361b8ccb0da546c91da7c98cd040fdaf918/main.nf#L122

allyhawkins:
Closed by #1333

PR #1333, opened 2026-08-18
Add multiome barcode list

allyhawkins:
Closes #1329 

Here I am updating the dictionary of single-cell technologies and associated barcodes to include the barcode list for Multiome kits (ATAC + GEX) v1. The biggest decision here is what to name the technology. I decided to use `10Xv3.1_multiome_v1`. I wanted to keep the v3.1 in there since the chemistry and UMI/barcode length is the same as 3.1, the only difference is the barcode list. I also included v1, since that's the version specified in the docs here: https://kb.10xgenomics.com/s/article/115004506263-What-is-a-barcode-inclusion-list-formerly-barcode-whitelist

It looks like the only two versions that exist both use this same setup and are both compatible with the barcode list and use v3.1 chemistry. 

As part of this I also added the new list to the references bucket, which can be found in `s3://scpca-nf-references/barcodes/10X/`.


main.nf
```diff
@@ -125,6 +125,7 @@ workflow {
     '10xv2_5prime': '737K-august-2016.txt',
     '10xv3': '3M-february-2018.txt',
     '10xv3.1': '3M-february-2018.txt',
+    '10xv3.1_multiome_v1': '737K-arc-v1.txt.gz',
```
jashapiro, line 128: Just to be clear, it uses 3.1 geometry but only 737K barcodes? Fascinating.

jashapiro (approved):
LGTM!

main.nf
allyhawkins: Yes that is correct.

PR #1334, opened 2026-08-19
Add timeout for stub runs

jashapiro:
While reviewing #1333, I noticed that the stub run again timed out for unknown reasons, which it seems to do periodically and stochastically. So here I am making that happen much faster. I considered adding some retry logic, but that seemed like overkill.

I also added a little fix so the workflow doesn't error again when uploading logs if not all steps completed (the error from the failed step will still surface!).
Finally, I bumped the nextflow version.


allyhawkins (approved):
LGTM

PR #1335, opened 2026-08-19
Update tag and add to changelog for release

allyhawkins:
Towards #1331 

Here I'm updating the tags and then adding a brief Changelog entry for what's being included in this release. Once this goes in I'll file a PR to `main` cherry-picking the commit from this PR and #1333 and then do the release.

PR #1336, opened 2026-08-19
Merge changes for v0.10.4 into main

allyhawkins:
Follow up to #1335, this adds in the changes for the new multiome barcode list and the tag updates so we can do the release. Note I also included the version pinning updates for Nextflow since that seemed helpful to have and showed up when I had to resolve a conflict.


jashapiro (approved):
LGTM

PR #1340, opened 2026-09-02
Account for scenarios with all unclassified cells for a given cell typing method

allyhawkins:
When I was working on generating test data with `--perform_celltyping` on, I was getting a weird error that the faceted UMAPs couldn't be created because there needed to be at least one factor level. It turns out one of the test libraries is a small library with only a few cells and all the cells were getting labeled as "Unclassified". Currently, we remove cells that are "Unclassified" for a single method when making the plots in the report. What this means is that if a cell is "Unclassified" in `SingleR` but it has a label for another method, it gets removed from all the plots. I think what we mean to do is to just not consider that cell for `SingleR` but still include it elsewhere. 

So I updated the cell type report to not remove these, but set them to `NA`. I then updated the plots to not show any cells that are `NA`. 

The other thing I changed, which was the actual cause of this error, was account for cases where all cells are unclassified for a given method and there's nothing to plot. In this case, no plot is shown. 

Here's a copy of the report for a sample where `CellAssign` had some unclassified cells so you can see that the plots still render correctly. 
[SCPCL000822_celltype-report.html](https://github.com/user-attachments/files/31757787/SCPCL000822_celltype-report.2.html)

Once this goes in, I can regenerate the Portal test data and return the missing cell type reports to that file structure.

### DIFF
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -451,7 +451,7 @@ if (length(all_unclassified_methods) > 0) {
     "\n  All cells in this library were unclassified by {all_unclassified_string}, so no tables or plots are shown for {method_word}."
   )
 } else {
-  skippted_text <- ""
+  skipped_text <- ""
 }
 
 # print out a warning about unclassified cells
--- a/templates/qc_report/celltypes_qc.rmd
+++ b/templates/qc_report/celltypes_qc.rmd
@@ -21,25 +21,6 @@ has_celltype_annotations <- function(df, column) {
   column %in% colnames(df) && !all(is.na(df[[column]]))
 }
 
-#' Convert a factor level to NA
-#'
-#' Values with the given level become NA and the level itself is dropped, so the
-#'  cells remain in the data frame but are excluded from that column's tables and
-#'  plots. This is a base R stand-in for `forcats::fct_na_level_to_na()`, which
-#'  requires forcats >= 1.0.0.
-#'
-#' @param f Factor (or character vector, which will be coerced to a factor)
-#' @param level The level to convert to NA, as a string
-#'
-#' @return Factor with `level` converted to NA and removed from the levels
-level_to_na <- function(f, level) {
-  if (!is.factor(f)) {
-    f <- factor(f)
-  }
-  levels(f)[levels(f) == level] <- NA
-  return(f)
-}
-
 #' Create tables of cell type annotation counts
 #'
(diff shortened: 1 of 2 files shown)


templates/qc_report/celltypes_qc.rmd
```diff
@@ -8,14 +8,53 @@ This file is meant to be run as a child report within either `main_qc_report.rmd
...
+#' @param level The level to convert to NA, as a string
+#'
+#' @return Factor with `level` converted to NA and removed from the levels
+level_to_na <- function(f, level) {
+  if (!is.factor(f)) {
+    f <- factor(f)
+  }
+  levels(f)[levels(f) == level] <- NA
+  return(f)
+}
```
sjspielman: I think you should be able to use `forcats::fct_recode()` instead of needing this function. Example below based on their docs https://forcats.tidyverse.org/reference/fct_recode.html

```
> x <- factor(c("apple", "bear", "banana", "dear"))
> x <- forcats::fct_recode(x, NULL = "dear") 
> levels(x)
[1] "apple"  "banana" "bear"  
> x
[1] apple  bear   banana <NA>  
Levels: apple banana bear
```

templates/qc_report/celltypes_qc.rmd
```diff
@@ -300,75 +350,139 @@ umap_facet_point_size <- umap_points_sizes[2]
 ```
 
 
-```{r, eval = (has_cellassign || has_singler || has_scimilarity) && (is_supplemental || !has_consensus), results='asis'}
+```{r, eval = (has_cellassign || has_singler || has_scimilarity) && (is_supplemental || !has_consensus)}
 unclassified_methods <- c()
+all_unclassified_methods <- c()
```
sjspielman: I feel like this is going to get tricky to maintain, with these two vectors being so similarly named, but not necessarily a huge deal? I'll leave it to you to decide how big a deal (aka if you wanted to rename one of them), but at a minimum I would definitely add some comments about what each means and how it functions later in the notebook.

sjspielman (commented):
Something seems to be up with the cellassign - there's a facet labeled `NA` now which probably is because we changed the unclassified string -> NA?

Image

allyhawkins:
Alright that issue with the NA facet has now been fixed and I went through the other plots and made modifications where necessary to make sure we don't have any NA's showing up. I also removed the redundant function. This should be ready for another look. 

 
[SCPCL000822_celltype-report.html](https://github.com/user-attachments/files/31760044/SCPCL000822_celltype-report.3.html)

sjspielman (approved):
looks good!

PR #1341, opened 2026-09-03
Make sure methods without any classified cells are not included in the Jaccard heatmaps

allyhawkins:
As a follow up to #1340, I ran into one more error where we were still trying to make a Jaccard matrix using a matrix with 0 cells as input. This adds a fix to make sure that when we make the heatmap comparing automated celltype annotations to consensus, we don't include any methods where all cells are considered "unclassified". These cells are those that were skipped when re-processing because we don't repeat cell typing. I've never seen this in practice but have only seen it with the test data that we have setup since the numbers are small. 

Once this goes in, I will again try to regenerate test data, which can only be done using the `main` or `development` branches.


jaclyn-taroni (approved):
👍🏻

PR #1343, opened 2026-09-08
only create rename if there are extra columns to rename

jashapiro:
When submitting https://github.com/AlexsLemonade/ScPCA-admin/issues/1317, I ran into an error for the case where there are no extra columns in the submitter cell type data. This is because the `paste0()` always returns at least one result, even if the one argument is empty. So here I am fixing that with a quick check before creating the named vector. 

I have a feeling this will require another point release, but we can get to that afterward.


allyhawkins (approved):
LGTM, thanks for fixing this!

PR #1344, opened 2026-09-09
Fix remaining bug with submitter renames

jashapiro:
Since I had not really tested the script changes I had made, I missed some places a variable now hidden in the `if` was being used. I fixed that now, and even tested the changes in https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/2WSR1vxFbDmrh3/tasks

(Got through this script at least; we'll see if that is everything needed)


allyhawkins (approved):
LGTM

PR #1345, opened 2026-09-09
Revert accounting for all unclassified cells

allyhawkins:
Last week I made some changes to account for a case with the test data where none of the cells in the checkpoints folder overlapped with the cells in the object resulting in the cell types being all "Unclassified cell". I was hitting some more snags, which was making me suspicious, so I went back and looked and we had copied over the wrong checkpoints for that specific test dataset. This makes sense since the likelihood of getting 0 cells that overlap upon a re-run seems very slim. We also probably want to make sure that's an error so we can look into. 

Here I'm just reverting the changes I had started to make. Note that this should hopefully help with whatever test runs @jashapiro is working on for the submitter cell types, since the last PR had introduced a breaking change that I had not yet fixed.


jashapiro:
> we had copied over the wrong checkpoints for that specific test dataset.

Do you happen to have an idea how that might have happened? We would want to fix it so we don't do that again, if possible.

allyhawkins:
> > we had copied over the wrong checkpoints for that specific test dataset.
> 
> Do you happen to have an idea how that might have happened? We would want to fix it so we don't do that again, if possible.

This was from me copying things over for the new flex libraries test data back in June. In general, we have been manually copying over `checkpoints` files for generating the portal test data, since adding new libraries doesn't happen very often and we have to change ID numbers. See https://github.com/AlexsLemonade/ScPCA-admin/tree/main/portal-test-data#test-data-provenance. 

I think in the future we could automate this somehow so we have a process to add any new checkpoints files, but we just haven't set that up yet.

jashapiro:
> > > we had copied over the wrong checkpoints for that specific test dataset.
> > 
> > 
> > Do you happen to have an idea how that might have happened? We would want to fix it so we don't do that again, if possible.
> 
> This was from me copying things over for the new flex libraries test data back in June. In general, we have been manually copying over `checkpoints` files for generating the portal test data, since adding new libraries doesn't happen very often and we have to change ID numbers. See https://github.com/AlexsLemonade/ScPCA-admin/tree/main/portal-test-data#test-data-provenance.
> 
> I think in the future we could automate this somehow so we have a process to add any new checkpoints files, but we just haven't set that up yet.

So this was specific to the **portal** test data, not the general sandbox data for testing the workflow?

jashapiro (approved):
Seems good to me.

allyhawkins:
> So this was specific to the portal test data, not the general sandbox data for testing the workflow?

Yes, that's correct.

PR #1348, opened 2026-09-21
Remove last show_celltype variable

allyhawkins:
This should fix the current error that's being reported in https://cloud.seqera.io/orgs/CCDL/workspaces/ScPCA/watch/5iDaE1tqeO0UoN/logs where the `show_singler` variable can't be found. Looks like I missed a spot when reverting some of the changes that I had made.


jashapiro (approved):
👍🏼

